PR #1, opened 2019-07-23
LICENSE: add year and licensee ("Contributors")

StefanKarpinski:
This just fills in the two Apache License template fields: [year] and [name of copyright owner]. You could also put the names of the original authors "and Contributors" which is what we do for Julia itself.


emstoudenmire:
Thanks! Good catch

PR #3, opened 2019-07-23
Index documentation improvements

MasonProtter:
This is a PR where I'll start adding some documentation improvements and some basic code cleanup (i.e. so far there was a redundant `Index` constructor I removed). 

Suggested edits would be greatly appreciated.


mtfishman:
Looks good to me, thanks!

PR #7, opened 2019-07-24
Overload `show` for IndexSet

nickrobinson251:
Let me know if this should print differently from this :) Should be easy to update 

Before:
```julia
julia> k = Index(4,"k");

julia> l = Index(4,"l");

julia> T = randomITensor(k,l);

julia> inds(T)
IndexSet(Index[(4|id=429|k), (4|id=0|l)])
```

After:
```julia
julia> k = Index(4,"k");

julia> l = Index(4,"l");

julia> T = randomITensor(k,l);

julia> inds(T)
IndexSet((4|id=589|k), (4|id=986|l))
```


emstoudenmire:
Big improvement! Thanks we needed this

PR #8, opened 2019-07-24
update keyword argument list for svd doc; some changes to truncation!

YiqingZhouKelly:
Hi,
I updated the keyword argument list in SVD documentation. And also made some small changes to truncate!()

Best,
Yiqing


emstoudenmire:
Thanks these are all very good improvements

PR #10, opened 2019-07-24
Fix travis file and add codecov?

kshyatt:
Science is incredible

Let's wait and see if codecov submission works before merging this!


kshyatt:
OK, codecov is working!

PR #13, opened 2019-07-24
Reduce dependencies

nickrobinson251:
AFAICT `KrylovKit` and `Permutations` are not used right now, and `Combinatorics` is just needed for tests so can be test-only dependency


emstoudenmire:
Agreed this is a very helpful improvement thanks

PR #15, opened 2019-07-24
Improve errors and add tests

kshyatt:


src/itensor.jl
```diff
@@ -34,7 +34,7 @@ ITensor(::Type{T},inds::Index...) where {T<:Number} = ITensor(T,IndexSet(inds...
 
 function ITensor(::UndefInitializer,
                  inds::IndexSet)
-  return ITensor(inds,Dense{Float64}(undef,dim(inds)))
+  return ITensor(inds,Dense{Float64}(Vector{Float64}(undef,dim(inds))))
```
kshyatt: You need this `Vector` here to avoid an error btw

emstoudenmire: Good fix!

PR #16, opened 2019-07-24
Array-like functions

mcabbott:
This makes a few simple functions from Base act on `ITensor`s, and overloads `randn(i,j)` for `i::Index`. Some duplicate existing functions, like `ndims == order`. 

You may not want all of these, but I wonder what the rule is. And more generally, why `ITensor <: AbstractArray` is not true?


MasonProtter:
It'd be good to add some basic tests for these functions.

src/itensor.jl
```diff
@@ -71,11 +71,18 @@ dims(T::ITensor) = dims(inds(T))
 
 dim(T::ITensor) = dim(inds(T))
 
+Base.ndims(A::ITensor) = order(inds(A))
+size(A::ITensor) = dims(inds(A))
+size(A::ITensor, d) = d in 1:ndims(A) ? dim(inds(A)[d]) : 
+  d>0 ? 1 : error("arraysize: dimension out of range")
```
kshyatt: The error type here is not consistent with `Base`

src/itensor.jl
```diff
@@ -71,11 +71,18 @@ dims(T::ITensor) = dims(inds(T))
...
+
 isNull(T::ITensor) = (store(T) isa Dense{Nothing})
 
 copy(T::ITensor) = ITensor(copy(inds(T)),copy(store(T)))
 
 Array(T::ITensor) = storage_convert(Array,store(T),inds(T))
+Base.Matrix(A::ITensor) = ndims(A) == 2 ? Array(A) : error("expected ndims=2")
```
kshyatt: Should also be `ArgumentError` here

mtfishman:
This seems good, though is there a particular reason to define the `randn(::Index...)` type constructors? That would maybe make me think it makes an Array based on the dimensions of the Indices, it seems like `randomITensor` is more clear.

emstoudenmire:
I agree with the comments above. Let's please make the following changes:
* add unit tests & fix any errors
* please remove the `randn` overloads

src/itensor.jl
mcabbott: While I was just being lazy, now that I check I see Base throws the same error here, somewhat surprisingly: 
```
@test_throws ErrorException size(rand(2),-42)
```

mtfishman:
And to answer why `ITensor` is not `<: AbstractArray`, it is because in general the goal of an ITensor is to have a memory-independent interface, which is at least conceptually at odds with the AbstractArray interface. It could work, but we would want to check that making ITensor a subtype of AbstractArray adds some value. For now we will just add our own ITensor functions and keep it in mind as a possibility.

mcabbott:
OK I tidied up a bit, added some tests. I've deleted the `randn` overload. I added some docstrings, and put indents into several others. 

One of these is `mul!(A,x,B)`. I wonder if this should be `scale!`, as `LinearAlgebra.mul!` is matrix multiplication. But I did not rename it. 

I do still wonder about `ITensor <: AbstractArray`, as this might make it easier to compose with other things outside this package. (For instance, to use your SVD in a calculation which is otherwise not very tensor-contraction heavy). It would not imply a fixed storage type, only the `eltype` and `ndims` need to be part of the type. I had a go in another branch but ran into problems with the way you do `TR + TC` by mutating `copy(TR)`, added `some promote_type(eltype...` things but far from finished.

src/itensor.jl
mcabbott: Is `DimensionMismatch` also OK? If not I can change it.

test/test_itensor.jl
```diff
@@ -369,7 +383,7 @@ end
     @testset "Test polar decomposition of an ITensor" begin
       U,P = polar(A,(k,l))
       @test A≈U*P
-      #Note: this is only satisfied when left dimensions
```
mcabbott: I see that my editor ate the trailing spaces, sorry about the extra noise.

MasonProtter:
I think it's somewhat important to make `ITensor <: AbstractArray`. `AbstractArray` makes no promises about memory layout at all. One of the biggest advantages of julia is generic code reuse, so if we do a good job of implementing `ITensor <: AbstractArray` then one can plug it into all sorts of generic `AbstractArray` code and have it 'just work'.

On this topic, I actually think `rand` should replace `randomITensor`. The common interface for random instances of a type `T` is `rand(T, args...)` so we could just replace `randomITensor(args...)` with  `rand(ITensor, args...)`. This will make generic interfaces easier and not cost any verbosity.

mtfishman:
> 
> One of these is `mul!(A,x,B)`. I wonder if this should be `scale!`, as `LinearAlgebra.mul!` is matrix multiplication. But I did not rename it.

I went back and forth about what to call this. I wasn't sure if `scale!` was great since I would think that would be more like `scale!(A,x)` to scale `A` in-place by `x`. In the end, we would prefer to use broadcast notation for this, such as:

```julia
A .= x .* B
```

so hopefully `mul!` will just become an internal function anyway.

> 
> I do still wonder about `ITensor <: AbstractArray`, as this might make it easier to compose with other things outside this package. (For instance, to use your SVD in a calculation which is otherwise not very tensor-contraction heavy). It would not imply a fixed storage type, only the `eltype` and `ndims` need to be part of the type. I had a go in another branch but ran into problems with the way you do `TR + TC` by mutating `copy(TR)`, added `some promote_type(eltype...` things but far from finished.

It is something to consider, I think we would want to see a particular use case before we made a design decision like that. To me it makes more sense to make the subtypes of `TensorStorage <: AbstractArray` since their interface fit better.

mtfishman:
> I think it's somewhat important to make `ITensor <: AbstractArray`. `AbstractArray` makes no promises about memory layout at all. One of the biggest advantages of julia is generic code reuse, so if we do a good job of implementing `ITensor <: AbstractArray` then one can plug it into all sorts of generic `AbstractArray` code and have it 'just work'.

I think it would be good to see a particular use case of where that would be useful, but it is something I would be open to. For now we are focused on implementing all of the functions we want for ITensors, which in general are pretty custom and I believe generally orthogonal to the functionality of the AbstractArray interface (for example, it is generally not necessary to call `permutedims` on an ITensor, since permutations are handled automatically by ITensor functions). One use case could be that it makes it easier to implement broadcasting over ITensors, but I need to dig into that more.

> 
> On this topic, I actually think `rand` should replace `randomITensor`. The common interface for random instances of a type `T` is `rand(T, args...)` so we could just replace `randomITensor(args...)` with `rand(ITensor, args...)`. This will make generic interfaces easier and not cost any verbosity.

`rand(ITensor, args...)` could be a good alternative interface, I think the single function `randomITensor(...)` is still good for simplicity. Again, I think we would want to see a situation where there is a clear advantage before we start adding more functions (that seem to do the same thing).

mcabbott:
On rand, it's a little confusing to know what the ideal spelling is... `rand(Float32, 3,4) isa Matrix` but `rand(Matrix{Float32}, 3,4)` is an error. So while `rand(ITensor, i,j,k)` looks nice (to me!) it may not allow much to be made generic. Another possible spelling is `randn!(ITensor(i,j,k))`, I guess, like `rand!(Matrix{Float32}(undef, 3,4) )`. 

Supporting broadcasting would be neat, and a move towards being generic. Am not sure I see the conflict between `scale!(A,x)` and `scale!(A,x,B)`, like `add!` you could add as many methods as seem useful, but they all do similar element-wise things, if I understand right. Whereas `mul!` is the friendly spelling of part of  `gemm!`.

emstoudenmire:
One serious issue with inheritance as an approach to generic programming is that you can only inherit from one thing, or else you have to deal with multiple-inheritance issues. And there are other issues.

But it could definitely be good to let ITensors be pluggable into existing abstract array code.

So how about the following compromise approach: we could (just as an experimental thing for a while) make a wrapper type that holds an ITensor and inherits from AbstractArray. It wraps all of the AbstractArray functions like `ndims` by calling the appropriate function on the ITensor it holds.

Then we could make a function `treatAsArray` which takes an ITensor and returns it wrapped into this wrapper object. So then (hypothetically) one could do things like `sort!(treatAsArray(T))` where T is an ITensor.

Feel free to destroy this idea!

MasonProtter:
Making `ITensor <: AbstractArray` is quite unlikely to make anyone's life easier from within ITensors.jl and will likely not benefit any usecases we can easily foresee right now. The advantage of generic code is to users who are doing things with the package that we don't expect. 

I think that if ITensors.jl is well written enough, the `ITensor` type can become a lingua franca for tensor based operations in julia. For instance, someone writing a neural network may think 

>Hey, `Array`s are kinda annoying for the sort of things I want to do with my neural network. I just want these indices to correspond to a weight and this index to correspond to a bias and I want general contractions between my tensors to be smart enough to respect that and I don't want to worry about the actual details of the memory layout.

An `ITensor` may indeed be a better option for that person, but that dream may shatter if they can't take an `ITensor` and stick it into a package that wasn't designed knowing about `ITensor`s. The easiest way to do that is to respect and support common interfaces. 

Regarding `permutedims`: yes, ITensor users may not need to use that, but it still makes sense to implement it, afterall it's just a shuffling of indices and then any code that relies on `permutedims` (say TensorCast.jl) can straightforwardly use an ITensor. None of this means that we have to give up the current ITensor interface, it just means that if we support more of the `AbstractArray` interface, this code will be more useful to more people. 

For instance, the DifferentialEquations.jl and the Flux.jl people probabably never dreamed that they'd one day be supporting shoving eachothers code together, but mere days after a NeuralODE paper dropped, they were able to write a couple hundred lines of code and suddenly they had their own NeuralODE implementation https://julialang.org/blog/2019/01/fluxdiffeq .

mtfishman:
> On rand, it's a little confusing to know what the ideal spelling is... `rand(Float32, 3,4) isa Matrix` but `rand(Matrix{Float32}, 3,4)` is an error. So while `rand(ITensor, i,j,k)` looks nice (to me!) it may not allow much to be made generic. Another possible spelling is `randn!(ITensor(i,j,k))`, I guess, like `rand!(Matrix{Float32}(undef, 3,4) )`.

Yes, I don't know what the signature would be for `randn` to construct an ITensor. We should definitely support `randn!(ITensor(i,j,k))` anyway (probably `randomITensor(i,j,k)` would just call that). We will also have similar functions like `randn!(::MPS)` and `randn!(::MPO)` if we don't already.

> 
> Supporting broadcasting would be neat, and a move towards being generic. Am not sure I see the conflict between `scale!(A,x)` and `scale!(A,x,B)`, like `add!` you could add as many methods as seem useful, but they all do similar element-wise things, if I understand right. Whereas `mul!` is the friendly spelling of part of `gemm!`.

Maybe `mul!` isn't the right choice given how it is already used in Julia, it just seems like it reads well as "multiply these things to together, then store into the preallocated ITensor". In that sense `mul!(A,x,B)` is nearly the same as `mul!(A,ITensor(x),B)` (which should also work), in the sense that `mul!` of ITensors is the contraction of ITensors, so `mul!(A,x,B)` is like the contraction of an order-0 tensor with an order-n tensor. I think of `scale!(A,x)` as involving only as scaling in-place, while `mul!(A,x,B)` doesn't really do that (it scales another ITensor, then stores the result in a pre-allocated ITensor). I don't think it is too bad here to diverge a bit from the Julia definition, I think it is pretty clear what it means in the context of ITensors.

mcabbott:
It already almost supports the `AbstractArray` interface, which is just `size`, `ndmis`, `getindex` I think. Of course this thinks of indices being integers in a fixed order, but this need not be storage order. `inds(A)` has an order, and `A[2,3]` works. The hurdle to my #array branch passing tests was `real + complex` being implemented with a mutation (of the container not the storage). 

A potential upside is that `@code_warntype` might give less bright red `Any`s. I haven't dug enough to know whether that matters for performance. (Although I guess that's just about the type of ITensor not its `<:` relations.)

A potential downside is that if you intend to hack broadcasting such that `ITensor .+ ITensor` goes only by the named indices, then it could complicate your life: I guess `ITensor .+ Matrix  isa Matrix`, but would  `ITensor .= ITensor .+ Vector` be an error, if their orders disagree? 

But anyway maybe this question is beyond this little PR.

emstoudenmire:
@MasonProtter: I am very sympathetic to this idea and definitely excited about the vision of generic programming. Here's why I'm being conservative about the <: AbstractArray thing though: why AbstractArray? Why not some other type we haven't heard of yet but which could be closer in spirit to ITensor? (<: UnorderedArray or <: AbstractTensor etc.). So I worry that if we put <: AbstractArray and people start writing code that depends on that, we will get locked in to this choice and if we change it it later, it will be a breaking change. So we may do it but I don't want to do it hastily.

On a related note, is there a way to write an "is_array" trait in Julia as an alternative to inheriting from AbstactArray?

MasonProtter:
Since the base language already exports `AbstractArray`, it has already kinda solidified as the common super-type for tensor like ideas and if something like `UnorderedArray` or `AbstractTensor` appeared, they would likely be `<:AbstractArray` as well. In fact, we would probably benefit from making our own type `AbstractTensor <: AbstractArray` and have `ITensor <: AbstractTensor`. 

I think you'll find that the `AbstractArray` interface actually makes very few assumptions about what's inside it. Basically all it assumes is index-ability and some notion of dimensionality / rank as @mcabbott said.

mtfishman:
I think one concern with using the `AbstractArray` interface would be that it would automatically define certain functions on ITensor that we might not want to implement right now  (i.e. something that changes the storage, but doesn't do the correct corresponding thing with the indices, such as a reshape or slicing). This could become a bit of a rabbithole of fixing unintended overloaded behavior. 

I definitely appreciate that it would be good to make ITensor play nicely with other packages (for example, we want ITensor to work with packages for iterative solvers). However, because of the possible unintended consequences and development overhead, I think it is best to wait for a good application where we need `ITensor <: AbstractArray` (if you come up with one, we would be happy to hear about it!).

MasonProtter:
As long as the corresponding basic functions `getindex`, `setindex!`, etc. are defined properly, I'm not sure how functions composed of those would end up abusing the index structure.

emstoudenmire:
@MasonProtter: although we don't do this very much, in principle functions which operate on ITensors can dynamically return the tensor indices in a different order and even ITensors with different numbers of indices. So this could make AbstractArray code that relies on the order not changing very brittle to use in conjunction with ITensor. One could protect against this by inserting lines of code that make sure the indices are in a certain, expected order though. This could actually be another argument in favor of my `treatAsArray` function idea in fact: it could require that you put the indices in the order you want when making the array wrapper object.

mcabbott:
Re `mul!`, sorry my mistake, I was misunderstanding this function. It's like `add!(A,0,x,B) = A .= x .* B` not like `scale!(A,x,B) = A .*= x .* B`. Which is then OK, since `mul!(rand(3), 5, ones(3)) == [5,5,5]` does exactly this. Docs corrected now, and one more test. 

Re `reshape`, here's what happens (on my #array branch):
```
julia> A = randomITensor(i,j)
ITensor{2} (2|id=293|i) (4|id=509)
Dense{Float64}:
 0.458029  0.427099  -0.0382075  -0.767498
 0.225983  0.806718  -1.02337     0.868276

julia> B = reshape(A, 4,2)
4×2 reshape(::ITensor{2,Float64,Dense{Float64}}, 4, 2) with eltype Float64:
 0.458029  -0.0382075
 0.225983  -1.02337  
 0.427099  -0.767498 
 0.806718   0.868276 

julia> B isa ITensor
false
```
i.e. you fall back to having something else `<: AbstractArray`, which is ignorant of indices. It's only touching the ITensor through `A[2,3]` etc. That seems OK to me, in that you do support indexing directly like this. But I agree the potential space of designs here is quite large. If you want maximum freedom then perhaps you should not expose integer indexing at all? 

On my other computer I was making some convenience macros for wrapping & unwrapping the data, which is another way to go back and forth to ordinary code. If that sounds interesting I will make a PR.

mcabbott:
Here's the 5-minute version of the wrapper idea, note that `permute(x, ijk...)` isn't lazy
```
struct WrappedTensor{T,N} <: AbstractArray{T,N}
    tensor::ITensor
    WrappedTensor(x::ITensor) = new{eltype(x), order(x)}(x)
end

Base.AbstractArray(x::ITensor) = WrappedTensor(x)
Base.AbstractArray(x::ITensor, ijk::Index...) = WrappedTensor(permute(x, ijk...))

ITensors.ITensor(x::WrappedTensor) = x.tensor

Base.size(x::WrappedTensor) = dims(x.tensor)
Base.getindex(x::WrappedTensor, i...) = x.tensor[i...]
Base.setindex!(x::WrappedTensor, y, i...) = setindex!(x.tensor, y, i...)
```

emstoudenmire:
Actually there are a lot of nice things in this pull request (I like having things like `ndims` work for ITensors, for example, and there are docstrings) so I'm reopening. 

I'll probably make a few small changes in the process of merging the pull is all.

PR #18, opened 2019-07-25
Add basic DMRG test AND a test for tagset

kshyatt:
Just make Miles review infinite PRs...

I also added a kwarg to make DMRG quieter. `isNull` should match between `TagSet` storage and `ITensor` imo. I changed `setindex` -> `setindex!` because I'm pretty sure code earlier in the file calls `setindex!` and not the immutable version.

### DIFF
--- a/src/mps/dmrg.jl
+++ b/src/mps/dmrg.jl
@@ -14,6 +14,7 @@ function dmrg(H::MPO,
   position!(PH,psi0,1)
   energy = 0.0
 
+  quiet::Bool = get(kwargs,:quiet,false)
   for sw=1:nsweep(sweeps)
     sw_time = @elapsed begin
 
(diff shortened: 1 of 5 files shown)


emstoudenmire:
No problem about the PRs. Great to have these improvements & especially more tests!

kshyatt:
The DMRG test doesn't take very long which is why I left it like that.

`setindex!` isn't really modifying, no. This code path isn't used yet either way. I can strip that part out and we can come back to it later?

emstoudenmire:
Thanks. Yes how about if you just take out that one setindex -> setindex! redefinition and then I’d approve of this PR. (Also feel free to pull it then if you want!) Having the quiet keyword for DMRG will be very useful in some contexts.

kshyatt:
Done!

PR #21, opened 2019-07-25
Add some siteset tests and make error type better

kshyatt:


test/test_siteset.jl
```diff
@@ -7,9 +7,22 @@ using ITensors,
 
   @testset "Star in operator strings" begin
     sites = spinHalfSites(N)
+    @test_throws ArgumentError op(sites, "Sp")
```
emstoudenmire: Hi Katie, not sure what this is intended to test. There isn't a method `op(::SiteSet,::String)`; it requires an integer argument at the end too. So right now it's throwing a MethodError and failing the test.

kshyatt: Ah derp, sorry about that! Early AM testing 😰

kshyatt:
Updated! Better now? Fixed some of the storage issues for complex ops too...

PR #23, opened 2019-07-25
Tests for IndexSet and eliminate double return

kshyatt:


test/test_indexset.jl
```diff
@@ -10,12 +10,24 @@ using ITensors,
   j = Index(jdim,"j")
   k = Index(kdim,"k")
   l = Index(ldim,"l")
+  @test size(IndexSet(i, j)) == (2,)
```
mtfishman: Maybe we need to think about what `size` of an IndexSet means (and possibly just remove it, since the main things we want are `dims(::IndexSet)` and `order(::IndexSet)`/`length(::IndexSet)`). At first I thought this should be `(2,3)`, i.e. the same as `dims(::IndexSet)`.

kshyatt: I think this `size` is really `ndims`. Shall I change it?

mtfishman: I think I only defined `size` on IndexSet since I thought that it was required for iteration, but maybe it is not since `size` isn't defined for tuple (but of course those are iterable). I think maybe we should just remove this test for now since it is something we should look into removing.

kshyatt: Goodbye, test!

PR #27, opened 2019-07-25
Fix many MPS/MPO bugs, more tests

kshyatt:
I decided to have `show` just print the indices of the MPS/MPO. I think we can define another function (like a `MIME` one) if you want to see all the elements.


emstoudenmire:
Agree - that sounds good!

MasonProtter:
Thanks for doing this, I had only just noticed the problems I introduced with those MPO / MPS changes and was going to fix them but you scooped me! Apologies for the sloppy code.

kshyatt:
No worries :). Gotta get that GREEN BADGE

PR #33, opened 2019-07-25
Use concrete storage in ITensor, MPS, MPO

MasonProtter:
Currently, MPS, MPO and ITensor are not fully type inferrable because `ITensor` stores the abstract type `TensorStorage`. This PR is a first step towards making everything parametric on the storage type. There's a lot to fix still and the tests fail currently but this is really important to the performance of the package.

### DIFF
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -208,15 +208,19 @@ end
 /(A::ITensor,x::Number) = A*(1.0/x)
 
 -(A::ITensor) = -one(eltype(A))*A
-function +(A::ITensor,B::ITensor)
-  C = copy(A)
-  add!(C,B)
-  return C
-end
-function -(A::ITensor,B::ITensor)
-  C = copy(A)
-  add!(C,-1,B)
-  return C
+
+
++(A::ITensor,B::ITensor) = add!(copy_promote(A, B),B)
+-(A::ITensor,B::ITensor) = add!(copy_promote(A, B),-1,B)
+
+function copy_promote(A::ITensor,B::ITensor)
+  if eltype(A) == eltype(B)
+    C = copy(A)
+  else
+    CT = promote_type(eltype(A),eltype(B))
+    C = ITensor(inds(A), convert(Dense{CT}, store(A))) # TODO promote_type on storage?
+  end
+  C
 end
 
 #TODO: Add special case of A==B
@@ -264,7 +268,8 @@ Add ITensors B and A and store the result in B.
 B .+= A
 """
 function add!(B::ITensor,A::ITensor)
-  B.store = storage_add!(store(B),inds(B),store(A),inds(A))
+    B.store = storage_add!(store(B),inds(B),store(A),inds(A))
+    B
 end
 
 """
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -14,12 +14,11 @@ export ITensor,
        scalar,
        store
 
-mutable struct ITensor
+mutable struct ITensor{T <: TensorStorage}
   inds::IndexSet
-  store::TensorStorage
+  store::T
   #TODO: check that the storage is consistent with the
   #total dimension of the indices (possibly only in debug mode);
-  ITensor(is::IndexSet,st::T) where T = new(is,st)
 end
 
 ITensor() = ITensor(IndexSet(),Dense{Nothing}())
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -15,10 +15,12 @@ export ITensor,
        store
 
 mutable struct ITensor{T <: TensorStorage}
-  inds::IndexSet
-  store::T
-  #TODO: check that the storage is consistent with the
-  #total dimension of the indices (possibly only in debug mode);
+    inds::IndexSet
+    store::T
+    #TODO: check that the storage is consistent with the
+    #total dimension of the indices (possibly only in debug mode);
+    ITensor{T}(inds, store) = new{T}(inds, store)
+    ITensor(inds, store)    = new{TensorStorage}(inds, store)
 end
 
 ITensor() = ITensor(IndexSet(),Dense{Nothing}())
@@ -48,6 +50,8 @@ function ITensor(A::Array{S},inds::IndexSet) where {S<:Number}
   return ITensor(inds,Dense{float(S)}(float(vec(A))))
 end
 ITensor(A::Array{S},inds::Index...) where {S<:Number} = ITensor(A,IndexSet(inds...))
+
+
 
 # Convert to complex
 complex(T::ITensor) = ITensor(inds(T),storage_complex(store(T)))
--- a/src/mps/mpo.jl
+++ b/src/mps/mpo.jl
@@ -1,14 +1,14 @@
 export MPO,
        randomMPO
 
-struct MPO
+struct MPO{T}
   N_::Int
-  A_::Vector{ITensor}
+  A_::Vector{ITensor{T}}
 
-  MPO() = new(0,Vector{ITensor}())
+  MPO() = new{Dense{Nothing}}(0,Vector{ITensor{Dense{Nothing}}}())
 
-  function MPO(N::Int, A::Vector{ITensor})
-    new(N,A)
+  function MPO(N::Int, A::Vector{ITensor{T}})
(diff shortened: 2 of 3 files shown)


MasonProtter:
The first issue I'm currently trying to solve is that this change somehow makes it so that now
```
i = Index(2,"i")
j = Index(2,"j")
TC = randomITensor(ComplexF64,i,j)
TR = randomITensor(Float64,i,j)
```
```julia
julia> TR + TC
ERROR: InexactError: Float64(-1.1452677317808255 - 0.2248844701998709im)
Stacktrace:
 [1] Real at ./complex.jl:37 [inlined]
 [2] convert at ./number.jl:7 [inlined]
 [3] setindex! at ./array.jl:780 [inlined]
 [4] copyto!(::IndexLinear, ::Array{Float64,1}, ::IndexLinear, ::Array{Complex{Float64},1}) at ./abstractarray.jl:807
 [5] Array at ./abstractarray.jl:799 [inlined]
 [6] convert at ./array.jl:488 [inlined]
 [7] Dense at /Users/mason/.julia/dev/ITensors/src/storage/dense.jl:7 [inlined]
 [8] convert at /Users/mason/.julia/dev/ITensors/src/storage/dense.jl:22 [inlined]
 [9] setproperty! at ./Base.jl:21 [inlined]
 [10] add!(::ITensor{Dense{Float64}}, ::ITensor{Dense{Complex{Float64}}}) at /Users/mason/.julia/dev/ITensors/src/itensor.jl:267
 [11] +(::ITensor{Dense{Float64}}, ::ITensor{Dense{Complex{Float64}}}) at /Users/mason/.julia/dev/ITensors/src/itensor.jl:213
 [12] top-level scope at REPL[6]:1
```
whereas before this problem did not show up.

mcabbott:
My attempt to fix exactly this test looked like this: 
https://github.com/mcabbott/ITensors.jl/blob/array/src/itensor.jl#L220

MasonProtter:
Clever. Are you also working on making the storage concrete?

mcabbott:
No, I stopped and must do other things!

MasonProtter:
My new problem: 
```julia
m,n = rand(0:20,2)
i = Index(m)
j = Index(n)
realData = rand(m,n)
complexData = realData+ zeros(m,n)*1im
A = ITensor(realData, i,j)
B = ITensor(complexData, i,j)

julia> A≈B
true
```
yet
```
realDataT = Array(transpose(realData))
C = ITensor(realDataT, j,i)
julia> C ≈ A
true

julia> C ≈ B
false
```

Wat.

emstoudenmire:
Hi Mason, thanks for the suggestion but we are not planning to parameterize the ITensor type by its storage at this time. I don’t think it actually results in much of a measurable speed benefit (except that doing it at one time in the past did help with catching crucial type-stability issues in some code). On the other hand, it would totally rule out certain more aggressive and impactful optimizations we would like to try in the future, such as delayed/lazy evaluation of ITensor storage which is triggered to evaluate by a call to setindex! for example.

Another drawback to parameterizing on the storage means there are certain conveniences we could not provide users, such as changing the storage type in-place in certain cases. Whether we will add these conveniences or not is something we’re debating, but it’s a good option to leave on the table right now.

emstoudenmire:
Also I agree that the adding code could be simplified a bit. 👍 But some of the complexity is due to the number of cases it needs to handle both now and in the future. (Some parts of Julia ITensors.jl may appear a bit over-engineered at the moment but it’s really because the C++ code does a lot of other things we haven’t finished bringing over yet.)

kshyatt:
As I recall when we *stopped* parameterizing the ITensor in the way suggested by this PR, we benchmarked it and did not see a big performance difference. @mtfishman is that what you remember?

emstoudenmire:
@MasonProtter: Also just wanted to say I hope you and @mcabbott don’t think we are just being stubborn and saying ‘no’ often to be disagreeable! You are actually touching on some very critical design decisions we have discussed at length within our core development team. Some of the decisions we made long ago and they have worked out very well for us over the better part of a decade. Others we have made recently as we got into Julia to ensure that we can pursue many of our future design goals.

MasonProtter:
Hm, I'm not sure I understand how parameterizing the type would cause problems with lazy evaluation or in place storage type changes. As long as the type changes are inferrable, there shouldn't be a problem right? Even if the type changes depend on some run-time properties and hence aren't inferrable, I'd think that the parametric type performance might 'merely regress to the same performance as the abstract case. But I'm no expert on this. 

> MasonProtter: Also just wanted to say I hope you and mcabbott don’t think we are just being stubborn and saying ‘no’ often to be disagreeable! You are actually touching on some very critical design decisions we have discussed at length within our core development team. Some of the decisions we made long ago and they have worked out very well for us over the better part of a decade. Others we have made recently as we got into Julia to ensure that we can pursue many of our future design goals.

Not at all! Similarly, I hope you all aren't annoyed by my constant suggestions to change your long thought out design. If you'd like me to back off on the design changes and only focus on the sort of suggestions that are in line with your current direction, do let me know. I realize that constantly justifying your decisions is a resource drain.

mtfishman:
Yes, I had tested parametrizing ITensor by the storage and didn't see a noticeable difference in performance, at least for larger tensors (I think for smaller tensors this would lead to some overhead).

One thing I was thinking @emstoudenmire was that I think we can parametrize ITensor by the storage type and still allow the storage type to switch if we want, by making the parameter default to the abstract TensorStorage type:

```julia
abstract type TensorStorage end

mutable struct ITensor{S <: TensorStorage}
  store::S
  ITensor(st) = new{TensorStorage}(st)
  ITensor{T}(st) where T = new{T}(st)
end

struct DenseReal <: TensorStorage
  data::Vector{Float64}
end

struct DenseComplex <: TensorStorage
  data::Vector{ComplexF64}
end

# Make an ITensor of type ITensor{TensorStorage}, which
# allows the type to switch
T = ITensor(DenseReal([1,2]))
T.store = DenseComplex([1,2])

# I know I don't want the type to switch, so I specify the
# concrete type
A = ITensor{DenseReal}(DenseReal([1,2]))
A.store = DenseComplex([1,2]) # gives an error
```

This could be a good compromise: if people want type stability and won't use the features that allow type switching, they can explicitly construct ITensors with a concrete type. However, this would make it so that we want to in some way factorize ITensor code in a way that basic functionality allows for type stability, and only certain optional parts allow for type switching (which may not be a bad design anyway).

emstoudenmire:
@mtfishman: I like that compromise idea. The only downsides (which are very minor!) I can see are that it makes the ITensor type definition a bit harder to understand for new users, and that people might overuse the concrete type feature and think the library is inflexible.

But indeed it could be a very valuable tool for calling specialized code without resorting to a second-level function call on the storage itself. Also I could imagine a benefit when putting many (thousands) of ITensors into a large array, so that all the memory would get packed together rather than it being an array of pointers to non-contiguous memory blocks.

MasonProtter:
@emstoudenmire 
> Also I could imagine a benefit when putting many (thousands) of ITensors into a large array, so that all the memory would get packed together rather than it being an array of pointers to non-contiguous memory blocks.

Shouldn't we already be hitting that problem with large MPS, since they store a `Vector{ITensor}`? I noticed before that `inner(::MPS, ::MPS)` was perhaps surprisingly slow so maybe that's part of it?

emstoudenmire:
As a technical note,  I wanted to mention that it's not a given that parameterized types and concrete types are faster. There are situations where they could actually be *slower*. 

One situation would be searching through a long array of ITensors, using a search criterion based only on the indices. If the storage was "wired" directly into the ITensor, then a lot of data which is irrelevant to the search would get "pulled in" to each cache line, requiring a lot more memory movement to perform the search.

On the other hand, if the storage type is not concrete, then it is just a pointer (64 bits) so it wouldn't take up as much space in the cache.

So I just wanted to say that, to mention that there's no rule or guarantee that concrete types are faster in all situations than abstract types. It depends on the context and the algorithm.

Finally, as you all know, even if they are faster, but are faster only for a part of the code that takes less than 1% of the running time then making them faster will not have a big overall effect (the change would be a micro-optimization, and these have a cost in terms of code complexity while not having a corresponding benefit).

PR #35, opened 2019-07-26
Tighten storage_scalar and more tests for Dense

kshyatt:


test/test_itensor.jl
```diff
@@ -98,6 +98,11 @@ end
   copyto!(A, B)
   @test A == B
   @test data(store(A)) == vec(N)
+  A = ITensor(M,i,j)
+  B = ITensor(N,j,i)
```
kshyatt: Want to test `storage_copyto!` when the permutation is not trivial here

emstoudenmire (approved):

PR #40, opened 2019-07-26
Implement swapprime for IndexSet, test it for IndexSet and ITensor

kshyatt:
I didn't want to change the function signature for `ITensor` but I think this should probably take two `IndexSet`s or `Index`s specifically, rather than varargs.


kshyatt:
Oh I tested `setprime` too, and `prime` and `noprime` explicitly for `IndexSet`.

kshyatt:
Dang what the heck, this passed locally. I'll try to fix.

kshyatt:
I think codecov is unhappy because I force-pushed. Is this ok to go?

emstoudenmire:
No problem. Actually I reimplemented it, because swapprime actually has a different definition! It takes a pair of integer prime levels and switches from the old to new primelevel given.

PR #44, opened 2019-07-26
One working and one not-working test for permuted-C contract

kshyatt:
The second case is broken (some of the indices on C are in the wrong place, specifically `beta` and `i`) and I'm unsure of how to fix it because of how incredibly complex `compute!` is. The first one passes, though! Because of the way we implemented `*` and `mul!`, it seems we weren't hitting these code paths for permuted output matrices.

### DIFF
--- a/src/storage/contract.jl
+++ b/src/storage/contract.jl
@@ -143,7 +143,6 @@ function compute!(props::CProps,
                   B::Array,
                   C::Array)
   compute_perms!(props)
-
   #Use props.PC.size() as a check to see if we've already run this
   length(props.PC)!=0 && return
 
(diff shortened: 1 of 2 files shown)


mtfishman:
Thanks for taking a look at this. I was facing similar problems with the contraction code and just picked examples and worked my way through them. Hopefully it can be caught and fixed in a simple case, e.g. `C_ik .= A_ij*B_jk` vs. `C_ki .= A_ij*B_jk` along with permutations of the inputs and indices.

mtfishman:
A lot of bugs were fixed for in-place contractions, so I think this is resolved.

PR #46, opened 2019-07-26
add macro indices

Roger-luo:
This would make things (at least the interface) more convenient and DRY

```jl
@indices i j k
```

equivalent to

```jl
i = Index(1, "i")
j = Index(1, "j")
k = Index(1, "k")
```

### DIFF
--- a/src/index.jl
+++ b/src/index.jl
@@ -18,7 +18,8 @@ export Index,
        ind,
        setprime,
        sim,
-       val
+       val,
+       @indices
 
 const IDType = UInt64
 
@@ -179,3 +180,15 @@ prime(iv::IndexVal,inc::Integer=1) = IndexVal(prime(ind(iv),inc),val(iv))
 adjoint(iv::IndexVal) = IndexVal(adjoint(ind(iv)),val(iv))
 
 show(io::IO,iv::IndexVal) = print(io,ind(iv),"=$(val(iv))")
+
+function indicem(xs::Symbol...)
+  ex = Expr(:block)
+  for (k, x) in enumerate(xs)
+    push!(ex.args, :($(esc(x)) = Index($k, $(string(x)))))
+  end
+  return ex
+end
+
+macro indices(xs...)
+  return indicem(xs...)
+end


mcabbott:
I was toying with something similar, which was going to allow `@Index (b,+)  (c,-)  5`:

https://github.com/mcabbott/ITensors.jl/blob/macros/src/util/macros.jl

mtfishman:
Hey Roger.

I am not sure how useful this is as it stands, most of the time we are not making indices of size 1. Also, most of the time the tags are not the same as the binding of the Index, for example:

```julia
s1 = Index(4, "Site, n=1, Electron")
l2 = Index(2, "Link, l=2")
```

Additionally, indices will have more complicated constructors later, when we need to pass quantum number information to them when there are conserved quantities.

It would definitely be nice to not have to write `Index` over and over for a bunch of indices. Maybe more of a code block would be nice, like:

@Index begin
  i = 4 | Site, n=1, Electron
  j = 2 | Link, l = 2
end

Again, though, this wouldn't handle the case when we need to pass more information to the indices.

Cheers,
Matt

mtfishman:
> I was toying with something similar, which was going to allow `@Index (b,+) (c,-) 5`:
> 
> https://github.com/mcabbott/ITensors.jl/blob/macros/src/util/macros.jl

I am a bit confused by the notation here, what index is made with `@Index (b,+)`? I think you want something to separately specify the name of the binding, the tags, and the size like my answer above.

I do like the notation:
```julia
A′ = @IT  A[i,j]
```
That seems neat, it could be something to consider.

mtfishman:
Though with the notation:
```julia
A′ = @IT  A[i,j]
```
I think it is better for this to only work if `i` and `j` already exist, since many times you want to go and define another ITensor with the same indices. Also, I think it is more clear having a separate part of the code where the indices are created and then later used.

mcabbott:
Like Roger I presume, I was picturing `@Index` saving typing when you have many similar indices, and can accept them being bound to their letters. Right now `@Index  i,Spin  j,Spin  2` binds an index with two tags to each of `i` & `j`, the bracket for `(b,+)` or `(c,n=1)` is a bit ugly but otherwise these would be much harder to parse. These don't work but could be made to: 
```
@Index  μ ν σ  4          # expand greek?
@Index  M N M' N'  10     # normalise primes somehow?
@Index  a b c d  3,colour # share size and some tags?
```
But as you say for very complicated indices there's no typing to be saved. 

The array macro `@IT` accepts many arrays, with the idea that their shared indices can then be defined once: 
```
AT,BT,CT = @IT A[i,j,k] B[j,k,l] C[i,k,n]  # defines i,j,k,l,n
```
If it doesn't define indices at all, then there is no gain really from `AT = ITensor(A, i,j,k)`; the point was to avoid writing `i = Index(size(A,1), "i")` too often. At the moment it'll silently re-use an existing index of the right name & size, which may or may not be a great idea... it does mean there's no difference putting `CT = @IT C[i,k,n]` on the next line.

MasonProtter:
@mtfishman 

> Again, though, this wouldn't handle the case when we need to pass more information to the indices.

What other kind of information would one need to pass to the indices?

Roger-luo:
Oh I see. Indices seem to contain a lot information. What about pairs? 

@indices 3=>i 4=>j

I think more information can be passed by having (infos…)=>i syntax. It saves you from repeated writing indices names at least.

Roger-luo:
I thought the first number is the index id. But it seems you generate that yourself. It would be nice if we could directly make use of gensym. But I'm not sure if this is fittable.

mtfishman:
@mcabbott:

> Like Roger I presume, I was picturing `@Index` saving typing when you have many similar indices, and can accept them being bound to their letters. Right now `@Index i,Spin j,Spin 2` binds an index with two tags to each of `i` & `j`, the bracket for `(b,+)` or `(c,n=1)` is a bit ugly but otherwise these would be much harder to parse. These don't work but could be made to:
> 
> ```
> @Index  μ ν σ  4          # expand greek?
> @Index  M N M' N'  10     # normalise primes somehow?
> @Index  a b c d  3,colour # share size and some tags?
> ```
> 
> But as you say for very complicated indices there's no typing to be saved.

I think most of the time, the variable names of the indices are not exactly the same as the tags. However if we find this is the case a lot it could be good to consider a more compact notation.

Also keep in mind that many times indices are created and stored in a vector, so indices are obtained by indexing into the vector, not from an individual variable (an example is that you may have a grid of tensors, and so you first define a grid of indices that you might store in a vector or matrix and then define the ITensors from those indices).

> 
> The array macro `@IT` accepts many arrays, with the idea that their shared indices can then be defined once:
> 
> ```
> AT,BT,CT = @IT A[i,j,k] B[j,k,l] C[i,k,n]  # defines i,j,k,l,n
> ```
> 
> If it doesn't define indices at all, then there is no gain really from `AT = ITensor(A, i,j,k)`; the point was to avoid writing `i = Index(size(A,1), "i")` too often. At the moment it'll silently re-use an existing index of the right name & size, which may or may not be a great idea... it does mean there's no difference putting `CT = @IT C[i,k,n]` on the next line.

I see, I suppose having to write `i = Index(size(A,1), "i")` over and over would be annoying. Again, though, it seems like this is assuming that the user wants the variable name and tag to be the same, which is not necessarily the case. Also, in the example of:
```julia
AT,BT,CT = @IT A[i,j,k] B[j,k,l] C[i,k,n]
```
do you propose that the variables `i,j,k,l,n` also come into scope? It is useful to have the indices available for later use (for defining other ITensors, priming indices of an ITensor, etc.).

mtfishman:
@MasonProtter:

> > Again, though, this wouldn't handle the case when we need to pass more information to the indices.
> 
> What other kind of information would one need to pass to the indices?

ITensor indices can also be divided into direct sums. If an Index is dimension 10, we may give a label or meaning to the range `1:5` and another label or meaning to the range `6:10`. This is useful for more easily defining block-sparse tensors, which show up a lot in quantum mechanics.

mtfishman:
@Roger-luo:

> I thought the first number is the index id. So it is bond dimension? But it seems you generate that id yourself. It would be nice if we could directly make use of gensym. But I'm not sure if this is fittable.
> 

Yes, the first number is the bond dimension.

I believe that I looked into `gensym` for this and it wasn't quite right (you have to keep in mind the example of reading and writing to disc: you want the ids to get saved properly and not clash with new indices you might make).

> Update: it can be dangerous to use random number as your id hash number. I'll post another PR when I'm back.

ITensor has been using random ids for quite a while at this point and I am not aware of it causing any problems (keep in mind that you only need to avoid clashing locally in a tensor network diagram, so the odds of clashing are quite small and we could always increase the integer type if that is a concern). Maybe @emstoudenmire can comment more about this.

MasonProtter:
> ITensor has been using random ids for quite a while at this point and I am not aware of it causing any problems (keep in mind that you only need to avoid clashing locally in a tensor network diagram, so the odds of clashing are quite small and we could always increase the integer type if that is a concern). Maybe @emstoudenmire can comment more about this.

I guess the worry there is that you're introducing a stochastic source of bugs, which would generally be very hard to locate. Every once in a while (even if it's very rare), one could get some strange error related to indexing and then just as suddenly as the bug appears, it'd disappear.

mcabbott:
@mtfishman
> ```julia
> AT,BT,CT = @IT A[i,j,k] B[j,k,l] C[i,k,n]
> ```
> do you propose that the variables `i,j,k,l,n` also come into scope? It is useful to have the indices available for later use (for defining other ITensors, priming indices of an ITensor, etc.).

Currently yes, `i,j,k,l,n` will all become defined, unless they are already `Index`s. (And if any existing ones are the wrong length, it should error.)

I'm not certain this is a useful thing at all, mostly I was just messing about while finding out what this package does.  But maybe someone finds a use case.

mtfishman:
@MasonProtter @Roger-luo 

> > ITensor has been using random ids for quite a while at this point and I am not aware of it causing any problems (keep in mind that you only need to avoid clashing locally in a tensor network diagram, so the odds of clashing are quite small and we could always increase the integer type if that is a concern). Maybe @emstoudenmire can comment more about this.
> 
> I guess the worry there is that you're introducing a stochastic source of bugs, which would generally be very hard to locate. Every once in a while (even if it's very rare), one could get some strange error related to indexing and then just as suddenly as the bug appears, it'd disappear.

It is reasonable to bring this up (and we appreciate getting more eyes on the code design as a check that the design is as good as possible), but I believe in practice it would be extremely rare to the point of not being a practical concern. Here is a simulation of index clashing I just tried:

```julia
const IDType = UInt64

struct Index
  id::IDType
  Index() = new(rand(IDType))
end

Base.isequal(i1::Index, i2::Index) = (i1.id == i2.id)

function check_clash(ncon,n=50)
  # make initial set of indices
  # to be contracted
  v = [Index() for i = 1:n]

  # check that initially there are no
  # clashes
  @assert allunique(v)

  # simulate contractions
  for i = 1:ncon
    # make a new index, simulating index generation
    # (from e.g. an svd)
    rand_pos = rand(1:n)
    v[rand_pos] = Index()

    # check there are still no clashes
    @assert v[rand_pos] ∉ v[1:rand_pos-1]
    @assert v[rand_pos] ∉ v[rand_pos+1:end]
  end
end

check_clash(1_000_000_000)
```

I chose 50 for the number of indices involved in our "contraction" since that is a pretty good upper bound on the number of indices that could practically be involved in a contraction on modern computers (e.g. think of a vector of size 2^50).

I ran this with a billion "contractions" and didn't see any clashes, and I doubt any single algorithm will involve a billion different contractions. We could move to `UInt128` if it is really a concern. 

Also remember that indices need to have equal tags and prime levels to be considered equal (and unequal indices will generally have different tags and prime levels in practice), so accidental clashes would be even more rare.

emstoudenmire:
Yes, just to chime in that’s a helpful check Matt. Since moving to 64 bit integers we haven’t had any random collision related bugs for I think about 5 years now. It was rare even with 32 bit, but did occasionally come up (like once or twice a year, depending on what people were doing) so that’s why we switched to 64. At the time Steve White did a similar calculation to Matt’s to predict the likelihood of collision and it’s astronomically low.

Roger-luo:
I think it is OK, if this is just a research package for someone's own use, from convenience aspect.

If it is a general package for users, it would be better to have more reliable implementation. The behaviour of this implementation is not deterministic (unless you ask the user set random seed), which also means once someone hit this, it is hard to re-produce error and it eventually become a heisenbug (even the probability is small, since the package is supposed to be used for general users, it is hard to tell that someone lucky that will hit this)

And on the other hand, it is not hard to do it. Just add a global constant counter would do the job

```julia
module ITensors

const _id_count = 1

### blabla

end
```

PR #47, opened 2019-07-26
Example with basic operations

nlw0:


examples/basic_ops/basic_ops.jl
```diff
@@ -0,0 +1,54 @@
+using ITensors
+
```
emstoudenmire: Would be a good idea to put a `let` ... `end` block around all the code to avoid global scope (which can cause some surprises to new Julia users)

emstoudenmire:
Great! I really like this

examples/basic_ops/basic_ops.jl
nlw0: Would a `main()` function be ok?

PR #48, opened 2019-07-26
Parameterize ITensors, MPS and MPO by their storage type. Default to abstract parameter 

MasonProtter:
Supersedes #33. This PR again uses parametric types in an attempt to improve efficience but is much less radical in that the concrete storage is *opt in*.


MasonProtter:
One thing I implemented here because I needed it to make the tests work was 
```julia
(==)(m::MPO, n::MPO) = (m.N_ == n.N_) && (m.A_ == n.A_)
(==)(m::MPS, n::MPS) = (m.N_ == n.N_) && (m.A_ == n.A_) && (m.llim_ == n.llim_) && (m.rlim_ == n.rlim_) 
``` 
otherwise, certain comparisons between identical `MPS` and `MPO` give the wrong answers. If needed, I can factor that change out into a separate PR.

emstoudenmire:
Thanks for the hard work, Mason, but I'm going to close this since we are pursuing a different design. The main reason we are sticking with un-parameterized ITensors is that it allows the storage type to change in-place, which can be very useful sometimes, and keeps open the possibility of some advanced optimization ideas we want to explore in the future.

We actually did some measurements of the difference in speed between an unparameterized and parameterized design, and found that the difference is only discernible when the 'payload' of an ITensor is very small, like a 10x10 matrix or so perhaps. Even then the difference is very small in absolute time terms.

PR #49, opened 2019-07-27
Got basic applyMPO working

kshyatt:
Some funniness with the prime levels could probably be fixed, but this appears to work!


emstoudenmire:
Awesome!

PR #50, opened 2019-07-29
Initial forays into nmultMPO

kshyatt:
Implement `denmatDecomp` (though not with noise, yet) and `nmultMPO` so that we can do some PEPS stuff! Does this test seem reasonable?

This code could easily be improved upon, I just wanted something to try out.


kshyatt:
Oh lol whoops I left a skeleton of `fitApplyMPO` chillin' in there. I can get rid of that after CI completes.

mtfishman:
Hey Katie,

Thanks for getting started on this!

I believe that the `denmat` function you made does the same thing as the `factorize` function (https://github.com/ITensor/ITensors.jl/blob/master/src/decomp.jl#L213). It does the following factorizations:

```julia
T = U*(S*V)   # "fromleft"
T = (U*S)*V   # "fromright"
T = (U*sqrt(S))*(sqrt(S)*V)   # "center"
```

It also allows you to choose an eigendecomposition or an svd for the factorization.

Cheers,
Matt

kshyatt:
Yeah, seems like it, since the tests pass this way! Thanks for pointing that out :)

kshyatt:
Is this mergeable? Working off it, I implemented `sum` for `MPS` and `MPO` as well...

src/mps/mpo.jl
```diff
@@ -258,3 +293,66 @@ function densityMatrixApplyMPO(A::MPO, psi::MPS; kwargs...)::MPS
...
+    maxdim::Int = get(kwargs,:maxdim,max(maxDim(A), maxDim(B)))
+    mindim::Int = max(get(kwargs,:mindim,1), 1)
+    n = length(A)
+    n != length(B) && throw(DimensionMismatch("lengths of MPOs A ($n) and B ($(length(B))) do not match"))
+    A_ = copy(A)
+    position!(A_, 1)
+    B_ = copy(B)
+    position!(B_, 1)
+
+    links_A = findinds.(A.A_, "Link")
```
emstoudenmire: Just a stylistic (minor) thing, but let's make an accessor method for MPO (and MPS) that returns A_ instead of writing .A_  directly.

src/mps/mpo.jl
```diff
@@ -258,3 +293,66 @@ function densityMatrixApplyMPO(A::MPO, psi::MPS; kwargs...)::MPS
...
+
+    for i in 1:n
+        if length(commoninds(findinds(A_[i], "Site"), findinds(B_[i], "Site"))) == 2
+            A_[i] = prime(A_[i], "Site")
+        end
+    end
+    res = similar(A_)
+    # otherwise we will end up modifying the elements of A!
+    res.A_ = deepcopy(A_.A_)
+    # need to reindex link indices of res
```
emstoudenmire: Maybe would be good to make this a separate function: seems like something we could use again in the future

src/mps/mpo.jl
```diff
@@ -258,3 +293,66 @@ function densityMatrixApplyMPO(A::MPO, psi::MPS; kwargs...)::MPS
...
+    sites_A = [setdiff(findinds(x, "Site"), findindex(y, "Site"))[1] for (x,y) in zip(A_.A_, B_.A_)]
+    sites_B = [setdiff(findinds(x, "Site"), findindex(y, "Site"))[1] for (x,y) in zip(B_.A_, A_.A_)]
+    res[1] = ITensor(sites_A[1], sites_B[1], commonindex(res[1], res[2]))
+    for i in 1:n-2
+        if i == 1
+            clust = A_[i] * B_[i]
+        else
+            clust = nfork * A_[i] * B_[i]
+        end
+        lA = commonindex(A_.A_[i], A_.A_[i+1])
```
emstoudenmire: Is getindex not defined for MPO? If so I think we should define and use it here: `A_[i]` instead of `A_.A_[i]`

test/test_mpo.jl
```diff
@@ -52,6 +52,21 @@ using ITensors,
     badpsi = randomMPS(badsites)
     @test_throws DimensionMismatch applyMPO(K,badpsi)
   end
+
+  @testset "nmultMPO" begin
+    psi = randomMPS(sites)
+    K = randomMPO(sites)
+    L = randomMPO(sites)
+    @test maxDim(K) == 1
+    @test maxDim(L) == 1
```
emstoudenmire: Later would be good to test nmult for MPOs with > 1 bond dimension. We can use AutoMPO for example to get some non-trivial MPOs

src/index.jl
```diff
@@ -34,7 +34,7 @@ associated with an index, i.e. the index leg is directed into or out of a given
 Reverse direction of a directed `Arrow`. Will throw an error for an input of `Neither`.
 """
 function -(dir::Arrow)
-  dir==Neither && throw(ArgumentError("Cannot reverse direction of Arrow direction 'Neither'"))
+  dir==Neither && return Neither #throw(ArgumentError("Cannot reverse direction of Arrow direction 'Neither'"))
```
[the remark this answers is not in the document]
kshyatt: Forgot to update the docstring, though!

src/mps/mpo.jl
```diff
@@ -159,7 +167,7 @@ end
 function show(io::IO, W::MPO)
   print(io,"MPO")
   (length(W) > 0) && print(io,"\n")
-  @inbounds for (i, w) ∈ enumerate(inds.(W.A_))
+  @inbounds for (i, w) ∈ enumerate(inds.(tensors(W)))
```
emstoudenmire: Hm does `inds.` here allocate a new array just to iterate? I'm sure it's not a real bottleneck or anything but it just seems like a thing we should avoid doing unless necessary. (I.e. a "premature pessimization".)

src/mps/mps.jl
```diff
@@ -80,22 +80,23 @@ function randomMPS(sites)
...
+similar(m::MPS) = MPS(m.N_, similar(tensors(m)), 0, m.N_)
 
 eachindex(m::MPS) = 1:length(m)
 
 
 function show(io::IO, M::MPS)
   print(io,"MPS")
   (length(M) > 0) && print(io,"\n")
-  @inbounds for (i, m) ∈ enumerate(inds.(M.A_))
+  @inbounds for (i, m) ∈ enumerate(inds.(tensors(M)))
```
emstoudenmire: Similar comment here about the iteration

emstoudenmire: Alternative could be:
```
for i in eachindex(tensors(M))
println(io,"$i $(inds(M[i]))")
end
```
if I have that correct...

emstoudenmire: Or:
```
for (i,A) in enumerate(tensors(M))
println(io,"$i $(inds(A))")
end
```

src/mps/mpo.jl
kshyatt: Should we instead enumerate the tensors, and within the loop get their indices?

emstoudenmire: Yes, I think that would be better. That is to do something like:
```
for (i,A) in enumerate(tensors(M))
```
then inside the loop body use `inds(A)` to grab just the indices of A.

kshyatt: Awesome, on it!

PR #51, opened 2019-07-30
rm rand in indices

Roger-luo:
this is a very tiny patch.

### DIFF
--- a/src/index.jl
+++ b/src/index.jl
@@ -22,6 +22,8 @@ export Index,
 
 const IDType = UInt64
 
+const __id_count__ = Ref(zero(UInt64))
+
 """
    Arrow
 `enum` type that can take three values: `In`, `Out`, or `Neither`, representing a directionality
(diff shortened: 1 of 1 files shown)


mtfishman:
Hi Roger,

How would this handle the situation of reading and writing to disk? Say I make indices:

```julia
i = Index(2) # id 1
j = Index(2) # id 2
k = Index(2) # id 3
A = randomITensor(i,j,k)
save("myfile.jld","itensor",A)
```

Now someone else starts a new session, does the global variable id start back at 1? Then we would have a clash if we tried to load an ITensor from previous sessions:

```julia
x = Index(2) # back to id 1?
B = randomITensor(x)
A = load("myfile.jld","itensor")
A*B  # Indices x and i would contract, even though they shouldn't
```

This would seem to require saving the global id whenever any ITensor object is saved and loading it when they are loaded, which sounds a bit unreasonable.

Perhaps @emstoudenmire also has some reasons why a globally incremented id hasn't been used.

Cheers,
Matt

emstoudenmire:
Yes this was one of the key issues random ids solved. A related one was serializing ITensors to send in messages like in MPI. 

Combining those key benefits with the fact that our current system has no known drawbacks we should keep the current id system as it stands.

PR #53, opened 2019-07-30
Implement adding MPOs and MPSs (without QNs)

kshyatt:
I changed the default direction of `Index` -- really not sure why it should be `Out` in the first place? Is there a reason for it to be? Having it as `Neither` allows the addition methods to work properly.


emstoudenmire:
Great to see so much functionality coming over from C++ so quickly. 

One change to this pull though: let’s rename + to sum, so that one can pass keyword arguments such as cutoff and maxdim. We had deprecated a plus operator overload before in C++ mainly for this reason.

emstoudenmire:
Oh also, having a default `Out` arrow for Index becomes very useful for parts of the code where arrows matter. So like for site sets one can just make Index's and not have to write `Out` for all of them. Then to get `In` Indices one just puts `dag`. 

It has the additional advantage that the same code in non-arrow-checking contexts "just works" when arrows become involved.

It's roughly similar to how Array functions (like `fill`) in Julia all make column vectors by default (or is it row vectors?) and then you have to do transpose to get a row vector, versus like if it made some thing that was neither a column nor a row vector or forced to you state that every time.

kshyatt:
The problem is that the default link indices in the MPO/MPS then all have `Out` direction, even if the MPO/MPS itself has no quantum numbers. Should I change the way the default c-tor works for these? Currently the constructors `MPS(sites)` and `MPO(sites)` have no way to detect QNs, I think...

emstoudenmire:
Yes, I was thinking that the Index objects will have an Out direction by default (this would be same as C++) and then the current code would just work in a way that is QN / Arrow agnostic. Ideally we would put `dag` in the same places in the current code as in the C++ to make it "future proof" but those will be easy to debug when we test with QNs soon.

kshyatt:
OK, it might, but I have to turn off the check in `plussers` then and we'll have to revisit this code once QNs are implemented

kshyatt:
All done :)

PR #62, opened 2019-08-01
Two small missing tests for MPO

kshyatt:
Call `position!` with a site argument `> 1`. Also test the error throw in `linkindex`.


kshyatt:
High speed merging before CI completes??? Danger zone Thursday here at CCQ.

PR #63, opened 2019-08-01
Actually use phys siteset tests, add more

kshyatt:
If you don't add the new test file to `runtests.jl`, then it won't be part of the coverage run! 😛


emstoudenmire:
D’oh!

PR #67, opened 2019-08-02
Implement combiner tensors

kshyatt:
Some of this code could be cleaned up but it appears to work.

### DIFF
--- a/src/storage/combiner.jl
+++ b/src/storage/combiner.jl
@@ -12,59 +12,58 @@ function storage_contract(CSstore::CombinerStorage,
     cind = Cis[1]
     if hasindex(dis, cind) # has combined index, uncombine
         cpos = findindex(dis,cind)
-        ninds = vcat(inds(dis)[1:cpos-1],inds(Cis)[2:end],inds(dis)[cpos+1:end])
-        Nis = IndexSet(ninds)
+        dinds = inds(dis)
+        Cinds = inds(Cis)
+        Nis = IndexSet(vcat(dinds[1:cpos-1], Cinds[2:end], dinds[cpos+1:end]))
         return Nis, Dstore
     else # lacks combined index, combine
         # dis doesn't have cind, replace
         # Cis[1], Cis[2], ... with cind, may need to permute
-        j_1_pos = indexpositions(dis, Cis[2])[1]
+        j_1_pos = findindex(dis,Cis[2])
         j_1_pos < 1 && throw(ArgumentError("no contracted indices in combiner-tensor product"))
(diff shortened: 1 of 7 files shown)


test/runtests.jl
```diff
@@ -8,11 +8,11 @@ using ITensors, Test
         "test_indexset.jl",
         "test_itensor.jl",
         "test_contract.jl",
+        "test_combiner.jl",
         "test_trg.jl",
         "test_ctmrg.jl",
         "test_dmrg.jl",
         "test_siteset.jl",
-        "test_phys_sitesets.jl",
```
emstoudenmire: Did you intend to remove this line?

kshyatt: what the

src/storage/combiner.jl
```diff
@@ -0,0 +1,89 @@
+export Combiner
+
+struct CombinerStorage <: TensorStorage
+    ci::Index
+    inds::IndexSet
```
emstoudenmire: Did the code end up using `ci` and `inds` stored inside CombinerStorage? If not, perhaps CombinerStorage can just hold no data.

test/runtests.jl
kshyatt: no I don't know why it vanished :(. I'll put it back

src/storage/combiner.jl
kshyatt: It doesn't yet but it might be nice to have the Combiner carry around which index is the combined one

emstoudenmire: True, but let’s make sure we have a concrete application of this choice. I think a good one would be a function `combinedIndex` which acts on an ITensor with combiner storage, and returns `ci`. We have that function in C++ but the way it works relies on the first index being the combined one, which is a brittle way to do it. 

So how about we keep `ci` but remove `inds`. We can always add `inds` back in later when it is needed without breaking any existing code.

kshyatt: Sure!

kshyatt:
Better now?

emstoudenmire:
I just pushed some more edits, trying to compress some of the loops to make the new IndexSet. Sometimes I'm on the fence about whether explicit for loops are better (you can see exactly what they are doing!) or things like `vcat` are better (shorter code, expressing intent).

kshyatt:
Compact, readable code is the best!

emstoudenmire:
Not sure how readable my `vcat` lines are... maybe if we had slicing for IndexSets it would be a lot more readable.

I just pushed another change to make them more readable I think.

Also found some unused variables! I guess Julia doesn't warn about these things? Would be good to have a code analyzer for Julia.

PR #70, opened 2019-08-05
Create Diag storage

mtfishman:
Diag storage with ITensor constructors, getting and setting elements, and contraction with Dense storage.


src/itensor.jl
```diff
@@ -409,11 +485,13 @@ function summary(io::IO,
   print(io," \n",typeof(store(T)))
 end
 
+# TODO: make a specialized printing from Diag
+# that emphasizes the missing elements
```
emstoudenmire: I like this idea 👍

src/itensor.jl
```diff
@@ -52,8 +58,78 @@ function ITensor(A::Array{S},inds::IndexSet) where {S<:Number}
...
-# Convert to complex
-complex(T::ITensor) = ITensor(inds(T),storage_complex(store(T)))
+#
+# Diag ITensor constructors
+#
+
+"""
+diagITensor(::Type{T}, is::IndexSet)
+
+Make an ITensor with Diag storage with element type T.
```
emstoudenmire: What do you think of saying:
"Make a sparse ITensor with non-zero elements only along the diagonal. The storage will have Diag type."
...or similar. Just to make it friendlier to newcomers who may not appreciate what Diag storage does.

src/itensor.jl
mtfishman: Julia has nice printing for their Diagonal matrix type:

```julia
julia> Diagonal([1,2,3])
3×3 Diagonal{Int64,Array{Int64,1}}:
 1  ⋅  ⋅
 ⋅  2  ⋅
 ⋅  ⋅  3
```

It could be nice to generalize that.

src/itensor.jl
mtfishman: Yeah that sounds good, I can change it.

mtfishman:
I'll wait for Travis to finish and then merge.

There are still other things we need with Diag (making a version where all elements are the same, making the delta constructor use Diag, making SVD output a diag, etc.) but I will do that in another pull request/open an issue to track things we still need to do.

emstoudenmire:
Great. And regarding issues, I made that milestone called "Dense ITensor features" but I really meant like "non-QN ITensor features" so I'd include things like Diag storage in that.

PR #71, opened 2019-08-05
More Diag functionality

mtfishman:
This adds: Diag+Diag, Diag+Dense, Diag*Diag. Delta now uses Diag (but with vector storage, we still need to make a version of Diag where all of the elements are the same).

### DIFF
--- a/src/decomp.jl
+++ b/src/decomp.jl
@@ -79,6 +79,9 @@ function truncate!(P::Vector{Float64};
   return truncerr,docut
 end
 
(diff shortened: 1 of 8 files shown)


mtfishman:
It looks like maybe something is messed up with the set up of Travis on Windows? Anyway, I will merge this.

emstoudenmire:
We can take the windows build out of the .travis.yml file. It’s a new build type so maybe it’s buggy?

PR #72, opened 2019-08-06
Modify position to use SVD sometimes

kshyatt:
The way we have this now, you can't truncate when you call position and so the bond dimension can grow pretty badly in `sum`, `applyMPO`, and friends.

### DIFF
--- a/src/mps/mpo.jl
+++ b/src/mps/mpo.jl
@@ -442,3 +442,49 @@ function nmultMPO(A::MPO, B::MPO; kwargs...)::MPO
     end
     return res
 end
+
+function position!(M::Union{MPS,MPO}, 
+                   j::Int)
+  while leftLim(M) < (j-1)
+    (leftLim(M) < 0) && setLeftLim!(M,0)
+    b = leftLim(M)+1
+    linds = uniqueinds(M[b],M[b+1])
+    Q,R = qr(M[b], linds)
+    M[b] = Q
+    M[b+1] *= R
+    setLeftLim!(M,b)
+    if rightLim(M) < leftLim(M)+2
+      setRightLim!(M,leftLim(M)+2)
+    end
+  end
+
+  N = length(M)
+
+  while rightLim(M) > (j+1)
+    (rightLim(M) > (N+1)) && setRightLim!(M,N+1)
+    b = rightLim(M)-2
+    rinds = uniqueinds(M[b+1],M[b])
+    Q,R = qr(M[b+1], rinds)
+    M[b+1] = Q
+    M[b] *= R
+    setRightLim!(M,b+1)
+    if leftLim(M) > rightLim(M)-2
+      setLeftLim!(M,rightLim(M)-2)
+    end
+  end
+end
+
+@doc """
+position!(M::MPS, j::Int)
+
+Move the gauge position, or orthogonality center, 
+to site j of an MPS. No observable property of 
+the MPS will be changed, and no truncation of the 
+bond indices is performed. Afterward, tensors 
+1,2,...,j-1 will be left-orthogonal and tensors 
+j+1,j+2,...,N will be right-orthogonal.
+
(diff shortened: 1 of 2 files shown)


src/mps/mpo.jl
```diff
@@ -189,17 +189,26 @@ function linkindex(M::MPO,j::Integer)
   return li
 end
 
-function position!(M::MPO,
-                   j::Integer)
+function position!(M::MPO, j::Integer; kwargs...)
+  default_method = (rightLim(M) - leftLim(M) > 2) ? "qr" : "svd"
```
mtfishman: Hi Katie,

Thanks for looking at this.

This logic seems a bit subtle here. Say we have the following situation:

-A-A-X-X-X-X-B-B-B-B-

(where A denotes a left orthogonal MPS tensor, B denotes a right orthogonal tensor, and X denotes a general tensor). Then we have that `leftLim(M) == 2`, `rightLim(M) == 7`. Say that someone calls `position!(M,10)`. Should this do a "qr" or an "svd"? While the gauge center is moved from sites 3 to 6, no truncation can occur, but after that it can. I am actually not really sure what is supposed to happen here (if "svd" is chosen, then between sites 3 to 6 we might be doing some truncation, but it is not right in general to truncate in that region).

Maybe Miles can chime in on how this situation is meant to be handled? It seems like the "proper" way to do it is that if `(rightLim(M) - leftLim(M) > 2)` then a separate `orthogonalize!` function should be called that orthogonalizes the region between `leftLim(M)` and `rightLim(M)`. Then after that, the gauge center can be moved with or without truncation (and "svd" or "qr" can be chosen freely by the user, with "svd" as the default since I believe it is more reliable).

Cheers,
Matt

src/mps/mpo.jl
```diff
@@ -310,15 +327,14 @@ function densityMatrixApplyMPO(A::MPO, psi::MPS; kwargs...)::MPS
...
-
     psi_c = dag(copy(psi))
     A_c   = dag(copy(A))
     prime!(psi_c, rand_plev)
     prime!(A_c, rand_plev)
-
     for j in 1:n-1
         unique_site_ind = setdiff(findinds(A_c[j], "Site"), findindex(psi_c[j], "Site"))[1]
-        A_c[j] = setprime(A_c[j], 1, unique_site_ind)
+        pl = id(unique_site_ind) == id(commonindex(A_c[j], psi_c[j])) ? 1 : 0
```
mtfishman: I don't quite get the logic here. What was wrong with the previous version? Can this be more like the C++ version (not in terms of "Site" tags, but instead using `siteIndex` and `uniqueSiteIndex` functions)?

I had been thinking that all of this logic could be replaced with a single function, maybe called `primeLinkInds!(A, psi, plev)` or `primeInternalInds!(A, psi, plev)`, that primes all "internal" indices of the network made by MPO `A` and MPS `psi`, i.e. all of the link indices of `A` and `psi` as well as the common site indices between the two. I think that is a pretty general code pattern that happens a lot using ITensor: you want to do something to all internal indices or external indices of a network made up of a set of ITensors.

I think in the end it is better to do this in terms of "similar" indices instead of a random prime level, since randomizing the id's seems more fitting than randomizing the prime level. My plan for the C++ version had been to write this function in terms of a `simLinkInds!(A_c, psi_c)` function, but the problem was that required a more general version of eigen which wasn't available in C++. 

Anyway, I guess if it works for your use case for now than it is fine, and we can look into other code logic here later.

src/mps/mpo.jl
emstoudenmire: I was thinking something like this: we check to see whether any truncation parameters have been passed (maxdim, cutoff, ...) and if so, we use SVD. Otherwise we use QR since it is faster. I don't believe QR should blow up the bond dimension if done properly and on one site tensor at a time, correct? (I could be mistaken about this.)

PR #73, opened 2019-08-07
add rng keyword argument to recursive svd with tests

YiqingZhouKelly:
Hi,

I added a rng keyword argument to the recursive SVD just to make it easier to reproduce results. This doesn't change the accuracy of the algorithm; I just find it helpful when debugging my own code. If this seems unnecessary, feel free to just close this PR. Thanks!

Best,
Yiqing

### DIFF
--- a/src/storage/dense.jl
+++ b/src/storage/dense.jl
@@ -254,11 +254,11 @@ function storage_svd(Astore::Dense{T},
   utags::String = get(kwargs,:utags,"Link,u")
   vtags::String = get(kwargs,:vtags,"Link,v")
   fastSVD::Bool = get(kwargs,:fastSVD,false)
-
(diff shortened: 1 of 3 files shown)


emstoudenmire:
Hi Yiqing, I think the better approach here might be to just pass a random number seed (an integer) instead of the whole random generator type. But mainly I think this is a pretty rare case to need to set the seed since the randomization here is just a way to deal with lack of orthogonality in a robust way, and need not be reproducible down to every number. Rather all that's required to be reproducible is the typical accuracy of the resulting SVD approximation.

YiqingZhouKelly:
Hi Miles,
Yes, I agree. And I'm also concerned that adding an extra parameter will make other parts of the code (e.g. functions that calls SVD) more complicated. It might be better to just leave it as it is now. So I've closed this PR. 
Thanks,
Yiqing

PR #74, opened 2019-08-09
WIP: Start making Diag with uniform diagonal value (for delta tensor)

mtfishman:
This is a work in progress for making a version of Diag with one value along the diagonal (the case in C++ with `d.allsame() = true`).

The strategy here is to allow Diag to have different storage types. The normal version is Diag having `Vector` storage. When Diag has scalar storage (i.e. `<: Number`) it is treated as having a uniform value on the diagonal. This allows us to use multiple dispatch when necessary, instead of having runtime checks like in the C++ version. It also lets Diag storage only store what it needs (and store other types, such as `CuVector`).

Work that has to be done is making a shortcut for index replacement (i.e. `A*\delta(i,j)`). Also, the storage code is a bit of a mess right now and needs to be cleaned up, I think it could be made a lot simpler (for example, there are many mixtures of cases of adding or contracting various storage types together, and some of the logic could be handled with better generic programming and type promotion).


emstoudenmire:
I agree this is a lot better design than in the C++ version. Some of the code there got sphagetti-like due to too many runtime cases to check.

PR #77, opened 2019-08-12
replace digit parsing with subtraction

JeffreySarnoff:
You can avoid the parse machinery here because the function is examining one decimal digit at a time. Conveniently, subtracting the character code for '0' from the character code for another digit gives the value of that digit as an integer.

### DIFF
--- a/src/smallstring.jl
+++ b/src/smallstring.jl
@@ -71,7 +71,7 @@ function Base.parse(::Type{Int}, s::SmallString)
   n = length(s)
   int = 0
   for j = 1:n
-    int = int*10 + parse(Int,Char(s[j]))
+    int = int*10 + (s[j] - 0x30)  # 0x30 === UInt8('0'), '9'-'0' == 9
   end
   return int
 end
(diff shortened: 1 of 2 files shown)


emstoudenmire:
Hi Jeffrey, thanks for the pull request. However, I will need some help understanding the improvements this brings, and am leaning toward continuing to use `parse(Int,Char)` since it's a base function which should be well-tested and fast (enough).

Is this an improvement for speed reasons? If so, can the speed difference be measured and how many small string objects have to be created for the speed difference to add up significantly? Thanks -

Miles

JeffreySarnoff:
Miles,

The PR is less a request that you pull it than a post-it note about how that specific task has been done whenever the conversion of '0'..'9' to 0..9 happened within a time-critical path or within a short section of code that is executed very heavily.  I have no information about the frequency with which this parsing happens, and even if it occur many times -- it may be of no consequence because it is swamped by other activity.  The relative throughput is much improved, if it matter. So, no worries.
Best,
Jeffrey

Issue #75, opened 2019-08-11
`dmrg` throws error when `quiet=false`


orialb:
The cause seems to be that `maxDim` was renamed to `maxLinkDim`, but not updated in `dmrg`.
I guess all the DMRG tests run with `quiet=true` so this was not seen.

PR #78, opened 2019-08-12
replace call to maxDim with maxLinkDim

orialb:
`maxDim(::MPS)` was renamed to `maxLinkDim`. updated `dmrg` to call the right function.
this fixes https://github.com/ITensor/ITensors.jl/issues/75 .


mtfishman:
Thanks for catching that!

PR #81, opened 2019-08-14
preserve bond index tags in replaceBond!

orialb:
this PR fixes a bug where the tags of the bond index of an MPS were not preserved by `replaceBond!`.

### DIFF
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -203,10 +203,11 @@ function replaceBond!(M::MPS,
                       b::Int,
                       phi::ITensor;
                       kwargs...)
-  FU,FV,u = factorize(phi,inds(M[b]); which_factorization="automatic", kwargs...)
-  replacetags!(FU,tags(u),tags(linkindex(M,b)), u)
-  replacetags!(FV,tags(u),tags(linkindex(M,b)), u)
+  FU,FV = factorize(phi,inds(M[b]); which_factorization="automatic",
+                        tags=tags(linkindex(M,b)), kwargs...)
   M[b]   = FU
   M[b+1] = FV
 end
+
+
 
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -203,7 +203,9 @@ function replaceBond!(M::MPS,
                       b::Int,
                       phi::ITensor;
                       kwargs...)
-  FU,FV = factorize(phi,inds(M[b]); which_factorization="automatic", kwargs...)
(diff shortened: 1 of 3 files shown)


src/mps/mps.jl
```diff
@@ -203,7 +203,9 @@ function replaceBond!(M::MPS,
                       b::Int,
                       phi::ITensor;
                       kwargs...)
-  FU,FV = factorize(phi,inds(M[b]); which_factorization="automatic", kwargs...)
+  FU,FV,u = factorize(phi,inds(M[b]); which_factorization="automatic", kwargs...)
```
mtfishman: Hi,

Thanks for pointing out this issue. In the C++ version, the way we handled this (which comes up all over in MPS functions) was to pass `tags = tags(linkindex(M,b))` into the `factorize` function. I think that is a slightly better way to go (and a code pattern we used a lot in the C++ version). Would you mind doing it that way?

Cheers,
Matt

orialb: Ok, I could do that.
So you would like it to be handled in each of the internal factorization functions that factorize can steer to (`_factorize...`) ? Or should `factorize` first get the results from those functions and then replace the tags ?

mtfishman: If you pass the keyword arg `tags = ...` to `factorize` it should properly propagate to all of the internal factorizations that are used. If it doesn't, that would be a bug in the internal factorization functions.

mtfishman: For example, this works right now:

```julia
i = Index(2,"i")
j = Index(2,"j")
A = randomITensor(i,j)

F,G,l = factorize(A,i;tags="x")

@assert hastags(l,"x")
```

So if you make the call to factorize:

```julia
FU,FV,u = factorize(phi,inds(M[b]); which_factorization="automatic", tags = tags(linkindex(M,b)), kwargs...)
```

that should be sufficient. That is also better because if a user defines `tags = ...` explicitly in their call to `replaceBond!` then it will overwrite the use of `tags = tags(linkindex(M,b))`, so people can set custom tags if they want.

orialb: > If you pass the keyword arg tags = ... to factorize it should properly propagate to all of the internal factorizations that are used. If it doesn't, that would be a bug in the internal factorization functions.

oh, I didn't realize from your first comment that it was already implemented. ok, then I'll update the PR.

mtfishman:
Looks good, thanks!

PR #88, opened 2019-08-20
Created truncate! function for MPS and MPO

emstoudenmire:
`truncate!` performs a global truncation of every bond of an MPS or MPO, using the `factorize` function internally on the backward, truncating pass.


mtfishman:
In the `orthogonalize` function in C++, it creates the density matrix and then diagonalizes the density matrix, right?

I like this version since it is very simple. Why didn't the C++ version just do it this way? Also, should we allow `position!` to accept keyword arguments to choose what factorization to use? I thought there were sometimes numerical stability issues when using QR decompositions.

emstoudenmire:
Hm, good question about the C++ version. I don't have such a good answer for that! 

I did put keyword arguments back into position! (now renamed orthogonalize!), but right now they don't do anything. Yes, we should allow switching methods if QR causes problems; I wasn't aware of that. Is it a fundamental thing about QR, you think, or more of just a quality of implementation issue with LAPACK / Julia QR?

mtfishman:
Perhaps I am mistaken, I thought @YiqingZhouKelly came across some problem using QR in her simulation. Julia just uses LAPACK, unless exotic data types like BigInt are used. We can just see how this version works out, and give different factorization options like `factorize` does.

emstoudenmire:
Yes,  I can definitely believe there are some instabilities in LAPACK QR. (We've seen them in LAPACK SVD too, and I know Yiqing did recently.)

Maybe we should make a fallback QR, which does a Gram-Schmidt with randomization in case of degenerate spaces. This would be a lot like what the code in storage/svd/svd.jl `orthog!` does and it would just compute R also (it sort of is implicitly now anyway).

YiqingZhouKelly:
Hi Miles and Matt,
I don't remember I got an issue with QR but I did encounter once when bulit-in SVD crashed. I might have asked you about the numerical stability of both QR and SVD, which might leave an impression that QR has some numerical stability issue.

emstoudenmire:
Thanks Yiqing, that's useful to know -

mtfishman:
Thanks for chiming in, Yiqing.

The version in this PR seems good enough for now, let's see how it goes.

emstoudenmire:
The one thing I wish was better about it at the moment though, Matt, is how it handles tags. I would need to study the approach you used in the C++ version to do it well. But I figured it's an enhancement we can put in later.

test/test_mps.jl
```diff
@@ -117,3 +117,74 @@ using ITensors,
   replaceBond!(psi,1,phi)
   @test tags(linkindex(psi,1)) == bondindtags
 end
+
+# Helper function for making MPS
+function basicRandomMPS(N::Int;dim=4)
+  sites = SiteSet(N,2)
+  M = MPS(sites)
+  links = [Index(dim,"n=$(n-1),Link") for n=1:N+1]
```
mtfishman: I know this is just a test function, but I think we should try to be consistent with link indices having the tags "Link, l=$n". It is useful for site indices to have "n=$n" tags and link indices to have "l=$n" tags so that to grab the right link of site `n`, you can just do `findindex(M[n],"l=$n")` and not have to also list the "Link" tag. (I think you know this, just writing it out since this was something that was inconsistent at some point in the C++ version and I kept seeing it pop up).

mtfishman:
I just always defaulted to making sure that the new tags on the links were the same as the old tags on the links (I did this throughout the MPS functions). At first there were quite a few places to track down to make sure it worked correctly, but in the end it was just a matter of finding where the factorizations were happening, grabbing the link tags before the factorization, and then passing the tags to the factorization function.

Issue #86, opened 2019-08-18
possible bug in `position!`


orialb:
Hi,

It seems to me that `position!` is not shifting the orthogonality center properly, in the code below the last test fails. Am I misunderstanding the functionality of `position!` or missing something else here? 
I can try to find out the issue if it is indeed a bug. 
```julia
using ITensors,LinearAlgebra, Test
N = 20
sites = spinOneSites(N)
ampo = AutoMPO(sites)
for j=1:N-1
    add!(ampo,"Sz",j,"Sz",j+1)
    add!(ampo,0.5,"S+",j,"S-",j+1)
    add!(ampo,0.5,"S-",j,"S+",j+1)
end
H = toMPO(ampo)
psi0 = randomMPS(sites)
sweeps = Sweeps(1)
maxdim!(sweeps, 10)
cutoff!(sweeps, 1E-10)
energy, psi = dmrg(H,psi0, sweeps)

position!(psi,5)
@test inner(psi,psi) ≈ 1.
@test dot(psi[5],psi[5]) ≈ 1.
```

orialb:
I looked at it a bit more, and it seems that the problem is actually in `replaceBond!` which does not update the MPS left/right orthogonality limit. 
With the definition of `replaceBond!` below, everything seems to be fine.  Does it make sense to handle this in that way? 
```julia
function replaceBond!(M::MPS,
                      b::Int,
                      phi::ITensor;
                      kwargs...)

  FU,FV = factorize(phi,inds(M[b]); which_factorization="automatic",
                        tags=tags(linkindex(M,b)), kwargs...)
  M[b]   = FU
  M[b+1] = FV

  dir::String = get(kwargs,:dir,"center")
  if dir=="fromright"
    M.llim_ = b-1
    M.rlim_ = b+1
  elseif dir=="fromleft"
    M.llim_ = b
    M.rlim_ = b+2
  end
end
```

emstoudenmire:
Hi Ori, thanks for taking a look at this. I think there were numerous issues with some interplay between them. One that Matt identified today when I was talking with him was that setindex!(M::MPS,...) was not updated the .llim_ and .rlim_ internal variables of the MPS as it should be. I just fixed that on Katie's ksh/morempo pull request branch.

[...]

Issue #32, opened 2019-07-25
Operator tracker for DMRG


kshyatt:
In the C++ code, you can include operators you want DMRG to measure as it sweeps back and forth, to see how the entropy/magnetization/whatever are doing. It would be nice to have this feature in the Julia code too.

emstoudenmire:
Yes, we do need to add this back

Here’s the design we currently have in C++ and how it would work in Julia approximately:
* you can optionally pass some object to DMRG that is the “observer” or “tracker” object (the name is not so important!)
* this object needs to support two methods: `measure` and `checkDone`. We can provide a default implementation of each of these in some way (is there a way in Julia to see if a method is defined, and call it, otherwise call a fallback method? Inheriting from some abstract type can let you do this by defining the method for the abstract type but then it means you have to inherit - not the worst thing in this use case)

As an alternative it might possible to just pass the measure and checkDone *functions* themselves in Julia, and if you want them to have access to some data use functional programming tricks like currying with a lambda or similar to wrap the data inside them (we could provide simple examples).

kshyatt:
This seems like it might be a good project for the Hackathon tomorrow...

emstoudenmire:
Yes! Great idea

orialb:
If no one is working on this I can have a go at this. In my own MPS code I pass an observer object and I have a tendency to prefer this way.  
I didn't fully understand from the design suggestion whether you want to support the two alternatives (object vs. passing functions directly),  or prefer choosing one or the other.

emstoudenmire:
Thanks Ori. I think that would be great. Let's prefer choosing just one design. My comment about passing functions was too vague, and more to say that maybe Julia gives us a chance to revisit the design we have in C++.

But upon further thought, the design we have in C++ seems good to me. It's just that subtyping classes in C++ is a pain and very verbose, whereas in Julia it's quite straightforward. So the design gets that much better as we port it to Julia.

Looking forward to seeing what you develop -

emstoudenmire:
Addressed in pull requeest #89 - thanks to Ori Alberton @orialb for contributing this.

PR #89, opened 2019-08-20
dmrg observer & fix for replaceBond!

orialb:
Hi,

This PR includes two things (I could split to 2 different PRs if you prefer):
1. perform update of `llim_, rlim_`  in `replaceBond!` (as far as I understood from discussion in https://github.com/ITensor/ITensors.jl/issues/86 , `replaceBond!` should do that). 
2. my suggestion for a dmrg observer implementation (https://github.com/ITensor/ITensors.jl/issues/32) .

I implemented a basic `DMRGObserver` object which allows user to specify local operators to measure during the sweep, tracks energy and can stop it if some convergence criterion is reached. I also added an example showing the observer usage. 

Currently the option for tracking entropy and truncation error is missing because the factorization functions do not return the singular values and the error.

Note that (1) is necessary for (2) to work, as I'm calling `position!` inside the observer `measure!` function and it is causing an error in `dmrg` without the fix.

let me know what you think.

### DIFF
--- a/examples/dmrg/dmrg_with_observer.jl
+++ b/examples/dmrg/dmrg_with_observer.jl
@@ -1,0 +1,43 @@
+# In this example we show how to pass a DMRGObserver to
+# the dmrg function which allows tracking energy convergence and
+# convergence of local operators.
+using ITensors
+
+
+"""
+get MPO of transverse field Ising model Hamiltonian with field strength h
+"""
+function tfimMPO(h,sites)
+    # Input operator terms which define a Hamiltonian
+    ampo = AutoMPO(sites)
+    for j=1:length(sites)-1
+        add!(ampo,"Sz",j,"Sz",j+1)
+        add!(ampo,h,"Sx",j)
+    end
+    add!(ampo,h,"Sx",length(sites))
+    # Convert these terms to an MPO tensor network
+    return toMPO(ampo)
+end
+
+N = 20
+sites = spinHalfSites(N)
+psi0 = randomMPS(sites)
+sweeps = Sweeps(5)
+maxdim!(sweeps, 10,20,100,100,200)
+cutoff!(sweeps, 1E-10)
+
+#=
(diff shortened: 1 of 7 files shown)


emstoudenmire:
Thanks for the high-quality PR! We will review this soon; hopefully by end of next week.

emstoudenmire:
Ok pulled! The biggest change I made was to remove the DMRGStepInfo struct in favor of passing keyword args to measure!. One reason is that it will make the design more easily extensible, where we can pass more things from the dmrg code without having to maintain the DMRGStepInfo struct. Another reason is that it's just less code and works equally well.

The measurement system you came up with is great by the way. I like the interface and the customizability of that. It's something we had wanted for a while in the C++ version but never quite got the design right (partly because basic data structures in the C++ std lib such as hash tables are a real mess and/or just missing).

I'm not sure if there's some clean way to put in support for two-site measurements too. Something to think more about. Ultimately though, users can just design custom observers for these sorts of things with minimal effort.

orialb:
> The biggest change I made was to remove the DMRGStepInfo struct in favor of passing keyword args to measure!

I also considered that (and actually thought you might prefer that looking at the code base). The main reason I thought having a type for this is a good idea is because it makes it clear to the user exactly what data `measure!` receives, without having to look at the DMRG code. But I guess this can also just be specified in the documentation of the abstract observer type as a guide to users who want to implement their own observer.

emstoudenmire:
That's good you considered that. Thanks for bringing it up because we want to make sure this new design can offer the same thing. I think it can though, because a user could make a new observer type and then when they first define `measure!` for their observer, they can just print out `kwargs` to see what information it holds. So I think it will be discoverable that way. Also, yes, through documentation also!

orialb:
I forgot to write before: 
thanks for reviewing and giving feedback!

emstoudenmire:
No problem! Thanks for the contribution

PR #91, opened 2019-08-23
WIP: New syntax for specifying tags and a prime level

mtfishman:
This implements the idea of using the syntax ("x",2) instead of "x,2" to specify tags along with a prime level. This can be used throughout the library, when tags are being specified to create a new index or for pattern matching.

### DIFF
--- a/src/index.jl
+++ b/src/index.jl
@@ -57,21 +57,21 @@ struct Index
   tags::TagSet
 end
 
-Index() = Index(IDType(0),1,Neither,TagSet("0"))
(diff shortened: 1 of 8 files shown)


mtfishman:
I am good with how this is right now, @emstoudenmire let me know if we should pull the trigger on this.

emstoudenmire:
I think it's a good improvement. I especially like how the prime stays as a true integer and we can explain more simply how TagSets work and which tags are allowed.

Issue #60, opened 2019-08-01
errMPOProd implementation needed


emstoudenmire:

emstoudenmire:
See: http://itensor.org/docs.cgi?vers=cppv3&page=classes/mps_mpo_algs

emstoudenmire:
Added in pull #92, thanks @alvarorga

Issue #45, opened 2019-07-26
MPO-MPS and MPO-MPO methods


kshyatt:
Would be nice to have `applyMPO`, `sum(::MPO, ::MPO0`, `nmultMPO`, and friends!

emstoudenmire:
Yes! And they are all 1-indexed even in the C++ code. It’s almost as if we knew Julia was coming.

mtfishman:
Yes, that was on my short to-do list. I just rewrote those fairly recently in C++ so it shouldn't be too hard.

kshyatt:
I'm trying applyMPO now

kshyatt:
We still need `errorMPOProd` if someone wants to claim it before I do

emstoudenmire:
Looks like we have most of these now or are very close! Thanks Katie & others for working hard on this.

PR #92, opened 2019-08-25
Added errMPOProd

alvarorga:
Hi,
This addresses issues #60 and #45.

I first added a new inner method `inner(B::MPO, y::MPS, A::MPO, x::MPS)` to compute `<By|A|x>`. Then I used that method to define the distance in `errMPOProd(y::MPS, A::MPO, x::MPS)` as `real(inner(y, y) - 2*inner(y, A, x) + inner(A, x, A, x))`.

You may find in `inner` that my way of swapping the prime levels is not very elegant (I use a temporal 3 level prime). If you find a better approach I'll be happy to change it.

### DIFF
--- a/src/mps/mpo.jl
+++ b/src/mps/mpo.jl
@@ -2,7 +2,7 @@ export MPO,
        randomMPO,
        applyMPO,
        nmultMPO,
-       errMPOProd,
+       errorMPOProd,
        maxLinkDim,
        orthogonalize!,
        truncate!
@@ -36,8 +36,8 @@ mutable struct MPO
     end
     new(N,v,0,N+1)
   end
- 
-  function MPO(sites::SiteSet, 
+
+  function MPO(sites::SiteSet,
                ops::Vector{String})
     N = length(sites)
     its = Vector{ITensor}(undef, N)
@@ -81,17 +81,17 @@ tensors(m::MPO) = m.A_
 leftLim(m::MPO) = m.llim_
 rightLim(m::MPO) = m.rlim_
 
-function setLeftLim!(m::MPO,new_ll::Int) 
+function setLeftLim!(m::MPO,new_ll::Int)
   m.llim_ = new_ll
 end
 
-function setRightLim!(m::MPO,new_rl::Int) 
+function setRightLim!(m::MPO,new_rl::Int)
   m.rlim_ = new_rl
 end
 
 getindex(m::MPO, n::Integer) = getindex(tensors(m), n)
 
-function setindex!(M::MPO,T::ITensor,n::Integer) 
+function setindex!(M::MPO,T::ITensor,n::Integer)
   (n <= leftLim(M)) && setLeftLim!(M,n-1)
   (n >= rightLim(M)) && setRightLim!(M,n+1)
   setindex!(tensors(M),T,n)
@@ -181,7 +181,7 @@ Get the maximum link dimension of the MPS or MPO.
 """
 function maxLinkDim(M::T) where {T <: Union{MPS,MPO}}
   md = 0
-  for b ∈ eachindex(M)[1:end-1] 
+  for b ∈ eachindex(M)[1:end-1]
     md = max(md,dim(linkindex(M,b)))
   end
   md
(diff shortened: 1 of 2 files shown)


alvarorga:
Hi again, two things:

1. When I made the tests for `inner` I reproduced the same kind of test as was done for `inner(y::MPS, A::MPO, x::MPS)` ([here](https://github.com/ITensor/ITensors.jl/blob/master/test/test_mpo.jl#L34)), which is very similar to the [code in the method](https://github.com/ITensor/ITensors.jl/blob/master/src/mps/mpo.jl#L213). Maybe the tests for `inner` would benefit if the actual computation was manually done, like
```julia
@testset "inner <y|A|x>" begin
    L = 5
    sites = spinHalfSites(L)
    
    x = randomMPS(sites)
    y = randomMPS(sites)
    A = randomMPO(sites)
    
    O = 1.
    for j ∈ eachindex(x)
        xj = reshape(Array(x[j]), 2)
        yj = reshape(Array(y[j]), 2)
        Aj = reshape(Array(A[j]), 2, 2)
        O *= (yj'*transpose(Aj)*xj)[]
    end

    @test O ≈ inner(y, A, x)

    badSites = spinHalfSites(L+2)
    badx = randomMPS(badSites)
    @test_throws DimensionMismatch inner(y, A, badx)
end

@testset "inner <By|A|x>" begin
    L = 5
    sites = spinHalfSites(L)
    
    x = randomMPS(sites)
    y = randomMPS(sites)
    A = randomMPO(sites)
    B = randomMPO(sites)
    
    O = 1.
    for j ∈ eachindex(x)
        xj = reshape(Array(x[j]), 2)
        yj = reshape(Array(y[j]), 2)
        Aj = reshape(Array(A[j]), 2, 2)
        Bj = reshape(Array(B[j]), 2, 2)
        O *= ((transpose(Bj)*yj)'*transpose(Aj)*xj)[]
    end

    @test O ≈ inner(B, y, A, x)

    badSites = spinHalfSites(L+2)
    badx = randomMPS(badSites)
    @test_throws DimensionMismatch inner(B, y, A, badx)
end
```

2. Now that the `inner` function is available, maybe it is a good idea to write functions for the variance / standard deviation of a measurement.

emstoudenmire:
Thanks for the high quality PR! 

This is already good enough to pull, but if you don't mind here are a couple changes right off the bat that would make this better:
1. could you please rename the function to `errorMPOProd` ? The reason is just that it is the name we have used in the C++ version (the bug / feature request may have misleadingly said otherwise)
2. could you also refactor the algorithm in `errorMPOProd` to reproduce the following code from our C++ version (also visible here: https://github.com/ITensor/ITensor/blob/v3/itensor/mps/mpo.cc). The main change is that the error gets normalized by <x|A^dag A|x> so that it becomes independent of the norm of A, I believe. (Someone earlier sent us a pull request in our C++ version making this change because they had found cases where a simpler definition of the error could be misleadingly large or small.) Here's the code --- note that innerC in C++ can just be replaced with inner in Julia since in Julia inner can return a real or complex number from the same function:

```
Real
errorMPOProd(MPS const& y,
             MPO const& A, 
             MPS const& x)
    {
    auto err = real(innerC(y,y));
    err += -2.*real(innerC(y,A,x));
    err /= real(innerC(A,x,A,x));
    err = std::sqrt(std::abs(1.0+err));
    return err;
    }
```

alvarorga:
Great! I renamed the function to `errorMPOProd` and divided by the norm of `A|x>`. I also added to the `inner` methods test the manual contractions I described earlier.

emstoudenmire:
Great - one last thing is, in the test for errorMPOProd, could you add a test that does applyMPO on some MPS and MPO and uses errorMPOProd to make sure the error is close to zero? Thanks!

alvarorga:
Sure, I just added it to the test of `errorMPOProd`. One thing that confused me a little is that I have to use `applyMPO` with `maxdim=1`, otherwise I don't always get a 0 distance. I don't know whether that behavior is expected. See this little code:
```julia
using Test, ITensors

L = 7
sites = spinHalfSites(L)
    
x = randomMPS(sites)
A = randomMPO(sites)
Ax = applyMPO(A, x)
B = applyMPO(A, x, maxdim=1)
    
@test inner(x, A, x) ≈ inner(x, B) # Always passes.
@test inner(x, A, x) ≈ inner(x, Ax) # Mostly fails.
```
Sometimes it passes and sometimes it doesn't. The relative distance between `inner(x, A, x)` and `inner(x, Ax)` is not small when the test fails.

emstoudenmire:
Looks good thanks! I just reviewed it in detail; before I was going to suggest exactly that you please implement `inner(MPO,MPS,MPO,MPS)` and rewrite `errorMPOProd` that way, but I see you've now done exactly that.

PR #95, opened 2019-09-03
Redesign of internals of SiteSet

emstoudenmire:
This redesign changes the Site types (which label the "physics" or semantic meaning of a site) to no longer store an Index inside. Instead the SiteSet just stores an array of (Index,Site) tuples.

The goals of this redesign are to simplify the Site types and to allow generic code to update the indices in a site set. A notable future use case is removing the QNs from indices in a site set.

I was going to use Pair instead of Tuple for the SiteSet storage but Pair is also immutable, like Tuple, so I didn't see the benefit.


emstoudenmire:
Instead of the method `siteType` for getting the site object stored with the Index, would a better name be `siteObject`? The reason is that technically `siteType` does not return a type, but an instance of a type.

mtfishman:
Looks like a pretty minimal change, it seems pretty natural to design it this way.

Is there a reason why the function that gets overloaded for the SiteTypes is called `operator` but the one for SiteSets is `op`?

Also, it is kind of small, but I prefer the signature `operator(::SiteType,::Index,::Int)` instead of `operator(::Index,::SiteType,::Int)`. It feels more natural to have the type that is used for dispatch first, I think that is fairly common in Julia.

Regarding `siteType` vs. `siteObject`, can we just use the types instead of the instances, or are we anticipating types that might have other data in them?

mtfishman:
Another question about interface. Say that I have my own SiteSet that has all SpinOnes, and then later on I want to add a SpinHalf at the edge. Is this the interface?
```julia
sites = SpinOneSites(N)
sites[1] = Index(2,"Site,SpinHalf,n=1")
setSite!(sites,1,SpinHalfSite())
```

If so, I am wondering if we can do:
```julia
sites = SpinOneSites(N)
sites[1] = (SpinHalfSite(),Index(2,"Site,SpinHalf,n=1"))
```
(I think we can do this with multiple dispatch, not sure if it is a huge improvement but I am just thinking out loud about what we can do with this design).

mtfishman:
Another question, how do we deal with default operators like we do in C++ (i.e. the "Id", "F" and projector operators)?

emstoudenmire:
Closing this for now, as we're probably going to pursue a different design of this system.

PR #96, opened 2019-09-05
Add fitApplyMPO

kshyatt:
Made some slight modifications to `Davidson` and `ProjMPO` for this. I think our equivalent of `denmatDecomp` doesn't have noise support yet, so `localH` isn't doing much ... for now.

### DIFF
--- a/src/iterativesolvers.jl
+++ b/src/iterativesolvers.jl
@@ -65,16 +65,14 @@ function davidson(A,
   maxsize = size(A)[1]
   actual_maxiter = min(maxiter,maxsize-1)
 
-  if dim(inds(phi)) != maxsize
-    error("linear size of A and dimension of phi should match in davidson")
-  end
+  dim(inds(phi)) != maxsize && throw(DimensionMismatch("linear size of A ($maxsize) and dimension of phi $(dim(inds(phi))) should match in davidson"))
 
   V = ITensor[copy(phi)]
-  AV = ITensor[A(phi)]
+  AV = ITensor[noprime(A*phi)]
 
   last_lambda = NaN
   lambda = dot(V[1],AV[1])
-  q = AV[1] - lambda*V[1];
+  q = AV[1] - lambda*V[1]
 
   M = fill(lambda,(1,1))
 
@@ -98,7 +96,7 @@ function davidson(A,
     end
 
     push!(V,copy(q))
-    push!(AV,A(q))
+    push!(AV,noprime(A*q))
 
     newM = fill(0.0,(ni+1,ni+1))
     newM[1:ni,1:ni] = M
--- a/src/mps/mpo.jl
+++ b/src/mps/mpo.jl
@@ -279,10 +279,53 @@ end
 
 function applyMPO(A::MPO, psi::MPS; kwargs...)::MPS
     method = get(kwargs, :method, "DensityMatrix")
-    if method == "DensityMatrix"
-        return densityMatrixApplyMPO(A, psi; kwargs...)
-    end
+    method == "DensityMatrix" && return densityMatrixApplyMPO(A, psi; kwargs...)
+    method == "Fit"           && return fitApplyMPO(A, psi; kwargs...)
     throw(ArgumentError("Method $method not supported"))
+end
+
+function fitApplyMPO(A::MPO, psi::MPS, guess_psi::MPS; kwargs...)::MPS
+    N  = length(A)
+    N != length(psi) && throw(DimensionMismatch("lengths of MPO ($N) and MPS ($(length(psi))) do not match"))
+    nsweeps::Int    = get(kwargs, :nsweep, 1)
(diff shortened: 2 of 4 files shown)


src/iterativesolvers.jl
```diff
@@ -65,16 +65,14 @@ function davidson(A,
...
   actual_maxiter = min(maxiter,maxsize-1)
 
-  if dim(inds(phi)) != maxsize
-    error("linear size of A and dimension of phi should match in davidson")
-  end
+  dim(inds(phi)) != maxsize && throw(DimensionMismatch("linear size of A ($maxsize) and dimension of phi $(dim(inds(phi))) should match in davidson"))
 
   V = ITensor[copy(phi)]
-  AV = ITensor[A(phi)]
+  AV = ITensor[noprime(A*phi)]
```
mtfishman: Hi Katie,

Thanks for taking a crack at fitApply. I think the previous version of this was a bit more general, since the operator `A` might not map indices `(i,j,k,...) -> (i',j',k',...)`, which this new version seems to implicitly assume. So I think the logic of mapping the indices back to the proper ones should be handled in a custom function (in the previous version it was the `(pm::ProjMPO)(v::ITensor)`, and I think it wasn't `*` on purpose since `*` would seem to imply that the resulting indices are the `uniqueinds(A,phi)`, instead of the indices of `phi`). 

-Matt

emstoudenmire: Yes, it's not a huge issue but I think Matt's points are all ones to be careful about for the long-term generality of the code. So I'd agree with going back to `A(phi)`.

(Though as a side question, I can't remember whether I defined `A(phi)` or not. Is that overload due to standard usage or compatibility with other Julia libraries? Otherwise just naming that function something like `product` could be even clearer.)

src/mps/mpo.jl
```diff
@@ -279,12 +279,55 @@ end
...
+            localH     = E[b] * A[b] * A[b+1] * E[b + 3]
+            wf_A       = noprime(dag(wf_A))
+            dir        = ha == 1 ? "fromleft" : "fromright"
+            replaceBond!(guess_psi, b, wf_A; dir=dir, kwargs...)
+            if ha == 1
+                E[b+1] = noprime(left_wf_A) * guess_psi[b]
+            else
+                E[b+2] = noprime(right_wf_A) * guess_psi[b+1]
+            end
+            ha > 2 && break
```
emstoudenmire: Another picky thing, but I believe `sweepnext` should only take `ha` through the values 1 and 2, so this condition should never be needed. It would be best to check this though. But I think it's safe to just remove this line; if so please do.

src/mps/projmpo.jl
```diff
@@ -48,7 +48,8 @@ function product(pm::ProjMPO,
   return noprime(Hv)
 end
 
-(pm::ProjMPO)(v::ITensor) = product(pm,v)
+(pm::ProjMPO)(v::ITensor)  = product(pm,v)
+*(pm::ProjMPO, v::ITensor) = product(pm,v)
```
emstoudenmire: If you agree with Matt's comments above, please remove the `*` overload here.

emstoudenmire (changes requested):
Please take a look at the comments, but otherwise looks ready to go to me! If there are some issues with tags and/or robustness in terms of prime levels or indices of the input MPS, those are gradual changes we can introduce later. But this code already looks pretty solid!

src/iterativesolvers.jl
kshyatt: The reason I changed it was that there's currently no way to `Davidson` an `ITensor` directly, because it doesn't implement `product` (yet...)

mtfishman: Miles: it looks like someone implemented `A(phi)` manually, that is not automatically defined (https://github.com/ITensor/ITensors.jl/blob/master/src/mps/projmpo.jl#L51).

I see... Yeah in the C++ version it required creating a small wrapper around an ITensor that implements `product` and `size` (https://github.com/ITensor/ITensor/blob/v3/unittest/iterativesolvers_test.cc#L11).

I am not sure if it is better to require people to make a small wrapper class like that or overload functions directly on `ITensor`. One concern I have about allowing people to directly pass an ITensor is that something like `size(ITensor)` seems fairly ambiguous, though maybe we can just come up with a different name for it.

I would say that maybe the best approach is for us to provide a standard `ITensorMap` wrapper for people (where it takes an ITensor with indices `(i,j,k,...,i',j',k',...)` and treats it like a matrix). Then you can just call `davidson(ITensorMap(A),...)` for an ITensor `A` than those functions are already available. We can make sure that object works with all of our iterative solver implementations (and perhaps other functions). It could even let us pass some extra type information, like `ITensorMap(A,Hermitian)`, so that we can have a general `eigs` function like Julia has that dispatches on `arnoldi` or `davidson` depending on the symmetry (just something to consider).

Issue #82, opened 2019-08-16
factorization with `eigen` does not respect `tags` keyword


orialb:
Hi,
It seems that the issue of bond tags not being conserved when `replaceBond!` is not fully fixed by https://github.com/ITensor/ITensors.jl/pull/81, because the factorization functions using `eigen` don't respect the `tags` keyword. My tests where unfortunately such that only SVD was used. 

I think a small change in `storage_eigen` fixes that (see https://github.com/orialb/ITensors.jl/commit/91515529ee5e5b8ad0ee8fd5742cb2cff6d46311) . 
I wanted to ask if this is how you meant it to be handled, because I saw that in the case of svd factorization the tags replacement is done in the `_factorize...` function and not in the `storage_svd` function.

mtfishman:
Sorry for the late response about this. That looks like the right fix to me.

orialb:
this was fixed during the re-design of the Tensor storage system. Seems to be working properly now.

PR #97, opened 2019-09-10
fix storage_eigen such that it uses tags keyword

orialb:
This is a small fix for https://github.com/ITensor/ITensors.jl/issues/82. 
Also added some tests to make sure `replaceBond!` preserves bond indices for all possible options of `which_factorization` and `dir`.


emstoudenmire:
Thanks for this, but it’s now been superseded by Matt’s reworking of the ITensor storage system (which is now based on an entirely, or nearly separate tensor library without ‘intelligent’ indices).

PR #99, opened 2019-09-14
docs for randomITensor; IndexSet indexing

christopherdavidwhite:


src/indexset.jl
```diff
@@ -52,8 +52,9 @@ function Base.show(io::IO, is::IndexSet)
   end
 end
 
-getindex(is::IndexSet,n::Integer) = getindex(is.inds,n)
+getindex(is::IndexSet,n::Any) = getindex(is.inds,n)
```
emstoudenmire: Could you please elaborate on why this type should be `Any` instead of `Integer`? Thanks!

src/itensor.jl
```diff
@@ -358,6 +358,11 @@ end
 
 const Indices = Union{IndexSet,Tuple{Vararg{Index}}}
 
+"""
+randomITensor([S,] inds)
```
emstoudenmire: Please add 4 spaces at the beginning of the line, so the Markdown processor will display this line as code instead of text. Minor thing but helpful & conventional.

emstoudenmire:
Thanks Christopher! Doc pull requests are very welcome

src/indexset.jl
christopherdavidwhite: I want to do

    fst = inds(A)[1]
    rst = inds(A)[2:end] 

(for context, I'm contructing a tree tensor network in kind of a funny way.) If getindex only takes Integers, that second line---which seems to me a pretty reasonable thing to do---doesn't work.

Like I said in the commit message: question is how much array-like behavior you folks want to support. If you change the type to Any, like I have here, you're letting the users go wild---but this means that if you ever change the underlying is.inds to anything *besides* an array, you risk breaking somebody's code.

Also: I've spent the last ~20 min. playing with this, and I've realized that there's a bit of PEBKAC here. You've already got perfectly workable Array semantics, near as I can tell, and I can do what I want with the (slightly ugly, but workable)

    fst = first(inds(A))
    rst = setdiff(inds(A), [fst])

If I expect everything to have Array semantics just 'cause I mostly use arrays, that's my problem, not yours. (Having said this, might be worth noting this in the docs somewhere; I'm happy to go write a little note in a docstring for the constructor.)

src/itensor.jl
christopherdavidwhite: Makes sense. Done.

src/indexset.jl
mtfishman: In fact, we are discussing other types of storage besides `Vector{Index}` for the IndexSet (for type inference, it is useful to know the order of the ITensor at "compile" time, so we might make it an MVector or even an SVector/NTuple). I think for now it may be better to be conservative about the indexing and keep it as just `Int`. In most use cases we have, we use "set" semantics like you show at the end of your comment, and don't rely on knowing the locations of specific indices. Ideally, indexing with ranges would not come up very often.

A question would also be, if we take a range of indices from an IndexSet, does it return a vector or an IndexSet? Questions like that make we want to avoid that route for now...

mtfishman: I realized that for the use case you discuss, we could provide functions like `pop!` and `popfirst!` to remove the last and first indices of an IndexSet. I think these are generally useful, and the advantage over the range indexing version is that the length of the output IndexSet is inferable (which we will care about once IndexSet's are parametrized by their size).

christopherdavidwhite: Yeah, this'd be helpful. (This is more or less what I did for my own use.)

Also, UI suggestion---if you don't want to provide full array-like behavior, might be better not to expose getindex( :: Integer) at all. Could be a little misleading.

mtfishman: Even though the UI is mostly based on set functions, I don't see a big downside to having `getindex` (since it is not really a set, it will be more like a Tuple/MVector with mostly a set interface). This seems like more of a documentation issue (and there is a reason why IndexSet is not an AbstractArray or AbstractSet, so that should tip people off that it has unique behaviour). 

The reason for not providing `getindex(::IndexSet,::UnitRange)` is mostly for type stability once we make the size of the IndexSet static (Base defines `getindex(::Tuple,::UnitRange)` but it is not type stable, which I find a bit odd. StaticArrays avoids this issue by returning an Array, which allocates so I also don't like that for IndexSet).

emstoudenmire: Good points Chris. I wish we could avoid having `getindex` for the reason you said, but thinking about it some more I think we need it as a developer-facing interface, so we can refactor the internals of IndexSet without having to change lots of other code. Even though IndexSet has set-like semantics, the reality is the storage is a contiguous array for efficiency reasons, and we need to treat it as such in lower-level code. So I think `getindex` has to stay.

One minor comment about the code you pushed: I think it's better to leave off the explicit `::Any` when having an argument accept any possible input. Instead just put no type at all. The reason is that the JIT can specially compile different versions of functions to adapt to the different types actually passed in practice, whereas with `::Any` written there, it might prevent this behavior. (Maybe the JIT works optimally either way, but I'd be concerned about the `::Any` case. It's something that would be good to test using `@code_warntype`.)

PR #100, opened 2019-09-16
exp function for ITensor (with dense storage)

orialb:
This PR adds a function `exp(A::ITensor, Lis::IndexSet)` which computes the exponential of A where it is treated as a matrix A_{lr} with l running over indices in Lis and r over the other indices of A. 

The use case I have in mind is creating time evolution gates from bond Hamiltonians (or some other bond operators).

### DIFF
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -4,6 +4,7 @@ export ITensor,
        combinedindex,
        delta,
        δ,
+       exp,
        replaceindex!,
        inds,
        isNull,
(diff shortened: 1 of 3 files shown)


mtfishman:
Looks good, maybe we also want `expHermitian` that calls `exp(Hermitian(...))` internally?

orialb:
Are you thinking of  a use case where the tensor is supposed to be hermitian but `ishermitian` will return `false` due to numerical errors ? I guess it makes sense to add this method , I'll add this.  
For the case where the matrix is exactly hermitian (which I think is what will happen when building Hamiltonians) I don't really see a performance difference between `exp(...)` and `exp(Hermitian(...))` )

emstoudenmire:
Hi Ori, yes that's right. Agreed that there won't be a performance difference when the matrix is exactly Hermitian, because Julia's exp function will use efficient Hermitian matrix routines internally and automatically.

But that's a big "if". In practice we will encounter many matrices which are only approximately Hermitian but which we know are supposed to be Hermitian if no approximations had been made, so we prefer the code to only use the upper triangle (as exp(Hermitian(...)) will do). 

Personally I think this is a mild mis-design of Julia's linear algebra system; one of those places where the ideal is nice in principle, but reality is usually different (due to the reality of floating point, roundoff error, etc.). In practice, I've seen older Julia code where people had to numerically symmetrize their matrices to "trick" the linear alg routines to calling the right internal thing. Fortunately since then they added the Hermitian wrapper type.

orialb:
Another option is to have an `ishermitian` keyword in `exp` so `exp(A::ITensor,Lis, ishermitian = true)` will call `exp(Hermitian(...))` internally. 
Maybe this will enable writing more general code in some situations as compared to `expHermitian` ? 
Or we can have both `expHermitian` and the keyword option. 
Any preferences?

emstoudenmire:
Good suggestion - I was thinking the same thing.

How about you add the keyword argument, then make expHermitian just a wrapper which plugs in `ishermitian=true` into the keyword arguments? That way people could use whichever is more convenient, and I agree the keyword version is much better within generic code.

orialb:
I added the `hermitian` keyword (thought it would be better to avoid collision with `ishermitian` function from LinearAlgebra)  and the `expHermitian` wrapper.

emstoudenmire:
Thanks Ori!

PR #101, opened 2019-09-16
Working on applyMPO bug

kshyatt:
So `applyMPO`, which I wrote (😰) doesn't work with `linkDim > 1`. The tests I've included for such cases will fail. I've compared with the C++ but clearly something is still wrong.


kshyatt:
Thanks @mtfishman for suggesting this was probably cutoff error. I added some larger link dimension tests for `inner` too.

kshyatt:
CI passed, life is amazing

PR #102, opened 2019-09-18
SmallString improvements

emstoudenmire:
The main changes in this PR are:
1. change internals of SmallString so it becomes an isbits type (and it no longer stores its length). This was needed to use Val{Tag("TagName")} in some experimental features.
2. it's now easier to construct SmallStrings/Tags, from Strings for example
3. getindex(TagSet,...) now returns Tag instead of IntSmallString

I made some other small code improvements & cleanup too, and added a few tests.

### DIFF
--- a/src/smallstring.jl
+++ b/src/smallstring.jl
@@ -1,5 +1,5 @@
 export convert,
-       push,
+       #push,
        setindex
 
 const IntChar = UInt8
@@ -41,17 +41,17 @@ end
 
 isNull(s::SmallString) = @inbounds s[1] == IntChar(0)
 
-function StaticArrays.push(s::SmallString,val)
-  newlen = 1
-  while newlen <= smallLength && s[newlen] != IntChar(0)
-    newlen += 1
-  end
-  if newlen > smallLength
-    throw(ErrorException("push: SmallString already at maximum length"))
-  end
-  icval = convert(IntChar,val)
-  return SmallString(setindex(s.data,icval,newlen))
-end
+#function StaticArrays.push(s::SmallString,val)
+#  newlen = 1
+#  while s[newlen] != IntChar(0) && newlen <= smallLength
+#    newlen += 1
+#  end
+#  if newlen > smallLength
+#    throw(ErrorException("push: SmallString already at maximum length"))
+#  end
+#  icval = convert(IntChar,val)
+#  return SmallString(setindex(s.data,icval,newlen))
+#end
 
 function SmallString(i::IntSmallString)
   mut_is = MVector{1,IntSmallString}(ntoh(i))
@@ -73,7 +73,7 @@ end
 
 function isint(s::SmallString)::Bool
   ndigits = 1
-  while ndigits <= smallLength && s[ndigits] != IntChar(0)
+  while s[ndigits] != IntChar(0) && ndigits <= smallLength
     cur_char = Char(s[ndigits])
     !isdigit(cur_char) && return false
     ndigits += 1
@@ -113,21 +113,21 @@ Base.isless(s1::SmallString,s2::SmallString) = isless(s1.data,s2.data)
 
 #######################################################
 
-function Base.String(s::SmallString)
-  res = ""
-  n = 1
-  while n <= smallLength && s[n] != IntChar(0)
-    res *= Char(s[n])
-    n += 1
-  end
-  return res
-end
+#function Base.String(s::SmallString)
+#  res = ""
+#  n = 1
+#  while s[n] != IntChar(0) && n <= smallLength
+#    res *= Char(s[n])
+#    n += 1
+#  end
+#  return res
+#end
 
 #Base.convert(::Type{String}, s::SmallString) = String(s)
 
 function Base.show(io::IO, s::SmallString)
   n = 1
-  while n <= smallLength && s[n] != IntChar(0)
+  while s[n] != IntChar(0) && n <= smallLength
     print(io,Char(s[n]))
     n += 1
   end
--- a/src/smallstring.jl
+++ b/src/smallstring.jl
@@ -4,13 +4,9 @@ export convert,
 
 const IntChar = UInt8
 const IntSmallString = UInt64
-const maxTagLength = 8
-const SmallStringStorage = SVector{maxTagLength,IntChar}
-const MSmallStringStorage = MVector{maxTagLength,IntChar} # Mutable SmallString storage
-
-function initSmallStringStore()
-  return SmallStringStorage(ntuple(_->IntChar(0),Val(length(SmallStringStorage))))
-end
+const smallLength = 8
+const SmallStringStorage = SVector{smallLength,IntChar}
+const MSmallStringStorage = MVector{smallLength,IntChar} # Mutable SmallString storage
 
 struct SmallString
   data::SmallStringStorage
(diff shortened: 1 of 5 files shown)


src/smallstring.jl
```diff
@@ -1,66 +1,86 @@
...
   end
-end
 
-function SmallString(sv::SmallStringStorage)
-  sv[1] == IntChar(0) && return SmallString(sv,0)
-  len = 1
-  while len < length(SmallStringStorage) && @inbounds sv[len+1] ≠ IntChar(0)
-    len += 1
+  function SmallString(str::String)
+    sv = initSmallStringStore()
```
mtfishman: Could this function be done by creating an `MSmallStringStorage` and then filling it with the values with `setindex!`, then converting to a `SmallStringStorage`/`SmallString`? That should still avoid allocations, and I believe I have seen that pattern was a bit faster.

src/smallstring.jl
```diff
@@ -1,66 +1,86 @@
...
 
 isNull(s::SmallString) = @inbounds s[1] == IntChar(0)
 
 function StaticArrays.push(s::SmallString,val)
-  len = length(s)
-  return SmallString(setindex(s.data,val,len+1),len+1)
+  newlen = 1
+  while newlen <= maxTagLength && s[newlen] != IntChar(0)
+    newlen += 1
+  end
```
mtfishman: The reason I put the length into `SmallString` (as I am sure you know) was to avoid having to do checks like this. Was the reason to remove it because we want `SmallString` to be `isbits`? It seems like we can still have that:

```julia
struct SmallString
  data::SVector{8,UInt8}
  length::Int
end

isbitstype(SmallString)   # I get true
```

Did that not work with the previous definition? I think structs of isbits types are still isbits.

src/smallstring.jl
```diff
@@ -1,66 +1,86 @@
...
+  if newlen > maxTagLength
+    throw(ErrorException("push!: SmallString already at maximum length"))
```
mtfishman: Should be "push: ...".

src/tagset.jl
```diff
@@ -57,15 +58,15 @@ end
...
+  if !isNull(t)
     if isint(t)
       error("""Cannot use a bare integer as a tag.
             If you are looking to set the prime level, use the syntax ("x",1)
             (instead of "x,1")""")
       #plev ≥ 0 && error("You can only make a TagSet with one prime level/integer tag.")
       #plnew = parse(Int,t)
     else
-      ntags = _addtag_ordered!(ts, ntags, tag)  
+      ntags = _addtag_ordered!(ts, ntags,IntSmallString(tag))
```
mtfishman: Why is this constructor call needed? `tag` is already `IntTag = IntSmallString`.

src/smallstring.jl
emstoudenmire: Good feedback! Ok hopefully improved in the latest commits. Please take a look.

src/smallstring.jl
emstoudenmire: Yes, I did jump the gun here. I thought the layout on master wasn't isbits but I was wrong about that.

There's a tradeoff here between memory and computation to be sure. Including the length would make the memory footprint exactly twice as large (since the length is also a 64 bit integer). But it would still meet the key design goal of not allocating.

I had forgotten, but this was the reason I actually capped the length of small strings at 7 in the C++ version. Then the last character could always be guaranteed to be equal to IntChar(0), through the constructor's behavior, and then it is safe for loops to only check equivalence to IntChar(0) and not have to also compare to the max length.

Would you be ok to cap the length at 7 (still keeping the storage at 8 chars like it is now)? Then we could have both 64 bits of memory and speed both! The only downside is having to keep tags shorter.

src/smallstring.jl
emstoudenmire: Fixed

src/tagset.jl
emstoudenmire: Yes, this was a mistake I made when changing other calls to `_addtag_ordered!`. I'll fix it - good catch!

src/smallstring.jl
mtfishman: Yes I guess that makes the most sense. I wasn't thinking about memory, but an ITensor could end up with quite a lot of tags. Too bad we can't have the tag "Electron".

emstoudenmire: That is a shame huh. Let's think about this some more. I was thinking the other day that one might want to use `Electron` operators for spinful fermions which aren't electrons (e.g. protons are one key example). So maybe there's a better name anyway. But SFermion is also 8 characters! Is there another good name? Elec? SpinF? Hm...hard to think of one.

(Sidebar: If we wanted to get really low-level with this, I just read that ASCII characters only use their first 7 bits anyway. So in principle one can pack the bits to the front and get some at the end back even with 64 bits. There's even an experimental Julia code that does something like this! https://github.com/tpapp/PackedASCIITypes.jl  But it's probably overkill. And it causes extra cpu steps to do things like retrieving a single character.)

mtfishman: In the C++ version we use "Elec", I can't really think of anything better either. Protons are a good example, maybe with the new tag-based system we can think of an easy way to let people use multiple names to call the same operators? (You could always just overload `op` and `state` on a ProtonType to call the ElectronType functions). It may be nice to define both "Elec" and "Proton" indices even if they use the same operators.

I also just realized the length doesn't need to by an Int64, just a UInt8, since of course it is always positive and not larger than UInt8. So maybe it is not too bad to store the length?

That package seems quite close to what we want, too bad accessing elements is not very good, we don't want to convert to strings all the time.

emstoudenmire: One possibility is we keep it at 8, don't put the length back in, and then just allow the extra computation of comparing to the maximum length to happen. It's not perfectly ideal, but I think in practice it will be only a very little bit extra. I'm willing to bet that comparisons of SmallStrings are really still how they mostly get used, and those remain fast.

Doing a survey of the code as it stands, the double comparison (to both IntChar(0) and smallLength) only happens in `push`, `String(SmallString)`, `show` and `isint`. However, push and the conversion to String aren't used anywhere in the code right now! The show function for small strings doesn't happen in any speed-critical loops. Quite probably `isint` never gets called in such loops either.

mtfishman: Yeah that seems fine. I guess the key decision we want to make now is how long the tag names should be, because that would be an annoying breaking change to make later on if we set them at 8 and then switched it back down to 7. But with 8, it looks like we have two different design options that we can test out for efficiency, so it doesn't seem to lock us into a single design. It is too bad `isint` needs to get called every time a tag is made, perhaps we can even loosen that restriction or make that a debug check.

`push` was there when I was A/B testing whether it was better to fill an `MTagSetStorage` when creating a TagSet or replace with a new `TagSetStorage` as tags got added, I am fine with removing it. The first version seemed to be faster, but it is something to keep in mind if we want to revisit optimizing the TagSet code later on.

emstoudenmire: I think keeping it at 8 is the right call so we can use good names.

Thanks for that info about push. Regarding isint, yes we could perhaps turn it off, especially now that primes are passed separately in a tuple (though it probably remains a good idea to disallow integer tags).

There may be one nice way to speed up isint I thought of today: just fully unroll the loop and write a function of 10 lines like this:
```Julia
is_int = false
is_int = (s[1] != IntChar(0) && isdigit(Char(s[1])))
is_int = (s[1] != IntChar(0) && isdigit(Char(s[1]))) 
is_int = (s[1] != IntChar(0) && isdigit(Char(s[1]))) 
```

mtfishman: Interesting idea. We can probably write a macro for that. I guess the over all conclusion though is that it is probably fast enough as is, so we can go with 8 characters and play with optimizations later if needed.

emstoudenmire: Ok yes I'm good with that conclusion too. Based on that then, do you think this is ok to pull at this point?

mtfishman:
Looks good to me!

PR #103, opened 2019-09-19
"Local physics" tags, replacing SiteSet system

emstoudenmire:
This PR is a pretty big redesign, or replacement rather, of the "site set" system. Merging this branch would make the following major changes:
* The `SiteSet` type is removed in favor of just using `Vector{Index}`
* The `op` function for getting physical operators and the `state` function for getting IndexVals from strings still work, but they work by recognizing special Index Tags
* To define one of these special Tags and the related `op` or `state` overloads, one writes code like the following (showing what a user would write outside of the ITensors module):
```Julia
using ITensors
import ITensors.op    #want to overload this
import ITensors.state #want to overload this

const QubitSite = makeTagType("Qubit")

function state(::QubitSite, st::AbstractString)
  if st=="0"
    return 1
  elseif st=="1"
     return 2
  else
  throw(ArgumentError("State string \"$st\" not recognized for Qubit site"))
end

function op(::QubitSite, s::Index, name::AbstractString)
  Op = ITensor(s,s')
  if name=="Z"
    Op[s(1),s(1)'] =  1.0
    Op[s(2),s(2)'] = -1.0
  elseif
     # ...
  end
  return Op
end
```


emstoudenmire:
A key advantage of this change is that we don't need to get the design of the `SiteSet` type right anymore; it was getting complicated. Instead `Vector{Index}` is easy to deal with and no new concepts are needed there.

Another key advantage is no matter how many functions deep you are in the stack, if you have an Index that represents a physical site, you can call `op` on it to get an operator if there is a matching tag.

mtfishman:
Definitely this is a very minimal interface for users and requires very little new code, it looks quite nice! What a way to take advantage of Julia's type system.

I know this is a high level design (and doesn't involve things that should be used in hot loops), but I am curious if there is any noticeable efficiency difference between the previous design and the old design (since the old design was probably easier on type inference and didn't involve searching through the tags). Like if someone is generating a lot of operators (maybe to make a quantum chemistry hamiltonian with a lot of terms), I think it would be good to check that this design doesn't cause some unexpected noticeable slowdown.

mtfishman:
I guess a quick test is to profile `op` and make sure that a majority of the time it takes to call `op` is spent allocating the ITensor itself, in which case the efficiency of this system is not an issue.

emstoudenmire:
Glad you're enthusiastic about this Matt. I am too, because it's so lightweight in terms of its intrusiveness on the rest of ITensor and the number of lines of code it involves. And yet it adds capabilities we didn't have before with the site set system.

Good suggestion to test the efficiency, because yes I could imagine use cases where someone generates a lot of operators in a loop (like maybe a quantum circuit sort of application? or doing a lot of measurements of a very small bond-dimension but very long MPS). Anecdotally I used this with AutoMPO and didn't see any speed difference. But yes I will time it (compared to just making the same operators "by hand") and let you know.

emstoudenmire:
Maybe there are other speed tests we should do, but I just did some timings by making an operator over and over again in a loop calling `op` with the new system, versus hand-written code which makes the same operator.

I consistently find that the new system is 4x (more like 3.5x) slower than the manual approach. I think that's acceptable for the convenience.

In absolute units, making a 2x2 operator ITensor takes 10 microseconds. So when using the `op` system it would be 35 microseconds per operator. For 5000 operators that's roughly 0.05 seconds with the manual approach versus roughly 0.20 seconds for the new `op` approach in this PR.

mtfishman:
That seems like an acceptable slowdown.

I keep trying to think of downsides to this approach, and the only potential ones are:

1. Speed (which doesn't seem to be a problem)
2. Magic (maybe users will be surprised to see tags used in this way for dispatch)

However, I don't think this is so magical and it is pretty easy to explain. Over all this seems like the best design for this system we have discussed, I am fine going ahead with this.

examples/dmrg/2ddmrg.jl
```diff
@@ -10,13 +10,13 @@ let
...
 
-  ampo = AutoMPO(sites)
+  ampo = AutoMPO()
   for b in lattice
     add!(ampo,0.5,"S+",b.s1,"S-",b.s2)
     add!(ampo,0.5,"S-",b.s1,"S+",b.s2)
     add!(ampo,    "Sz",b.s1,"Sz",b.s2)
   end
-  H = toMPO(ampo)
+  H = toMPO(ampo,sites)
```
mtfishman: I definitely prefer having the sites passed to the MPO constructor instead of to the AutoMPO! 

Should we consider to calling the constructor just `MPO` instead of `toMPO`? I think we only used `toMPO` in the C++ version because we wanted to pass keyword arguments to the constructor but couldn't because of C++.

emstoudenmire: Yes, the only downside is giving a way to check if the AutoMPO input is out of bounds or not. But this does seem like the right move. 

I’m good will calling it MPO again yes! Let’s keep toMPO also to assist people already used to C++.

src/physics/site_types/electron.jl
```diff
@@ -0,0 +1,107 @@
...
+  Up    = s(2)
+  UpP   = sP(2)
+  Dn    = s(3)
+  DnP   = sP(3)
+  UpDn  = s(4)
+  UpDnP = sP(4)
+
+  Op = ITensor(dag(s), s')
+
+  if opname == "Nup"
```
mtfishman: This may be a bit of an aside from this pull request, but can we use a similar trick to use dispatch to choose the operator instead of a long list of if statements?

For example, we could do:

```julia
function op(::ElectronSite, s::Index, opname::AbstractString)
  Op = ITensor(dag(s), s')
  op!(Op, s, ElectronSite(), Val(Symbol(opname)))
  return Op
end

function op!(Op::ITensor, s::Index, ::ElectronSite, ::Val{:Nup})
  Op[s(2),s'(2)] = 1.
  Op[s(4),s'(4)] = 1.
  return Op
end
```

The advantage of this is avoiding a very long set of `if ... elseif ...` statements. Also, this allows people to add their own operators to an existing site type by overloading `op!` (in case it is something complicated we may not want to include, but is useful for their own use case).

If we don't like `Val` in the signature, it may be nicer to use a type alias like `const OpType{T} = Val{T}` so then the signature is `OpType{:Nup}`. We could also use Tag/SmallString instead of a symbol, but right now SmallString doesn't allow unicode (and of course has length limits).

examples/dmrg/2ddmrg.jl
mtfishman: I guess bounds checking can still be done when creating the MPO from the AutoMPO and the sites, right?

src/physics/site_types/electron.jl
emstoudenmire: Yes I was thinking about something like this too, or at least using this trick more broadly. I wish this kind of thing was more natively built into the language hmm. 

I think we could always add this later without having to do any breaking changes, though. So let's definitely keep it in mind. There are some details to think about, like whether to use a SmallString or a Symbol (good thing about symbol is it can be long, but it forces a somewhat advanced concept on users).

Probably a good solution is to create a helper function that returns Val(Symbol(opname)) so users can just call that.

examples/dmrg/2ddmrg.jl
emstoudenmire: Yes - ok good point! It doesn't hurt to delay it a little if it gets done eventually.

Also, in a way it gives more flexibility to AutoMPO. Like one could use sites past N for making infinite MPO's, then it's up to the conversion function to figure out whether it wants to allow that or not.

emstoudenmire:
Thanks for the feedback. I agree about the magic part, but I think the opaqueness of each version of SiteSet I've come up with ended up feeling too magical too. So it's hard to get rid of the magic altogether! This does nicely "unmagic" some things like what a site set is (just a vector of indices). And it decouples this system a lot more from things like MPS and MPO, which I like.

PR #108, opened 2019-09-20
Begin adding capability to read from C++ ITensor files

emstoudenmire:
This pull adds the ability to read dense and combiner ITensors, as well as Tag, TagSet, Index, and IndexSet from binary files written by the C++ read/write system.

We ought to think of a way to unit test this. One way could be to include a sample C++ code that generates output files and a Makefile, and some companion Julia unit test code, but just not include it in runtests.jl. Instead we could run it manually once in a while.

### DIFF
--- a/src/index.jl
+++ b/src/index.jl
@@ -190,7 +190,9 @@ function Base.read(io::IO,::Type{Index}; kwargs...)
     tags = read(io,TagSet;kwargs...)
     id = read(io,IDType)
     dim = convert(Int64,read(io,Int32))
-    dir = read(io,Arrow)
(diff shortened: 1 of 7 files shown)


emstoudenmire:
Still needed are reading of Diag{T}. It should be easy to add...

emstoudenmire:
@kshyatt do you happen to know of a way to include this without really dropping our code coverage? The problem is, to test it we need binary files for it to read. But these files are not portable across platforms the way the C++ write-to-disk works. And it may be tricky or undesireable to make our tests require the C++ version of ITensor.

No need to implement it, but just wondering if you have any advice. Thanks!

Issue #106, opened 2019-09-20
Add syntax `A[i=>1,j=>2]` as syntax for getting/setting elements of an ITensor


mtfishman:
This should just involve making a new `IndexVal` constructor that takes a `Pair{Index,Int}`, then extending the definition of `getindex(::ITensor,...)`

mtfishman:
Something I just noticed while randomly playing around with the REPL:

```julia
julia> =>(i::Int,j::Int) = i+j
=> (generic function with 1 method)

julia> 1=>2
3
```

so we can directly overload the `=>` operator. This could give us a simpler way to implement this, since we could just define:

```julia
=>(i::Index,val::Int) = IndexVal(i,val)
```

and avoid having to use `Pair` anywhere. The only downside I see is that it removes that notation if you actually _want_ a `Pair{Index,Int}`, but you could always just call the `Pair` constructor directly.

emstoudenmire:
I think this is great for most, if not all purposes. It's definitely nice in that it's a minimal amount of extra code to get a great feature. And there's something to be said for not having more types flying around. (Meaning that we don't have to explain that `Pair{Index,Int}` means the same thing as an `IndexVal`.)

The only downside I can see is that we are sort of breaking, or hijacking a feature of the language. Meaning that if someone *did* expect i=>2 to give a Pair and then it didn't, they could be surprised and possibly spend too long trying to figure out what happened. 

So I currently still vote we do it this way, but let's think about if that downside I just mentioned is acceptable.

mtfishman:
Closed by #227 (it was easy enough to do without defining `=>(i::Index,val::Int) = IndexVal(i,val)`, so I avoided that).

PR #111, opened 2019-09-24
Pair indexval

emstoudenmire:
Implements #106 

Allows the notation `T[i=>2,j=>1]` for an ITensor with indices i,j.

### DIFF
--- a/src/index.jl
+++ b/src/index.jl
@@ -177,8 +177,16 @@ ind(iv::IndexVal) = iv.ind
 ==(i::Index,iv::IndexVal) = (i==ind(iv))
 ==(iv::IndexVal,i::Index) = (i==iv)
 
+==(i::Index,pin::Pair{Index,Int}) = (i==pin.first)
+==(pin::Pair{Index,Int},i::Index) = (i==pin)
+val(pin::Pair{Index,Int}) = pin.second
+
 plev(iv::IndexVal) = plev(ind(iv))
 prime(iv::IndexVal,inc::Integer=1) = IndexVal(prime(ind(iv),inc),val(iv))
 adjoint(iv::IndexVal) = IndexVal(adjoint(ind(iv)),val(iv))
 
 show(io::IO,iv::IndexVal) = print(io,ind(iv),"=$(val(iv))")
+
+function IndexVal(ivp::Pair{Index,Int})
+  return IndexVal(ivp.first,ivp.second)
+end
--- a/src/index.jl
+++ b/src/index.jl
@@ -177,9 +177,13 @@ ind(iv::IndexVal) = iv.ind
 ==(i::Index,iv::IndexVal) = (i==ind(iv))
 ==(iv::IndexVal,i::Index) = (i==iv)
 
-==(i::Index,pin::Pair{Index,Int}) = (i==pin.first)
-==(pin::Pair{Index,Int},i::Index) = (i==pin)
+ind(pin::Pair{Index,Int}) = pin.first
 val(pin::Pair{Index,Int}) = pin.second
+==(i::Index,pin::Pair{Index,Int}) = (i==ind(pin))
+==(pin::Pair{Index,Int},i::Index) = (i==pin)
+
+ind(v::AbstractVector{IndexVal}) = ind(v[1])
+val(v::AbstractVector{IndexVal}) = val.(v)
 
 plev(iv::IndexVal) = plev(ind(iv))
 prime(iv::IndexVal,inc::Integer=1) = IndexVal(prime(ind(iv),inc),val(iv))
@@ -190,3 +194,4 @@ show(io::IO,iv::IndexVal) = print(io,ind(iv),"=$(val(iv))")
 function IndexVal(ivp::Pair{Index,Int})
   return IndexVal(ivp.first,ivp.second)
 end
+
--- a/src/indexset.jl
+++ b/src/indexset.jl
@@ -453,6 +453,25 @@ function calculate_permutation(set1, set2)
   return p
 end
 
+function permute_map(set1, set2, map)
+  L1 = length(set1)
+  L2 = length(set2)
+  L1==L2 || throw(DimensionMismatch("Mismatched input sizes in permute_map: l1=$l1, l2=$l2"))
+  res = zeros(Int,L1)
+  for i1 = 1:L1 
+    found = false
+    for i2 = 1:L2
+      if set1[i1]==set2[i2]
+        res[i1] = map(set2[i2])
+        found = true
+        break
+      end
+    end
+    !found && error("Sets aren't permutations of each other")
+  end
+  return res
+end
+
 function compute_contraction_labels(Ai::IndexSet,Bi::IndexSet)
   rA = order(Ai)
   rB = order(Bi)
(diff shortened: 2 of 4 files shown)


src/index.jl
```diff
@@ -177,8 +177,21 @@ ind(iv::IndexVal) = iv.ind
 ==(i::Index,iv::IndexVal) = (i==ind(iv))
 ==(iv::IndexVal,i::Index) = (i==iv)
 
+ind(pin::Pair{Index,Int}) = pin.first
+val(pin::Pair{Index,Int}) = pin.second
+==(i::Index,pin::Pair{Index,Int}) = (i==ind(pin))
+==(pin::Pair{Index,Int},i::Index) = (i==pin)
+
+ind(v::AbstractVector{IndexVal}) = ind(v[1])
```
mtfishman: This seems like a strange definition, what is this for?

src/index.jl
```diff
@@ -177,8 +177,21 @@ ind(iv::IndexVal) = iv.ind
...
+val(v::AbstractVector{IndexVal}) = val.(v)
```
mtfishman: Can't we just rely on broadcasting here, i.e. just call `val.(v)` directly?

emstoudenmire:
Thanks Matt. So those are a bit strange I agree. The idea, though, was that then the `map` calls in the fancier versions of `getindex` and `setindex!` for ITensor could just call `ind` to grab the Index and `val` to grab the value, and not have to check `isa IndexVal`. That was getting tricky, because now the inputs could either be `IndexVal` or `Pair{Index,Int}`.

But I'm open to a better solution for sure!

One thing I don't like about this solution is that it makes a function `val` with the behavior of `val.`

emstoudenmire:
One idea could just be to move these `val` and `ind` overloads for `AbstractVector{IndexVal}` *inside* of `getindex` and `setindex!`, so that they don't get exported.

mtfishman:
I think a better way would be to split apart the `getindex` and `setindex!` functions into a basic one that accepts `IndexVal` or `Pair{Index,Int}` (either with a union, or leaving it generic and then converting `Pair{Index,Int} -> IndexVal`, and then another one that accepts `AbstractVector{IndexVal}` (for now we could just not have that version work with `Pair{Index,Int}`?).

emstoudenmire:
Ok I will give that a go!

You know, maybe it's also better to make a wrapper for the overloads taking `Pair{Index,Int}`. The wrapper would then convert them into `IndexVal`s and call that version. Thinking about it more, it's weird to be able to mix notations: `T[i(1),j=>2]`

mtfishman:
One way you could do that with a generic function is to have a signature `getindex(::ITensor,::T...) where {T}` which would mean that all of the inputs have to be the same type. You could also change it to `... where {T<:Union{IndexVal,Pair{Index,Int}}` to make it more specific.

mtfishman:
The wrapper version is fine too, I am just thinking that it may be good to come up with a nice generic pattern for this so that any function accepting `IndexVal` also accepts `Pair{Index,Int}`, like how I made it so that any function accepting `TagSet` also accepts `AbstractString`.

emstoudenmire:
Good suggestions.

Just now, more out of curiosity than anything, I just tried *defining* IndexVal as Pair{Index,Int}. It actually works with very minimal changes needed (and no changes at all to any ITensor functions).

But that approach has some disadvantages:
* seeing IndexVal as a type is nice for understanding the purpose of it (it's an Index paired with a value). If it's just a "typedef" i.e. `const IndexVal = Pair{Index,Int}` then when you do typeof on an IndexVal it just shows `Pair{Index,Int}`.
* the constructor would not check if the value is valid (e.g. greater than the Index dimension)

One question just for my own education is, is the definition of (i::Index, c::Colon) as returning an array of IndexVal's a choice of ours? Or is that the standard way one slices arrays in Julia? 

I ask because it seems a cleaner approach could be to just make something like `Pair{Index,Colon}` and have `getindex` and `setindex!` accept `Pair{Index,T}` where T could be `Int` and `Colon`, but also `UnitRange{Int}` which is the `typeof(2:4)` for example. Then we could be ready to allow Index slicing in the future when we want to get into that later.

mtfishman:
I was also thinking that maybe we could define `const IndexVal = Pair{Index,Int}`, but I think your argument about wanting a size check in the constructor is a good reason not to do it.

This is the definition here: https://github.com/ITensor/ITensors.jl/blob/master/src/index.jl#L173.

I believe this definition was added by Katie at some point to help with a slicing operation. As you say, I think I would prefer to pair an Index with a range to represent slicing.

I think that idea of dispatching on `Pair{Index,T}` where `T` is `Int`, `Colon`, `UnitRange`, etc. is a good way to go. In that case, it seems like we may want a more general `IndexRange` object like the following:

```julia
IndexRange{T}
   i::Index
   r::T
end
```

that gets constructed from `Pair{Index,T}`. If `T===Int` it is treated as an `IndexVal` (we could define `const IndexVal = IndexRange{Int}`). Then slicing would involve `IndexRange{Colon}`, `IndexRange{UnitRange}`, and even non-contiguous ranges with `IndexRange{Vector{Int}}`.

We would need to think of course about what the result of slicing an ITensor is. I think it should be a view of the tensor data (i.e. share the storage of the original tensor). It also seems like the best approach is to have the new ITensor have indices that are themselves the sliced indices of the original ITensor (otherwise slices of ITensors with the same indices wouldn't recognize each other, which doesn't seem good). In that case, a sliced ITensor would have `IndexRange` indices.

However, this all makes me think that maybe instead of having a separate `Index` and `IndexVal/IndexRange`, it could all be merged into just `Index` that is parametrized over a slice type:

```julia
struct Index{T}
  id::IDType
  dim::Int
  dir::Arrow
  tags::TagSet
  slice::T
end
```

A standard `Index` would be when `T===Colon`. We could then define `const IndexVal = Index{Int}`, and `const IndexRange = Index{UnitRange}`. I am not sure what the implications of that would be, but it might be nice to just have a single base type `Index` that can get passed around.

emstoudenmire:
Hi Matt, yes I think it’s good to discuss this in depth this week. I do want to get the design right in terms of having all these capabilities (different types of Index ranges, etc.). One other consideration is where do quantum numbers go when we add that in? A good design could be to have `Index{T}` where T is a type that holds extra information like that, then use `IndexRange` to add additional slicing information on top of either `Index{Nothing}` or `Index{QNInfo}` etc. (just making up some names there). 

But an alternative to `IndexRange{T}` could just be `Pair{Index,T}` because the validity of the range will eventually be checked in lower-level code for sure. Then that way it’s easier for users to understand some code since they don’t have to understand a new type. And it lightens the code complexity. It’s a tradeoff for sure, though, because `IndexRange` is more self-documenting and gives us the opportunity to produce better (higher-level) error messages. Hm.

mtfishman:
Ignoring the issues of `IndexRange` for now (by not allowing `Pair{Index,Colon}`), would this be good to go after conflicts are resolved?

emstoudenmire:
I think I'll actually re-do this pull request instead, pulling out the best parts but leaving off some of the changes. (But let's leave it open for now so I can remember to do it!)

PR #113, opened 2019-09-27
import Base.iterate to define SweepNext iterator

orialb:
Hi,
This is a super small fix to allow using `sweepnext(N)` iterator outside of ITensors.
I just added `import Base.iterate` so it will overload `iterate` properly.

### DIFF
--- a/src/mps/sweeps.jl
+++ b/src/mps/sweeps.jl
@@ -66,6 +66,8 @@ end
 
 sweepnext(N::Int)::SweepNext = SweepNext(N)
 
+import Base.iterate
+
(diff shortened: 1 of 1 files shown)


orialb:
I checked again and it looks like Base.iterate is already imported on latest master, so not sure why I encountered an error when trying to call `sweepnext(N)` in the REPL (which I don't get anymore on master) and why adding this line changed anything. I guess there was some mess with my local ITensors clone which seems to be gone now.

emstoudenmire:
Thanks for keeping an eye on this!

PR #114, opened 2019-09-30
Add QN type

emstoudenmire:
This pull request adds a QN type that's very similar to the one in the v3 branch of the C++ version. Internally an SVector storage is used similar to TagSet.


mtfishman:
Over all this seems reasonable to me.

What do you think of using `q(2)` instead of `q[2]` for creating a `QNVal`? It looks a bit nicer to me...

I also just realized that we could use `Pair{QN,Int}` to create a `QNVal` (like we will do for `IndexVal`). I think this could make creating indices with QNs look pretty nice:

```julia
i = Index(QN("Sz",-1) => 1,
          QN("Sz",+1) => 1,
          "SpinHalf,Site,n=1")
```

We could add this interface later, just bringing it up as food for thought.

Besides that I am fine with this going in (and it will be nice to have as we go foreword with implementing the block-sparse code). We could always adjust some things, since there is no reason to use this until the storage code is done.

emstoudenmire:
Hi Matt, great thanks for taking a look. Which part of the code do you mean regarding `q[n]` to create a `QNVal`? Do you mean the `getindex` method defined on `QN` which retrieves one of the `QNVal`'s stored inside the `QN`?

mtfishman:
Oh sorry I misread the code. I thought the `QNVal` was the pairing of the `QN` with it's size. I am still wondering if:

```julia
i = Index(QN("Sz",-1) => 1,
          QN("Sz",+1) => 1,
          "SpinHalf,Site,n=1")
```

might be nice syntax for constructing a QN Index, however.

mtfishman:
But again, that is a bit separate from this pull request. Go ahead and merge when you are ready.

PR #118, opened 2019-10-07
Small tests for diag ITensors

kshyatt:


kshyatt:
Odd that this isn't calling the `*(x::Number, D::Diag)` one...

PR #119, opened 2019-10-11
bugfix in swapprime!

christopherdavidwhite:
Previously swapprime mapped pl1 -> pl2, but did not map pl2 -> pl1 (essentially duplicating mapprime). This commit changes swapprime so it maps pl1 -> pl2, pl2 -> pl1. It also
    - uses mapprime rather than duplicating code.
    - changes the API (single match rather than varargs, for convenience---so we can call mapprime---and consistency with mapprime.


mtfishman:
Thanks for catching this and fixing it!

src/indexset.jl
```diff
@@ -3,6 +3,7 @@ export IndexSet,
        hasinds,
        hassameinds,
        findindex,
+       lastindex,
```
mtfishman: Was this included by accident?

mtfishman:
It looks like some of the tests are failing now (ones related to `swapprime`, which was clearly wrong before, so maybe the tests themselves are wrong).

src/indexset.jl
christopherdavidwhite: yep

christopherdavidwhite:
Well, some of them are and some of them aren't: e.g. the test on line is wrong, but the test on line 111 is _not_---that ought to work as is. 

I'll poke around.

christopherdavidwhite:
So two things were going on here:
    - the test on line 111 hit an edge case in my code---test was right, code was wrong.
    - Most of the rest of the tests were wrong: they were written assuming swapprime! worked exactly like mapprime!. (Which may be what you folks intended, I don't know, but seems a little strange.)

I've updated the code to deal with the edge case 111 hit, and updated the rest of the tests to reflect the desired behavior.

I also added a couple more that more that more directly probed swaps of tagged indices.

emstoudenmire:
Great: thanks for catching this. Many of those tests were in error - you're right.

I think it's ready to pull. I wasn't 100% sure which convention we were deciding on for the optional arguments, but most of the functions in indexset.jl use `match=nothing` so I think it's good for swapprime to be consistent with that, yes.

PR #122, opened 2019-10-18
Add autoMPO four-site test, fix MPOTerm  for > 3 Ops

### LINKED ISSUE #121: Possible bug in MPOTerm, proposal for fix
Hi,

I noticed that when I run 
`
sites = spinHalfSites(4);
ampo = AutoMPO();
add!(ampo, "S+", 1, "S-", 2, "S+", 3, "S-", 4);
H = MPO(ampo, sites)
`
the error  "BoundsError: attempt to access 4-element Array{SiteOp,1} at index [5]" is thrown.
It seems to me like this should not happen and that it's due to lines 85-87 in autompo.jl
`
for n=1:2:length(ops);
    vop[2+n] = SiteOp(ops[n],ops[n+1]);
end
`
as n grows to length(ops) while length(vop) = length(ops) / 2.
A possible fix might be:
`
for n = 1:div(length(ops),2);
    vop[2+n] = SiteOp(ops[2*n-1],ops[2*n]);
end
`

I'll make a pull request with this fix and a four site Ising test that catches the issue. If I've misunderstood something and this is all expected behaviour then please ignore!

MichaelSven:
This pull request is to address four site operators and should fix #121.

### DIFF
--- a/src/physics/autompo.jl
+++ b/src/physics/autompo.jl
@@ -82,8 +82,8 @@ function MPOTerm(c::Number,
   vop = OpTerm(undef,2+div(length(ops),2))
   vop[1] = SiteOp(op1,i1)
   vop[2] = SiteOp(op2,i2)
-  for n=1:2:length(ops)
-    vop[2+n] = SiteOp(ops[n],ops[n+1])
+  for n = 1:div(length(ops),2)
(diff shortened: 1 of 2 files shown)


emstoudenmire:
Appreciate you sending a fix as well - thanks!

PR #123, opened 2019-10-22
Add tests for similar, get function signatures to work

kshyatt:
The old signatures would cause errors if you tried `similar(delta(...))`.

### DIFF
--- a/src/tensor/diag.jl
+++ b/src/tensor/diag.jl
@@ -31,13 +31,13 @@ Base.eltype(::Type{<:Diag{T}}) where {T} = eltype(T)
 # Deal with uniform Diag conversion
 Base.convert(::Type{<:Diag{T}},D::Diag) where {T} = Diag{T}(data(D))
 
-Base.similar(D::Diag{T}) where {T} = Diag{T}(similar(data(D)))
+Base.similar(D::Diag{T}, ::Type{T}) where {T} = Diag{T}(similar(data(D)))
 
 # TODO: write in terms of ::Int, not inds
-Base.similar(D::Diag{T},inds) where {T} = Diag{T}(similar(data(D),minimum(dims(inds))))
-Base.similar(D::Type{<:NonuniformDiag{T}},inds) where {T} = Diag{T}(similar(T,length(inds)==0 ? 1 : minimum(dims(inds))))
+Base.similar(D::Diag{T},inds::IndexSet) where {T} = Diag{T}(similar(data(D),minimum(dims(inds))))
+Base.similar(D::Type{<:NonuniformDiag{T}},inds::IndexSet) where {T} = Diag{T}(similar(T,length(inds)==0 ? 1 : minimum(dims(inds))))
 
-Base.similar(D::UniformDiag{T}) where {T<:Number} = Diag{T}(zero(T))
+Base.similar(D::UniformDiag{T}, ::Type{T}) where {T<:Number} = Diag{T}(zero(T))
 Base.similar(::Type{<:UniformDiag{T}},inds) where {T<:Number} = Diag{T}(zero(T))
 
 # TODO: make this work for other storage besides Vector
--- a/src/tensor/diag.jl
+++ b/src/tensor/diag.jl
@@ -31,11 +31,12 @@ Base.eltype(::Type{<:Diag{T}}) where {T} = eltype(T)
 # Deal with uniform Diag conversion
 Base.convert(::Type{<:Diag{T}},D::Diag) where {T} = Diag{T}(data(D))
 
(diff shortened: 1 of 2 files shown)


src/tensor/diag.jl
```diff
@@ -31,13 +31,13 @@ Base.eltype(::Type{<:Diag{T}}) where {T} = eltype(T)
 # Deal with uniform Diag conversion
 Base.convert(::Type{<:Diag{T}},D::Diag) where {T} = Diag{T}(data(D))
 
-Base.similar(D::Diag{T}) where {T} = Diag{T}(similar(data(D)))
```
mtfishman: Don't we still want this version?

kshyatt: Do we? The tests passed without it...

mtfishman: Hmm interesting. I think we still want it anyway though. The goal is for the subtypes of `TensorStorage` to act like `AbstractVector` (I am thinking of defining `TensorStorage` that way). My guess is that this version failed for `delta` since `data(D)` is just a number, we can have a branch here for `NonuniformDiag` and `UniformDiag`.

kshyatt: OK, I'll add it back :)

src/tensor/diag.jl
```diff
@@ -31,13 +31,13 @@ Base.eltype(::Type{<:Diag{T}}) where {T} = eltype(T)
...
+Base.similar(D::Diag{T}, ::Type{T}) where {T} = Diag{T}(similar(data(D)))
 
 # TODO: write in terms of ::Int, not inds
-Base.similar(D::Diag{T},inds) where {T} = Diag{T}(similar(data(D),minimum(dims(inds))))
-Base.similar(D::Type{<:NonuniformDiag{T}},inds) where {T} = Diag{T}(similar(T,length(inds)==0 ? 1 : minimum(dims(inds))))
+Base.similar(D::Diag{T},inds::IndexSet) where {T} = Diag{T}(similar(data(D),minimum(dims(inds))))
```
mtfishman: Was there a reason for adding `IndexSet` here? The goal is actually to have all of the code in the `tensor/` directory independent of other `ITensor` types (like ITensor, Index, IndexSet, etc.). Eventually we want to turn it into a stand-alone library, so `inds` here is meant to be generic (either an IndexSet with no QNs, a `NTuple{N,Int}`, or anything else with `dims` defined on it).

kshyatt: Without it, the code tries to call this function for `::Diag{Float64}, Float64` as args and `Float64` has no method `dims`.

mtfishman: That sounds like a `UniformDiag` case, again there it sounds like we should have a dispatch branch in `similar` for `UniformDiag` and `NonuniformDiag`.

kshyatt: Although with the new method, it's not necessary anymore it seems, so I can remove that type requirement too.

kshyatt:
Tests passed - ok to go?

src/tensor/diag.jl
```diff
@@ -32,12 +32,13 @@ Base.eltype(::Type{<:Diag{T}}) where {T} = eltype(T)
 Base.convert(::Type{<:Diag{T}},D::Diag) where {T} = Diag{T}(data(D))
 
 Base.similar(D::Diag{T}) where {T} = Diag{T}(similar(data(D)))
+Base.similar(D::Diag{T}, ::Type{T}) where {T} = Diag{T}(similar(data(D)))
 
 # TODO: write in terms of ::Int, not inds
 Base.similar(D::Diag{T},inds) where {T} = Diag{T}(similar(data(D),minimum(dims(inds))))
 Base.similar(D::Type{<:NonuniformDiag{T}},inds) where {T} = Diag{T}(similar(T,length(inds)==0 ? 1 : minimum(dims(inds))))
 
-Base.similar(D::UniformDiag{T}) where {T<:Number} = Diag{T}(zero(T))
```
mtfishman: Just like above, it seems like we would still want this version. Was it causing a problem?

src/tensor/diag.jl
```diff
@@ -32,12 +32,13 @@ Base.eltype(::Type{<:Diag{T}}) where {T} = eltype(T)
 Base.convert(::Type{<:Diag{T}},D::Diag) where {T} = Diag{T}(data(D))
 
 Base.similar(D::Diag{T}) where {T} = Diag{T}(similar(data(D)))
+Base.similar(D::Diag{T}, ::Type{T}) where {T} = Diag{T}(similar(data(D)))
```
mtfishman: I am also kind of confused by this definition, it seems strange to have a similar method only defined on making the same type (i.e. the second input doesn't seem to do anything). What is this for?

Issue #124, opened 2019-10-27
bug in `mul!(w::ITensor, a, v::ITensor)` when `w= similar(v)` and v is complex


orialb:
The current implementation of `mul!(w,a,v)` returns `add!(w,0,v,a)` which is a problem when `w` has `NaN` values, as it means that the result of `mul!` will also have NaN values in it. 
This situation can occur for example when `v` is an ITensor with complex storage and `w = similar(v)` (see code below). 
Due to this behavior it is currently not possible to use KrylovKit functions with ITensors (which is how I found out about this problem).

A simple fix would be to just use `apply!` directly in the definition of `mul!` instead of relying on `add!`. I can make a quick PR if you approve this solution.

here is a minimal code to see the problem (at least on my machine):
```julia
using ITensors, LinearAlgebra
i = Index(1000,"i")
v = randomITensor(ComplexF64,i)
w = similar(v)
mul!(w,0.1,v)
any(x->isnan(x),data(store(w)))
```

orialb:
> A simple fix would be to just use apply! directly in the definition of mul! instead of relying on add!. I can make a quick PR if you approve this solution.

I just went ahead and created a PR which can be used in case this solution is ok (I can also confirm that this fixes the problems I'm seeing with KrylovKit).

mtfishman:
Interesting, thanks for the catch and for fixing it! Seems to make sense to do it the way you propose.

PR #125, opened 2019-10-27
fix bug in mul!

orialb:
This small PR fixes https://github.com/ITensor/ITensors.jl/issues/124 by defining `mul!` directly in terms of `apply!` instead of relying on `add!`.

I also added a test to verify that `mul!(w,a,v)` works when `w` contains NaNs.


emstoudenmire:
Looks good - thanks Ori

PR #134, opened 2019-11-04
isless for complex coefficient in autompo, with test

christopherdavidwhite:
The coef of each MPOterm is complex, but before this commit, line 61 was
    coef(t1) < coef(t2)
So if execution ever reached this line (which it did not in any of the tests, AFAICT), this line would fail: there is no method `isless` for those `Complex{Float64}` variables.

This commit replaces that line with a check for a "lexicographic ordering", and adds a test that should trigger the bug in the old code.

Incidentally, none of this code will check for or gracefully handle terms with coef NaN.

### DIFF
--- a/src/physics/autompo.jl
+++ b/src/physics/autompo.jl
@@ -58,7 +58,10 @@ end
 
 function Base.isless(t1::MPOTerm,t2::MPOTerm)::Bool
   if !isapprox(coef(t1),coef(t2))
-    return coef(t1) < coef(t2)
+    ct1 = coef(t1)
+    ct2 = coef(t2)
+    #"lexicographic" ordering on  complex numbers
+    return real(ct1) < real(ct2) || (real(ct1) == real(ct2) && imag(ct1) < imag(ct2))
   end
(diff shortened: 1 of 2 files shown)


christopherdavidwhite:
Should've mentioned: this is all in autompo.jl

emstoudenmire:
Hi Christopher, did you mean to include the commits with `real_if_close` and other changes into this pull request? Is that feature related to the isless and AutoMPO fix? Thanks -

christopherdavidwhite:
Nope. Those are for the next pull request (at the moment `sample` can't handle complex MPSs, so I went and fixed that.)

I knew if I couldn't have two open at the same time, but I didn't realize that GitHub would automatically add the new stuff to the old pull request. Sorry about that

christopherdavidwhite:
Did a reset and push --force. Should be down to only that one commit now.

emstoudenmire:
Thanks for this Christopher. Good catch. I will make a small change to the unit test after pulling to make it more robust to internal changes to AutoMPO, but appreciate the PR and the test.

PR #137, opened 2019-11-12
Fix combiner and add tests

kshyatt:
Fixes this unpleasantness:
```
julia> using ITensors

julia> i = Index(2, "i"); j = Index(3, "j"); k = Index(4, "k"); a = Index(5, "a");

julia> A = randomITensor(i, j, k, a);

julia> cmb, ci = combiner(i, j, k, tags="CMB");

julia> Ac = A*cmb;

julia> cmb*Ac == Ac*cmb
false
```
and adds tests to keep it from happening again. The `combiner` tests pass, other tests are broken on master, at least on rusty...

### DIFF
--- a/src/indexset.jl
+++ b/src/indexset.jl
@@ -58,11 +58,6 @@ IndexSet(inds::IndexSet,i::Index) = IndexSet(inds...,i)
 IndexSet(i::Index,inds::IndexSet) = IndexSet(i,inds...)
 IndexSet(is1::IndexSet,is2::IndexSet) = IndexSet(is1...,is2...)
 
-function IndexSet(inds::Vector{Index})
-  N = length(inds)
-  return IndexSet{N}(ntuple(i->inds[i],N))
-end
-
 # This is used in type promotion in the Tensor contraction code
 Base.promote_rule(::Type{<:IndexSet},::Type{Val{N}}) where {N} = IndexSet{N}
 
--- a/src/indexset.jl
+++ b/src/indexset.jl
@@ -58,6 +58,11 @@ IndexSet(inds::IndexSet,i::Index) = IndexSet(inds...,i)
 IndexSet(i::Index,inds::IndexSet) = IndexSet(i,inds...)
 IndexSet(is1::IndexSet,is2::IndexSet) = IndexSet(is1...,is2...)
 
+function IndexSet(inds::Vector{Index})
+  N = length(inds)
+  return IndexSet{N}(ntuple(i->inds[i],N))
+end
+
 # This is used in type promotion in the Tensor contraction code
 Base.promote_rule(::Type{<:IndexSet},::Type{Val{N}}) where {N} = IndexSet{N}
 
--- a/src/tensor/combiner.jl
+++ b/src/tensor/combiner.jl
@@ -49,16 +49,40 @@ function contract!!(R::Tensor{<:Number,NR},
     # TODO: handle the case where inds(R) and inds(T1)
     # are not ordered the same?
     # Could just use a permutedims...
-    return Tensor(store(T2),inds(R))
+    Alabels,Blabels = compute_contraction_labels(inds(T2),inds(T1))
+    final_labels    = contract_labels(Blabels, Alabels)
+    final_labels_n  = contract_labels(labelsT1,labelsT2)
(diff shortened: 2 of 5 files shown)


mtfishman:
It looks like some of the tests are using the function `isActive`, but it was renamed to `isactive`.

mtfishman:
Also `isFermionic` -> `isfermionic` and `makeTagType("MySite")` -> `TagType"MySite"`. Not sure how those passed before.

kshyatt:
```
test_phys_site_types.jl: Error During Test at /mnt/home/khyatt/.julia/dev/ITensors/test/runtests.jl:4
  LoadError: LoadError: UndefVarError: @TagType_str not defined
  Stacktrace:
   [1] top-level scope
   [2] include at ./boot.jl:328 [inlined]
   [3] include_relative(::Module, ::String) at ./loading.jl:1105
   [4] include(::Module, ::String) at ./Base.jl:31
   [5] include(::String) at ./client.jl:432
   [6] top-level scope at /mnt/home/khyatt/.julia/dev/ITensors/test/runtests.jl:28
   [7] top-level scope at /mnt/home/khyatt/software/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1180
   [8] top-level scope at /mnt/home/khyatt/.julia/dev/ITensors/test/runtests.jl:4
   [9] top-level scope at /mnt/home/khyatt/software/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
   [10] top-level scope at /mnt/home/khyatt/.julia/dev/ITensors/test/runtests.jl:4
   [11] include at ./boot.jl:328 [inlined]
   [12] include_relative(::Module, ::String) at ./loading.jl:1105
   [13] include(::Module, ::String) at ./Base.jl:31
   [14] include(::String) at ./client.jl:432
   [15] top-level scope at none:6
   [16] eval(::Module, ::Any) at ./boot.jl:330
   [17] exec_options(::Base.JLOptions) at ./client.jl:271
   [18] _start() at ./client.jl:468
  in expression starting at /mnt/home/khyatt/.julia/dev/ITensors/test/test_phys_site_types.jl:160
  in expression starting at /mnt/home/khyatt/.julia/dev/ITensors/test/test_phys_site_types.jl:160

```

mtfishman:
Hmm, `TagType_str` is defined here: https://github.com/ITensor/ITensors.jl/blob/25175fd808611282c68dad040207adf2fcf93545/src/physics/tag_types.jl#L20

kshyatt:
It's not exported, that was the problem. I added some more tests for this failing SVD issue which is grinding my gears 😡

kshyatt:
OK, tests pass, including SVD, but this is probably *not* what @mtfishman wants this code to look like in the end... Feel free to push better versions to this branch, as long as tests pass

emstoudenmire:
Looks good Katie - thanks for these fixes.

Interesting about using combiners for index replacement, but actually I would argue that creating a combiner with just one index should be an error (raise an exception). This is because it's out of the scope of what combiners are for and also one can use delta tensors to replace indices.

mtfishman:
Actually for quantum numbers, the single index combiner is quite non-trivial and very useful. In the C++ version, a single index combiner makes a new index with the blocks merged, which Peter has used to speed up his VUMPS code (since the prescribed way of increasing the bond dimension in VUMPS can introduce many small QN blocks). Perhaps this should be a different functionality, but apparently it was already working once the general combiner function was working.

kshyatt:
Are we ok to merge? Feel free to squash this also

mtfishman:
I am fine with it. I haven't looked at it very closely and will probably go over it later, but I think we just want this code working right now. The only thing is that I would prefer if the `IndexSet(::Vector)` constructor was removed (since I don't think we want to encourage using that, since it is not type stable). Beyond that, go ahead and merge when you think it is ready.

emstoudenmire:
Great I'll remove that and then leave it be. I was just trying to use tuples all the way through, but unfortunately we're still missing some tools to make that clean (without essentially writing those tools into the combiner code itself). But I know you are planning to work on that very thing down the line so this code will get nicer over time.

kshyatt:
Thanks to you both for the improvements on this!

Issue #104, opened 2019-09-20
Return truncation error and singular values from replaceBond and factorize


orialb:
Currently `replaceBond`, `factorize`, `svd` and `eigen` don't return the truncation error which we would like to track in many cases. It is also necessary in some cases to have the actual singular values from `replaceBond` and `factorize`.  
It would be simple to just propagate this through all these functions, but might make the code a little more cumbersome in cases when one doesn't need this information (although not much more).
Another option is to have a factorization object returned from `factorize`, `svd` and `eigen` functions which will contain all the information.  
A third option is to return this information from `factorize`, `svd` and `eigen` only when it is asked for by the caller with a keyword argument. 

Any thoughts?

mtfishman:
It seems like a good idea. We already pass the newly introduced indices from those functions, which you can grab or not just by including them in your output list:

```julia
U,S,V = svd(A,(i,j))
U,S,V,u,v = svd(A,(i,j))
```

We can just add the truncation error as another output after the indices, it wouldn't affect users who don't want it. So people could get it like this:

```julia
U,S,V,u,v,truncerr = svd(A,(i,j))
```

The list could get a bit long, but I think that is ok (and we can always pack things after `U,S,V` into an object if we end up outputting a lot of stuff).

emstoudenmire:
I second the idea of putting things into an object beyond a certain point. A good question would be whether the `u` and `v` inds should be included in that object too.

In C++ ITensor there is a Spectrum object returned by the SVD which includes not only the truncerr, but all of the singular values (these are already in memory anyway, so no overhead incurred by returning a reference to them). It could also return the number of singular values kept and discarded (these can be recomputed but it's just convenient to have them at hand).

orialb:
Yes, it would be good to return also the singular values so one can track entanglement entropy during sweeps.

A `Spectrum` object would be useful for that, since, depending on the internal factorization used, factorize would return either the singular values, the square of the singular values or the square root of the singular values. Having a dedicated object will allow to store the information about the type of the returned spectrum, so user could just call `entropy(spec)` (for example) without having to think whether they need to square/sqrt the spectrum before. But this might be also overcomplicating things. 

Otherwise we could just transform the spectrum such as to always return singular values (or square of them if that's better), but I think this might incur some extra overhead depending on the internal factorization used.

emstoudenmire:
Thanks, these are some good considerations.

[...]

PR #139, opened 2019-11-13
return spectrum and truncation error from factorization functions

orialb:
This PR implements a `Spectrum` object which is used to return the density matrix eigenvalue spectrum and truncation error from  `svd`, `eigenHermitian`, `factorize` and `replaceBond!` (similar to the C++ version) as discussed in https://github.com/ITensor/ITensors.jl/issues/104.

I also added tracking of the maximal truncation error in each sweep to the built-in DMRGObserver.

In https://github.com/ITensor/ITensors.jl/issues/104 Miles suggested to also add the number of discarded eigenvalues (which is otherwise not accessible from the spectrum because it is already truncated in place in `truncate!`), but I saw this is not available in the C++ version. Is this actually something that someone would like to track, should I add it?

### DIFF
--- a/test/test_decomp.jl
+++ b/test/test_decomp.jl
@@ -21,12 +21,12 @@ A = randomITensor(i,j)
   S ./= norm(S)
   A = ITensor(U*diagm(0=>S)*V', i,j)
 
-  _,_,_,_,_,spec = svd(A,i)
+  spec = svd(A,i).spec
 
   @test eigs(spec) ≈ S .^2
-  @test truncErr(spec) == 0.0
+  @test truncerror(spec) == 0.0
 
-  _,_,_,_,_,spec = svd(A,i; maxdim=length(S)-3)
-  @test truncErr(spec) ≈ sum(S[end-2:end].^2)
+  spec = svd(A,i; maxdim=length(S)-3).spec
+  @test truncerror(spec) ≈ sum(S[end-2:end].^2)
 
 end
--- a/test/test_decomp.jl
+++ b/test/test_decomp.jl
@@ -11,3 +11,22 @@ i = Index(2,"i")
 j = Index(2,"j")
 A = randomITensor(i,j)
 @test_throws ArgumentError factorize(A, i, dir="fakedir")
+
+
+@testset "Spectrum" begin
+  i = Index(100,"i")
+  j = Index(100,"j")
+
+  U,S,V = svd(rand(100,100))
+  S ./= norm(S)
+  A = ITensor(U*diagm(0=>S)*V', i,j)
+
+  _,_,_,_,_,spec = svd(A,i)
+
+  @test eigs(spec) ≈ S .^2
+  @test truncErr(spec) == 0.0
+
+  _,_,_,_,_,spec = svd(A,i; maxdim=length(S)-3)
+  @test truncErr(spec) ≈ sum(S[end-2:end].^2)
+
+end
--- a/examples/ctmrg/ctmrg.jl
+++ b/examples/ctmrg/ctmrg.jl
@@ -145,7 +145,7 @@ function leftright_move!(T,(Clu,Cru,Cld,Crd),(Al,Ar,Au,Ad);dir="left",maxdim=5)
       elseif dir=="up" || dir=="down"
         utags = "$dir,link,x=$iy,y=$ixp"
       end
-      U,S,Vh,u,v = svd(ρ, (li,si); utags=utags, vtags="tmp", maxdim=maxdim, cutoff=0.0)
+      U,S,Vh,spec,u,v = svd(ρ, (li,si); utags=utags, vtags="tmp", maxdim=maxdim, cutoff=0.0)
       V = dag(Vh)
       U *= δ(u,v)
       invsqrtS = S
--- a/src/decomp.jl
+++ b/src/decomp.jl
@@ -68,7 +68,7 @@ function svd(A::ITensor,
   Lis = commoninds(inds(A),IndexSet(Linds...))
   Ris = uniqueinds(inds(A),Lis)
   Lpos,Rpos = getperms(inds(A),Lis,Ris)
-  UT,ST,VT = svd(tensor(A),Lpos,Rpos;kwargs...)
+  UT,ST,VT,spec = svd(tensor(A),Lpos,Rpos;kwargs...)
   U,S,V = ITensor(UT),ITensor(ST),ITensor(VT)
   u₀ = commonindex(U,S)
   v₀ = commonindex(S,V)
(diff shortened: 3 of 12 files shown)


orialb:
I realized that it would probably make more sense to have `svd` and `eigenHermitian` just return a truncation error and move the `Spectrum` functionality one level up to `factorize`, since the signular/eigen values spectrum is already returned from `svd` and `eigenHermitian` in form of a `Diag` ITensor.

The reason I initially put it at the lower level is that it was suggested by Miles that `Spectrum` should always contain the density matrix spectrum, and the squaring of singular values in the SVD case is already done in `svd` before truncating. 
If the `Spectrum` object is constructed by `factorize` then we have to square the singular values there again. This introduces a small overhead which should be completely negligible, except maybe for very small matrices. 

Unless there are other thoughts about it I will go ahead and update the PR to move the functionality one level up to `factorize`.

test/test_decomp.jl
```diff
@@ -11,3 +11,22 @@ i = Index(2,"i")
...
+
+@testset "Spectrum" begin
+  i = Index(100,"i")
+  j = Index(100,"j")
+
+  U,S,V = svd(rand(100,100))
+  S ./= norm(S)
+  A = ITensor(U*diagm(0=>S)*V', i,j)
+
+  _,_,_,_,_,spec = svd(A,i)
```
mtfishman: This is a bit unfortunate. Maybe we want svd to return a struct, and then you can get the spectrum with `spec = svd(A,i).spec`? I think you can still use the syntax `U,S,V = svd(A,i)` if it returns a struct, for example Julia does this:

```julia
julia> A = randn(2,2)
2×2 Array{Float64,2}:
 2.45284  -0.276589
 2.73764   0.779181

julia> F = svd(A)
SVD{Float64,Float64,Array{Float64,2}}([-0.6508728185978933 -0.7591867846652984; -0.7591867846652984 0.6508728185978933], [3.6978397331941464, 0.72161329507932], [-0.9937883641647395 -0.11128650974296625; -0.11128650974296625 0.9937883641647395])

julia> U,S,V = F;

julia> U == F.U
true
```

emstoudenmire: Yes I agree we want to keep the `U,S,V = svd(....)` syntax working if possible.

Does having the above behavior of unpacking a custom struct require iteration to be defined for that struct?

mtfishman: I would guess that is how you define that behavior.

orialb: This is how it's done for the SVD factorization in LinearAlgebra
```julia
# iteration for destructuring into components
Base.iterate(S::SVD) = (S.U, Val(:S))
Base.iterate(S::SVD, ::Val{:S}) = (S.S, Val(:V))
Base.iterate(S::SVD, ::Val{:V}) = (S.V, Val(:done))
Base.iterate(S::SVD, ::Val{:done}) = nothing
```
So I could implement similar factorization structs for `svd` and `eigenHermitian` which 
will also contain the truncation information (could be called `TruncSVD`,`TruncEigen` or something like this).

As a side note - would you agree with my comment above that `svd` should actually return only truncation error (in addition to U,S,V etc.) and not the whole `Spectrum` object?

orialb: Also, I've noticed that while `svd(A::ITensor,...)` returns `U,S,V,u,v,spec`, `u` and `v` seem to never be used in the code base (it is always called as `U,S,V = svd(...)`). 
So maybe the order of return parameters (with or without a factorization struct) should be changed such that `U,S,V,spec = svd(...)`  works?

Actually is it necessary to return `u,v` from `svd` ? I find the logic cleaner if the task of finding the common indices is done by the caller (which seems to occur anyway in practice).

mtfishman: We can return the indices last (or not return them, but I think there is not much downside in returning them). I just thought it was neat to return them since it is much harder to do it in C++.

I don't have too much of an opinion on what object should be output, but Spectrum might give some other useful information (like for BlockSparse tensors, maybe it can output the truncation per block?). Maybe Miles has more to say on that.

orialb:
I implemented the `TruncSVD` and `TruncEigen` factorization types discussed in the code review and also changed the order of return values such that common indices are returned after the spectrum (to allow `U,S,V,spec =svd(...)`)

orialb:
p.s I wanted to rebase the PR on top of the latest master changes instead of merging master but I didn't know if it preserves the code review comments on the rebased commits. Do you know if that would have preserved the comments?

mtfishman:
Great, thanks for adding that! I think the comments get preserved, and github just marks that they relate to old code if the lines of code change.

orialb:
Slightly off-topic, but looking at the code for `factorize` I am confused about something- shouldn't the singular-values be re-scaled after truncation such that the wave-function stays normalized? Does it happen at some point and I missed it? Or is it actually not necessary?

mtfishman:
`factorize` is meant to be a general purpose matrix factorization operation, so wouldn't really be handling any normalization. In DMRG, I believe that the `davidson` routine normalizes the wavefunction (if it is not done explicitly somewhere else).

orialb:
makes sense about `factorize`, thanks.
I find the fact that  `replacebond!` doesn't keep the wavefunction normalized by default surprising though, but maybe I don't fully understand the meaning of `replacebond!` as well :)

orialb:
Just wanted to clarify that this is ready from my side for merge.
I am interested in implementing a quantum trajectories computation based on ITensors  and this feature would be very helpful as one needs to track the decay of the wave-function norm after each gate application. 
Of course if you are busy at the moment with higher priority work I could just use my fork in the meantime, so no worries.

emstoudenmire:
Hi Ori, this is really great thanks. I’ll go ahead and pull because it’s really good quality code!

We may change a few things but probably very minor ones. One change is to return the entire spectrum, not the truncated one, in the Spectrum object. 

One other change I’m toying around with would be to just embed the spectrum information into the TruncateSVD and TruncateEigen objects themselves. Functions like entropy (which is great to have) could then operate on either one. But that’s just a possible change; probably the way you have it set up is the best.

orialb:
thanks Miles!

> One other change I’m toying around with would be to just embed the spectrum information into the TruncateSVD and TruncateEigen objects themselves. Functions like entropy (which is great to have) could then operate on either one. But that’s just a possible change; probably the way you have it set up is the best.

But I guess you would probably still want to return the `Spectrum` object from `replacebond!` ? Or would you just have `replacebond!` return the whole `TruncateSVD`object to propagate the truncation error and spectrum information?

emstoudenmire:
Yes, definitely let’s keep the Spectrum return from replacebond!. So that’s a good argument right there for why it’s good to have a distinct Spectrum object, separate from TruncateSVD etc.

Also, to answer your question about what replacebond! does, the idea is that you give it a tensor B which has the same index structure as contracting psi[i]*psi[i+1] for some MPS psi, then it decomposes it into tensors compatible with the MPS structure. There are many applications of this function outside of DMRG, time evolution, or even physics, so we treat it as a lower-level (maybe mid-level) function and leave it up to the calling code to decide whether B should be normalized or not. For example, one could imagine a case where one wants to compute H|psi> and decompose it back into an MPS. Then this new MPS won’t be normalized to 1, but one might still want to use replacebond! throughout the algorithm to compute H|psi>

Thanks again for the pull request -

PR #140, opened 2019-11-15
sample for complex MPSs; extra goody real_if_close

christopherdavidwhite:


src/itensor.jl
```diff
@@ -695,3 +696,44 @@ function Base.read(io::IO,::Type{ITensor};kwargs...)
...
+# and typically IME it already has real tensors
+# ---but only up to 1e-13 or something)
+# (CDW)
+"""
+    real_if_close(A :: ITensor; tol = 1e-10)
+
+Return the real part of an itensor `A` if all the elements of A are nearly real, i.e. `all((x .|> imag .|> abs) < tol)`; errors otherwise.
+
+"""
+real_if_close(A :: ITensor; tol = 1e-10) = ITensor( real_if_close.(A |> array, tol=tol)  , inds(A))
```
mtfishman: In general we don't want to convert to an Array, since that will be bad for sparse storage types (my vision for the `array` function was that it turned an ITensor into it's closest dense array version, so for Diag and BlockSparse it would fill with zeros. I call it `array` to specify that if the ITensor is `Dense` already, the storage won't get copied, which is a common Julia convention).

I can see the point of this function, since you might want to take the real part only under certain conditions. Is there a "better" norm for determining if the complex part should be thrown away, as opposed to an element-wise check? What about `norm(imag(A)) < 1e-10`?

src/mps/mps.jl
```diff
@@ -257,7 +257,7 @@ function sample(m::MPS)
       projn = ITensor(s)
       projn[s[n]] = 1.0
       An = A*projn
-      pn = scalar(dag(An)*An)
+      pn = scalar(dag(An)*An) |> real_if_close
```
mtfishman: Did you see a problem occur here if you just use `real(scalar(dag(An)*An))`? That result should definitely be real (up to floating point precision). It seems like checking for `imag(pn) < tol` should be more like a debug check.

src/itensor.jl
mtfishman: Also, your current implementation points to the fact that we should have broadcasting defined for ITensors, which is definitely something we plan to do. Then you could just do `real_if_close.(A,tol=tol)` for the elementwise version.

mtfishman: Actually, I forgot that I already defined broadcast on the Tensor type, so you may be able to just do:
```julia
ITensor(real_if_close.(tensor(A), tol=tol))
```
(assuming we end up going with the elementwise version).

src/mps/mps.jl
christopherdavidwhite: No, that'd work; it just makes me nervous to give up the check. If some value fails to make sense I want the program to choke early, rather than silently continuing and (maybe) giving nonsensical results. Especially for something like this, where the symptom's going to be that some average just ... doesn't quite look right; tracking that down to find some subtle bug like accidentally producing an imaginary part sounds miserable.

On the other hand, I can't imagine a case where `scalar(dag(An)*An))` _isn't_ real (up to numerical precision), provided `dag` is working as it should.

On the gripping hand, my imagination is limited...

src/itensor.jl
christopherdavidwhite: Aha! This is much better. This seems like a decent, compact idiom; honestly, I'd be happy to just let the user do this any time she needs the functionality.

mtfishman: Also could we add a branch `isreal(A) && return A` so this is a no-op for real ITensors? If we don't have an `isreal` function for ITensors, we should add it.

christopherdavidwhite: Good thinking; my impulse is to do it with `eltype(A) <: Real && return A`. (If the tensors were parametrized over type we could do it at compile time with

    real_if_close(A :: ITensor{<:Real, N}, tol) = A

. They're not, are they? I didn't just miss that?)

mtfishman: They are not, so you will have to do it at runtime with a function (either `isreal(A)` or `eltype(A) isa Real`, but really we would just define `isreal(A) = eltype(A) isa Real`). 

The reason is that ITensors are supposed to be flexible enough to change their data type "in-place", which would not be possible if they are parametrized by the element type. This allows for more generic code that handles in-place operations more automatically. For example, `scale!(A,3+im)` just works for an ITensor, whether the ITensor has complex or real storage (in the case of real storage, a new complex storage is made and replaces the real storage). It adds a bit of runtime overhead, but as long as performance critical code is within a type-stable function (the function barrier technique), we have seen that the overhead is minimal. That is actually part of the goal of the `Tensor` design: `ITensor`s pass their storage to `Tensor` functions, which are then type stable and performant, so the only overhead is an extra runtime dispatch to choose which `Tensor` function to dispatch on.

mtfishman:
Hey Christopher,

I was held up about pulling this since I wasn't totally convinced that the `real_if_close` function was appropriate for including in ITensors.jl (I haven't come across cases so far where I felt like that kind of functionality was necessary, at least within the library). Would you mind removing it from this pull request and just using `real` within the sample function? I think that use case is a place where proper testing would catch any issues. `real_if_close` could be something that you define for yourself in your own use cases, and we can keep it in mind as functionality to add when the need arises.

-Matt

christopherdavidwhite:
Just made a new commit nixing real_if_close.

mtfishman:
Thanks! Again, we can consider putting back in the future (we could just add it as an unexported function if it turns out to be useful).

PR #142, opened 2019-11-15
Updated README to use latest function names

emstoudenmire:
Also added test_readme.jl to make sure README code is correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,35 +23,37 @@ Development of ITensor is supported by the Flatiron Institute, a division of the
 
 ### Basic Overview
 
-Here is a basic intro overview of making 
-ITensors, setting some elements, contracting, and adding
-ITensors. See further examples below for detailed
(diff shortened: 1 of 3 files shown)


mtfishman:
Could we have `test_readme.jl` automatically pull in the code from the README file?

emstoudenmire:
It's a good idea, I'm just not sure of how to do it. Best I can think of would be code in test_readme.jl which opens README, parses it to get just the code blocks, then uses Julia's `eval` to run these.

mtfishman:
Yeah that sounds reasonable (or instead of `eval`, just generate a new file with just the code in it and `include` that file at the end of `test_readme.jl`). I guess the tricky part may be automatically inserting test statements.

mtfishman:
Although it looks like we could just directly replace `@show` with `@test`.

PR #143, opened 2019-11-15
Compatibility for GPU/PEPS stuff

kshyatt:
Bumped the tensor warn order because in the PEPS code we often have tensors of rank 12.

The change to randomizing phi lets the GPU code use the same Davidson method. Plussers has the change it does because it needs to be overloaded by the GPU code to avoid a very slow transfer of the plusser to the device.

Changed the sites mechanism bc the old version was giving me errors when I did `multMPO` with MPOs where the site indices aren't primed versions of each other.

### DIFF
--- a/src/ITensors.jl
+++ b/src/ITensors.jl
@@ -51,7 +51,7 @@ import Base.adjoint,
 #####################################
 # Global Variables
 #
-const warnTensorOrder = 10
+const warnTensorOrder = 14
 const GLOBAL_TIMER = TimerOutput()
 
 #####################################
--- a/src/iterativesolvers.jl
+++ b/src/iterativesolvers.jl
@@ -77,7 +77,9 @@ function davidson(A,
 
   nrm = norm(phi)
   if nrm < 1E-18 
-    phi = randomITensor(inds(phi))
+    phi_ = similar(phi)
+    randn!(phi_)
+    phi = phi_ 
     nrm = norm(phi)
   end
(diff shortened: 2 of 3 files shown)


mtfishman:
Nice to see that those are pretty small changes. I was thinking that the entire `plussers` code should just be Tensor level functions anyway (it is really doing a partial direct sum), that would avoid having to rely on dispatch at this level. It would be nice to have all of the tag-independent index logic in the MPS and MPO code like we have in C++, this is telling me that was definitely worth the effort. So I think both of these parts we can do nicer and more generally, but we can go over it later if this is important to have for your code.

kshyatt:
Being able to specialize this stuff on the storage type is basically all I need (and not having `davidson` assume CPU storage of `phi`).

mtfishman:
For other storage types like BlockSparse this will need to be entirely different code anyway, so we really should have it as a Tensor function anyway.

kshyatt:
Aaaaaaaaaaah there's something up with a `multMPO` test... anyone want to dive in?

mtfishman:
What is the advantage of the new version over the old version? They both seem to be using the "Site" tag anyway.

kshyatt:
@mtfishman for whatever reason (I didn't dive too deeply into this), the old version would end up with all the `Site` indices on the last tensor, and this new version would put them in the correct place. I'll come up with a MWE.

kshyatt:
Superseded by #144

PR #144, opened 2019-11-19
Updated compatibility 

kshyatt:
The old version of `multMPO` didn't work if you had a test case like the one I added. Now it should for both the original style and this new one.


mtfishman:
Does this entirely replace  #143?

kshyatt:
Yes.

mtfishman:
Ok, I will merge this and we can open up some issues for cleaning up plussers and multMPO.

PR #150, opened 2019-12-02
noise term for dmrg

christopherdavidwhite:
Critical bits are in `src/mps/projmpo.jl: noisetensor()` and `src/decomp.jl: deltaRho(), _factorize_from_{left,right}_eigen()`. Have a look and make sure I understood the paper/the old implementation/etc. correctly


src/decomp.jl
```diff
@@ -142,13 +142,21 @@ function _factorize_from_right_svd(A::ITensor,
   return FU,FV,spec,commonindex(FU,FV)
 end
 
+# Make the perturbation to the density matrix used in "noise term" DMRG
+# This assumes that A comes in with no primes
+# If it doesn't, I expect A² += drho later to fail
+function deltaRho(A :: ITensor, nt :: ITensor, is)
```
mtfishman: Could we rename this to `deltarho`? We are trying to follow the Julia convention of all lower case names.

christopherdavidwhite: I wondered what the story was with `lproj` etc. Should be fixed in the latest.

src/mps/sweeps.jl
```diff
@@ -60,6 +63,20 @@ function cutoff!(sw::Sweeps,cutoffs::Float64...)::Nothing
...
+function noise!(sw::Sweeps,noises::Float64...)::Nothing
+  Nm = length(noises)
+  N = min(nsweep(sw),Nm)
+  for i=1:N
+    sw.noise[i] = noises[i]
+  end
+  for i=Nm+1:nsweep(sw)
+    # this is different behavior from cutoff!, min/maxdim!, etc.,
+    # but I (CDW) think it's the right thing:
+    # if the noise isn't specified, it should be 0
```
emstoudenmire, line 75: Good to think about this. However, first, if the user wants to set the noise on all remaining sweeps to zero, they can just append an explicit 0.0 to the argument list. Second, this is the behavior in the C++ version so it's also good to be consistent with that unless there's a strong reason otherwise. Finally, having the same behavior as the other functions like `maxdim!` is less surprising to the user. So I will change this to act like the others before merging.

emstoudenmire:
Thanks for your patience on this & for the contribution. I'm working on merging this before the v0.1 release.

PR #151, opened 2019-12-05
Add tree algo for sum and tests (failing atm)

kshyatt:
I added the vectorized method from C++ (it's nice to have imo) also. And some tests for higher dim.


kshyatt:
My original guess for wrong tests was bad, but now it seems that truncating `M` causes big discrepancies? Is that supposed to happen? Am I just crazy jet lagged? 🤔

Issue #152, opened 2019-12-05
Mistake in Spin 1/2 operator Sy


jtschneider:
https://github.com/ITensor/ITensors.jl/blob/72b0bd9b2c9c2f37d311a4e3958f20d08ad7c069/src/physics/site_types/spinhalf.jl#L39-L45

There is a non-canonical definition of the Pauli matrix `\sigma_y`; it has a sign error with respect to the canonical definition [1,2]:

<img width="215" alt="image" src="https://user-images.githubusercontent.com/9451057/70255514-3e366f00-1787-11ea-8932-de90bc88f928.png">

Obviously, this is only noticeable in operators which are odd in `\sigma^y` and thus this might be a mild bug. However, I wish to argue that we should fix the sign in order to not throw the commutation relations 

<img width="128" alt="image" src="https://user-images.githubusercontent.com/9451057/70255041-67a2cb00-1786-11ea-8bd1-88a632c17f9c.png">

over board as they indeed **define** the Pauli matrices in a basis-independent way. NB: Anti-commutation relations are still valid under `\sigma^y -> -\sigma^y` (even in `\sigma^y`).
Currently, the commutation relations **do not hold**.

Moreover, `\sigma^+` and `\sigma^-` would have to change role as

<img width="161" alt="image" src="https://user-images.githubusercontent.com/9451057/70255167-9e78e100-1786-11ea-8063-13580e17fe43.png">

https://github.com/ITensor/ITensors.jl/blob/72b0bd9b2c9c2f37d311a4e3958f20d08ad7c069/src/physics/site_types/spinhalf.jl#L32-L35

Cheers,
Jan

[1] W. Pauli, „Zur Quantenmechanik des magnetischen Elektrons", Zeitschrift für Physik, Bd. 43, 1927, S. 601, DOI: [10.1007/BF01397326](https://link.springer.com/article/10.1007%2FBF01397326)
[2] [Pauli matrices, Wikipedia (en)](https://en.wikipedia.org/wiki/Pauli_matrices), visited: 05/12/2019 at 17:00UTC+1

jtschneider:
I propose a fix with Pull Request https://github.com/ITensor/ITensors.jl/pull/153 .

emstoudenmire:
Thanks Jan, however I believe you are reading the matrices in a transposed way relative to how the code operates. With the ITensor system, the line of code:
`Op[Dn, UpP] = -0.5*im`
has exactly the same effect as the line of code
[...]

PR #153, opened 2019-12-05
fixed sign error in pauli matrix sigma^y

jtschneider:
See issue https://github.com/ITensor/ITensors.jl/issues/152


emstoudenmire:
Please see my comment on issue #152

PR #155, opened 2019-12-06
WIP: Initial BlockSparse code

mtfishman:
This includes a start to the BlockSparseTensor type. So far, BlockSparseTensors support a variety of constructors, printing, converting to DenseTensor, getindex, setindex!, and permutedims.

This also includes a variety of other large code reorganizations, so it would be nice to include this soon before other work starts clashing too much with it. The major change is separating all of the Tensor code and tests into separate directories and wrapping all of it in a Tensors module in preparation for splitting it off into a separate repository.

This should not be breaking at all besides moving around files and changing file names, so I will work on fixing any conflicts and merge once that is done.


emstoudenmire:
Looks good Matt. I think it’s good to go. I really like the BlockSparse interface. One small nitpick is that the name nnz is a little cryptic. Maybe nzlength instead or similar?

Nice to see how much the tupletools cleans up the code and should be very fast too.

mtfishman:
I based it off of Julia's SparseArrays interface: https://docs.julialang.org/en/v1/stdlib/SparseArrays/#SparseArrays.nnz

I also found it a bit cryptic, we could always add a different name that is more verbose, but I think it is good to support `nnz` for those who are familiar with the Julia interface.

Issue #154, opened 2019-12-05
Printing and showing ITensor in non-transposed format


jtschneider:
In view of closed issue https://github.com/ITensor/ITensors.jl/issues/152, I propose a change in the printing and passing mechanism of an ITensor.
The issue is briefly summarised with the following code example:

```julia
julia> using ITensors

julia> sites = ITensors.siteinds("S=1/2", 1)
1-element Array{Index,1}:
 (2|id=90|S=1/2,Site,n=1)

julia> Sy = op(sites[1], "Sy");

julia> @show Sy
Sy = ITensor ord=2 (2|id=90|S=1/2,Site,n=1) (2|id=90|S=1/2,Site,n=1)'
Dense{Complex{Float64},Array{Complex{Float64},1}}
2×2 Tensor{Complex{Float64},2,Dense{Complex{Float64},Array{Complex{Float64},1}},IndexSet{2}}:
  0.0+0.0im  0.0+0.5im
 -0.0-0.5im  0.0+0.0im
ITensor ord=2 (2|id=90|S=1/2,Site,n=1) (2|id=90|S=1/2,Site,n=1)'
Dense{Complex{Float64},Array{Complex{Float64},1}}

julia> matrix(Sy)
2×2 Array{Complex{Float64},2}:
  0.0+0.0im  0.0+0.5im
 -0.0-0.5im  0.0+0.0im
```

`@show {ITensor}` and `matrix({ITensor})` prints and returns, respectively, the transposed version of a conventional matrix, as it begins with the unprimed (ket) index and then the primed (bra) index, rather than the other way around. Indeed, thinking of `Sy` as a matrix mapping (a vector to its left and a dualvector to is right) to the complex numbers, would expect the indices to be permuted.
This has proven to be confusing in at least one case 🙋‍♂️.

What are your thoughts?

Cheers,
Jan

jtschneider:
Addendum:
For the sake of completeness and to immediately see why I was confused, here is the definition of `Sy` for `S=1/2`:

<img width="210" alt="image" src="https://user-images.githubusercontent.com/9451057/70319160-eb5dc580-1821-11ea-9c41-4a3985bdce54.png">

orialb:
> @show {ITensor} and matrix({ITensor}) prints and returns, respectively, the transposed version of a conventional matrix, as it begins with the unprimed (ket) index and then the primed (bra) index, rather than the other way around. 

[...]

PR #157, opened 2019-12-07
Changed index ordering in operator definitions

emstoudenmire:
Although the operators defined in the site_type files (electron.jl, spinhalf.jl, etc) are correctly defined, they use an index ordering which is the reverse of the one conventionally used when writing these operators as matrices. So to keep this from being confusing, this PR flips the order to be the conventional one. 

Thanks to user @jtschneider for pointing out this situation!


emstoudenmire:
This PR also fixes a small bug where the `norm` function wasn't being exported.

emstoudenmire:
This PR references bug #154 though it doesn't quite implement it, so I'm not marking it as closed by this PR.

PR #164, opened 2019-12-20
update examples with new siteinds syntax

orialb:
This PR updates the DMRG examples to use the correct function (`siteinds`) when building the site index set.


emstoudenmire:
Thanks very much

PR #170, opened 2020-01-13
Fix errorMPOprod and allow scaling MPOs

kshyatt:
`errorMPOprod` couldn't handle the case where the `MPO` has different site indices (e.g. it has `t` outputs and `s` inputs, rather than `s'` and `s`). Also added a little routine so that you can do `2*A` where `A` is an `MPO`.

### DIFF
--- a/src/mps/mpo.jl
+++ b/src/mps/mpo.jl
@@ -6,7 +6,8 @@ export MPO,
        maxlinkdim,
        orthogonalize!,
        truncate!,
-       sum
+       sum,
+       overlap
 
 mutable struct MPO
   N_::Int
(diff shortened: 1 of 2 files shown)


kshyatt:
@mtfishman any thoughts?

mtfishman:
Hi Katie,

Thanks for the contributions. I need to look over the changes to `inner` more closely.

What was the purposed of adding the `overlap` function? That function was deprecated in the C++ version of ITensor in favor of `inner` and `trace`.

Also, to me it seems like a better definition of scalar multiplication with an MPS/MPO is not scaling each tensor by the scalar, but instead scaling just one tensor (for example the center site). In the current implementation, if you scale every tensor by the scalar `a`, than you are scaling the norm of the MPS by `a^N` if there are `N` sites, which doesn't seem great. The same goes for negating an MPS/MPO. What do you think?

-Matt

kshyatt:
I needed `overlap` for part of the PEPS code and (IIRC) `inner` and `trace` don't exist yet for MPOs, so I added the one that was simplest. Delete it if you want.

The scaling thing makes sense, I'll change it.

mtfishman:
I see, in that case it probably makes sense to remove it here (to not add confusion) and you can keep it as a definition in your own project until the `inner` and `trace` functions in ITensors.jl catch up.

src/mps/mpo.jl
```diff
@@ -558,6 +569,18 @@ function Tensors.truncate!(M::Union{MPS,MPO}; kwargs...)
 
 end
 
+function Base.:*(x::Number,M::Union{MPS,MPO})
+    N = deepcopy(M)
+    N.A_[div(size(N), 2)] = N.A_[div(size(N), 2)] .* x
```
emstoudenmire: Any reason not to use N[div(size(N),2)] here instead of N.A_? Generally we'd like to avoid directly accessing data members unless necessary.

src/mps/mpo.jl
```diff
@@ -558,6 +569,18 @@ function Tensors.truncate!(M::Union{MPS,MPO}; kwargs...)
...
+    return N
+end
+
+function Base.:-(M::Union{MPS,MPO})
+    N = deepcopy(M)
+    N.A_[div(size(N), 2)] = -1.0*N.A_[div(size(N), 2)]
```
emstoudenmire: See comment above ^

kshyatt:
Added some tests, caught some problems before they were merged ... you love to see it

PR #171, opened 2020-01-15
Define QN Index, basic operations with QN ITensor working

mtfishman:
This pull request starts the process of bringing QN support to ITensors.

It extends the definition of Index to be parametrized over a "space" field, which store the information about the space that the Index lives in. For example, for a regular index, the space is the integer dimension (so `Index{Int}`), and for a QN Index, the space is the QNs and sizes of the blocks (so `Index{Vector{Pair{QN,Int}}}`).

It also adds new ITensor functionality like constructing an ITensor given a desired flux. Basic functionality like contraction of QN ITensors and scalar multiplication appears to be working (see the new tests added).

The main things that still need to be added are the QN combiner and factorizations of QN ITensors, those will be addressed in other pull requests.


mtfishman:
Some tests are failing related to QN comparison. The definitions of `==(::QN,::QN)` and `<(::QN,::QN)` previously had issues, and I only fixed them enough to get ITensor functionality working. I need to go back over them.

@emstoudenmire, what is the definition of `<(::QN,::QN)`?

emstoudenmire:
This is great, Matt. About `<(::QN,::QN)`, roughly speaking it's a series of comparisons which check whether the first QNVal is equal, and if so goes to the next one, and then the next until either one hits all of the "active" (non-trivial) ones, or until one compares less than the other in terms of the value they have.

I'll take a look at the error and see what could be happening. Also at a glance, the < comparison for QNVal seems to only be comparing the name, not the value - not sure if that's intentional or a mistake. I'll check.

mtfishman:
Also, what do you think about changing the Arrow convention such that `In=1` and `Out=-1`, and Indices default to having `In` arrows?

I like that convention better, since to me it is more natural when drawing tensor network diagrams. For example, for a gauged MPS, with that convention the arrows follow the "renormalization flow" direction, which is in towards the center (the same would go for a MERA, where to me it makes sense for the arrows to point in the coarse-grained direction). Also, the Arrow directions then line up more naturally with how I draw isometries (I find it natural to draw isometries with arrows on the indices from the larger space to the smaller space).

PR #179, opened 2020-02-01
Add QN combiner

mtfishman:
This adds functionality for a QN combiner.


mtfishman:
@emstoudenmire, from my end this is ready to merge.

There are definitely possible optimizations to make in the combiner logic, but I think right now it is important to have the functionality in place and we can profile and optimize as needed.

emstoudenmire:
Great! Congrats. Tests look very thorough from what I can see. I agree about circling back to the performance as needed. We can always do rounds of profiling, as long as you think this is a reasonably fast algorithm to build on. 

Please merge if you think it's ready.

PR #181, opened 2020-02-03
Add constructor for QN ITensor with no flux

mtfishman:
This adds the constructor:

```julia
i = Index([QN(0)=>1,QN(1)=>2],"i")
A = ITensor(i,dag(i'))
```

It makes an ITensor with BlockSparse storage but with no blocks. `flux(A)` returns nothing, and blocks can be added by setting elements (see the tests).


emstoudenmire:
Great that this already passes the unit tests!

Issue #184, opened 2020-02-09
`norm(v::ITensor)` became very slow (performance regression)


orialb:
Hi,
I noticed that my TDVP code became extremely slow with latest master (edit: also DMRG is extremely slow, see comment below). After some profiling I think the problem is related to `norm(v::ITensor)` becoming extremely slow.
I did some git bisecting and I believe this regression was introduced in https://github.com/ITensor/ITensors.jl/pull/171. 

Using the following code to benchmark:
```julia
using ITensors, BenchmarkTools
i = Index(1000)
v = randomITensor(i)
@btime norm(v)
```
I get on master (https://github.com/ITensor/ITensors.jl/commit/b6279695d3824f5554cf9b2f349c970a460bba90): 
```julia
julia> @btime norm(v)
  2.501 ms (18025 allocations: 344.16 KiB)
```
While on https://github.com/ITensor/ITensors.jl/commit/6ae3e7b89fd18c4a1c3046fbd42717df908ac265 I get:
```julia
julia> @btime norm(v)
  3.976 μs (3 allocations: 64 bytes)
```

So it seems that `norm` became 3 orders of magnitude slower and is allocating a lot. 
I'll try to keep digging into it, but if you have some idea of what can be the reason let me know.

orialb:
Not sure if that is the problem, but it also seems that `getindex` for `DenseTensor` got slower (and in the profiling graph I saw that the majority of time of `norm` is spent in `getindex`).

```julia
using ITensors, BenchmarkTools
i = Index(1000)
t = tensor(randomITensor(i))
@btime t[1]
```
I get `1.189 μs (7 allocations: 112 bytes)` on master, and `24.920 ns (1 allocation: 16 bytes)` on 6ae3e7b .

mtfishman:
Thanks a lot for catching and fixing this. There are a lot of changes going on in the `Tensors` module as we work on the BlockSparseTensor functionality, so I am not surprised that issues like this are showing up. I will investigate the regression in `getindex`, I think it is likely that there are a lot of performance improvements to make for basic `Tensor` functionality.

This definitely points to a need to have systematic benchmarks that can be run for each pull request (and note that it may be good to be careful about following along with master at this time).

orialb:
> This definitely points to a need to have systematic benchmarks that can be run for each pull request

[...]

PR #187, opened 2020-02-11
Fix to restore performance of norm(t::DenseTensor) and DMRG 

orialb:
As I reported in https://github.com/ITensor/ITensors.jl/issues/184,  for some reason norm(t::DenseTensor) became very slow, which also severely affects DMRG performance.  
This PR fixes that by overloading `norm(T::DenseTensor)` and calling `norm` on the storage.
To the best of my understanding that achieves the same effect as before (i.e treating the tensor as a vector and computing the vector's norm). 

I still can't figure out why `norm(t::DenseTensor)` itself became so much slower, but this seems to solve the problem at least. 

running the `1d_Heisenberg_dmrg.jl` example before the change:
```
After sweep 1 energy=-137.889012787989 maxlinkdim=9 time=11.974
After sweep 2 energy=-138.934952940994 maxlinkdim=20 time=22.388
After sweep 3 energy=-138.940077788262 maxlinkdim=91 time=189.390
```
with this PR performance is restored:
```
After sweep 1 energy=-137.791153491960 maxlinkdim=9 time=11.132
After sweep 2 energy=-138.933183889406 maxlinkdim=20 time=1.227
After sweep 3 energy=-138.940071627956 maxlinkdim=94 time=5.043
After sweep 4 energy=-138.940085926733 maxlinkdim=100 time=12.213
After sweep 5 energy=-138.940086053788 maxlinkdim=95 time=13.604
```

### DIFF
--- a/src/Tensors/dense.jl
+++ b/src/Tensors/dense.jl
@@ -626,6 +626,8 @@ function permute_reshape(T::DenseTensor{ElT,NT,IndsT},
   return reshape(T,newinds)
 end
 
+LinearAlgebra.norm(T::DenseTensor) = norm(store(T))
+
 # svd of an order-n tensor according to positions Lpos
 # and Rpos
 function LinearAlgebra.svd(T::DenseTensor{<:Number,N,IndsT},


emstoudenmire:
Thanks for catching this, Ori. We were starting from letting Julia’s built in AbstractArray functions handle a lot of cases, but sometimes a custom overload is needed like this one to get optimal performance. So there may be a few more cases like this which we plan to catch by profiling. Keep us posted also about the performance you are seeing and any drops you observe.

PR #195, opened 2020-02-13
More concise uncombinedinds and tests

kshyatt:
`uncombinedinds` can be a one-liner, I think. Just a few missing tests from coverage...


mtfishman:
Looks good, thanks!

PR #196, opened 2020-02-13
initialise MPS with certain data type T<:Number

jtschneider:
Perhaps a `function MPS(::Type{T},sites) where {T<:Number}` might be appreciated in case of initialising an MPS with e.g. `ComplexF64` data type.


mtfishman:
Definitely useful, thanks. What if we define `MPS(sites) = MPS(Float64,sites)` to avoid code duplication?

jtschneider:
Sounds quite reasonable. Will do.

orialb:
It will be also very useful to have this feature in `productMPS` which I guess is one of the most common ways for users to initialize an `MPS` for physics applications.

src/mps/mps.jl
```diff
@@ -26,30 +26,32 @@ mutable struct MPS
...
-      v[ii] = ITensor(l[ii], s)
-    elseif ii == N
-      v[ii] = ITensor(l[ii-1], s)
-    else
-      v[ii] = ITensor(l[ii-1],s,l[ii])
-    end
-  end
-  return MPS(N,v,0,N+1)
-end
+function MPS(::Type{T},sites) where {T<:Number}
```
mtfishman: The formatting of this function looks off, could you please clean it up? Thanks!

jtschneider: Yes, don't know what happened there...

mtfishman:
Looks good, thanks!

Issue #198, opened 2020-02-14
performance regression `dag(::ITensor)`


orialb:
Seems that `dag(t::ITensor)` became slower. using 
```
i = Index(1000)
t = randomITensor(ComplexF64, i)
@btime dag($t)
```
I get ` 1.272 ms (8023 allocations: 156.91 KiB)` on master, and `2.596 μs (10 allocations: 16.06 KiB)` on f92d4ab . This seems to have large effect on my TDVP code as I see that Krylov based exponentiation using `KrylovKit.exponentiate` is spending most of its time in `dag` (I'm not using this code in production yet, just wanted to point out that this slowness of `dag` can have a large effect in user codes).

I think the problem is similar to #184 where calling `conj(tensor(t))` became slower. Indeed 
On master:
```
julia> @btime conj(tensor(t));
  1.272 ms (8023 allocations: 156.91 KiB)
```
on f92d4ab:
```
julia> @btime conj(tensor(t))
  2.502 μs (7 allocations: 15.94 KiB)
```

Also here calling `conj(store(t))` (on master) is much faster compared to `conj(tensor(t)` and I get `1.747 μs (1 allocation: 15.75 KiB)` . Underlying problem might be related to `getindex` regression I pointed out in #184. 
I could make a PR with a fix for this (overloading `conj`).

mtfishman:
Thanks for pointing this out, sorry this issue keeps popping up. Definitely need to look into why `getindex` seems to be so slow for Tensor types. Your suggestion of overloading `conj` is a good intermediate solution.

orialb:
fixed by #201 .

PR #201, opened 2020-02-15
overload conj for Tensor and tensor storage

orialb:
This is a temporary fix for https://github.com/ITensor/ITensors.jl/issues/198.

One thing to point out is that  `conj(::DenseTensor{<:Real,N})` is non-copying (because it doesn't have to do anything) while `conj(::DenseTensor{<:Complex,N})` is creating a copy (this is also true without this fix when the generic fallback from Base is used). This true also for base Julia (i.e with `Array{Float64,N}` vs. `Array{ComplexF64,N}`). 
This might lead to undesired behavior of `dag` - `dag(t::ITensor)` will contain the same dense storage as `t` when the elements of `t` are real but not when they are complex. I guess either one of those behaviors is OK, but there might be a problem with the inconsistency.

### DIFF
--- a/src/Tensors/dense.jl
+++ b/src/Tensors/dense.jl
@@ -60,6 +60,9 @@ Base.similar(D::Dense,::Type{T}) where {T<:Number} = Dense(similar(data(D),T))
 Base.copy(D::Dense) = Dense(copy(data(D)))
 Base.copyto!(D1::Dense,D2::Dense) = copyto!(data(D1),data(D2))
 
+Base.conj(D::Dense{<:Real,VecT}) where {VecT} = D
+Base.conj(D::Dense{<:Complex,VecT}) where {VecT} = Dense(conj(data(D)))
+
 Base.fill!(D::Dense,v) = fill!(data(D),v)
 
 # TODO: should this do something different for SubArray?
--- a/src/Tensors/diag.jl
+++ b/src/Tensors/diag.jl
@@ -32,6 +32,9 @@ Base.complex(D::Diag) = Diag(complex(data(D)))
 
 Base.copy(D::Diag) = Diag(copy(data(D)))
 
+Base.conj(D::Diag{<:Real, VecT}) where {VecT} = D
+Base.conj(D::Diag{<:Complex, VecT}) where {VecT} = Diag(conj(data(D)))
+
 Base.eltype(::Diag{ElT}) where {ElT} = ElT
 Base.eltype(::Type{<:Diag{ElT}}) where {ElT} = ElT
 
--- a/src/Tensors/blocksparse/blocksparse.jl
+++ b/src/Tensors/blocksparse/blocksparse.jl
@@ -80,6 +80,8 @@ end
 # TODO: this could be a generic TensorStorage function
 Base.complex(D::BlockSparse{T}) where {T} = BlockSparse{complex(T)}(complex(data(D)),
                                                                     blockoffsets(D))
+Base.conj(D::BlockSparse{<: Real}) = D
+Base.conj(D::BlockSparse) = BlockSparse(conj(data(D)), copy(blockoffsets(D)))
 
 Base.eltype(::BlockSparse{T}) where {T} = eltype(T)
 # This is necessary since for some reason inference doesn't work
--- a/src/Tensors/tensor.jl
+++ b/src/Tensors/tensor.jl
@@ -38,6 +38,8 @@ Base.copy(T::Tensor) = Tensor(copy(store(T)),copy(inds(T)))
 
 Base.complex(T::Tensor) = Tensor(complex(store(T)),copy(inds(T)))
 
+Base.conj(T::Tensor) = Tensor(conj(store(T)), copy(inds(T)))
+
 Random.randn!(T::Tensor) = (randn!(store(T)); return T)
 
 #function Base.similar(::Type{<:Tensor{ElT,N,StoreT}},dims) where {ElT,N,StoreT}
--- a/test/Tensors/blocksparse.jl
+++ b/test/Tensors/blocksparse.jl
@@ -75,6 +75,7 @@ using ITensors,
   A = BlockSparseTensor(ComplexF64,locs,indsA)
   randn!(A)
   @test conj(data(store(A))) == data(store(conj(A)))
+  @test typeof(conj(A)) <: BlockSparseTensor
 
   @testset "BlockSparseTensor setindex! add block" begin
     T = BlockSparseTensor([2,3],[4,5])
--- a/test/Tensors/blocksparse.jl
+++ b/test/Tensors/blocksparse.jl
@@ -72,6 +72,10 @@ using ITensors,
     @test A[I] == Ap[permute(I,(2,1))]
   end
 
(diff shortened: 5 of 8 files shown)


orialb:
One test checking that DMRGObserver is correctly capturing truncation errors was failing (just on mac os, linux was fine). When I ran it locally on my mac it passed. 
I triggered CI again to see if it resolves itself (MPS is initialized randomly),  but I don't know if the fact that the test (line 82 in "test/dmrg.jl") failed once should be a cause for concern or not.

src/Tensors/dense.jl
```diff
@@ -626,6 +626,9 @@ function permute_reshape(T::DenseTensor{ElT,NT,IndsT},
   return reshape(T,newinds)
 end
 
+Base.conj(T::DenseTensor{<:Real,N}) where {N} = T
+Base.conj(T::DenseTensor) = Tensor(Dense(conj(store(T))), copy(inds(T)))
```
mtfishman: It seems like we could have a more generic definition:
```julia
Base.conj(T::Tensor) = Tensor(conj(store(T)), copy(inds(T)))
```
and then every storage type must define there own version of `conj`. Each version should return a storage object of the same type, so the `Dense(...)` constructor should not be necessary.

mtfishman: And just as an FYI, for generic function like that (where it seems like there is a reasonable definition for Tensors of most storage types), I have been trying to put them in the file `src/Tensors/tensor.jl`.

mtfishman:
Hi Ori,

Thanks for contributing a fix.

It is a good question of whether or not we should follow the base convention here. I think in general we are leaning towards the convention that if a function leads to a copy in certain cases, it should copy in all cases since it could be surprising to get a view of the data in some cases but not others (which could lead to unexpected bugs when modifying data). I am actually a bit surprised by the base behavior for this. I would say that if the storage is real, it should make a copy. @emstoudenmire, do you have an opinion on this?

I think I have also seen random failures in that test, can we just fix the seed or use a non-random starting MPS?

-Matt

src/Tensors/dense.jl
orialb: Indeed this could be done. I thought about it but was not sure whether you want this fix to be applied also for the block sparse tensors. But I guess it's not a big deal to change it afterwards in any case. I will update the PR.

orialb:
>I think in general we are leaning towards the convention that if a function leads to a copy in certain cases, it should copy in all cases since it could be surprising to get a view of the data in some cases but not others (which could lead to unexpected bugs when modifying data) ... I would say that if the storage is real, it should make a copy.

I agree with the fact that surprises should be avoided. On the other hand there could be a possible performance gain in generic tensor network algorithms for TNs with real elements (by avoiding unnecessary copies when calling`dag`).

> I think I have also seen random failures in that test, can we just fix the seed or use a non-random starting MPS?

I could update the test. should I open a new PR for it or is it fine to do it in this PR?

emstoudenmire:
Let's discuss this further (about copying versus not). My feeling right now is that it would be sufficient to document the `dag` function for ITensors by saying "dag is *not* guaranteed to make a copy of the underlying data", and then go on to explain to users what to do if a copy is needed.

However this raises a second issue of: if one wants to guarantee a copy in generic code, and the data happens to be complex, will two copies now happen if one calls `copy` and then `dag`? 

If there's no way to avoid that two-copy issue in generic code, then I'd actually switch my vote to have `dag` always copy.

mtfishman:
It seems like one of the questions is how easy it is to "opt out" of whatever behavior we choose. The tricky part with `dag` is that it both complex conjugates and daggers the indices, so it is hard "opt out" of the copying. For example if we made `dag` always copy, and you want to not copy if the ITensor is real, this would not work for QN ITensors:
```julia 
A = randomITensor(i,j)
Adag = iscomplex(A) ? dag(A) : A
```
so we would have to think of a special syntax like `daginds` that just daggers the indices but doesn't copy the storage, like:
```julia
Adag = iscomplex(A) ? dag(A) : daginds(A)
```

Going the other way, if `dag` only sometimes copies, and you really wanted a copy since you are modifying elements in-place afterwards, you would do:
```julia
Adag = dag(A)
Adag = isreal(Adag) ? copy(Adag) : Adag
```
or something similar (this would get around the double copying issue, as long as the rule is consistent that complex data gets copied and real data does not, which I think can always be the case).

An alternative is that we can have a keyword argument in `dag` that chooses whether or not you want `dag` to always copy:
```julia
Adag = dag(A; always_copy=false)
Adag = dag(A; always_copy=true)
```
and choose the default that we think will be used most. I think in general people will want `always_copy=false` since I suspect in-place modifications are relatively rare in ITensor code (for example, a common case of `dag` is calculating an expectation value, like `dag(A)*Sz*A`, so `dag(A)` is temporary and may as well not copy). Then, the syntax `dag(A; always_copy=true)` can be used instead of the second example above, which is I think is nicer.

Anyway, for the sake of this pull request, I would lean towards `conj` following the Julia convention, and we can pick the behavior of `dag` at the ITensor level.

mtfishman:
And Ori, I think a separate PR for the test would be great, thanks!

src/Tensors/dense.jl
mtfishman: Yes it's no big deal to change it, and we can always overload the general definition for other Tensor/storage types.

emstoudenmire:
I like the idea of the `always_copy` keyword argument. Then we can have `dag` not copy real data, which should normally cause no issue and not require too careful of handling in generic code. But then if someone has code logic which requires a copy to be correct, they can flag `always_copy=true` and get at most one copy, not having to worry about two copies happening.

(By the way, this kind of stuff is why the ref counting approach in C++ ITensor is more elegant on a logical level, but I think it comes with its own possible performance hits.)

src/Tensors/diag.jl
```diff
@@ -32,6 +32,9 @@ Base.complex(D::Diag) = Diag(complex(data(D)))
 
 Base.copy(D::Diag) = Diag(copy(data(D)))
 
+Base.conj(D::Diag{<:Real, VecT}) where {VecT} = D
+Base.conj(D::Diag{<:Complex, VecT}) where {VecT} = Dense(conj(data(D)))
```
mtfishman: I think you meant to use `Diag` on the right side.

mtfishman:
Thanks for the update! Looks good, once you resolve the comment I made.

orialb:
> once you resolve the comment I made.

oops. I fixed it now. Interesting that it didn't break any of the unit tests.
I added also type stability tests for `DenseTensor` and `BlockSparseTensor`.
Currently it seems that `DiagTensor` doesn't have any unit tests if I understand correctly, should I add a test file for `test/Tensors/diag.jl` and add a similar test there?

mtfishman:
That would be great. I was mostly testing the `DiagTensor` functionality through the ITensor interface, but they should have devoted tests as well. Not to surprised it didn't break any tests, I think the tests for complex Tensors/ITensors are a bit lacking.

orialb:
Added the test.

mtfishman:
Looks good, thanks again!

PR #203, opened 2020-02-19
fix the seed in DMRGObserver test

orialb:
The truncation error test was failing on some (not so common) occasions due to a bad random initial state. We agreed that it is better to fix the seed to remove randomness in the unit test.


mtfishman:
Looks good to me, thanks for the contribution.

PR #209, opened 2020-02-29
Test eigen and add some necessary methods for non-hermitian problem

kshyatt:
the `eigen` method in `decomp` wasn't tested and turned out to need some more methods defined ... hopefully this is ok.


kshyatt:
closing and reopening to re-run CI

kshyatt:
OK re-ran `pr` ci passed. OK to merge?

Issue #136, opened 2019-11-12
Some improvements to priming/tagging


mtfishman:
Various people have asked for functions like `primeExcept(A,"i")` (at least in the C++ version). I was hesitant to add it in order to avoid a proliferation of priming/tagging functions. I think a nice solution could be syntax like the following:

```julia
i = Index(2,"i")
j = Index(2,"j")
k = Index(2,"k")

A = randomITensor(i,j,k)

Ap = prime(A,not("i"))
```

This is useful in cases where you want to contract an ITensor over only one specified Index, so is quite a common operation.

`not` could just create a simple type that wraps the input (so creates an object `Not{TagSet}` that just stores the TagSet `"i"`), and then the function [indexpositions](https://github.com/ITensor/ITensors.jl/blob/master/src/indexset.jl#L337) could dispatch on `Not{T}` objects. One could of course make a `not(i)` or `not(i,j)` to avoid priming an index or set of indices.

This could of course all work in the C++ version too. This syntax could even be extend to more complicated logic, like `prime(A,or("i","j"))`. Just some food for thought.

mtfishman:
This is closed by #214.

PR #214, opened 2020-03-04
Add prime(A,not(i)) syntax

mtfishman:
This implements the idea in #136 to add syntax like `prime(A,not(i))` to prime/tag the complement of the specified indices.

For example:

```julia
i = Index(2,"i")
j = Index(2,"j")
k = Index(2,"k")
A = randomITensor(i,j,k)

prime(A,not("j"))  # has indices (i',j,k')

addtags(A,"x",not("k"))  # has indices (addtags(i,"x"),addtags(j,"x"),k)

setprime(A,2,not(i,j))  # has indices (i,j,k'')
```
This can help clean up some priming and tagging code.

A further possibility is to add other kinds of logic, for example `prime(A,or("i","j"))` to prime indices containing either the tags "i" or "j", but we can wait to see if that turns out to be useful.

### DIFF
--- a/src/indexset.jl
+++ b/src/indexset.jl
@@ -5,6 +5,7 @@ export IndexSet,
        findindex,
        findinds,
        indexposition,
+       not,
        swaptags,
        swaptags!,
        swapprime,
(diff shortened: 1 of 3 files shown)


emstoudenmire:
Nice to see how little codes this requires

Issue #189, opened 2020-02-12
Make IndexSet parametrized by Index type


mtfishman:
Now that Index is a parametric type (depending on having QN storage or just the dimension for a non-QN Index), I think it would be useful to make IndexSet a parametric type. This is mostly for better code organization, to allow for type dispatch since many functions that accept IndexSet have different functionality and implementations depending on whether or not they have QNs.

Also, we may want to think about the storage type of IndexSet. I still can't decide if it should be a tuple or a mutable structure. Right now it is a `SizedVector`, which is similar to but implemented quite differently from an `MVector` (it is a wrapper type around a `Vector` with the size as a type parameter). It can't be an `MVector` because `setindex!` doesn't work for an `MVector` that stores dynamically sized objects, like our QN storage. For example:
```julia
julia> using StaticArrays

julia> v = MVector(([1,2],[3,4]))
2-element MArray{Tuple{2},Array{Int64,1},1,2} with indices SOneTo(2):
 [1, 2]
 [3, 4]

julia> v[1]
2-element Array{Int64,1}:
 1
 2

julia> v[1] = [4,5]
ERROR: setindex!() with non-isbitstype eltype is not supported by StaticArrays. Consider using SizedArray.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] setindex!(::MArray{Tuple{2},Array{Int64,1},1,2}, ::Array{Int64,1}, ::Int64) at /mnt/home/mfishman/.julia/packages/StaticArrays/1g9bq/src/MArray.jl:109
 [3] top-level scope at REPL[18]:1
```
Making it a Tuple storage may be annoying, since I believe it is quite common, at least internally, to replace indices in an IndexSet by setting values.

emstoudenmire:
Agreed - I’ve already found some important use cases for this

mtfishman:
This is closed by #216.

PR #216, opened 2020-03-04
Parametrize IndexSet by the Index type

mtfishman:
This implements parametrizing the IndexSet type by the Index type. This will help with code organization, so that we can define `const QNIndexSet{N} = IndexSet{N,QNIndex}` and then dispatch on `QNIndexSet`. Luckily the changes are pretty minimal to get this working, and all tests pass as they are.

This addresses issue #189.

### DIFF
--- a/src/indexset.jl
+++ b/src/indexset.jl
@@ -21,20 +21,22 @@ export IndexSet,
        permute,
        hasqns
 
-struct IndexSet{N}
-  inds::SizedVector{N,Index}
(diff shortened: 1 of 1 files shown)


mtfishman:
@emstoudenmire, let me know if you think we should discuss this design move more (I don't see much downside), otherwise this is ready to go.

emstoudenmire:
I’m all for this change. For one thing, it will be useful for handling fermions more efficiently (since we can make sure that logic only runs for QN IndexSets and doesn’t penalize non QN code).

PR #222, opened 2020-03-09
QN diagITensor

mtfishman:
This PR adds QN `diagITensor` and `delta` tensors, based on a new `DiagBlockSparse` TensorStorage type.

Right now, `DiagBlockSparse` supports getting and setting elements and contraction with `BlockSparse` tensors. Tensor decompositions like `svd` and `eigen` now return singular values/eigenvalues with `DiagBlockSparse` storage.


emstoudenmire:
Looks good! One question I have is about the flux arguments to the constructor functions (`diagITensor` functions). Must the flux provided be the one that corresponds to the diagonal blocks? Meaning diagonal in the sense of if one converted to dense, then the blocks which contain the diagonal elements of the tensor? 

Or is the tensor diagonal in the sense that each of the blocks allowed to be non-zero according to the given flux then a diagonal block?

mtfishman:
> Looks good! One question I have is about the flux arguments to the constructor functions (`diagITensor` functions). Must the flux provided be the one that corresponds to the diagonal blocks? Meaning diagonal in the sense of if one converted to dense, then the blocks which contain the diagonal elements of the tensor?
> 
> Or is the tensor diagonal in the sense that each of the blocks allowed to be non-zero according to the given flux then a diagonal block?

What the `diagITensor` constructor does is searches just the diagonal blocks (i.e. for an order 3 tensor, (1,1,1), (2,2,2), (3,3,3), etc.)), checks which ones are nonzero according to the flux, and just stores those blocks. So a `diagITensor` cannot have non-zero blocks off of the diagonal, but doesn't necessarily have to store all of the diagonal blocks (so if none of the diagonal blocks have the flux that is provided, the tensor would have no blocks).

mtfishman:
Not sure if that answers your question... All of the blocks themselves only have nonzero elements on their diagonals. So a `diagITensor` has no off-diagonal elements, if the tensor is interpreted as a dense tensor (i.e. T_ijk = 0 if !(i==j==k)).

emstoudenmire:
Great! Ok that answers my question. It may be that the docstrings can say some of this in more detail, but otherwise it looks good to go to me. 

The tests should already be pretty rigorous, since we already have extensive tests of the various tensor decompositions (svd, eigen, etc.) which now involve this storage.

mtfishman:
Yes, the main goal for now was to get those tests passing. We can add more functionality in the future (for example, `removeqns`, conversion to `Dense`, etc.).

PR #223, opened 2020-03-09
+ syntax for AutoMPO

### LINKED ISSUE #107: Small change to AutoMPO syntax
An alternative syntax we could use for AutoMPO is the following:

```julia
ampo = AutoMPO(sites)
for j=1:N-1
  ampo += ("Sz",j,"Sz",j+1)
  ampo += (0.5,"S+",j,"S-",j+1)
  ampo += (0.5,"S-",j,"S+",j+1)
end
```

as opposed to:

```julia
ampo = AutoMPO(sites)
for j=1:N-1
  add!(ampo,"Sz",j,"Sz",j+1)
  add!(ampo,0.5,"S+",j,"S-",j+1)
  add!(ampo,0.5,"S-",j,"S+",j+1)
end
```

This would get use very close to the C++ syntax. It should just require overloading `Base.:+(::AutoMPO,::Tuple)`.

@emstoudenmire, let me know what you think.

mtfishman:
Implements the idea in #107.

Closes #107.

### DIFF
--- a/src/physics/autompo.jl
+++ b/src/physics/autompo.jl
@@ -117,9 +117,12 @@ end
 
 struct AutoMPO
   terms::Vector{MPOTerm}
-  AutoMPO() = new(Vector{MPOTerm}())
+  AutoMPO(terms::Vector{MPOTerm}) = new(terms)
 end
+AutoMPO() = AutoMPO(Vector{MPOTerm}())
 terms(ampo::AutoMPO) = ampo.terms
+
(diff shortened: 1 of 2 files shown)


emstoudenmire:
Nice to see this. Too bad about the copying but I get that’s just how Julia works without extra tricks; also shouldn’t be a bottleneck except for very large inputs such as chemistry Hamiltonians, for which we can recommend the `add!` function.

mtfishman:
Yeah, we can play around with using `ampo .+= ("Sz",j)` as an alternative syntax.

PR #225, opened 2020-03-10
Add some broadcasting syntax for ITensors

mtfishman:
This adds some broadcasting syntax to ITensors, such as:
```julia
i = Index(2,"i")
A = randomITensor(i,i')
B = randomITensor(i',i)
c = 2
B .= c .* A
```
to overwrite `B` with `c*A`, without any allocations. Internally, this will just call `mul!(B,c,A)`. Take a look at the tests for a more complete list of example operations that are supported now.

@emstoudenmire, this made me think that we could abuse the broadcasting syntax a bit to get lazy ITensor contractions. For example, it would be simple enough to make:
```julia
i = Index(2,"i")
A = randomITensor(i,i')
B = randomITensor(i',i'')
C = randomITensor(i,i'')
c = 2
C .= c .* A  .* B
```
do an in-place multiplication of `A` and `B` into `C`, scaling by `c`, without any allocations. This would be a bit of an abuse of the notation, since for Julia Arrays, this operation would not do a matrix multiplication but instead does an elementwise multiplication (a Hadamard product). However, ITensors act differently enough from Julia Arrays that maybe this is ok.

### DIFF
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -548,12 +548,12 @@ Base.:/(A::ITensor,x::Number) = A*(1.0/x)
 Base.:-(A::ITensor) = ITensor(-tensor(A))
 function Base.:+(A::ITensor,B::ITensor)
   C = copy(A)
-  C = add!(C,B)
+  add!(C,B)
   return C
 end
 function Base.:-(A::ITensor,B::ITensor)
   C = copy(A)
-  C = add!(C,-1,B)
+  add!(C,-1,B)
   return C
 end
 
@@ -650,11 +650,11 @@ B .+= A
 B .+= α .* A
 ```
 """
-add!(R::ITensor,T::ITensor) = add!(R,1,T)
+add!(R::ITensor,T::ITensor) = apply!(R,T,(r,t)->r+t)
 
-function add!(R::ITensor{N},α::Number,T::ITensor{N}) where {N}
-  return apply!(R,T,(r,t)->r+α*t)
-end
+add!(R::ITensor,α::Number,T::ITensor) = apply!(R,T,(r,t)->r+α*t)
+
+add!(R::ITensor,T::ITensor,α::Number) = add!(R,α,T)
 
 """
     add!(A::ITensor, α::Number, β::Number, B::ITensor)
@@ -664,9 +664,7 @@ Add ITensors α*A and β*B and store the result in A.
 A .= α .* A .+ β .* B
 ```
 """
-function add!(R::ITensor{N},αr::Number,αt::Number,T::ITensor{N}) where {N}
-  return apply!(R,T,(r,t)->αr*r+αt*t)
-end
+add!(R::ITensor,αr::Number,αt::Number,T::ITensor) = apply!(R,T,(r,t)->αr*r+αt*t)
 
 function apply!(R::ITensor{N},T::ITensor{N},f::Function) where {N}
   perm = getperm(inds(R),inds(T))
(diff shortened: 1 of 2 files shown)


emstoudenmire:
Very nice. It’s great to see this can be added with a fairly small bit of code, covering all storage types! On the other hand, it still seems unfortunate to me that Julia works this way, namely that one can’t get this kind of behavior already with the standard, undotted notation.

One question about this feature: I suppose the reason for the exceptions in the last unit tests is the ordering of indices, correct? And that the broadcasted operations requires the same index ordering?

emstoudenmire:
I think this code is good to go -

mtfishman:
> Very nice. It’s great to see this can be added with a fairly small bit of code, covering all storage types! On the other hand, it still seems unfortunate to me that Julia works this way, namely that one can’t get this kind of behavior already with the standard, undotted notation.

That is only partially true. Even for an operation like `B .= c .* A`, this adds something beyond doing `B = c * A` in the C++ version, since the two dots are "fused" into a single operation, so no temporaries are made (it actually works by lazily handling the `c .* A` operation).

> 
> One question about this feature: I suppose the reason for the exceptions in the last unit tests is the ordering of indices, correct? And that the broadcasted operations requires the same index ordering?

The broadcast operations here are done by overloading other general ITensor operations like `mul!` and `add!`, so they don't require the same index ordering (the permute is handled as usual, note that all of the other tests involve tensors that are transposes of each other). It is actually a bit more subtle, mostly just that those operations correspond to ones we don't have nice ITensor functions for right now. There is no reason we can't support them, I was just keeping this PR simpler and we can add more functionality later.

What do you think about using `C .= A .* B` as a syntax for contracting `A` and `B` and storing the result in `C`, i.e. without temporaries?

emstoudenmire:
Agreed that it's doing something we don't even have in C++. I was more saying I wish there was a better approach to this sort of thing that either Julia's or C++; there could be, I think, if language designers thought about this case ahead of time.  Maybe later on Julia will have a way to introspect into the garbage collecting system or do custom garbage collection to see if references are unique. What makes me nervous is that I don't want users to think they should try to dot everything just in case the undotted code is slow; but we can lead there by showing examples.

Regarding `.*` for ITensor contractions, I'm supportive of it, so let's go ahead. I suppose the only reason not to would be because it would mean a different thing for Julia arrays? But then like you said, we are already interpreting `*` differently for ITensors, so I think it's very sensible to have `.*` mean the broadcasted version of the same thing.

To offer an element-wise broadcasting operation, we could overload a different symbol for it. I believe there are a number of available mathematical symbols for overload, such as a circle with a dot in the center, etc. That might even be closer to the math tensor literature notation. And of course we could just have a function for that operation when it's not occurring in a broadcasting operator expression.

Thanks for explaining about the indices -

mtfishman:
> Agreed that it's doing something we don't even have in C++. I was more saying I wish there was a better approach to this sort of thing that either Julia's or C++; there could be, I think, if language designers thought about this case ahead of time. Maybe later on Julia will have a way to introspect into the garbage collecting system or do custom garbage collection to see if references are unique. What makes me nervous is that I don't want users to think they should try to dot everything just in case the undotted code is slow; but we can lead there by showing examples.

I agree it would be nice if there was better reference counting functionality in Julia, as well as the ability overloading `=` and `*=`. 

Definitely giving good examples will be useful, and eventually a lot of the usage of dots can be replaced by ITensor-level lazy operations (I think basically everything except for `.=`, `.*=` and `.+=` could be).

> 
> Regarding `.*` for ITensor contractions, I'm supportive of it, so let's go ahead. I suppose the only reason not to would be because it would mean a different thing for Julia arrays? But then like you said, we are already interpreting `*` differently for ITensors, so I think it's very sensible to have `.*` mean the broadcasted version of the same thing.

Yes, that is the only reason. There is some interpretation that the dot means an element-wise operation, and contraction is not quite that, but it seems ok to stretch the definition since contraction is clearly much more useful for ITensors.
 
> 
> To offer an element-wise broadcasting operation, we could overload a different symbol for it. I believe there are a number of available mathematical symbols for overload, such as a circle with a dot in the center, etc. That might even be closer to the math tensor literature notation. And of course we could just have a function for that operation when it's not occurring in a broadcasting operator expression.

Yeah, it is not something that has even really come up so far, but it would be pretty easy to overload another symbol for that operation. [Wikipedia](https://en.wikipedia.org/wiki/Hadamard_product_(matrices)#Definition) uses an open circle or an open circle with a dot for a Hadamard product.

> 
> Thanks for explaining about the indices -

PR #226, opened 2020-03-11
Add more broadcast operations

mtfishman:
This replaces uses of `add!`, `mul!` and `scale!` with the equivalent broadcast operations throughout the library. It also adds more broadcast operations, for example elementwise function operations (like `A .^ 2` to square each element of the ITensor).


emstoudenmire:
Looks good, especially the changes to iterativesolvers.jl

PR #233, opened 2020-03-23
Remove prime level from TagSet

mtfishman:
This removes the prime level from the TagSet and moves it into the Index.

There is also some code cleanup. Some notation is added to various functions to use keyword args to find/filter for indices with specified tags and prime levels, for example:
```julia
julia> i = Index(2; tags="Site", plev=2)
(dim=2|id=725|"Site")''

julia> j = Index(2; tags="Site", plev=1)
(dim=2|id=349|"Site")'

julia> is = IndexSet(i,j)
(dim=2|id=725|"Site")'' (dim=2|id=349|"Site")' 

julia> isp = prime(is; tags="Site", plev=2)
(dim=2|id=725|"Site")''' (dim=2|id=349|"Site")' 

julia> findindex(isp; plev=3)
(dim=2|id=725|"Site")'''
```

### DIFF
--- a/src/decomp.jl
+++ b/src/decomp.jl
@@ -9,7 +9,7 @@ function LinearAlgebra.qr(A::ITensor,
   tags::TagSet = get(kwargs,:tags,"Link,qr")
   Lis = commoninds(inds(A),IndexSet(Linds...))
   Ris = uniqueinds(inds(A),Lis)
-  Lpos,Rpos = getperms(inds(A),IndexSet(Lis...),IndexSet(Ris...))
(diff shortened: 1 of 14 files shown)


mtfishman:
The only breaking change here (should be) removing the notation `("Site",1)` that used to mean a TagSet with tag `"Site"` and prime level `1`, but I don't think that was used very much anyway.

emstoudenmire:
I think this is a good improvement in terms of better structuring of the code, and also makes user inputs to be a bit nicer (e.g. not having to think about the interplay of tags and primes). Good use of keyword arguments for the priming functions too. Thanks for the hard work on this!

PR #236, opened 2020-03-24
Replace davidson with eigsolve from KrylovKit in dmrg

### LINKED ISSUE #234: DMRG occasionally gives very wrong answer
I have found that DMRG occasionally gives a very wrong answer on a very simple Hamiltonian. I used the Hamiltonian H = -(sum from j = 1 to N)S_j^z (just a Pauli-z matrix on each particle, with no interactions between neighbors). The ground state energy is clearly -N, and usually the program gives this correctly, but the program will give an answer much lower than this (something like -1.5N to -3N) up to 20% of the time (for N = 5 to 20 it seems to be lower).

To show an example of this, I ran DMRG 100 separate times (all with different initializations, chosen using the command psi0 = randomMPS(sites)) for N = 100. Here is an example of what a correct run looked like for me:

Trial 2:
After sweep 1 energy=-84.934490501774 maxlinkdim=5 time=0.257
After sweep 2 energy=-95.364721763258 maxlinkdim=15 time=0.400
After sweep 3 energy=-99.349810066932 maxlinkdim=26 time=0.612
After sweep 4 energy=-99.958095815756 maxlinkdim=31 time=0.638
After sweep 5 energy=-99.994179503234 maxlinkdim=24 time=0.460
Final energy = -99.99417950323418
Expected energy = -100

And here is an example of what an incorrect run looked like for me:

Trial 3:
After sweep 1 energy=-84.296520611164 maxlinkdim=5 time=0.259
After sweep 2 energy=-92.686538471605 maxlinkdim=16 time=0.398
After sweep 3 energy=-96.351629915200 maxlinkdim=29 time=0.572
After sweep 4 energy=-98.323749683386 maxlinkdim=29 time=0.650
After sweep 5 energy=-240.380636427750 maxlinkdim=38 time=0.574
Final energy = -240.38063642775018
Expected energy = -100

One thing I've noticed is that sometimes the run goes haywire on an earlier sweep, and then somehow it magically gets fixed on the next sweep, such as here:

Trial 39:
After sweep 1 energy=-84.034964797986 maxlinkdim=5 time=0.260
After sweep 2 energy=-93.790455347323 maxlinkdim=17 time=0.406
After sweep 3 energy=-97.326033857518 maxlinkdim=31 time=0.664
After sweep 4 energy=-251.224945310846 maxlinkdim=38 time=0.905
After sweep 5 energy=-99.995652140835 maxlinkdim=27 time=0.639
Final energy = -99.99565214083535
Expected energy = -100

In fact, one thing I tried was to check for an "obviously wrong" answer and then run DMRG for an extra sweep if that was the case, and this always fixed the problem. Also, the choice of psi0 definitely has some role to play, because the DMRG algorithm doesn't have any randomness (I tried multiple runs with the same psi0 and got the exact same result every time).

Does someone know what might be causing this error? Any clarity would be greatly appreciated.

mtfishman:
This fixes #234. Besides that issue, I have seen that `eigsolve` and `davidson` give very similar results for DMRG, for example for `example/dmrg/1d_Heisenberg_dmrg.jl` (using a fixed seed):

This branch (dmrg using `eigsolve`)
```julia
After sweep 1 energy=-137.585233665598 maxlinkdim=9 time=0.152
After sweep 2 energy=-138.935109267445 maxlinkdim=20 time=0.354
After sweep 3 energy=-138.940079055055 maxlinkdim=92 time=1.360
After sweep 4 energy=-138.940086046771 maxlinkdim=100 time=3.040
After sweep 5 energy=-138.940086072140 maxlinkdim=96 time=3.309
Final energy = -138.940086072140
```
On master (dmrg using `davidson`)
```julia
After sweep 1 energy=-137.585233665598 maxlinkdim=9 time=0.148
After sweep 2 energy=-138.935109267445 maxlinkdim=20 time=0.430
After sweep 3 energy=-138.940079055055 maxlinkdim=92 time=1.312
After sweep 4 energy=-138.940086046771 maxlinkdim=100 time=3.415
After sweep 5 energy=-138.940086072140 maxlinkdim=96 time=3.808
Final energy = -138.940086072140
```
so it seems like a pretty safe replacement.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.1.0"
 
 [deps]
 HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
+KrylovKit = "0b1a1467-8014-51b9-945f-bf0ae24f4b77"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
@@ -22,4 +23,4 @@ Suppressor = "fd094767-a336-5f1f-9728-57cf17d0bbfb"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Combinatorics", "QuadGK","Suppressor","Test"]
+test = ["Combinatorics", "QuadGK", "Suppressor", "Test"]
--- a/examples/dmrg/1d_Heisenberg_dmrg.jl
+++ b/examples/dmrg/1d_Heisenberg_dmrg.jl
@@ -1,5 +1,6 @@
 using ITensors
 using Printf
+using Random
 
 # Use DMRG to solve the spin 1, 1D Heisenberg model on 100 sites
 # For the Heisenberg model in one dimension
--- a/src/ITensors.jl
+++ b/src/ITensors.jl
@@ -6,7 +6,8 @@ using Random,
       StaticArrays,
       TimerOutputs,
       Reexport,
-      HDF5
+      HDF5,
+      KrylovKit
 
 # TODO: move imports to individual files
 import Base.adjoint,
--- a/src/mps/dmrg.jl
+++ b/src/mps/dmrg.jl
@@ -10,8 +10,24 @@ function dmrg(H::MPO,
   quiet::Bool = get(kwargs,:quiet,false)
 
   # eigsolve kwargs
-  krylovdim = get(kwargs,:maxiter,2)+1
-  tol = get(kwargs,:errgoal,1E-14)
+  eigsolve_tol::Float64 = get(kwargs,:eigsolve_tol,1E-14)
+  eigsolve_krylovdim::Int = get(kwargs,:eigsolve_krylovdim,3)
+  eigsolve_maxiter::Int = get(kwargs,:eigsolve_maxiter,1)
+  eigsolve_verbosity::Int = get(kwargs,:eigsolve_verbosity,0)
+
+  # TODO: add support for non-Hermitian DMRG
+  ishermitian::Bool = true #get(kwargs,:ishermitian,true)
+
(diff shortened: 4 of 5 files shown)


mtfishman:
There are some questions this raises about the keyword arguments that get passed to `eigsolve`. 
The `maxiter` kwarg in `davidson` is nearly the same as the `krylovdim` kwarg in `eigsolve` (though it appears that they are off by one). I actually prefer the name `krylovdim` instead of `maxiter` since it is more descriptive. It is a bit confusing since `eigsolve` also has a `maxiter` kwarg which controls the number of restarts. 

I almost think we should be very explicit with the keyword arguments, for example name them `eigsolve_krylovdim`, `eigsolve_tol`, `eigsolve_maxiter`, etc (see [here](https://jutho.github.io/KrylovKit.jl/stable/man/eig/#Eigenvalues-and-eigenvectors-1) for the list of options). Then it is clear that these are keyword arguments that get passed to `eigsolve`.

mtfishman:
That sounds good to me, I can make those changes in this pull request. What if we remove `maxiter` right now, with an error directing people towards `eigsolve_krylovdim` (and explain the off-by-one issue), and discuss adding something like `cpp_maxiter` later?

mtfishman:
Probably most people aren't even setting `maxiter` themselves, so I doubt this will affect many people.

emstoudenmire:
I think that's a good plan. For the most part, people can either just learn the new approach right away or very easily migrate from the old set of arguments. (We can include this in the transition guide we should write on the website for migrating from C++ ITensor to Julia ITensor. Not too soon to start on that!)

mtfishman:
Ok, I've changed `errgoal` to `eigsolve_tol` and `maxiter` to `eigsolve_krylovdim`. This is good to go from my end if it looks good to you.

mtfishman:
By the way, is there a problem with Travis? Why aren't the checks showing up on pull requests?

emstoudenmire:
Not sure about the Travis issues but that's the second one I've seen. Another is that sometimes one build will succeed and the other fails (the Mac one fails). I did recently bump the Julia version to 1.4 so maybe that's partly the reason?

PR #238, opened 2020-03-25
Fix #231 combiner bug

### LINKED ISSUE #231: blockview issue arising in combiner
I am running some code which is a bit of an edge case, but a very important case. For background as to why it might be causing the bug below, it acts an MPO on a very simple product-state MPS (spinless fermion hopping Hamiltonian on a state with just a single fermion on one site).

At the end of doing naiveApplyMPO to act this MPO onto the MPS, the code calls `orthogonalize!` which puts the block-sparse tensor shown below into the `svd` function. The error then occurs at the beginning of `svd` where it makes the row and column indices with combiners.

The error is: "LoadError: Block must be structurally non-zero to get a view"

And the tensor going into the SVD is:

M[b] = ITensor ord=3

IndexSet{3,Index{Array{Pair{QN,Int64},1}}}
Dim 1: (dim=8|id=323|"CMB,Link") <Out>
 1: QN(("Nf",-1,-1)) => 2
 2: QN(("Nf",0,-1)) => 4
 3: QN(("Nf",1,-1)) => 2

Dim 2: (dim=2|id=904|"Fermion,Site,n=4") <Out>
 1: QN(("Nf",0,-1)) => 1
 2: QN(("Nf",1,-1)) => 1

Dim 3: (dim=5|id=235|"Link,u") <Out>
 1: QN(("Nf",1,-1)) => 2
 2: QN(("Nf",0,-1)) => 1
 3: QN(("Nf",1,-1)) => 2

BlockSparse{Float64,Array{Float64,1},3}
 8×2×5

Block: (2, 1, 2)
 [3:6, 1:1, 3:3]
[:, :, 1] =
 0.0
 1.0
 0.0
 0.0

Block: (1, 2, 2)
 [1:2, 2:2, 3:3]
[:, :, 1] =
 1.0
 0.0

Block: (2, 2, 3)
 [3:6, 2:2, 4:5]
[:, :, 1] =
 0.0
 0.0
 0.0
 0.0

[:, :, 2] =
 0.0
 0.0
 0.0
 0.0

Apparently there are some blocks which are numerically zero, but I'm unclear as to what is making the "pos" argument of `blockview` be `nothing`, which is what triggers the error condition in `blockview`. Thanks for the help with this!

mtfishman:
This fixes issue #231.

### DIFF
--- a/src/Tensors/blocksparse/blocksparsetensor.jl
+++ b/src/Tensors/blocksparse/blocksparsetensor.jl
@@ -1,5 +1,6 @@
 export BlockSparseTensor,
-       blockview
+       blockview,
+       addblock!
 
 #
 # BlockSparseTensor (Tensor using BlockSparse storage)
@@ -334,11 +335,6 @@ function combine_dims(blocks::Blocks{N},
   blocks_comb = Blocks{N-NC+1}(undef,nnzblocks(blocks))
   for (i,block) in enumerate(blocks)
     blocks_comb[i] = combine_dims(block,inds,combdims)
-    #slice = getindices(block,combdims)
-    #slice_comb = LinearIndices(nblcks)[slice...]
-    #block_comb = deleteat(block,combdims)
-    #block_comb = insertafter(block_comb,tuple(slice_comb),minimum(combdims)-1)
-    #blocks_comb[i] = block_comb
   end
   return blocks_comb
 end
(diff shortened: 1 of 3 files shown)


mtfishman:
@emstoudenmire, could you double check that this branch fixes your issue?

emstoudenmire:
Thanks for working on this; it will be helpful to have fixed for the work on fermions I'm doing.

Unfortunately, with this branch I'm getting a different error now "LoadError: DimensionMismatch("array could not be broadcast to match destination")"

The minimal code attached to #238 for example gives this error, as does my more extended code I based that code on.

emstoudenmire:
I should add that error is coming from the svd. Here is a stack trace:

Stacktrace:
 [1] check_broadcast_shape at ./broadcast.jl:509 [inlined]
 [2] check_broadcast_axes at ./broadcast.jl:512 [inlined]
 [3] instantiate at ./broadcast.jl:259 [inlined]
 [4] materialize! at ./broadcast.jl:823 [inlined]
 [5] _svd_no_truncate(::Tensor{Float64,2,BlockSparse{Float64,Array{Float64,1},2},IndexSet{2,Index{Array{Pair{QN,Int64},1}}}}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /Users/mstoudenmire/software/jitensor/src/Tensors/blocksparse/linearalgebra.jl:241
 [6] _svd_no_truncate at /Users/mstoudenmire/software/jitensor/src/Tensors/blocksparse/linearalgebra.jl:167 [inlined]
 [7] #svd#126 at /Users/mstoudenmire/software/jitensor/src/Tensors/blocksparse/linearalgebra.jl:23 [inlined]
 [8] svd(::Tensor{Float64,2,BlockSparse{Float64,Array{Float64,1},2},IndexSet{2,Index{Array{Pair{QN,Int64},1}}}}) at /Users/mstoudenmire/software/jitensor/src/Tensors/blocksparse/linearalgebra.jl:18
 [9] svd(::ITensor{3}, ::IndexSet{2,Index}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /Users/mstoudenmire/software/jitensor/src/decomp.jl:101
 [10] svd(::ITensor{3}, ::IndexSet{2,Index}) at /Users/mstoudenmire/software/jitensor/src/decomp.jl:87
 [11] orthogonalize!(::MPS, ::Int64; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /Users/mstoudenmire/software/jitensor/src/mps/mpo.jl:538

... (not the whole stack trace but just the top)

mtfishman:
Very strange, the test I added [here](https://github.com/ITensor/ITensors.jl/pull/238/files#diff-4aeb09dea7b467ec0adfee802e9ed635R972) was supposed to reproduce svding the tensor you posted in that issue. Is there something off about that test from the svd that is causing the error?

emstoudenmire:
Do you think it could be a mismatch somehow in the index of blocks from the storage (storage of tensor values)? That could explain why the tensor made from hand in the test could be in a different internal state from the one "in the wild". Let me print out more of the internals of that tensor and post to help with that

mtfishman:
One thing was that it didn't look like the index directions were printed in your example or how the indices were grouped in the svd, that would be useful. I tried multiple combinations and they all worked for me.

emstoudenmire:
Yes, I'd be happy to provide those if the error returns to being with that other tensor. Now the error is coming from a different tensor, and after studying this new tensor a bit, I've found that one of its non-zero blocks has a different flux from the others. So that's a problem right there, barring other errors that may come after. I'll see if I can find out what's causing it to have that incorrect property.

(Do you know of a good system for turning on more stringent run-time checking in Julia, such as a "debug mode" where we can say always check that each newly created block of an ITensor has the same flux as the others, no matter how it was created?)

emstoudenmire:
By the way, my new bug could be totally unrelated to your fix. It could have just uncovered another bug. So if the unit test you made *fails* with the master branch before this patch, and now clearly succeeds after, I'm happy to merge this patch and just start working on the new bug as a separate issue (which could be coming from something as silly as a problem with my hand-made MPO).

mtfishman:
Yes, this patch definitely fixes what was previously bug, and is a clear improvement on that part of the combiner code, so we should merge this and continue to investigate other issues separately.

PR #239, opened 2020-03-25
Make IndexSet storage immutable (use SVector storage)

mtfishman:
This changes the storage of IndexSet from SizedVector to SVector, which means that IndexSet is fully immutable now (both the type and the storage are immutable). It also means that they are now non-allocating:
```julia
julia> using ITensors

julia> i = Index(2,"i");

julia> j = Index(3,"j");

julia> @btime IndexSet($i,$j)
  0.089 ns (0 allocations: 0 bytes)
(dim=2|id=86|"i") (dim=3|id=4|"j") 
```
while on master it is:
```
julia> using ITensors

julia> i = Index(2,"i");

julia> j = Index(3,"j");

julia> @btime IndexSet($i,$j)
  158.176 ns (3 allocations: 256 bytes)
(dim=2|id=653|"i") (dim=3|id=556|"j") 
```
Using the SizedVector storage wasn't ideal, since it is just a wrapper around a Vector so it is technically still dynamically sized and allocated. MVector doesn't make sense as a storage since for QNs, the indices have dynamically sized containers (the QN storage), so MVector doesn't generically allow modifying the elements anyway.

The main disadvantage of this is that you can't set elements of an IndexSet using notation like `is[2] = j`, but this should mostly be something that is done internally and can easily be replaced by `is = setindex(is,j,2)`.

Note that indices of an ITensor can still be modified in-place, for example with functions like `prime!(A)`, by replacing the entire IndexSet of the ITensor (like how certain in-place operations of the ITensor are done by replacing the storage). So the ITensor interface should not change from this.

### DIFF
--- a/src/indexset.jl
+++ b/src/indexset.jl
@@ -7,11 +7,8 @@ export IndexSet,
        indexposition,
        not,
        swaptags,
-       swaptags!,
        swapprime,
-       swapprime!,
        mapprime,
-       mapprime!,
        commoninds,
        commonindex,
        uniqueinds,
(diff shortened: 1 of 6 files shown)


mtfishman:
Right. The only downside is that the way those functions are written now, it is looping over calls to `setindex`, which creates new IndexSets inside the loop. This was just a simplification for the sake of getting it implemented, but there are some options for optimizing it. We could go with the strategy of using a temporary MVector, modifying it, and then creating an IndexSet from the MVector at the end. Alternatively, there should be a way to make the IndexSet all in one go using the functionality introduced in #241. Just pointing out that the current implementation has some potential for improvement.

PR #242, opened 2020-03-27
Copying replace index functions

emstoudenmire:


emstoudenmire:
@mtfishman please let me know if this is superseded by #241, or if you could just include this change in that PR. (Since you are working on IndexSet functions there and renamed these too.)

mtfishman:
Hey Miles,

I actually think that this may not be the definition we want for `replaceinds`, since `copy(::ITensor)` copies the indices and the storage. I believed that we had settled on the behavior that functions that just manipulate the indices (like `prime(::ITensor)`) should not copy the storage data, for efficiency.

If that is the case (and I still agree with that approach), I think the better way to implement this is by having a `replacinds`/`replacinds!` function that works on the IndexSet level, and then the ITensor function calls that one and swaps out the modified IndexSet.

 #242 doesn't affect `replaceinds!`, so these should be fairly orthogonal, though #239 does change the implementation of `replaceinds!` so may affect this PR.

-Matt

emstoudenmire:
Agreed - I made these changes too hastily based on a superficial reading of our Slack conversation earlier. Appreciate the reminder that indeed we are not having these functions copy the storage. I'm going to just close this PR and resubmit a similar and better one after #241 is merged.

Issue #90, opened 2019-08-20
Factorize interface


emstoudenmire:
The factorize function doesn't always seem to generate a unitary U as the first return value. Hopefully I am not doing something wrong here, but here is some code to reproduce what I found:
```
l = Index(4,"Link")
s = Index(2,"Site")
T = randomITensor(l,s)
U,R = factorize(T,s)

@show U*prime(U,"Link")  # not equal to the identity
```

mtfishman:
Hey Miles,

I explained the `factorize` function wrong to you before. What `factorize` does is the following:

```julia
i = Index(2,"i")
j = Index(2,"j")
T = randomITensor(i,j)

# Make A = U and B = S*V
A,B = factorize(T,i,dir="fromleft")

# Make A = U*S and B = V
A,B = factorize(T,i,dir="fromright")

# Make A = V*sqrt(S) and B = sqrt(S)*U
A,B = factorize(T,i,dir="center")
```

Right now, the default is "center", so you are actually getting the correct result.

Then, `factorize` also has two modes, `which_factorization = "svd"` which uses an SVD to get `U,S,V`, and the `which_factorization = "eigen"` which, in the case of "fromleft" or "fromright", uses an eigendecomposition of the density matrix to get the `U` and `S*V` factors (or `U*S` and `V` factors).

The idea is that all of these are different versions of a low rank factorization of a matrix.

Cheers,
Matt

mtfishman:
Perhaps this is too much functionality packed into one function, and as we were discussing the "fromleft" and "fromright" may be superfluous because you could instead use the other set of indices to get the "fromright" factorization. I guess these two are essentially the same:
```julia
A,B = factorize(T,i,dir="fromright")
```
and
```julia
A,B = factorize(T,j,dir="fromleft")
```

emstoudenmire:
Yes, let's think about this some more, because I think it could become an easier function to use if we simplified it and even split it into separate functions perhaps.

One thing to observe is that, in addition to having the freedom of which indices to put in the arguments, one can also switch the order of the result tensors on the left. So given all of that freedom already, I think the keyword arguments could be dropped.

Then to switch between the fromright/fromleft functionality (which could just be combined into one call, where you switch between these two by which indices you put or how you order the return variables), and between the "center" functionality, we could just make the "center" one its own function. 

So what do you think of that idea? Two functions like this (but with a better names):

* `Q,B = factor(T,i)  # does Q = U, B = S*V`
[...]

PR #246, opened 2020-03-31
Change factorize keyword argument api

mtfishman:
This makes some changes to the `factorize` api based on issue #90.

The change in this pull request is the based on the final few suggestions:
```
A,B = factorize(T, i; ortho="left") # A = U , B = S*V, default
A,B = factorize(T, i, ortho="right")  # A = U*S , B = V
A,B = factorize(T, i; ortho="none")  # A = U*sqrt(S) , B = sqrt(S)*V, or some other factorization
```
The decomposition used is chosen with the `which_decomp` keyword argument (perhaps we could shorten it to `decomp`, suggestions are welcome). `which_decomp` can be `"svd"`, `"eigen"`, or `"automatic"`, in which case `"svd"` or `"eigen"` is chosen depending on the provided cutoff. If eigen is chosen, `ortho="none"` doesn't work, since it is not so natural to calculate that decomposition using eigendecompositions (I think it would require doing two eigendecompositions, at which point you may as well do an svd).

The idea is that this interface would allow for other factorizations to be available, for example `qr`, `lq`, `polar`, `cholesky`, `lu`, etc. These would be chosen consistently with the `ortho` keyword (some orthogonalize and some don't). Some automation could occur to allow for a smart choice of decomposition, for example if no truncation is requested and orthogonalization is requested, a `qr` could be used instead of `svd`/`eigen` since it is faster. Or if no orthogonalization is requested, an `lu` decomposition could be chosen by default, and if `ishermitian=true` then `cholesky` could be used.

### DIFF
--- a/src/decomp.jl
+++ b/src/decomp.jl
@@ -229,7 +229,7 @@ function factorize_svd(A::ITensor,
                        Linds...;
                        kwargs...)
   ortho::String = get(kwargs, :ortho, "left")
-  tags::TagSet = get(kwargs, :tags, "Link,u")
+  tags::TagSet = get(kwargs, :tags, "Link,fact")
   U,S,V,spec,u,v = svd(A, Linds...; kwargs...)
   if ortho == "left"
     L,R = U,S*V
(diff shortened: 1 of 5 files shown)


emstoudenmire:
Looks good I think; I'd say this is ready to merge. One minor improvement would be a discussion of the ortho keyword in the doc string for factorize. (And shifting the emphasis of the usage toward that over the which keyword, which is more advanced I would think.)

mtfishman:
Yes good point about that doc string, I changed it to emphasize `ortho` more. It could still be improved, but at least it has the right information now.

PR #250, opened 2020-04-01
always_copy keyword

### LINKED ISSUE #204: `dag` copying behavior
This is just a reminder for a discussion that occurred in a previous PR- https://github.com/ITensor/ITensors.jl/pull/201#issuecomment-587676621 . 
It was discussed that:
1. `dag` is only sometimes making a copy of the internal storage (when the storage is complex).
2. This should be noted in the docstring. 
3. It would be good to add a `always_copy` keyword argument to `dag` that can be used for generic code where a copy is needed.

emstoudenmire:
This PR adds the always_copy keyword to `Base.conj` methods for Tensors types and for the ITensor `dag` function. Fixes #204


src/Tensors/blocksparse/blocksparse.jl
```diff
@@ -72,8 +72,17 @@ end
...
-Base.conj(D::BlockSparse) = BlockSparse(conj(data(D)), copy(blockoffsets(D)))
+
+function Base.conj(D::BlockSparse{<: Real}; always_copy = false) 
+  if always_copy
+    return conj!(copy(D))
+  end
+  return D
+end
+
+Base.conj(D::BlockSparse; kwargs...) = BlockSparse(conj(data(D)), copy(blockoffsets(D)))
```
mtfishman, line 83: Looks good, thanks. One thing this made me wonder about is if this should copy the block-offset list. Copying the block-offset list, in the limit of small blocks, could be as expensive as copying the data. Perhaps it should be `opt-in` with the `always_copy` keyword, and removed from this version of `conj`.

emstoudenmire: Agreed; the latest commit below should fix

emstoudenmire: I ran the tests manually and they are passing

PR #254, opened 2020-04-03
Remove call to qr in orthogonalize! so it works with qns

mtfishman:


emstoudenmire:
Would this be a good use case for `factorize`? Because then later we could just upgrade `factorize` for QN ITensors to use QR internally and then this code will automatically speed up from that.

emstoudenmire:
Nice to see Travis is working again too for some reason

mtfishman:
Yes that's a good point, let's do that.

I was wondering if you did something to get it back, but I guess it was just taking a break.

mtfishman:
The test failed because it ran out of time on the mac build (not sure why it took so long...). Besides that, it is passing, I'll merge this.

PR #256, opened 2020-04-04
Bunch of misc. tests for ITensors and a bugfix

kshyatt:
Feel free to squash this!

Had to change the `exphermitian` export location because otherwise the method for `Tensor`s wasn't in the method table.


mtfishman:
Looks good, thanks Katie!

PR #260, opened 2020-04-06
CP-decomposition by alternating least-squares (ALS)

nlw0:
This is an implementation of the classic Alternating Least-Squares (ALS) algorithm for CP-decomposition (aka Canonical Polyadic/CANDECOMP/PARAFAC), as described in "Tensor Decompositions and Applications", Kolda & Bader 2009, Figure 3.3.

The test only demonstrates that it seems to work for random 2x2x2 matrices, and further tests may be necessary. The implementation should support generic tensor orders and sizes, though.

This code was based on a [prior Python+NumPy implementation](https://www.kaggle.com/nicw102168/rank-of-random-2x2x2-tensors), and can probably be vastly improved to make a better use of ITensor features.

### DIFF
--- a/src/cp_als.jl
+++ b/src/cp_als.jl
@@ -1,0 +1,80 @@
+export cp_als
+
+## The Alternating Least-Squares (ALS) algorithm for CP-decomposition with ITensors.jl
+# Based on "Tensor Decompositions and Applications", Tamara G. Kolda and Brett W. Bader. SIAM Review 51(3), 2009
+# Porting the code from https://www.kaggle.com/nicw102168/rank-of-random-2x2x2-tensors
+
+using ITensors
+using ITensors.LinearAlgebra
+
+
+function cp_als(X, R; maxiter=1000, test_period=1, tol=1e-10)
+    # initialize with spheric-random colunms
+    A = map([randomITensor(Iₙ, Index(R)) for Iₙ in X.inds]) do An
+        aλ = columnnorms(An)
+        rr = Index(R)
+        iλ = diagITensor(aλ.^-1, rr, An.inds[2])
+        replaceind(An, An.inds[2], rr) * iλ
+    end
+    cp_als_(X, R, A, maxiter=maxiter, test_period=test_period, tol=tol)
+end
+
+function cp_als_(X, R, A; maxiter=1000, test_period=1, tol=1e-10)
+    N = order(X)
+
+    rind = [An.inds[2] for An in A]
+    λ = nothing
+    iteration = nothing
+
+    for it in 1:maxiter
+        iteration=it
+
+        for n in 1:N
+            V = reduce((a,b)->a.*b, array(A[m])' * array(A[m]) for m in 1:N if m != n) ## is there a better, ITensor way to do that?
+            Vinv = ITensors.LinearAlgebra.pinv(V)
+
+            W = reduce(khatrirao, [A[m] for m in N:-1:1 if m != n])
+            Xn = unfold(X,n)
+            qq = diagITensor(ones(Xn.inds[2].space), Xn.inds[2], W.inds[1])
+
+            rr = Index(R)
+            newAn = Xn * qq * W * itensor(Vinv, W.inds[2], rr)
+
+            aλ = columnnorms(newAn)
+            λ = diagITensor(aλ, rind...)
+            iλ = diagITensor(aλ.^-1, rr, A[n].inds[2])
+
+            A[n] = newAn * iλ
+        end
+
+        if it % test_period==0
+            X̂ = prod(A) * λ  # calculate PARAFAC model
+
+            if norm(X - X̂) < tol
+                break
+            end
+        end
+    end
+
+    return λ, A, iteration
+end
+
+"""Mode-n matricization of tensor X, or X₍ₙ₎ in Kolda & Bader 2009."""
+function unfold(X, n)
+    matn,_ = combiner(uniqueinds(X, IndexSet(X.inds[n])))
+    X*matn
+end
+
+"""The Khatri-Rao product, or A ⊙ B."""
+function khatrirao(A, B)
+    K = A.inds[2].space
+    AB = A * B * diagITensor(ones(K), A.inds[2], B.inds[2], Index(K))
+    matk,_ = combiner(B.inds[1], A.inds[1])
+    matk * AB
+end
+
+function columnnorms(A)
+    aA = array(A)
+    R = size(aA)[2]
+    [norm(aA[:,r]) for r in 1:R]
+end
--- a/src/cp_als.jl
+++ b/src/cp_als.jl
@@ -10,11 +10,11 @@ using ITensors.LinearAlgebra
 
 function cp_als(X, R; maxiter=1000, test_period=1, tol=1e-10)
     # initialize with spheric-random colunms
-    A = map([randomITensor(Iₙ, Index(R)) for Iₙ in X.inds]) do An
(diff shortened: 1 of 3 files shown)


emstoudenmire:
Thanks! I have been eager to make ITensor more useful and connected to the wider applied math and machine learning community, so having an ALS CP code is very welcome. Please give us a bit of time to review your code and we'll get back to you. 

One quick comment is that it may not be necessary to explicitly unfold tensors into matrices within ITensor, if it's for the purpose of contracting them with other tensors (or other matrices or vectors), or for the purpose of computing matrix factorizations such as SVD or QR. (The reason is that ITensor automatically handles mappings to matrices and back for both types of operations.) But I could imagine an explicit unfolding function could be necessary if you need to do other matrix operations not yet provided at the level of the ITensor interface.

nlw0:
Hi Miles! I'm sure I'm far from being able to write idiomatic ITensor code, and I'm looking forward to your review. My priority was just porting my old code making sure it works, and refactor it later. I bet my uses of diagonal tensors and combiners can be completely rewritten. Perhaps I can comment the code some more to help rewriting it? While it is short, it's maybe a little "dense", especially if you are only used to matrix algorithms like myself...

Refactoring this will probably be very instructive to a newbie like me, once we finish I can go back and add what I didn't know to the basic examples.

src/cp_als.jl
```diff
@@ -0,0 +1,80 @@
...
+
+    rind = [inds(An)[2] for An in A]
+    λ = nothing
+    iteration = nothing
+
+    for it in 1:maxiter
+        iteration=it
+
+        for n in 1:N
+            V = reduce((a,b)->a.*b, array(A[m])' * array(A[m]) for m in 1:N if m != n) ## is there a better, ITensor way to do that?
```
mtfishman, line 33: Could you explain the equation you are trying to compute here?

src/cp_als.jl
```diff
@@ -0,0 +1,80 @@
...
+"""The Khatri-Rao product, or A ⊙ B."""
+function khatrirao(A, B)
+    K = inds(A)[2].space
+    AB = A * B * diagITensor(ones(K), inds(A)[2], inds(B)[2], Index(K))
+    matk = combiner(inds(B)[1], inds(A)[1])
+    matk * AB
+end
+
+function columnnorms(A)
+    aA = array(A)
```
mtfishman, line 77: We generally try to avoid converting to `array`.

You can directly do this in ITensor by projecting onto a state (unfortunately we don't have great slicing functionality at the ITensor level right now), so for example:
```julia
function slicenorm(A::ITensor, i::Index, val::Int)
  return norm(A * onehot(i => val))
end
slicenorms(A::ITensor, i::Index) = [slicenorm(A, i, n) for n in 1:dim(i)]
```

src/cp_als.jl
```diff
@@ -0,0 +1,80 @@
...
+end
+
+"""Mode-n matricization of tensor X, or X₍ₙ₎ in Kolda & Bader 2009."""
+function unfold(X, n)
+    matn = combiner(uniqueinds(X, IndexSet(inds(X)[n])))
+    X*matn
+end
+
+"""The Khatri-Rao product, or A ⊙ B."""
+function khatrirao(A, B)
```
mtfishman, line 69: Could you explain a bit more about what this function is doing?

src/cp_als.jl
```diff
@@ -0,0 +1,80 @@
...
+    K = inds(A)[2].space
```
mtfishman, line 70: It is best to avoid accessing internals like this, you can use the `dim` function to get the dimension of the space of an Index.

src/cp_als.jl
```diff
@@ -0,0 +1,80 @@
...
+        for n in 1:N
+            V = reduce((a,b)->a.*b, array(A[m])' * array(A[m]) for m in 1:N if m != n) ## is there a better, ITensor way to do that?
+            Vinv = ITensors.LinearAlgebra.pinv(V)
+
+            W = reduce(khatrirao, [A[m] for m in N:-1:1 if m != n])
+            Xn = unfold(X,n)
+            qq = diagITensor(ones(inds(Xn)[2].space), inds(Xn)[2], inds(W)[1])
+
+            rr = Index(R)
+            newAn = Xn * qq * W * itensor(Vinv, inds(W)[2], rr)
```
mtfishman, line 41: Could you explain what this operation is doing?

mtfishman:
Hi @nlw0, sorry for the slow response on this.

I think it would help to have an explanation of the equations that different steps of the code are trying to implement, so we can determine if there is a better way to translate them to ITensor code. I am not so familiar with details of solving the CP-decomposition with ALS.

As a general strategy, it is best to avoid calling internals, for example use `dim(i)` instead of `i.space`. Also, it is generally best to avoid grabbing indices in a certain position, like `inds(A)[1]`. Instead, you can almost always determine the index you want through set functions, like `commonind(A, B)` if the index you want is shared with another tensor `B` in your network (alternatively, you might know that it is not shared with some other tensors `B` and `C`, and use `uniqueind(A, B, C)`).

Also, something we could do is have this in a small external package, like `ITensorCP`, since it is a bit orthogonal to the other functionality in the codebase.

nlw0:
Hi @mtfishman ! No worries, I wrote this mostly as an exercise to learn more about ITensors and about the subject itself, but I hope this might be useful to others since it seems to be a classic algorithm in the field. Whether it belongs here along `iterativesolvers.jl`, as an example, test or separate package is totally up to you.

I'll do my best to try to remember how it all works. Please keep in mind that this code is potentially very bad and inefficient right now, and the proper implementation might end up looking very different.

src/cp_als.jl
nlw0: If memory serves me right, the way the algorithm works is that tensors get "matricized" in different ways, producing 2D matrices where each column contains "unrolled" slices of the tensor. Once you have these two matrices, you can compute this product. Following the Kolda-Bader paper, section 2.6, given inputs of IxK and JxK elements, we produce an output with (I*J)xK elements. Each column has all the element-wise products of the two corresponding input vectors, essentially and unrolled version of the outer products of the column pairs.

I believe I have native Julia and Python versions of this code available somewhere, perhaps it might help looking at that?

src/cp_als.jl
nlw0: I tried to closely follow the algorithm presented in Kolda&Bader, Figure 3.3 here https://www.kolda.net/publication/TensorReview.pdf

A(n) are matrices of InxR, where R is the order of the approximation. This loop goes through each dimension, and V is computed as the Hadamard product of the the matrix products of every other dimension, skipping the current `n`.

Therefore `[array(A[m])' * array(A[m]) for m in 1:N if m != n]` is this collection of the matrix products by themselves, skipping the current dimension. We take that, and compute the total Hadamard product. We could probably define a `hadamard_collection(xx) = reduce(.*, xx)` or something like that.

V is an array of

src/cp_als.jl
nlw0: Here we compute the second row in the loop of the Figure 3.3 algorithm. The `W` is the central part in the expression where we compute the Khatri-Rao products across the matrices, skipping the current dimenson, similar to before. `Vinv` and Xn are the other clear elements in that expression. The rest I'd have to think more to remember, it's just what I figured out was a way to get all of this together... I suppose it must be possible to write simply as `Xn * W * Vinv`, but I messed up the tensor shapes somehow.

src/cp_als.jl
nlw0: added the code

PR #264, opened 2020-04-09
Code reorganization and new docs

mtfishman:
This is mostly a code reorganization. It also incorporates some api changes and new doc strings.

The main change is that the Tensors module is included into ITensors with a `using` statement, instead of using `Reexport`. This means that functions and types from Tensors are not exported when someone calls `using ITensors` (right now, one would have to do `using ITensors.Tensors` to pull `Tensors` symbols into the namespace). There are a few reasons for this. First, the `Tensors` layer should only be an advanced feature when using ITensors, so I think people should have to explicitly pull it in if they know what they are doing. Second, it helps right now since it makes the "official" (exported) interface of ITensors.jl smaller, and people should expect more stability in ITensors than in Tensors. This is pretty breaking, but most problems should be easily fixed by just adding `using ITensors.Tensors` along with `using ITensors`.

There are a few other small api changes, like introducing the constructor `tensor(::TensorStorage, inds)` to make a Tensor that stores a view of the TensorStorage, and some other changes at the Tensors level (like changing the name of `block(::Tensor, ::Int)` to `nzblock(::Tensor, ::Int)` to get the `n`th block of the Tensor in the blockoffset list).


mtfishman:
This brings us closer to being ready to register ITensors.jl as well as Tensors.jl, and forces people to explicitly use the Tensors module if they want to interact with the Tensor layer. Perhaps this would be a good time to discuss a name for the Tensors module. @emstoudenmire, you had suggested:

- Titan
- Tent
- Hypercube

Some more descriptive names might be:

- NDTensors
- SparseTensors
- BlockTensors
- TensorBase
- Multilinear

I think Titan sounds good, and I like that it is short. I am generally more drawn to descriptive names, but none of the ones I can think of really stand out to me. NDTensors might be my favorite, and contrasts from the existing Julia package `Tensors.jl` which focuses more on symmetric low-dimensional tensors. Let me know what you think.

emstoudenmire:
I think this sounds good, and yes we do need to select another name besides Tensors.jl.

NDTensors is very good, because it's descriptive and reasonably catchy/concise. (Multilinear is my second favorite; more interesting name, but less literally descriptive. TensorBase is good but probably under-sells the library, which could be a good standalone library for many applications.) Titan.jl turns out to be taken already, at least on Github; not sure if it's registered in the Julia package system and appears to be defunct: https://github.com/malmaud/Titan.jl

I think you have the honors (or burden?) of selecting the name so I'll leave it up to you! I'm happy to act as a tie-breaker if you are stuck.

emstoudenmire:
I should point out there is an existing C++ project called NDTensor:
https://github.com/tschmidt23/NDTensor

It doesn't look very active though (only 109 commits; 5 stars; last commit in 2018).

mtfishman:
I don't think either of those are too concerning as conflicts (Titan.jl doesn't appear to be registered or actively developed, and NDTensor hasn't been updated in a while). Right now I am deciding between `Titan.jl` and `NDTensors.jl`. I'm leaning towards `NDTensors.jl`, I'll think about it over the weekend. Let me know if you are leaning heavily one way or another.

emstoudenmire:
Agreed, I don't think those projects are big enough that we should worry about the naming conflict.

mtfishman:
In the latest commits, I changed the name of the Tensors module to NDTensors. I also exported NDTensors from ITensors (just the module name, not all of its exports), so that when you call `using ITensors`, you can access NDTensors functions with the notation `NDTensors.contract`.

I also did some more code reorganization. The main one is introducing an `export.jl` file that stores all of the exports of the ITensors library, instead of spreading them out throughout the files (they are organized by file within `export.jl`). I think this is nice because it makes it easier to see if multiple files export the same function, and get an idea for the public interface in one place. I borrowed the idea from Julia's base library.

@emstoudenmire, from my end this PR is ready to go. Let me know if these changes are good with you. After this, I believe we are in a good position to register ITensors and NDTensors (besides online documentation), unless we have some last minute interface changes to make.

emstoudenmire:
Looks good! I'm supportive of the idea of exports.jl. I think all of the approaches (exports together, exports in files) have pros/cons but this one is probably the best now that we have the interface nearly set. Also agreed NDTensors is a very good choice of name for the inner library.

mtfishman:
Ok, I will merge this!

PR #274, opened 2020-04-13
Add documentation website build process with Documenter.jl

mtfishman:
This adds a build process for making a documentation website using Documenter.jl. It is a very rough draft right now, more docstrings were added in #264 which we can use once that it merged.

Note that in this PR, the README.md now just points to `docs/src/index.md`, which is the main page of the website that gets generated. It seems reasonable for these to match each other, and we should be able to have the examples run as tests as well.


mtfishman:
I'm going to merge this so that we can start filling out the documentation as we add more doc strings.

PR #277, opened 2020-04-15
Rename data storage of many types to `data`

mtfishman:
Rename the data storage field of `IndexSet`, `TagSet`, `MPO`, `MPS`, `QN` and `AutoMPO` to `data`, with a function to access it called `ITensors.data(...)` (for now it is not exported, to encourage people not to use it). This is breaking, if people were accessing the data storage for those types, but that would generally be discouraged anyway.

This also removes some exports, like `leftlim`, `setleftlim!`, etc. to discourage people from using them too much (perhaps we can bring back `leftlim`/`rightlim`, but people should think before setting the limits).


mtfishman:
I think this is good to go, I'll merge.

@emstoudenmire, I have looked over `exports.jl` to check that we are not exporting too many functions. If you have some time it would be nice if you could glance over it as well to see if there is anything you think we can remove. I am trying to err on the side of caution in terms of what we export, especially for functions that are mostly for developer level usage, in order to minimize the official interface and give us some more flexibility in the future.

emstoudenmire:
I just looked over exports.jl and I have to say I was actually surprised how few function names are exported, which I think is a very good thing. It looks like a reasonable set of functions & I don't think we are therefore committing to supporting too large an interface. (There's maybe a few we could quibble about if users really need them, such as lproj and rproj from projmpo.jl, but these also have rare names so shouldn't pollute anyone's namespace too much.)

mtfishman:
Thanks for taking a look. I agree. I think with functions like `lproj` and `rproj`, as long as we don't expect to change the behavior it is ok.

PR #278, opened 2020-04-15
Iterating of an Index iterates over the range of possible IndexVals

mtfishman:
Previously, iterating over an Index `i` iterated over the integers `1:dim(i)`. Instead, in this PR it iterates over the IndexVals `i(1)` to `i(dim(i))`. This seems more useful in the context of ITensors. This means you can do:
```julia
using ITensors
i = Index(2)
A = randomITensor(i,i')
for iv in i, ipv in i'
  @show A[iv,ipv]
end
```

I'm not sure how much this featured is used right now, but it would be nice to generally have better iteration features for iterating through IndexSets and ITensors as well (i.e. iterating over an IndexSet could iterate over all IndexVals of the IndexSet. I was even thinking we could define an `IndexValSet = IndexSet{<:IndexVal}`).

### DIFF
--- a/src/index.jl
+++ b/src/index.jl
@@ -387,12 +387,13 @@ Prime an Index using the notation `i^3`.
 """
 Base.:^(i::Index, pl::Int) = prime(i, pl)
 
-# Iterating over Index I gives
-# integers from 1...dim(I)
-# TODO: should this iterate through the IndexVals?
+"""
+Iterating over Index `I` gives the IndexVals
+`I(1)` through `I(dim(I))`.
+"""
 function Base.iterate(i::Index, state::Int = 1)
   (state > dim(i)) && return nothing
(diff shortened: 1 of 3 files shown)


mtfishman:
@emstoudenmire, this is good to go from my end. Let me know if you like this design. In theory we don't have to directly define iteration of an Index in this way at all, and could use a special function like `eachindexval(i::Index)` to create a special iterable object, but I think the notation is pretty clear and should be useful.

emstoudenmire:
I do like this design; it makes more sense than the original thing I put of just having the values be integers, because this way users can just plug the IndexVals straight into functions like setindex, and others. I'm good to merge this!

mtfishman:
Thanks for the input, I'll merge this.

PR #279, opened 2020-04-15
DMRG over sum of MPOs

emstoudenmire:
This PR adds a variant of DMRG which accepts a Vector of MPOs, treating them as if they were summed. The algorithm handles this summation efficiently through a type called `ProjMPOSum`, which wraps a collection of `ProjMPO` objects.


mtfishman:
Great, thanks for taking care of this. Looks like very minimal amount of code!

emstoudenmire:
Yes! Mostly just looping over the ProjMPO objects held inside the ProjMPOSum object.

PR #280, opened 2020-04-15
Make combiner just return ITensor, not (ITensor,Index)

mtfishman:
This changes the notation to just return an ITensor, not a tuple of the ITensor and combined Index.

This is pretty breaking, and a bit hard to give a good error message to catch it. Fortunately right now we haven't defined iteration over an ITensor, so the old notation gives the error:
```julia
julia> using ITensors

julia> i = Index(2)
(dim=2|id=564)

julia> C,c = combiner(i)
ERROR: MethodError: no method matching iterate(::ITensor{2})
Closest candidates are:
  iterate(::Core.SimpleVector) at essentials.jl:603
  iterate(::Core.SimpleVector, ::Any) at essentials.jl:603
  iterate(::ExponentialBackOff) at error.jl:253
  ...
Stacktrace:
 [1] indexed_iterate(::ITensor{2}, ::Int64) at ./tuple.jl:84
 [2] top-level scope at REPL[3]:1
```
We could overload `Base.iterate(::ITensor)` and for now give people an error message that says they may have been trying to use the old combiner notation and to explain the new interface.

### DIFF
--- a/src/decomp.jl
+++ b/src/decomp.jl
@@ -53,11 +53,13 @@ function LinearAlgebra.svd(A::ITensor,
   Lis = commoninds(A,IndexSet(Linds...))
   Ris = uniqueinds(A,Lis)
 
-  CL,cL = combiner(Lis...)
-  CR,cR = combiner(Ris...)
+  CL = combiner(Lis...)
+  CR = combiner(Ris...)
 
   AC = A*CR*CL
 
+  cL = combinedind(CL)
+  cR = combinedind(CR)
   if inds(AC) != IndexSet(cL,cR)
     AC = permute(AC,cL,cR)
   end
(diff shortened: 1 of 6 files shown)


emstoudenmire:
Looks good. Is there a new mechanism for getting the combined Index, such as a function one can call on the combiner ITensor afterward?

mtfishman:
The notation is `C = combiner(i); combinedind(C)` (that had already existed).

emstoudenmire:
Right, thanks. This looks good to go (though we may have to apologize to Katie once again!).

mtfishman:
Yeah... I was going back and forth about this, but at the end of the day it makes sense for an ITensor constructor to just output the ITensor, and this version is pretty clear.

Issue #286, opened 2020-04-16
Unclear Bug in Latest Version of ITensors


sujay-kazi:
I just downloaded the updated repository less than an hour ago (to be more precise, about 12:40 P.M. PDT). However, I noticed that some of my programs are now experiencing a bug that I have never seen before. In particular, when I tried to run my Schwinger model DMRG code, I got the following error:

```
ERROR: LoadError: MethodError: no method matching blockdim(::Index{Array{Pair{QN,Int64},1}}, ::Int64)
Closest candidates are:
  blockdim(::Array{Pair{QN,Int64},1}, ::Int64) at C:\Users\user\.julia\packages\ITensors\FjuEF\src\qn\qnindex.jl:10
Stacktrace:
 [1] qnblockdim(::Index{Array{Pair{QN,Int64},1}}, ::QN) at C:\Users\user\.julia\packages\ITensors\FjuEF\src\qn\qnindex.jl:136
 [2] (::ITensors.var"#defaultMat#313"{DataType})(::Index{Array{Pair{QN,Int64},1}}, ::Index{Array{Pair{QN,Int64},1}}, ::QN, ::QN) at C:\Users\user\.julia\packages\ITensors\FjuEF\src\physics\autompo.jl:481
 [3] #qn_svdMPO#300(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(ITensors.qn_svdMPO), ::AutoMPO, ::Array{Index{Array{Pair{QN,Int64},1}},1}) at C:\Users\user\.julia\packages\ITensors\FjuEF\src\physics\autompo.jl:487
 [4] qn_svdMPO at C:\Users\user\.julia\packages\ITensors\FjuEF\src\physics\autompo.jl:362 [inlined]
 [5] #MPO#330 at C:\Users\user\.julia\packages\ITensors\FjuEF\src\physics\autompo.jl:734 [inlined]
 [6] MPO(::AutoMPO, ::Array{Index{Array{Pair{QN,Int64},1}},1}) at C:\Users\user\.julia\packages\ITensors\FjuEF\src\physics\autompo.jl:732
 [7] top-level scope at D:\Sujay Google Drive\Sujay Data\MIT 2019-2020 (Junior)\Spring Semester\UROP Tensor Networks for Lattice QCD Calculations\Julia Files\Schwinger Model\Schwinger_DMRG.jl:64
[...]

PR #287, opened 2020-04-16
Fix missing NDTensors. in qnindex.jl

emstoudenmire:
This fixes an issue following the recent split of NDTensors.jl from ITensors.jl. I believe it fixes the bug #286 as well.

### DIFF
--- a/src/qn/qnindex.jl
+++ b/src/qn/qnindex.jl
@@ -133,7 +133,7 @@ dimension of the block of the QNIndex having
 QN equal to `q`. Assumes all blocks of `ind` 
 have a unique QN.
 """
-qnblockdim(ind::QNIndex,q::QN) = blockdim(ind,qnblocknum(ind,q))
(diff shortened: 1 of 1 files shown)


mtfishman:
It looks like the problem is that `blockdim` is an NDTensors function, but it is not always overloaded as `NDTensors.blockdim` elsewhere in `qnindex.jl`. I think the better way to handle this is to change those other overloads to `NDTensors.blockdim`.

mtfishman:
(So that we don't have two versions of `blockdim` running around, they will all be the one exported from `NDTensors` and we can just call it with `blockdim(...)`)

emstoudenmire:
That does sound better. The reason I hastily merged this PR is that there's some totally unrelated issue with the tests making the tests fail, so I went ahead and merged to deal with that issue in another PR.

I'll do another PR too right now to make qnindex.jl's blockdim functions just overloads of NDTensors.blockdim.

Thanks!

PR #288, opened 2020-04-16
Fix NDTensors.blockdim issue

emstoudenmire:
Fixes issues with qn/qnindex.jl blockdim definitions not overloading NDTensors.blockdim


mtfishman:
Thanks!

PR #290, opened 2020-04-17
Changes to MPS/MPO api

mtfishman:
This implements some of the proposed changes in issue #285. In particular, this includes:

- Changing `applympo` to `mul`/`*` 
- Changing `multmpo` to `mul`/`*`
- Changing `sum(::MPS/MPO, ::MPS/MPO)` to `add`/`+` (this leaves `sum(::Vector{MPS/MPO})` as is, which is more in line with the Julia convention that `sum` acts on a collection while `+` acts on a list).
- A keyword argument in `dot`/`inner` that lets you choose if the site indices get matched automatically, which is on by default.

Also, I noticed that `applympo` and `multmpo` did not have the same behavior as the C++ version. In the C++ version, `applympo(A, x) -> Ax` results in an MPS `Ax` with the site indices of `A` that are not common with `x`, and `multmpo` wasn't requiring that only one pair of site indices was common between the MPOs. Given the new API, these operations are now treated more literally as contractions between the input MPS/MPO. We can add APIs that do smarter things with the indices in a future PR, for example a function `product(::MPO, ::MPS)` that wraps `mul(::MPO, ::MPS)` and maps the indices back to the original ones, and `matmul(A::MPO, B::MPO)` that allows MPO inputs with all the same site indices and performs something like `mapprime(mul(prime(A), B), 2, 1)`.

This also adds some priming/tagging functions and docstrings for MPS/MPO functions, though more are needed.

### DIFF
--- a/docs/src/MPSandMPO.md
+++ b/docs/src/MPSandMPO.md
@@ -1,8 +1,60 @@
 # MPS and MPO
 
+## Types
+
+```@docs
+MPS
+MPO
+```
+
+## MPS Constructors
+
+```@docs
+MPS(::Int)
+MPS(::Type{<:Number}, ::Any)
+MPS(::Any)
+randomMPS
+productMPS
+```
+
+## MPO Constructors
+
+```@docs
+MPO(::Int)
+MPO(::Any, ::Vector{String})
+MPO(::Any, ::String)
+```
+
+## Properties
+
+```@docs
+length(::ITensors.AbstractMPS)
+maxlinkdim(::ITensors.AbstractMPS)
+```
+
+## Priming and tagging
+
+```@docs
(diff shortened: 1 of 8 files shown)


emstoudenmire:
This looks great

mtfishman:
Let me know if you are good with the API changes and I can merge this. I think we already discussed everything except changing `sum(::MPS, ::MPS)` to `add(::MPS, ::MPS)`/`+(::MPS, ::MPS)`. The reason is because Julia has a specific meaning for `sum`, which is summing the values of a collection (which we also use for `sum(::Vector{MPS})`). That seems like a good convention to me, it makes sense to distinguish between those two meanings.

PR #291, opened 2020-04-17
Add documentation building to Travis

mtfishman:
This should hopefully add a process for building the documentation and running the doctests with Travis (following https://juliadocs.github.io/Documenter.jl/stable/man/hosting/), we will see.


emstoudenmire:
Great! Looks like it works except that Documenter.jl needs to be added to the testing dependencies in Project.toml

PR #293, opened 2020-04-18
More fermion tests, exthubbard.jl DMRG example

emstoudenmire:


emstoudenmire:
I provided some more details about why the tests are failing in issue #295

PR #294, opened 2020-04-18
Add nothing values to unit tests

emstoudenmire:
This PR adds a line with just the value `nothing` to the end of each unit test file in ITensors.jl, for the purpose of running the tests individually in the REPL. As far as I can tell, these lines don't do anything to interfere with the tests, but they do prevent a very large output sometimes which can occur when the Test object gets returned after a test finishes running when being included in the REPL.

### DIFF
--- a/test/autompo.jl
+++ b/test/autompo.jl
@@ -746,3 +746,5 @@ end
   end
 
 end
+
+nothing
--- a/test/combiner.jl
+++ b/test/combiner.jl
@@ -234,3 +234,5 @@ end
 end
 
 end
+
+nothing
--- a/test/contract.jl
+++ b/test/contract.jl
@@ -257,3 +257,4 @@ end
 
 end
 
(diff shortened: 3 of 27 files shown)


mtfishman:
Seems reasonable, and all of that output is annoying. I do wonder if there is a reason to grab the Test object (like to pass it to something else or analyze it in some way), but if we find a reason we can deal with it then.

emstoudenmire:
Thanks, yes happy to change this later if we find a better approach

mtfishman:
Note that you can of course run it like `include("mpo.jl");` but I often forget to do that.

emstoudenmire:
Yes, actually that’s exactly what this change was for: when I run them like `include(“mpo.jl”)` sometimes I get a huge screenful of output which is the Test object being returned. So with this PR I just get a nice report of the passing or failing tests only

PR #297, opened 2020-04-18
Faster randomMPS

emstoudenmire:
This PR implements a faster, direct-instantiation, method for generating random MPS with typical properties. Right now it's only appropriate to use for dense-tensor MPS (the code checks for that).

One motivation for this is that the previous algorithm could take up to a second to run within some PEPS applications, which was too slow.

### DIFF
--- a/src/exports.jl
+++ b/src/exports.jl
@@ -158,6 +158,7 @@ export
 # mps/abstractmps.jl
   add,
   mul,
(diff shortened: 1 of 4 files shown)


emstoudenmire:
Seems like testing is broken (as in not even starting to run the test script correctly), so going to merge.

Issue #299, opened 2020-04-20
Update DMRG examples to not use `toMPO`


emstoudenmire:

emstoudenmire:
Closed by #301

PR #301, opened 2020-04-20
Change examples to use MPO instead of toMPO

emstoudenmire:
Fixes issue #299


emstoudenmire:
Tests are passing except known issue with one MPO test

PR #302, opened 2020-04-20
Throw exception when using wrong which_decomp with eigen_perturbation

emstoudenmire:


src/decomp.jl
```diff
@@ -273,6 +273,7 @@ function factorize_eigen(A::ITensor,
     Ris = uniqueinds(A,IndexSet(Linds...))
     A2 = A*prime(dag(A),Ris)
     if delta_A2 != nothing
+      println("Applying eigen perturbation")
```
mtfishman, line 276: Looks like this needs to be removed. Otherwise looks good to me.

emstoudenmire: Whoops! Good catch

mtfishman:
Failing tests should be fixed by #304, I'll merge.

PR #303, opened 2020-04-20
Fix some test errors

mtfishman:
Also export `\cdot` alias for `dot`/`inner`.


mtfishman:
These errors should be fixed by #304.

Issue #295, opened 2020-04-18
Issues with test/mpo.jl


emstoudenmire:
There seem to be some issues with the mpo.jl unit test file, but I am having trouble nailing them down. There are at least two issues:

- [ ] when doing julia runtests.jl on my computer, all of the tests pass. However, when running only julia mpo.jl (or doing include("mpo.jl") in the REPL), I get an error which has to do with the "." function (centered-dot symbol, not period, presumably an operator overload for either inner or do) not being defined. I can't actually find where it is defined in the source code, so I'm a bit confused about how these tests are passing in the first place.
- [ ] some of the numerical tests involving the `mul` function are failing, but only on one of the Travis machines. Perhaps the test is too numerically stringent? See PR #293 for an example of this. (I actually ran the tests twice for this PR to check that it wasn't a fluke.)

mtfishman:
That is referring to the symbol `\cdot`, which in Julia's LinearAlgebra is an alias for the `dot` function. We can fix this by exporting the `\cdot` unicode character from ITensors.jl

PR #304, opened 2020-04-20
Fix issue with chi=1 case of randomMPS

emstoudenmire:
Addresses one of the issues in #295, regarding `mul` tests failing.


mtfishman:
Not sure why the OSX tests are taking so long to run, but I'll merge this.

Issue #267, opened 2020-04-10
Option to print every step in DMRG


emstoudenmire:
We should add a keyword option to print the energy and bond dimension, etc., after each step of DMRG, not just after each sweep. (See #263 for one example where this is useful.)

It's worth checking if this is better done by supplying a default Observer through that system, or just with a keyword arg to the dmrg function directly. (Probably the latter to allow the printing to work orthogonally to the choice of observer.)

mtfishman:
This also made me wonder if we should pick a different energy that gets printed by default. I guess the last energy of the sweep makes sense, but maybe we could print both the last energy and the middle energy?

emstoudenmire:
Fixed by #309

PR #309, opened 2020-04-21
DMRG outputlevel

emstoudenmire:
Enhancement to the `dmrg` functions which replaces the quiet::Bool keyword with outputlevel::Int. Setting the output level to 0 means no output; 1 means only per-sweep output; 2 shows every step. The DMRGObserver type also makes limited use of the outputlevel, which gets passed through the keyword args. Addresses issue #267 

This PR includes a few other small tweaks and fixes:
- fix output in README doctest regarding showing noise level in sweeps
- change energy tolerance arg of DMRGObserver constructor to be a keyword arg

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -252,11 +252,11 @@ end
 # output
 
 sweeps = Sweeps
-1 cutoff=1.0E-10, maxdim=10, mindim=1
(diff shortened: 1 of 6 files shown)


mtfishman:
Looks good to me! You can merge when you are ready.

PR #310, opened 2020-04-21
Define `dag` for IndexVal and small printing tweak

emstoudenmire:
Defines dag(i(1)) to be the same as dag(i)(1).

Changes printing of IndexVals to use => notation (instead of = notation).

@mtfishman: would it be a good idea to define dag(::Pair{Index,Int}) too?

### DIFF
--- a/src/index.jl
+++ b/src/index.jl
@@ -451,6 +451,8 @@ plev(iv::IndexVal) = plev(ind(iv))
 prime(iv::IndexVal,
       inc::Integer = 1) = IndexVal(prime(ind(iv), inc), val(iv))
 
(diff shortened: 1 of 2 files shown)


mtfishman:
I would say yes. I think ideally people could use the `Pair` syntax anywhere they can use an `IndexVal`. We can just change the signature to `IndexValOrPairIndexInt`. I guess we should extend the type signature for `plev`, `prime` and `adjoint` as well.

I keep wondering if we should just define `const IndexVal{IndexT} = Pair{IndexT,Int}`, but perhaps that is a separate issue.

mtfishman:
Looks good to me!

PR #313, opened 2020-04-22
Add normalization step after truncation in DMRG

emstoudenmire:
Fixes an error in `dmrg` where the wavefunction can lose normalization due to truncation. Adds the `normalize::Bool` keyword arg to `replacebond!` which can be used by other algorithms such as time evolution. Also passes more keyword args to the observer checkdone! method for future observer customizations.

### DIFF
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -295,6 +295,7 @@ function replacebond!(M::MPS,
                       kwargs...)
   ortho = get(kwargs, :ortho, "left")
   which_decomp = get(kwargs, :which_decomp, "automatic")
+  normalize = get(kwargs,:normalize,false)
 
   # Deprecated keywords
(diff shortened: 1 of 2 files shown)


src/mps/mps.jl
```diff
@@ -310,9 +311,11 @@ function replacebond!(M::MPS,
   if ortho == "left"
     leftlim(M) == b-1 && setleftlim!(M, leftlim(M)+1)
     rightlim(M) == b+1 && setrightlim!(M, rightlim(M)+1)
+    normalize && (M[b+1] /= norm(M[b+1]))
```
mtfishman, line 314: When I merge #314 you can change this to `M[b+1] ./= norm(M[b+1])` (a slight optimization).

emstoudenmire: Great! Will do

PR #314, opened 2020-04-22
Add broadcast syntax for dividing by a scalar

mtfishman:
This adds the syntax `A ./= c` and `B .= A ./ c`.

Also, it adds a generic fallback error for broadcast operations that are not implemented yet, for example:
```julia
julia> i = Index(2)
(dim=2|id=484)

julia> A = randomITensor(i,i')
ITensor ord=2 (dim=2|id=484) (dim=2|id=484)' 
ITensors.NDTensors.Dense{Float64,Array{Float64,1}}

julia> A .= 2 ./ A
ERROR: The broadcasting operation you are attempting is not yet implemented for ITensors, please raise an issue if you would like it to be supported.
```

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -8,7 +8,8 @@ import Base.Broadcast: BroadcastStyle,
                        Broadcasted,
                        broadcastable,
                        instantiate,
-                       DefaultArrayStyle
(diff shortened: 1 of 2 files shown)


emstoudenmire:
The error message part is really helpful

PR #315, opened 2020-04-22
Replace fastsvd keyword argument in svd with alg

mtfishman:
This replaces the `fastsvd` keyword argument in `svd` with the `alg` keyword argument. The `alg` keyword argument has the options `"recursive"`, `"divide_and_conquer"`, and `"qr_iteration"` (the first is ITensor's custom svd, and the last two use LAPACK's gesdd and gesvd functions respectively).

Additionally, this adds the `svd_alg` keyword argument to `factorize`, with the same options.


mtfishman:
I also came across a bug in `replaceinds` which is fixed in this pull request.

mtfishman:
@emstoudenmire, let me know if you are good with the interface and names, otherwise this is ready to merge.

emstoudenmire:
Looks good!

PR #321, opened 2020-04-23
Add more docstrings

mtfishman:
This adds more documentation, mostly for ITensor constructors and other operations.

In the process, I was wondering about the use of default null constructors like `Index()`, `IndexVal()`, and `ITensor()`. I think that these are generally not necessary in Julia, with the use of a standardized `nothing` as well as constructors like `Vector{Index}(undef, N)`. I replaced the uses of them throughout ITensors.jl and those Julia standard worked well as drop-in replacements.


mtfishman:
As a specific example of using those, in `ProjMPO`, before the boundary Hamiltonian terms were initially constructed as a vector filled with default null ITensors, i.e. `fill(ITensor(), N)`. Then, it was checked if they had been assigned with the function `isnull(A::ITensor) = A == ITensor()`. Instead, one can define the initial boundary Hamiltonian terms as `LR = Vector{ITensor}(undef, N)`. Then, there is a Julia function `isassigned(LR, 3)` that checks if the 3rd boundary term is undefined. It is not a huge change, but I think it is nice to use the Julia construct here since it is pretty good and general (and doesn't require having a special null constructed object).

emstoudenmire:
I agree about using a built-in type (`nothing`) over having to define an isnull concept for ITensors and rely a lot on that. It's similar to how using things like tuples has been a big win for us over lots of single-use, obscure custom types in C++, such as in the AutoMPO code.

It's too bad that for making an undef vector of plain Index objects, one has to spell out `Index{Int}` though I understand why. Don't know if there's some solution for this other than making a type alias for `Index{Int}`. It's not so ideal to just put `Index` of course since it's an abstract type.

I'm ready to merge this if you are -

mtfishman:
Yeah I agree, it is unfortunate that the way we defined the type you have to spell out `Index{Int}` to get the non-QN concrete type. However, a vast majority of the time it is sufficient to just type `Index` and then specifiy `QNIndex` if needed, so I think the design works well in general. 

The alternatives are:
- Define the general type as `IndexT{T}` and then define the concrete type `const Index = IndexT{Int}` (which I don't like, since I think that would lead to using `IndexT` everywhere).
- As you say, make an alias for `Index{Int}`. Unfortunately I can't think of a good name, and it seems like it is very rare to need to use `Index{Int}` explicitly (even in the use case above, it is probably a micro-optimization and it wouldn't be too bad to use `Vector{Index}(undef, N)`).

I am going to continue adding more docs to this PR, I'll merge it when I am at a good stopping point.

PR #328, opened 2020-04-27
Fix empty combiner

### LINKED ISSUE #324: Empty combiner contraction
`combiner()` performs the wrong contraction. For example:
```julia
i = Index(2)
A = randomITensor(i)
@show A*combiner()
```
outputs something like:
```julia
A * combiner() = ITensor ord=1

IndexSet{1,Index{Int64}}
Dim 1: (dim=2|id=592)
ITensors.NDTensors.Dense{Float64,Array{Float64,1}}
 2-element

 6.9398952879376e-310
 6.93989528956526e-310
```
so is resulting in something like `similar(A)` instead of `copy(A)`.

mtfishman:
Fixes #324.

### DIFF
--- a/NDTensors/src/combiner.jl
+++ b/NDTensors/src/combiner.jl
@@ -61,7 +61,8 @@ function contract!!(R::Tensor{<:Number,NR},
                     T2::Tensor{<:Number,N2},
                     labelsT2::NTuple{N2}) where {NR,N1,N2}
   if N1 ≤ 1
-    #println("identity")
+    # Empty combiner, acts as multiplying by 1
+    R = permutedims!!(R, T2, getperm(labelsR, labelsT2))
     return R
   elseif N1 + N2 == NR
     error("Cannot perform outer product involving a combiner")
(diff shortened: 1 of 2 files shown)


emstoudenmire:
Good fix. Quick question: why does the empty combiner do a call to permutedims!! ? (At least that's how it looks from the file changes.) Does it reorder the indices somehow or is that just code that handles some other cases too?

mtfishman:
Since it is the in-place version, it is accounting for a case where the user may have provided an output ITensor with a different ordering than the input ITensor, for example:
```julia
i = Index(2)
A = randomITensor(i, i')
C = combiner()
R = ITensor(i', i)
R .= A .* C
```

emstoudenmire:
Thanks - that make sense!

I hadn't quite appreciated how the `.=` broadcasting gives us the capability to determine the ordering of all three tensors participating in a contraction, because of course one could do `permutes` of A and B before computing `C .= A .* B`.

Looks ready to merge to me -

PR #330, opened 2020-04-27
Fix QN MPO * MPS

### LINKED ISSUE #307: Calling scalar on QNITensor with no blocks gives error
Currently, if a scalar (order-0) QNITensor `T` has no blocks (`nzblocks(T)==0`), then calling `scalar(T)` or `T[]` results in an error. Would it not be better to have `scalar(T)` return 0.0 for this case?

Example to reproduce:
```
s = siteinds("S=1/2",4;conserve_qns=true)
A = ITensor(s[1]'',dag(s[1]'))
B = ITensor(s[1]',dag(s[1]''))
A[1,1] = 1.0
B[2,2] = 1.0
R = A*B
@show scalar(R) #gives an error
```

### LINKED ISSUE #320: _contract_densitymatrix not working correctly with QNITensors
There seems to be a bug with the density matrix MPO*MPS algorithm for QN-conserving tensors, probably just related to the proper use of dag and QNIndex arrows. I'm happy to fix it; just putting it here as a reminder.

mtfishman:
This fixes the `"densitymatrix"` version of MPO * MPS with QNs. Fixes #320.

I've also included a few other things in this PR (improvements I came across while rewriting `_contract_densitymatrx`):

- Fixed #307 (return 0 for an QN ITensor with no indices and no blocks).
- Add a lot of priming, tagging and similar functions for manipulating indices of MPOs and MPSs (this utilizes metaprogramming to make it easier to generically add new functionality, instead of the ad hoc way in the C++ version).
- Adds a keyword argument to `setindex!` for MPO and MPS called `set_limits` which defaults to `true`. When false, it turns off setting new limits when replacing one of the ITensors, if you know it doesn't change the orthogonality.


emstoudenmire:
I just pushed a small formatting improvement to the mismatched arrow error message. Otherwise looks good to merge. Nice to have these fixes!

mtfishman:
In the latest, I made the following changes:

- Rename `siteinds(A::MPO, B::MPS)` to `unique_siteinds(A::MPO, B::MPS)` to get a vector of the site indices of `A` that are not shared with `B` (the previous one was a bit ambiguous, and doesn't distinguish from `common_siteinds(A::MPO, B::MPS)`).
- Add `common_siteinds(A::MPO, B::MPS)` to get a vector of the site indices shared between `A` and `B`.
- Add functions like `prime_common_siteinds!(A::MPO, B::MPS)` to `prime` all site indices shared between `A` and `B` (works with all prime and tag related functions, and `sim`).
- Remove functions like `prime_linkinds!(A::MPO, B::MPS)` which primes the link indices of `A` and `B` as well as the site indices shared by `A` and `B`. I think that name it is a bit ambiguous given the functionality, and can be done with `prime_linkinds(A); prime_linkinds(B); prime_common_siteinds(A, B)`. We could bring back that functionality in a more compact way, but I would rather be on the conservative side right now and come up with a better name.

emstoudenmire:
Those changes all sounds really good. I agree that `prime_linkinds` is a confusing name if it also primes indices which could be considered "site". And similar with the other changes you mentioned. By the way, those functions which find all of the indices in common between two MPS or MPO's and do things like replace links with similar ones etc. are really useful for "hardening" MPO and MPS algorithms against all possible user inputs. They've been great to have!

PR #333, opened 2020-04-28
Use Strided for permutedims

mtfishman:
Jutho has a nice package called [Strided.jl](https://github.com/Jutho/Strided.jl) which has a custom implementation of Julia's `permutedims` that acts lazily and can fuse with other operations with broadcasting. This pull request uses it as a drop-in replacement for `permutedims[!]` in NDTensors (including replacing our custom version that applies a function). 

Here are some results running `examples/dmrg/exthubbard.jl`. On master:
```
After sweep 1 energy=-15.684191171846 maxlinkdim=50 time=0.593
After sweep 2 energy=-17.211812572484 maxlinkdim=100 time=1.379
After sweep 3 energy=-17.429068161255 maxlinkdim=200 time=2.681
After sweep 4 energy=-17.435370008361 maxlinkdim=400 time=5.339
After sweep 5 energy=-17.435415463390 maxlinkdim=630 time=8.128
After sweep 6 energy=-17.435415758946 maxlinkdim=654 time=6.943
```
and on this branch:
```
After sweep 1 energy=-16.351502376855 maxlinkdim=50 time=0.739
After sweep 2 energy=-17.342280545549 maxlinkdim=100 time=1.504
After sweep 3 energy=-17.434333146203 maxlinkdim=200 time=2.607
After sweep 4 energy=-17.435410320715 maxlinkdim=400 time=4.993
After sweep 5 energy=-17.435415710430 maxlinkdim=603 time=6.444
After sweep 6 energy=-17.435415760166 maxlinkdim=618 time=6.199
```
Not a huge speedup over all, but of course `permutedims` wasn't dominating the `dmrg` cost so it is pretty good.

For reference, here is the same run in C++ ITensor v3:
```
    Largest link dim during sweep 1/6 was 50
    Energy after sweep 1/6 is -13.654587063495
    Sweep 1/6 CPU time = 0.361s (Wall time = 0.266s)

    Largest link dim during sweep 2/6 was 100
    Energy after sweep 2/6 is -16.892217260698
    Sweep 2/6 CPU time = 5.572s (Wall time = 1.003s)

    Largest link dim during sweep 3/6 was 200
    Energy after sweep 3/6 is -17.409957867258
    Sweep 3/6 CPU time = 11.16s (Wall time = 1.947s)

    Largest link dim during sweep 4/6 was 400
    Energy after sweep 4/6 is -17.435221134464
    Sweep 4/6 CPU time = 18.04s (Wall time = 3.160s)

    Largest link dim during sweep 5/6 was 713
    Energy after sweep 5/6 is -17.435414564047
    Sweep 5/6 CPU time = 35.16s (Wall time = 6.151s)

    Largest link dim during sweep 6/6 was 659
    Energy after sweep 6/6 is -17.435415752219
    Sweep 6/6 CPU time = 26.14s (Wall time = 4.591s)
```
so it is not too far off. All of these runs are with MKL, letting MKL choose the number of threads.

There are still some things to play with, for example Strided.jl supports multithreaded `permutedims`. On a first test adding more threads made things slower, perhaps because it was competing with MKL BLAS threads. I tried turning off MKL BLAS threads and I still didn't see Strided.jl's multithreading helping much.

I did some profiling of the `exthubbard.jl` `dmrg` run, and focused first on `permutedims` since I saw it show up heavily in the profile. There were a few other things that showed up a bit too much, for example some contraction logic and block contraction logic, which we should look into. Calls to `svd_recursive` also showed up, so I added an `svd_alg` keyword argument to `dmrg`. Using `svd_alg = "divide_and_conquer"`, here are the results:
```
After sweep 1 energy=-15.134829206976 maxlinkdim=50 time=0.589
After sweep 2 energy=-17.068192318420 maxlinkdim=100 time=1.409
After sweep 3 energy=-17.423436286897 maxlinkdim=200 time=2.571
After sweep 4 energy=-17.435342471525 maxlinkdim=400 time=4.436
After sweep 5 energy=-17.435415113102 maxlinkdim=653 time=6.820
After sweep 6 energy=-17.435415754267 maxlinkdim=657 time=5.609
```
so another modest speedup.

### DIFF
--- a/NDTensors/src/NDTensors.jl
+++ b/NDTensors/src/NDTensors.jl
@@ -4,6 +4,7 @@ using Random
 using LinearAlgebra
 using StaticArrays
 using HDF5
+using Strided
 
 #####################################
 # DenseTensor and DiagTensor
--- a/NDTensors/src/dense.jl
+++ b/NDTensors/src/dense.jl
@@ -276,7 +276,9 @@ end
 function Base.permutedims!(R::DenseTensor{<:Number,N},
                            T::DenseTensor{<:Number,N},
                            perm::NTuple{N,Int}) where {N}
-  permutedims!(array(R),array(T),perm)
+  RA = array(R)
+  TA = array(T)
+  @strided RA .= permutedims(TA, perm)
   return R
 end
 
@@ -297,7 +299,9 @@ function permutedims!!(R::Tensor,
                        perm::NTuple{N,Int},
                        f::Function=(r,t)->t) where {N}
   if !is_trivial_permutation(perm)
-    R = permutedims!(R,T,perm,f)
+    RA = array(R)
+    TA = array(T)
+    @strided RA .= f.(RA, permutedims(TA, perm))
   else
     R = apply!(R,T,f)
   end
@@ -308,7 +312,7 @@ end
 function Base.permutedims(T::Tensor{<:Number,N},
                           perm::NTuple{N,Int}) where {N}
   Tp = similar(T,permute(inds(T),perm))
-  Tp = permutedims!!(Tp,T,perm)
+  Tp = permutedims!!(Tp, T, perm)
   return Tp
 end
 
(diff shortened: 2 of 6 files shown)


emstoudenmire:
Reminder to move the dependency to NDTensors (may have to happen post registration of packages)

Jutho:
Hi @mtfishman , that's great to see. I see a similar issue with multithreading in tensor contractions in my own packages TensorOperations.jl and TensorKit.jl. When benchmarking a single permutation, the multithreading helps, but in combination with contractions, it seems to slow it down, even though the BLAS threads should not be competing with the Strided/permutation threads, as these run at different times. Not sure what is causing this, but I should definitely find some time to update Strided to use the new multithreading infrastructure of Julia (though I am not sure this will solve it).

mtfishman:
@Jutho, helpful to hear that you have seen similar issues with multithreading and it wasn't just a mistake I was making. Looking forward to future versions of the package, please keep us updated and we would be happy to try out new features and provide benchmarks.

The way the package works fluidly with `permutedims` within broadcast calls is great, it will be very helpful for adding generic support of broadcasting with ITensors (where addition of ITensors generally involves permutations of dimensions).

Jutho:
Hi @mtfishman and @emstoudenmire . Strided has been updated to 1.0.0. I think problems with multithreading slowing things down should be much improved. Also, there is more control to set the number of threads in Strided, e.g. there is a `Strided.set_num_threads()`.

mtfishman:
@Jutho, that is great, thanks for the update! I'm looking forward to trying it out. I was wondering if there would be a good strategy for setting threads on a module-by-module basis, looks like you figured out a nice interface. That will be very helpful as we start adding multithreading to ITensors.jl.

Am I reading the README of Strided.jl correctly that you have a custom multithreaded matrix multiplication in that package that is within a few percent of OpenBLAS and MKL performance? If so that sounds like a really big deal!

Jutho:
But it still uses standard BLAS for the single threads right; it just takes the multithreading in its own hands (in fact, it requires the user to manually set `LinearAlgebra.BLAS.set_num_threads(1)`; maybe it should print a warning that this should be done).

It's not something like the truly amazing work of LoopVectorization.jl .

mtfishman:
I see. I guess the implementation is here? https://github.com/Jutho/Strided.jl/blob/439b79690a3098942794b02b859855236916bb38/src/abstractstridedview.jl#L141

Looks deceptively simple. Very cool that it has good performance, sounds very useful to allow Julia to take over the threading (in C++ we just turn off BLAS multithreading altogether when we use multithreading for block-sparse tensor contraction).

Jutho:
Yes that’s the one. But it should also be benchmarked a bit more carefully to conclude (with statistical significance) that it is on par with OpenBLAS/MKL's multithreading model. 

In Julia, it would be easy to do LinearAlgebra.BLAS.set_num_threads(1) before this runs and restore the number of BLAS threads at the end. The only problem is that there is no BLAS.get_num_threads() to actually query the number of threads that you want to restore at the end. But it is not that hard to write it yourself.

mtfishman:
Yes, I suppose the block size `1024` could also be played around with, and may be system dependent?

That sounds like it would be helpful (so someone can automatically use Julia threading using your matrix multiplication and BLAS threading for matrix decompositions), strange that function is not provided already. Maybe it is a pain to get correct for every system, version of BLAS, etc. but presumably they already dealt with all that when making `LinearAlgebra.BLAS.set_num_threads`.

Jutho:
You can express your support here:
https://github.com/JuliaLang/julia/pull/21105

Issue #331, opened 2020-04-27
`eigen` interface


mtfishman:
In `eigen`, in the default usage, it is easy to reproduce the original ITensor:
```julia
i = Index(2, "i")
j = Index(2, "j")
A = randomITensor(i, j, i', j')
A = A + swapprime(A, 0, 1)
U, D = eigen(A; ishermitian=true)
Ut = U'
B = U*D*dag(Ut)
@show norm(A - B)
```
However, in a more general use case, this isn't as easy:
```julia
i = Index(2, "i")
j = Index(2, "j")
Lis = IndexSet(i, j)
Ris = dag(sim(Lis))
A = randomITensor(Lis..., Ris...)
A = A + swapinds(A, Lis, Ris)  # Note: this function doesn't exist right now, we need to make it
U, D = eigen(A, Lis, Ris; ishermitian = true)
u = commonind(U, D)
# Annoying and error prone:
Ut = replaceinds(prime(U, u), Lis, Ris)
B = U * D * dag(Ut)
@show norm(A - B)
```
In general, we could return the ITensor `Ut` that has the proper indices such that `U*D*dag(Ut)` gives an approximation for the input (at least in the hermitian case). Returning `Ut` would only add the cost of the Index manipulations (while returning `dag(Ut)` would add the possible cost of the complex conjugation).

This came up for my when I was rewriting `_contract_densitymatrix`, and I found it useful to have the more general version of `eigen`, but I need both `U` and `Ut` and it was a bit annoying. The only downside I see is that it breaks code where someone was extracting the spectrum object from the `eigen` return, i.e.:
```julia
U,D,spec = eigen(A)
```
if we decide to make the return order `U,D,Ut,spec`. We could return `Ut` last, and people could extract it like:
```julia
F = eigen(A)
@show norm(F.U * F.D * F.Ut - A)
```
I'm not sure how common it will be to want `Ut` vs. the Spectrum object.

mtfishman:
Just want to point out this is also useful in the non-Hermitian case, since `Ut` is the ITensor needed to test the eigenvectors:
```julia
i = Index(2, "i")
j = Index(2, "j")
Lis = IndexSet(i, j)
Ris = sim(Lis)
A = randomITensor(Lis..., Ris...)
U, D = eigen(A, Lis, Ris)
u = commonind(U, D)
# Annoying and error prone:
Ut = replaceinds(prime(U, u), Lis, Ris)
@show norm(U*D - D*dag(Ut))
[...]

PR #334, opened 2020-04-29
New eigen interface

mtfishman:
This addresses #331.

~~The version proposed here is non-breaking. The api is the following:~~
```julia
i = Index(QN(0) => 2,
          QN(1) => 3,
          QN(2) => 4; tags="i")
j = settags(i, "j")
it, jt = sim(i), sim(j)
A = randomITensor(QN(), i, j, dag(it), dag(jt))
F = eigen(A, (i, j), (it, jt); lefttags = "x", righttags = "y")
U, D, spec = F
Ut = F.Ut
@show norm(A * Ut - U * D)
```
~~Here, the ITensor `Ut` that has the correct indices to contract with `A` can only by extracted from the `TruncEigen` object `F`, not through iteration.~~

~~We could consider returning it at the end of the iteration, or somewhere else (like after `D` and before `spec`). I guess the question is if people are more likely to use `Ut` or `spec`.~~

See later comments for the updated syntax.

### DIFF
--- a/NDTensors/src/blocksparse/diagblocksparse.jl
+++ b/NDTensors/src/blocksparse/diagblocksparse.jl
@@ -139,9 +139,11 @@ function Base.promote_rule(::Type{<:UniformDiagBlockSparse{ElT1,VecT1}},
 end
 
 function Base.promote_rule(::Type{BlockSparseT1},
-                           ::Type{<:NonuniformDiagBlockSparse{ElT2,VecT2}}) where {BlockSparseT1<:BlockSparse,
-                                                                                   ElT2<:Number,VecT2<:AbstractVector}
-  return promote_type(BlockSparseT1,BlockSparse{ElT2,VecT2})
+                           ::Type{<:NonuniformDiagBlockSparse{ElT2,VecT2,N2}}) where {BlockSparseT1<:BlockSparse,
+                                                                                      ElT2<:Number,
+                                                                                      VecT2<:AbstractVector,
+                                                                                      N2}
+  return promote_type(BlockSparseT1,BlockSparse{ElT2,VecT2,N2})
 end
 
 function Base.promote_rule(::Type{BlockSparseT1},
--- a/NDTensors/src/blocksparse/linearalgebra.jl
+++ b/NDTensors/src/blocksparse/linearalgebra.jl
@@ -21,7 +21,7 @@ function _truncated_blockdim(S::DiagMatrix,
 end
 
 """
-svd(T::BlockSparseTensor{<:Number,2}; kwargs...)
+    svd(T::BlockSparseTensor{<:Number,2}; kwargs...)
 
 svd of an order-2 BlockSparseTensor.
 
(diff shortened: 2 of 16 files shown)


mtfishman:
Here is the latest interface for `eigen`. `eigen` now outputs eigenvalues `D` and eigenvectors `V`. `V` stores the "right" eigenvectors of the input ITensor, in the sense that it has the "right" indices defined by `eigen(A, Linds, Rinds)`. In the simple case with pairs of unprimed and primed indices, `Rinds` are taken to be the unprimed indices and `Linds` are taken to be the primed indices.

For example, in the case where the ITensor has pairs of primed and unprimed indices:
```julia
i = Index(QN(0) => 2, QN(1) => 2; tags="i")
j = settags(i, "j")

A = randomITensor(ComplexF64, QN(), i', j', dag(i), dag(j))
F = eigen(A)
D, V = F
Vt = F.Vt

# V, Vt store the right eigenvectors of A
@show norm(A * V - Vt * D)

# In this case, Vt == V'
@show norm(A * V - V' * D)

# Index k runs over the eigenvectors
k = uniqueind(V, A)

@show hasinds(V, i, j, k)
@show hasinds(Vt, i', j', k')
```
and in the case of more general indices:
```julia
i = Index(QN(0) => 2, QN(1) => 2; tags="i")
j = settags(i, "j")

it, jt = sim(i), sim(j)

A = randomITensor(ComplexF64, QN(), it, jt, dag(i), dag(j))
F = eigen(A, (it, jt), (i, j))
D, V = F
Vt = F.Vt

# V, Vt store the right eigenvectors of A
@show norm(A * V - Vt * D)

# Index k runs over the eigenvectors
k = uniqueind(V, A)

@show hasinds(V, i, j, k)
@show hasinds(Vt, it, jt, k')
```

@emstoudenmire, let me know if this interface looks good to you (it basically follows the C++ convention, but accepts more general indices and the interface is unified for Hermitian and non-Hermitian ITensors).

emstoudenmire:
This looks great. I think this convention for what V is makes the most sense in terms of likely use cases, together with the ITensor approach to things. And it’s nice that the interface can be unified with the Hermitian case this much.

One question for my understanding is: is V’ in the first set of examples just the complex conj of V or does the overload of `adjoint` for ITensors also do a swap prime type of operation? (Or is V’ just a priming thing as it is for indices?)

Also in every case, D has the indices k and k’, correct? If so I do think that makes sense no matter what because those vector spaces have to have identical properties, correct?.

mtfishman:
`V'` in the first example is just the prime of `V`, no complex conjugate is performed. The check there is just showing the eigenvector equation `A v_k = d_k v_k` is satisfied (so no conjugation is needed). Perhaps the overload there of `adjoint(A::ITensor) = prime(A)` is a bit confusing, but it follows with the overload `adjoint(i::Index) = prime(i)`.

And yes, `D` has the indices `(k', k)` with same sizes and ids. I think it has to be the case in general that they have the same "spaces", so then in general they can be constrained to have the same `id`s. Of course there is a user interface for choosing the tags and prime levels of those two indices (so in general they could be `(l, r)` which have the same `id` but different tags and prime levels).

One thing that I wasn't sure about with that convention is that it seems to imply that even in the non-Hermitian case, the input ITensor has to have a flux of 0. If `D` is limited to having indices that have the same spaces, it has a flux of 0, and it seems like `D` has to have the same flux as the input ITensor. My understanding is that in general, ITensors with non-zero fluxes may not have well defined eigenvectors (for example, creation and annihilation operators). Does that sound right to you?

emstoudenmire:
Thanks for clarifying about those details. Yes, I see now about V' and that makes sense. (No problem about the adjoint overload, though it's something we should keep in mind as a possible source of confusion for people new to ITensor who use it for complex, matrix-like ITensors and expect it to work similarly as for Julia arrays.)

It's a good question about non-zero flux ITensors. I believe it's the case that such objects can't have well-defined or non-trivial eigenvectors. Say there is a flux non-zero tensor M, then because acting on some tensor T yields a new tensor R=M*T with flux(R) = flux(T)+flux(M), there's no such T such that T is proportional to R (an eigenvector), otherwise it would require flux(R)-flux(T) = 0 = flux(M).

One thing that's tricky about the above argument is that I can always make a flux-zero ITensor by attaching a dimension-1 index to M which cancel's M's flux. So then what happens? I'm not sure if such a case is a meaningful exception to the argument above It may just be that then M is no longer a matrix-like tensor (in the sense of having two sets of indices [row/column] with similar structure), so that eigenvectors can't even really be defined. But you may see why it's a bit tricky because since the new Index is dimension-1, it can be "hidden" by shifting the QN's of the other indices. 

In conclusion, I think we should ship version 0.1 with support only for the flux-zero case and we could always generalize the code later should we find a clear use-case.

mtfishman:
I agree with your analysis and that for now we can just limit `eigen` to only accept zero flux ITensors. As you say, we can always deal with more general cases if they come up.

I'll merge this once the conflicts are resolved.

PR #337, opened 2020-05-01
Replace ITensor function apply! with map!

mtfishman:
This replaces the function `apply!(::ITensor, ::ITensor, ::Function)` with `map!(::Function, ::ITensor, ::ITensor, ::ITensor)`. There are two main reasons for this:

1. `apply!` essentially was the same as the Julia `map!` function, but with a slightly different API.
2. The new API `map!(f, R, T1, T2)` for a 2-argument element-wise function `f` will more easily generalize, such that in general we can have `map!(f, R, Ts::Vararg{ITensor,N})` for an N-argument element-wise function `f`.

Also, this reorganizes the code such the `map!` is only used within definitions of broadcasting functionality, and then broadcasting is used to define other primitive operations (like `mul!(A, x, B) = (A .= x .* B)`). I think this can be a general code pattern was can use for in-place operations involving ITensors.


emstoudenmire:
Sounds good, though will there also be a version of the new `map!` function for functions taking only one argument? Or is the preferred solution there to use broadcasting?

mtfishman:
Good question, yes there will also be a new `map!` for functions of one argument, i.e. `map!(x -> 2*x, B, A)` which would be a backend for the broadcast call `B .= 2 .* A`. Officially, I think we should advertise the broadcast version more, since the broadcast version will call the `map!` version anyway.

The reason I didn't implement it yet is because `NDTensors.permutedims!!` right now only works with functions of two arguments. I think I will rework it to be more general (and maybe change the name to something like `NDTensors.mappermutedims!!` to make it clear that a function is being applied). I'm leaving that as a future project, since it is more of an NDTensors-level issue.

Issue #332, opened 2020-04-27
Add `swapinds`


mtfishman:
Add the function `swapinds`, like `replaceinds` but swaps the indices pairwise.

mtfishman:
Fixed by #338.

PR #338, opened 2020-05-03
swapinds added #332

amartyabose:


mtfishman:
Hi, thanks for the contribution. It turns out that I had implemented a version of `swapinds` in #334 which I just merged. I like your implementation better so I just resolved the conflicts with the new master branch by overwriting `swapinds` with your implementation. I'll merge when the tests pass.

PR #342, opened 2020-05-04
Misc improvements

emstoudenmire:


src/itensor.jl
```diff
@@ -955,12 +955,18 @@ Base.:/(A::ITensor, x::Number) = A*(1.0/x)
 Base.:-(A::ITensor) = itensor(-tensor(A))
 
 function Base.:+(A::ITensor, B::ITensor)
+  if order(A) != order(B)
```
mtfishman, line 958: We could instead make this a compile time check by defining this function as:
```julia
Base.:+(A::ITensor{N}, B::ITensor{N}) where {N}
```
then if we want we can add our own error message by defining the more general version:
```julia
Base.:+(A::ITensor, B::ITensor) = error("different number of indices when adding ITensors")
```
which it should call as a fallback.

emstoudenmire:
Good point. Before I do that, though, how should we prepare for having a feature like the `zeroITensor()` which does allow addition with different numbers of indices? (We could always adjust the code later if need be and not worry about it too much right now since it's not a breaking interface change.)

mtfishman:
That's a good question. I guess if we want a check at the ITensor level, that would require making it a runtime check, since we would only check the order if both of the ITensors have non-Zero storage. It seems like the Zero-storage might make some code a bit messier at the ITensor level, which is unfortunate.

mtfishman:
But yeah, we can punt that question until we make a PR for the Zero storage ITensor.

mtfishman:
I'm good to merge this (seems fine to leave it as a runtime check).

emstoudenmire:
I ended up liking the parse-time / compile-time version. As we discussed, though, it will be easy to change to whichever one is best in the future as we add the zeroITensor or similar.

PR #346, opened 2020-05-05
Documentation for AutoMPO, LatticeBond, etc.

emstoudenmire:
This PR is a work in progress adding docstrings to functions and types for which they are missing.


src/physics/autompo.jl
```diff
@@ -131,26 +150,48 @@ Base.copy(ampo::AutoMPO) = AutoMPO(copy(data(ampo)))
...
+add!(ampo::AutoMPO,op::String, i::Int) = add!(ampo,1.0,op,i)
+
+
+"""
+    add!(ampo::AutoMPO,
+         op1::String, i1::Int,
+         op2::String, i2::Int,
+         ops...)
 
+    add!(ampo::AutoMPO,
```
mtfishman, line 184: Maybe it would be nice to mention the alternative syntax `.+=` here as well?

emstoudenmire: Yes, I was actually wondering where to mention that. I’m not sure if people ever do doc strings for operators like ?+ so then the named version (such as add!) is a great place to mention it.

emstoudenmire: Do you have a good suggestion for how to document it though? Like we could put
```
+(ampo::AutoMPO, term::Tuple)
```
there as well but it’s a bit cryptic. Maybe that plus a doc example later on will help.

mtfishman: I think it is fine to put what you wrote, and then also doc examples.

src/mps/projmpo.jl
```diff
@@ -62,8 +110,16 @@ function Base.eltype(pm::ProjMPO)
...
-(pm::ProjMPO)(v::ITensor) = product(pm,v)
+"""
+    size(pm::ProjMPO)
 
+The size of a ProjMPO is the dimension
+of the space on which it acts as a linear
+operator. Thus if a ProjMPO maps from a space
+with indices `(a,s1,s2,b)` to the space 
+`(a',s1',s2',b')` then the size 
+is `dim(a)*dim(s1)*dim(s1)*dim(b)`.
```
mtfishman, line 121: Shouldn't this read that the size is `(d, d)` where `d = dim(a)*dim(s1)*dim(s1)*dim(b)`?

src/mps/projmpo.jl
```diff
@@ -1,19 +1,44 @@
 
+"""
+A ProjMPO object represents the projection of an
+MPO into a basis defined by an MPS, leaving a
+certain number of site indices of the MPO unprojected.
+Which sites are unprojected can be shifted by calling
+the `position!` method.
+
+Drawing of the network represented by a ProjMPO `P`, 
+showing the case of `nsite(P)==2` and `position!(P,4)`:
```
mtfishman, line 10: Isn't there only a function `position!(pm::ProjMPO, psi::MPS, pos::Int)`, not `position!(pm::ProjMPO, pos::Int)`?

src/mps/projmpo.jl
```diff
@@ -128,9 +196,18 @@ function position!(pm::ProjMPO,
 end
 
 # Return a "noise term" as in Phys. Rev. B 72, 180403
+"""
+    noiseterm(pm::ProjMPO,
+              phi::ITensor,
+              b::Int,
```
mtfishman, line 202: Looks like it doesn't require `b` now?

emstoudenmire: Thanks, fixed!

src/mps/projmpo.jl
emstoudenmire: Yes, fixed this in new commits

src/mps/projmpo.jl
emstoudenmire: Yes, I had missed that, thanks!

emstoudenmire:
I haven’t finished making docs pages for all of the new docstrings here, but it would help me to go ahead and merge this PR if there aren’t any more serious errors so I can see how things look so far (still getting the hang of Documenter but it’s really nice to use).

Issue #340, opened 2020-05-04
Have a base "zero ITensor" for summing ITensors?


emstoudenmire:
A pattern that can come up often is that one wants to sum up a set of ITensors in a for loop, but the code inside the loop is rather complicated and the tensor accumulating the results doesn't exist yet. For example (it's a fake example, so no need to press into the details):

```
Q,R = qr(v[1],s[1])
T = alpha*Q*w[1]

for n=2:N
  Q, R = qr(v[n],s[n])
  T += alpha*Q*w[n]
end
```

As can be seen above, the complicated code logic inside the loop has to be repeated in front of the loop to initialize T, causing code duplication and possible bugs etc.

In the C++ version of ITensor we allow T to be initialized as `auto T = ITensor()` and then other ITensors are allowed to be added into T, so that it acts as a kind of "universal zero" tensor. But maybe in Julia we can do something better to solve this problem? E.g. we could allow T to be initialized as `nothing` and then add to it?

emstoudenmire:
I tried it just now, and it requires very little code to implement:

```
Base.:+(A::Nothing,B::ITensor) = copy(B)
Base.:+(A::ITensor,B::Nothing) = copy(A)
```

Of course the question is whether it's a good idea.

mtfishman:
Yes, I was trying to think of a good interface for this. I think I prefer an interface where it is explicit that you are creating an ITensor with a special storage type. I would prefer to preserve plain constructors like `ITensor()` for simpler meanings, for example right now it creates a Dense ITensor with no indices and one element of zero, which is expected in many ways from other similar Dense constructors.

I don't really like using `nothing` for that purpose. There may be functions which return `nothing` or an ITensor, where you want operations with that `nothing` to fail and in order to catch that an ITensor was not returned.

It seems like we would want a special `Zero` storage type like in C++, the purpose of which is that it acts as an additive identity (and I guess could also act as multiplicative zero as well). We could give it an explicit sparse constructor function like `zeroITensor`, which matches with functions like `diagITensor` to create an ITensor with a `Diag` storage.

emstoudenmire:
Good point about the `nothing` approach leading to unwelcome surprises. I agree we should exercise caution here.

One thing that could be confusing, though, is if we have two kinds of zero ITensors both of which have no indices, but only one which is allowed to add to other ITensors with indices. Or would it be allowed to add both an ITensor() and a zeroITensor() to another ITensor?

mtfishman:
[...]

PR #348, opened 2020-05-05
QN ITensor interface improvements

mtfishman:
This adds documentation for `IndexVal` and various QN ITensor constructors.

In the process, it also made me start to think about some of the QN ITensor constructors, so this is a work in progress.

This PR will address #349 and #340.


mtfishman:
In the latest version of this PR, I've implemented the change proposed in #349.

This change meant that certain uses of `ITensor(::QNIndex...)` as an uninitialized ITensor was no longer valid. I introduced an initial version of `zeroITensor(::Index...)` which for QN indices it creates a `BlockSparse` storage without any blocks, and for non-QN indices it makes a `Dense` storage filled with zeros. This is in preparation for introducing a `Zero` storage.

I've also simplified and unified the ITensor constructor docstrings.

mtfishman:
In that latest, I changed the name of `zeroITensor` to `emptyITensor`. `emptyITensor` still just constructs ITensors with `Dense` or `BlockSparse` storage (filled with zeros and with no blocks, respectively). We still need to create an `Empty` storage type in NDTensors, though I am waiting for the NDTensors package to get registered and will add it in that new package.

mtfishman:
I've added an initial version for an `emptyITensor()` constructor. It is a little bit tricky, but it creates an ITensor of type `ITensor{Any}` with an IndexSet of type `IndexSet{Any, Union{}, Tuple{}`. This ITensor type is only used for the following case:
```julia
A = emptyITensor()
for i in 1:10
  A += randomITensor(i, j)
end
```
Making it type `ITensor{Any}` is meant to emphasize that it is quite a special ITensor, and the interpretation is that it can take on "any" order, based on the context (like addition).

I also realized with the current design where the ITensor type is parametrized by the order, this is not possible:
```julia
A = emptyITensor()
A[i => 1, j => 2] = 2.0  # This would give the ITensor A the indices (i, j)
```
since it would require changing the ITensor's type in-place (from type `ITensor{Any}` or whatever it is initially to type `ITensor{2}`). I don't think this is a big loss, since setting elements requires that you have access to the indices to begin with, so we can just require that you do:
```julia
A = emptyITensor(i, j)
A[i => 1, j => 2] = 2.0
```

mtfishman:
Also the current version is failing because of an issue in NDTensors (https://github.com/ITensor/NDTensors.jl/issues/3). NDTensors just got registered and I would prefer to fix that issue there instead of in the ITensors version. I may go ahead and remove the NDTensors module in this PR and add it as a dependency, so then the fix over there can be used here (and also we can use the `Empty` storage type being developed over there).

emstoudenmire:
Good to see how this is progressing. That’s too bad about the inability to infer the indices from setindex!, but I agree it’s not a big loss. The tradeoff of having the tensor order in the type definitely pays some dividends in other places.

That would be neat if you could go ahead and finish splitting off NDTensors in this PR, since I saw it’s now registered. (Congrats!) I’m not sure if that’s a rather easy thing to do, or pretty complicated.

emstoudenmire:
Also, just to clarify, the other use of `emptyITensor` besides for summations is for optimizing code, correct? Like where you don't want to allocate any memory because the code immediately after will start changing the tensor anyway?

mtfishman:
The other use case besides addition is what you said, and more generally having a generic way to create an uninitialized ITensor of any type. Basically, with the current design `ITensor(i, i')` for QN indices makes an ITensor of type `Float64` with flux `QN()`, so `emptyITensor(i, i')` would be the canonical way to create an ITensor with no blocks and undefined flux, where the flux gets determined once elements are set. So it basically is interpreted as an ITensor with the specified indices, but no storage yet. So the main use case is in functions like `op` where you want a generic way to initialize the ITensor at the beginning.

I am still deciding on some design choices, like if you call `getindex` on and `emptyITensor`, does it return `0.0`? I think we want to treat it as an additive zero, so I would say yes, but for now I'll only define `setindex!`.

mtfishman:
In the latest few commits, I removed the NDTensors module from ITensors.jl and added NDTensors.jl as a dependency. I set the compat version of NDTensors up to `v0.1.1` since this PR is now using the `Empty` storage type which I introduced in NDTensors `v0.1.1` (and is the latest registered version of that package).

This branch is ready to go, as long as we are good with the interface for `emptyITensor`. In the latest version, the current syntax for constructing an ITensor with `Empty` storage and `Any` number of indices is:
```julia
A = emptyITensor(Any)
order(A) == Any
# A[] = 2.0  # This would fail
i = Index(2)
for _ in 1:4
  A += randomITensor(i, i')
end
```
The `Any` represents that it is making an ITensor with `Any` number of indices. The reason I picked that interface was to distinguish between that usage of `emptyITensor` and creating an ITensor with `Empty` storage of order 0 (i.e. 0 indices, instead of `Any` number of indices). For example:
```julia
A = emptyITensor()
order(A) == 0
# A + randomITensor(i, i') # This fails, since they have different numbers of indices
A[] = 2.0  # Creates a scalar ITensor (which has Dense storage of length 1)
```
My reasoning is that an `emptyITensor(Any)` has quite exotic behavior and I expect will have very narrow use cases, so I wanted to make that very clear. Perhaps it is a strange abuse of notation, however.

emstoudenmire:
Great. I think this design makes sense, because it differentiates between the case of just wanting to delay creating an expensive storage (this is the use case I foresee for emptyITensor(i,i’); are there others?) versus wanting the kind of “Any” behavior. 

For people who want to be lazy about typing `empyITensor(Any)`, we could make an alias 
```
anyITensor() = emptyITensor(Any)
```
Only half-joking!

The only other thing I’d want to be sure about is whether using the `Any` type creates some kind of inefficiency with Julia’s type system or something like that. But I doubt it does in this case; I guess it mainly just defines another type of ITensor, alongside ITensor{0}, ITensor{1}, etc.

mtfishman:
I can't think of other uses of `emptyITensor(i, i')` besides a generic constructor where you don't know information about the storage you want yet (i.e. the element type or quantum number flux).

Another alias we could use for `emptyITensor(Any)` could be `ITensor{Any}()` (very similar to your suggestion `anyITensor()`). Additionally, we could consider defining `ITensor{3}()` as a way to create an ITensor with `Empty` storage of order `3` where you don't know what the indices are yet (or alternatively `emptyITensor(Val{3})`). I just realized this could allow us to get around the in-place operations limitation I had mentioned, for example this would work:
```julia
A = ITensor{2}() # ITensor with Empty storage and 2 unspecified indices
i = Index(2)
A[i => 1, i' => 2] = 2.5   # This makes an ITensor with indices (i, i')
```
Also, this would allow us to do loops with in-place addition, where you don't want to specify the indices beforehand:
```julia
A = ITensor{3}()
i = Index(2)
for _ in 1:5
  A .+= randomITensor(i,i',i'')
end
```
which wouldn't be possible starting from an `emptyITensor(Any)`/`ITensor{Any}()`. Just some food for thought, no rush to add it right now.

I was also wondering if there are any efficiency issues. I believe just using `Any` as a type parameter should be no different from using a number (or any other type) as a type parameter, since it is used purely as a type parameter for dispatch and the storage of the IndexSet is not abstract (if you look at the new definition of the `IndexSet` type you can see what I mean by that). If it is purely used as a type parameter, then it is just used in dispatch and I think is even treated as a "concrete" type, i.e.:
```julia
julia> isconcretetype(Any)
false

julia> struct T{N} end

julia> isconcretetype(T{Any})
true
```
An alternative is that we could make our own simple internal type `struct AnyOrder end` for this purpose instead of using Julia's `Any`. But `Any` already represents a very similar concept to what we are using it here for, i.e. it is like `ITensor{<:Any} = ITensor{N} where {N<:Any}`. However, that is an abstract type, so you can't make an instance of an ITensor of that type so the closest you can get is the type `ITensor{Any}`.

PR #353, opened 2020-05-11
Simplify output of Base.show for IndexSet types

emstoudenmire:


emstoudenmire:
I checked and this also works for `IndexSet{Any}`. It outputs "IndexSet{Any}" for that case.

PR #354, opened 2020-05-11
In-place permute! function for ITensor type

emstoudenmire:
It's convenient sometimes to have `permute!` for an ITensor versus only `permute`. So this pull request adds that function. The implementation simply calls `permute` and "steals" the store and inds references from the newly created ITensor. Perhaps there's a better way to do it?

### DIFF
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1001,6 +1001,14 @@ permute(T::ITensor,
         inds::Index...) = permute(T,
                                   IndexSet(inds...))
 
+"""
+    permute!(T::ITensors, inds)
+    permute!(T::ITensors, inds::Index...)
+
+Modify the ITensor T to have indices permuted according
+to the input indices inds. The storage of the ITensor
+is permuted accordingly.
+"""
 function Base.permute!(T::ITensor{N},new_inds) where {N}
   Tp = permute(T,new_inds)
   T.store = Tp.store
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1001,6 +1001,17 @@ permute(T::ITensor,
         inds::Index...) = permute(T,
                                   IndexSet(inds...))
 
+function Base.permute!(T::ITensor{N},new_inds) where {N}
+  Tp = permute(T,new_inds)
+  T.store = Tp.store
+  T.inds = Tp.inds
+  return T
+end
+
+Base.permute!(T::ITensor,
+              inds::Index...) = permute!(T,IndexSet(inds...))
+
+
 Base.:*(T::ITensor, x::Number) = itensor(x*tensor(T))
 
 Base.:*(x::Number, T::ITensor) = T*x
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1011,8 +1011,8 @@ is permuted accordingly.
 """
 function Base.permute!(T::ITensor{N},new_inds) where {N}
   Tp = permute(T,new_inds)
-  T.store = Tp.store
-  T.inds = Tp.inds
+  setstore!(T,store(Tp))
+  setinds!(T,inds(Tp))
   return T
 end
 
--- a/test/itensor.jl
+++ b/test/itensor.jl
@@ -609,6 +609,19 @@ end
     for ii ∈ 1:dim(i), jj ∈ 1:dim(j), kk ∈ 1:dim(k)
       @test A[k=>kk,i=>ii,j=>jj]==permA[i=>ii,j=>jj,k=>kk]
     end
+
+    Aorig = deepcopy(A)
+    permute!(A,k,j,i)
+    @test k==inds(A)[1]
+    @test j==inds(A)[2]
+    @test i==inds(A)[3]
+    for ii ∈ 1:dim(i), jj ∈ 1:dim(j), kk ∈ 1:dim(k)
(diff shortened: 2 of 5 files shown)


emstoudenmire:
@mtfishman please let me know if this looks ok to you before merging. Mostly it's for convenience, so a user doesn't have to write `T = permute(T,k,j,i)` and can just write `permute!(T,k,j,)`. Then later we could always modify the implementation if there's a better one to be had.

mtfishman:
This made me think about the copying behavior of `permute`. It seems like if you call `permute!(A, (i, i'))`, if `A` has indices ordered as `(i, i')` then it shouldn't doing anything to the storage. However, right now it copies the storage. We could put a check at this level for that. I was wondering if we should give the option for similar functionality for `permute`. We could do a similar thing as for `dag`, where by default it tries to do as little work as possible (and maybe returns a view), but we have a keyword argument `always_copy` that makes sure it is never a view (but doesn't do superfluous copies either).

src/itensor.jl
```diff
@@ -1001,6 +1001,25 @@ permute(T::ITensor,
...
+    permute!(T::ITensors, inds)
+    permute!(T::ITensors, inds::Index...)
+
+Modify the ITensor T to have indices permuted according
+to the input indices inds. The storage of the ITensor
+is permuted accordingly.
+"""
+function Base.permute!(T::ITensor{N},new_inds) where {N}
+  Tp = permute(T,new_inds)
+  T.store = Tp.store
```
mtfishman, line 1014: Could you rewrite this as `setstore!(T, store(Tp))`? Generally I prefer using functions to access/modify fields since it is a bit more robust.

test/itensor.jl
```diff
@@ -609,6 +609,19 @@ end
     for ii ∈ 1:dim(i), jj ∈ 1:dim(j), kk ∈ 1:dim(k)
       @test A[k=>kk,i=>ii,j=>jj]==permA[i=>ii,j=>jj,k=>kk]
     end
+
+    Aorig = deepcopy(A)
```
mtfishman, line 613: Is the `deepcopy` here necessary? I think our regular `copy(::ITensor)` is already a "deep copy" in the sense that it copies the indices and storage. Partially I'm asking because this started to make me wonder about the copying behavior of ITensors...

src/itensor.jl
emstoudenmire: Yes, good point!

test/itensor.jl
emstoudenmire: Will change, thanks

mtfishman:
Somehow it seems backwards to define `permute!` in terms of `permute` instead of the other way around, but I think that is the way to go in this case...

emstoudenmire:
Yes, I was wondering the same thing about if there should be a check for the case of a trivial permutation, in which case ideally the function would just do nothing. But I hadn't thought about how that brings back up the issue of whether one can be sure of a copy if that behavior is needed. So I agree having an alway_copy keyword argument is the best solution.

I guess it wouldn't hurt to just implement the permutation checking and always_copy logic online in `permute` and not separately in `permute!`? I think the setstore! and setinds! calls within `permute!` won't have any significant cost no matter what `permute` does.

mtfishman:
That sounds good to implement it only in `permute` and then `permute!` just uses the `always_copy = false` version. I think it is fine to always call `setstore!` and `setinds!` (I guess those should only be pointer manipulations).

A question is, does `permute` default to `alway_copy = true` or `false`? I would lean towards `false`.

mtfishman:
A more subtle issue comes up that for `always_copy = false`, storage types like `Diag` and `DiagBlockSparse` in general do not need to manipulate the storage, so I think could always return a view. That would require the logic of `always_copy` to reach down into `NDTensors` (which is fine, but more involved).

We could always implement those optimizations later, and emphasize that if `always_copy = false`, we can't guarantee that the view vs. copy behavior will remain the same over time (so we don't consider that to be a breaking change).

emstoudenmire:
Changed to Draft mode and removed 0.1 Milestone in order to work on getting the `always_copy` behavior settled, which is an important thing to think through. On the other hand, this function is not a blocker for 0.1.

mtfishman:
One potential breaking change that came up here is if we should have `permute` by default return a view if possible (i.e. default to `always_copy = false`). I would lean towards that behavior, in which case it would be breaking if we introduced it later.

emstoudenmire:
Yes, that's an important thing to decide about pre-0.1. I am in agreement with you, that `permute` is allowed to return a view. Besides the performance gains, one other nice thing about that is that if someone asks for a trivial permutation the function just succeeds and doesn't annoyingly have to raise a warning or an error. (There could be good reasons to ask for a trivial permutation, like just to make sure of a certain index ordering as a guarantee.)

Since we will go forward with that behavior (as we both agree), doesn't it become a non-breaking change? What I mean is that now the only statement in the docs is that `permute` might or might not return a view, and you can't rely on its behavior as it could be different under any number of circumstances. Of course later we might broaden the number of cases where it returns views, but in any circumstance users should not assume it will do this.

mtfishman:
Yes, I just meant that it would be breaking from the current behavior, so we should decide on it pre-0.1. Agreed that we can add features like `permute` for `Diag` always returning a view by default post-0.1 without considering it breaking.

I can make a PR for the simple behavior, that for a trivial permutation it returns a view by default.

PR #355, opened 2020-05-12
add read/write for MPS and MPO via HDF5

jtschneider:
Hello everybody,
I added a read/write function via the HDF5 package for an MPO and MPS, respectively.
I wanted this small convenience in my script and just copy/pasted the template from the read/write functions for an ITensor. I hope this suffices.

Should the test for reading/writing be adapted accordingly?

Cheers,
Jan


emstoudenmire:
Hi Jan, thanks - this looks great. The main changes I'd like to ask for are just these:
1. please do go ahead and add a unit test inside of test/readwrite.jl, along the lines of the other tests there (using the same filename for the output file, since it gets deleted by the test at the end)
2. please also write and read back the .llim and .rlim fields inside the MPS / MPO. Your implementation is already correct in the sense of not having any bugs per se, but if an MPS is canonicalized, which is what the llim and rlim record information about, then this information would be lost and extra work would have to be done to re-canonicalize the MPS after reading it back in.

PR #357, opened 2020-05-12
Performance regression in ITensor broadcasting operations

mtfishman:
There is a very strange performance regression with ITensor broadcasting operations. You can see it in the following example:
```julia
julia> i = Index(100);

julia> @btime axpy!(2, A, B) setup = (B = randomITensor(i, i'); A = randomITensor(i', i));
  498.087 μs (49533 allocations: 775.45 KiB)

julia> @btime map!((y, x) -> 2 * x + y, B, B, A) setup = (B = randomITensor(i, i'); A = randomITensor(i', i))
  9.657 μs (31 allocations: 1.83 KiB)
```
`axpy!(a, x, y)` is just defined [here](https://github.com/ITensor/ITensors.jl/blob/master/src/itensor.jl#L1172) in terms of the broadcast call `y .+= a .* x`, so that is testing the broadcasting performance.

The strange part about this is that the broadcast call `y .+= a .* x` is _itself_ defined in terms of the `map!` call above, as you can see [here](https://github.com/ITensor/ITensors.jl/blob/master/src/broadcast.jl#L256). It is therefore a complete mystery to me why the performance above is so different.

This regression is currently causing a huge slowdown in `dmrg`, for example running `examples/1d_Heisenberg_dmrg.jl`:
```
After sweep 1 energy=-138.837128197001 maxlinkdim=10 time=0.500
After sweep 2 energy=-138.937288615100 maxlinkdim=20 time=0.997
After sweep 3 energy=-138.940084627227 maxlinkdim=100 time=7.992
After sweep 4 energy=-138.940086090413 maxlinkdim=100 time=21.790
After sweep 5 energy=-138.940086113354 maxlinkdim=122 time=28.010
```
which is clearly quite slow.

The "fix" in this PR is to call `map!` directly in `axpy!`, instead of indirectly through the slow broadcast call (which, again, is defined in terms of `map!` anyway...). That improves the `dmrg` performance back to what we would expect:
```
After sweep 1 energy=-138.832801271759 maxlinkdim=10 time=0.216
After sweep 2 energy=-138.937263887382 maxlinkdim=20 time=0.297
After sweep 3 energy=-138.940084729709 maxlinkdim=100 time=1.676
After sweep 4 energy=-138.940086098620 maxlinkdim=100 time=4.005
After sweep 5 energy=-138.940086121925 maxlinkdim=122 time=5.478
```

It is a pretty dissatisfying "fix", and the broadcast call is still slow. I haven't benchmarked other broadcast calls, it would be interesting to see if this is a general problem and figure out how to make the broadcast call fast again.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -257,7 +257,7 @@ function Base.copyto!(T::ITensor,
     else
       bc_bc_bc = find_type(Broadcasted, bc_bc.args)
       if isnothing(α)
-        α = find_type(Number, bc_bc_bc.args)
+      #  α = find_type(Number, bc_bc_bc.args)
         B = find_type(ITensor, bc_bc_bc.args)
       else
         A, B = bc_bc_bc.args
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -257,12 +257,12 @@ function Base.copyto!(T::ITensor,
     else
       bc_bc_bc = find_type(Broadcasted, bc_bc.args)
       if isnothing(α)
-      #  α = find_type(Number, bc_bc_bc.args)
+        β = find_type(Number, bc_bc_bc.args)
         B = find_type(ITensor, bc_bc_bc.args)
       else
         A, B = bc_bc_bc.args
       end
-      mul!(T, A, B, α, 1)
+      mul!(T, A, B, β, 1)
     end
   else
     error("When adding two ITensors in-place, one must be the same as the output ITensor")
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1169,15 +1169,9 @@ end
 w .+= a .* v
 ```
 """
-function LinearAlgebra.axpy!(a::Number,
-                             v::ITensor,
-                             w::ITensor)
-  return map!((r, t) -> r + a * t, w, w, v)
-end
-
-#LinearAlgebra.axpy!(a::Number,
-#                    v::ITensor,
-#                    w::ITensor) = (w .+= a .* v)
+LinearAlgebra.axpy!(a::Number,
+                    v::ITensor,
+                    w::ITensor) = (w .+= a .* v)
 
 """
 axpby!(a,v,b,w)
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1169,9 +1169,15 @@ end
 w .+= a .* v
 ```
 """
-LinearAlgebra.axpy!(a::Number,
-                    v::ITensor,
-                    w::ITensor) = (w .+= a .* v)
+function LinearAlgebra.axpy!(a::Number,
+                             v::ITensor,
+                             w::ITensor)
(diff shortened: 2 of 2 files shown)


mtfishman:
As some more background, this regression was introduced in this commit: 7449c053ffc6022ae329fb3915a87778106b3d43. In that commit, the old ITensor function `apply!` was replaced by the function `map!`, which essentially did the same thing with a new interface. Before that commit (i.e. bf90e9ef9646360edf5117a8a95550ac1288c4a4) the problematic broadcast call was using `apply!`, without issue.

emstoudenmire:
I'm really glad you were able to find the issue, but I see why it's kind of mystifying. 

Do you think it's possible that there are different more or less specialized versions of `map!` which are being generated by the compiler, and one way of calling `map!` directly calls a more specialized, efficient version whereas calling it through `axpy!` was ultimately calling a less specialized version? I'm not sure; it's just a guess.

It would be interesting to call those two functions and see if exactly the same `map!` code (with exactly the same argument types) is being called.

mtfishman:
Ok, this performance issue is almost too strange to believe. In the latest version of this branch, I reverted the function `axpy!` to use the broadcast call `y .+= a .* x`. I only changed the broadcast call by commenting out a single line [here](https://github.com/ITensor/ITensors.jl/pull/357/commits/42fa3b71f811399acbafb02dbd1e30e0e09a4bc6#diff-2d8d5b5a5cc004c730be016fd93348d8R260) that is within a conditional statement that is not called by `axpy!` (I confirmed that it is not being called by putting in a print statement). Now on this branch, the performance issue is fixed:
```julia
julia> i = Index(100)
(dim=100|id=181)

julia> @btime axpy!(2, A, B) setup = (B = randomITensor(i, i'); A = randomITensor(i', i));
  11.983 μs (41 allocations: 2.08 KiB)

julia> @btime map!((y, x) -> 2 * x + y, B, B, A) setup = (B = randomITensor(i, i'); A = randomITensor(i', i));
  9.577 μs (31 allocations: 1.83 KiB)
```
I also confirmed that the `dmrg` performance is back to what we would expect. You just can't make this stuff up.

I'm sure we could rewrite that function in some other way to avoid all of those if-statements (for example we could use dispatch to split up that function). I can't even begin to guess how commenting out that line creates such a performance discrepancy. In my profiling, the actual low-level Array permutation is somehow slowed down, not anything within that function (both when I use `Strided` and our own `permutedims!` function). That all happens beneath many layers of function barriers. I wish we could make a minimal example of this issue, but I'm not sure how we could go about doing that.

mtfishman:
Note that the only test failure is a test of the broadcast call `C .= a .* A .* B`, which uses the branch of code with the line that is commented out.

mtfishman:
I'll merge this since it is a minimal fix, I'm sure this code will get rewritten in a better way at some point.

PR #361, opened 2020-05-13
Add BenchmarkCI support

mtfishman:
This (hopefully) adds support for running the benchmarks defined in `benchmark/benchmarks.jl` at every commit, using [BenchmarkCI.jl](https://github.com/tkf/BenchmarkCI.jl).

It is set up to compare against the latest master branch.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,21 @@
+name: Run benchmarks
+
+on:
+  pull_request:
+
+jobs:
+  Benchmark:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: 1.4
+      - name: Install dependencies
+        run: julia -e 'using Pkg; pkg"add PkgBenchmark BenchmarkCI@0.1.2"'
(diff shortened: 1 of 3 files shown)


emstoudenmire:
I'm really excited to see this. I wonder if you also noticed that the benchmark results appear 3 times. Is that just 3 different machines running them or is there some duplication going on?

mtfishman:
Looks like it worked!

I only see two results, the "Target result" (this PR) and the "Baseline result" (the master branch). The comparison is shown in the "Results" category, but the table is empty since there were no significant speedups or regressions.

emstoudenmire:
Ah ok yes I somehow counted 3 as I was scrolling up and down but now I see there's just the two and what they are. I was confused also as to why they were the same but I see it's just because things are only getting started and from now on we'll be seeing differences.

This is really great to have and to be an automated thing.

emstoudenmire:
At some point we should remove Travis from the CI settings, since Github Actions is running a good bit quicker most of the time. I guess it's still there mainly for Documenter?

mtfishman:
I think this will be really useful going forward, I'll merge this.

Yeah, right now Travis is building the documentation, but I agree we should get rid of it. There's a guide to build the documentation with Github actions, so we can just use that for everything.

PR #362, opened 2020-05-14
Remove travis, build docs with Github Actions

mtfishman:
This removes Travis (since it has been pretty slow) in favor of fully using Github Actions. Now documentation should be built by Github Actions, and I also added a macOS build to the Github Actions test.


mtfishman:
One thing I'm not sure about with this new setup is if this still includes code coverage, since it seemed like before it was set up for both Travis and Github Actions. It looks like it is included at the bottom of `.github/workflows/tests.yml` (which I didn't change). However, I don't see this PR on https://codecov.io/gh/ITensor/ITensors.jl/pulls and it didn't post a code coverage comment like it usually does. Perhaps it is just a temporary issue or we can look into it later.

emstoudenmire:
Great - yes in the Travis file there’s an explicit line turning on coverage, so perhaps the actions file needs that too (or a separate action maybe). We can add this back in another PR.

mtfishman:
The actions file looks like it has a line for code coverage already: https://github.com/ITensor/ITensors.jl/blob/master/.github/workflows/tests.yml#L32. I'm a bit confused which one was being used before. Oh well, we'll see going forward.

PR #365, opened 2020-05-15
Split make.jl into multiple scripts for deployment and local building of docs, with and without doctests

emstoudenmire:
This PR moves the ITensor-specific Documenter configuration into a file docs/settings.jl. It also introduces new scripts make_local_test.jl and make_local_notest.jl for building docs locally without deployment and also without running (time consuming) doc tests in the case of make_local_notest.jl.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,10 +1,5 @@
 include("settings.jl")
 
-makedocs(sitename = sitename;
-         modules = modules,
-         pages = pages,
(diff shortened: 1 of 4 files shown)


emstoudenmire:
@mtfishman I’m going to pull this so I can keep adding new docs pages. No problem if you want to change it around some more in later PRs (like changing the type of the settings object etc.)

PR #382, opened 2020-05-20
Fix the correct QNs for the indices introduced in eigen

mtfishman:
This fixes a bug brought up by Steve White where sometimes the new Index introduced by eigen had the wrong flux, causing errors later on related to some blocks having the wrong flux.

The fix is just to manually set the correct QNs of the new Index, based on the constraint that `flux(V) == QN()` and `flux(D) == QN()`.

### DIFF
--- a/src/decomp.jl
+++ b/src/decomp.jl
@@ -150,6 +150,12 @@ function LinearAlgebra.eigen(A::ITensor{N},
                              Linds,
                              Rinds;
                              kwargs...) where {N}
+  @debug begin
+    if hasqns(A)
+      @assert flux(A) == QN()
+    end
+  end
+
   NL = length(Linds)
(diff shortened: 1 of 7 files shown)


mtfishman:
Also note there are some `@debug` statements introduced, mostly for checking flux consistencies (in particular, a debug check is added to `flux(::ITensor)` that checks that all blocks indeed have the same flux). These can be turned on by typing:
```julia
julia> ENV["JULIA_DEBUG"] = ITensors
```
and turned off by typing:
```julia
julia> delete!(ENV, "JULIA_DEBUG")
```
Unfortunately, it prints something whether or not the code in `@debug` prints or throws an error, so it adds a lot of noise right now. Hopefully there is a way to suppress output to only what we want.

PR #384, opened 2020-05-21
Add norm, lognorm, loginner for MPS/MPO

mtfishman:
It turned out `norm` for MPS and MPO was not implemented, and was calling a generic Julia function which somehow actually ran (!) but output garbage (probably something like the sum of squared norms of the individual tensors). This implements `norm`, as well as `lognorm` and `loginner`/`logdot` for MPS/MPO that are not gauged, where generally the actual norm can diverge or go to zero but the log norm is well behaved.


mtfishman:
I believe that what is implemented here would be the natural definition of `inner(A::MPO, B::MPO)`, which would be the Frobenius inner product `tr(A^{\dagger} B)`. I convinced myself that this naturally translates in ITensor pseudocode to essentially `dag(A)*B` if `A` and `B` have all shared site indices. It isn't literally implemented as `dag(A)*B`, since that operation isn't defined if `A` and `B` share all site indices (only one pair of site indices, I think).

emstoudenmire:
Looks good 👍 Yes I agree that’s the norm we should be using by default for MPOs. For one reason, it’s the norm that would result when mapping it to an MPS. More importantly, I think it’s the norm that would relate to the L2 distance of thinking of the MPO as a single tensor and then viewing the data of this tensor as a “vector” for comparing distances between two MPOs.

emstoudenmire:
Seeing the failing test here makes me think the following: one good test to have would be of an MPO whose norm we can calculate analytically (e.g. the identity MPO, which has a norm that grows exponentially as d^N). Then the lognorm test would compare the output of lognorm of this particular MPO (identity MPO) to the value N*log(d), or similar for a different MPO with a simple enough structure. That way we would side-step any floating point issues, relying instead on the analytically known result.

mtfishman:
I'm already basically doing that, I'm taking `M = randomMPO(sites)` which has norm 1 and doing `for n in 1:N M .*= n end`, so the norm should be `n!`. I just forgot to put that at the top of the test, so one of the tests involves a norm 1 MPO (so the log ends up nearly zero). Fixing it now.

emstoudenmire:
Oh I see, great!

PR #387, opened 2020-05-22
First draft of advanced usage guide

mtfishman:
Here is a first draft for a set of advanced topics in the documentation. The plan is to cover things like:

 - [x] Installing and updating ITensors.jl.
 - [x] Writing code based on ITensors.jl.
 - [x] Developing ITensors.jl and Revise.
 - [ ] Style guide for contributing (as suggested here: #390).
 - [x] Making code and packages depending on ITensors.jl.
 - [x] Compiling the package with PackageCompiler.
 - [x] Writing performant code.
 - [x] ITensor type design and writing performant code (including discussions of type instability and function barriers).
 - [x] Benchmarking and profiling.
 - [x] Multithreading and Strided.jl.
 - [x] Discussing the relationship between NDTensors and ITensors.
 - [x] ITensor in-place operations and broadcasting.

This is a draft that needs to be cleaned up and expanded a bit.

### DIFF
--- a/docs/src/AdvancedUsageGuide.md
+++ b/docs/src/AdvancedUsageGuide.md
@@ -39,13 +39,13 @@ you can do `add ITensors#master`. You can switch back to the latest
 released version with `add ITensors`. Using the development/master
 branch is generally not encouraged unless you know what you are doing.
 
-## Writing code based on ITensors.jl
+## Using ITensors.jl in the REPL
 
 There are many ways you can write code based on ITensors.jl, ranging 
 from using it in the REPL to writing a small script to making a 
 package that depends on it.
 
-For example, you can just start the REPL from your command like:
+For example, you can just start the REPL from your command line like:
 ```
 ~ julia
 ```
@@ -63,7 +63,7 @@ julia> A = randomITensor(i, i')
 ITensor ord=2 (dim=2|id=355|"i") (dim=2|id=355|"i")'
 NDTensors.Dense{Float64,Array{Float64,1}}
 
-julia> @show A
+julia> @show A;
 A = ITensor ord=2
 Dim 1: (dim=2|id=355|"i")
 Dim 2: (dim=2|id=355|"i")'
@@ -136,8 +136,6 @@ NDTensors.Dense{Float64,Array{Float64,1}}
 A common place you might accidentally come across this is the 
 following:
 ```julia
-Reminder to add a section about this kind of "problem":
-```julia
 julia> N = 4;
 
 julia> sites = siteinds("S=1/2",N);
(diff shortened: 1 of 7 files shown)


mtfishman:
Reminder to add a section about this kind of "problem":
```julia
julia> N = 4;

julia> sites = siteinds("S=1/2",N);

julia> ampo = AutoMPO();

julia> for j=1:N-1
         ampo += ("Sz", j, "Sz", j+1)
       end
ERROR: UndefVarError: ampo not defined
Stacktrace:
 [1] top-level scope at ./REPL[10]:2
 [2] eval(::Module, ::Any) at /Applications/Julia-1.4.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 [3] eval_user_input(::Any, ::REPL.REPLBackend) at /Applications/Julia-1.4.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 [4] run_backend(::REPL.REPLBackend) at /Users/gtorlai/.julia/packages/Revise/AMRie/src/Revise.jl:1023
 [5] top-level scope at none:0
```
and provide some alternatives like `.+=`, `add!`, or surrounding the code in a function or let-block.

docs/src/AdvancedUsageGuide.md
```diff
@@ -0,0 +1,488 @@
...
+  for i in 1:length(A.data)
+    t += A.data[i]
+  end
+  return A
+end
+```
+
+Therefore, users should keep this in mind when they are writing 
+ITensors.jl code, and we warn that explicitly looping over large 
+ITensor by individual elements should be avoiding in performance
```
mtfishman, line 419: Maybe rephrase to "should be done with caution", since there may be ways to make it fast with type declarations.

emstoudenmire: That sounds good. What about also pointing to some alternatives to looping over ITensor elements, such as using broadcasted calls?

mtfishman: Yes of course, that is also discussed in the next section but can be pointed out here.

mtfishman:
There are a few more involved topics like:
 - Making a project that depends on ITensors.jl.
 - More details on contributing to ITensors.jl, such as:
   - Developing both NDTensors and ITensors simultaneously
   - A style guide.
 - Using PackageCompiler to compile ITensors.jl.
which are marked in the documentation as "Coming soon", but overall I think this is in a pretty good place to get merged so we can start pointing people towards it.

mtfishman:
@emstoudenmire, if you have time to glance through this and point out anything objectionable that would be great, otherwise I'll just merge and we can edit it as we go along.

emstoudenmire:
@mtfishman this is all excellent. It will be very helpful for users (and for me!).

The main change I'd suggest is about the wording of the "Make a project based on ITensors.jl" section. We should make it clearer that it's not necessary to develop a package in order to use ITensors.jl for a project, because for that one could just develop a set of scripts collected into a folder, which ought to be good enough in many cases. So then we should briefly explain that this is a more advanced approach and say what its advantages are, such as being able to do `using MyProject` in the REPL. (And other advantages?)

mtfishman:
That's a good suggestion, I'm emphasizing that now.

Issue #389, opened 2020-05-22
throw exception/debug warning if an index mismatch occurs at `product(P::ProjMPO,wf)`


orialb:
Currently if `product(P::ProjMPO,wf)` is called and there is an index mismatch between `P` and `wf` (due to the[ user not being careful ](https://github.com/ITensor/ITensors.jl/issues/386#issuecomment-632639197)and calling orthogonalize! for example) it will fail silently and just return an `ITensor` with more indices which could be confusing to debug (of course the iterative method later on would fail).

It could be nice to throw an exception or at least some debug warning to indicate that something went wrong, e.g. something like this:
```julia
function product_with_err(P::ProjMPO,
                          v::ITensor)::ITensor
  Hv = v
  if isnothing(lproj(P))
    if !isnothing(rproj(P))
      Hv *= rproj(P)
    end
    for j in P.rpos-1:-1:P.lpos+1
      Hv *= P.H[j]
    end
  else #if lproj exists
    Hv *= lproj(P)
    for j in P.lpos+1:P.rpos-1
      Hv *= P.H[j]
    end
    if !isnothing(rproj(P))
      Hv *= rproj(P)
    end
  end
  order(Hv)==order(v) || throw("Order of the ProjMPO-ITensor product P*v is not equal to the order of v,"*
                               "this is probably due to an index mismatch. Try rebuilding the `ProjMPO`.")
  return noprime(Hv)
end
```

It seems to not hurt much:
```julia
using BenchmarkTools
sites = siteinds("S=1/2",10)
ampo = AutoMPO()
for j=1:length(sites)-1
  add!(ampo,1,"Sz",j,"Sz",j+1)
  add!(ampo,1,"Sx",j)
end
add!(ampo,1,"Sx",length(sites))
H= MPO(ampo,sites)

psi = productMPS(sites, fill("↑",10))
P = ProjMPO(H)
position!(P,psi,3)
wf = psi[3]*psi[4]

@btime product(P,wf) # => 22.344 μs (356 allocations: 33.38 KiB)
@btime product_with_err(P,wf) # => 22.593 μs (356 allocations: 33.38 KiB)
```

I could make a PR if this is acceptable (error message itself could be improved).

emstoudenmire:
[...]

PR #390, opened 2020-05-23
throw error in case of order mismatch in ProjMPO-ITensor product

orialb:
This PR adds a check to `product(P::ProjMPO,wf::ITensor)` which will throw an informative error if the order of `P(wf)` doesn't match `wf`, as discussed in  https://github.com/ITensor/ITensors.jl/issues/389. 
I also implemented `NDTensors.inds(P::ProjMPO)`.

The error message (for the case `nsite(P)==2`) would look like this:
image

I should probably add some tests, but there was no `projmpo.jl` test file, should I create a new one or should the tests be in `dmrg.jl` test file?

Also there are some multi-line function calls there and I wasn't fully sure what formatting guidelines should be followed, so let me know if formatting is off (maybe worth adding some contributors guide section somewhere).

### DIFF
--- a/src/mps/projmpo.jl
+++ b/src/mps/projmpo.jl
@@ -69,7 +69,7 @@ as `v`. The operator overload `P(v)` is
 shorthand for `product(P,v)`.
 """
 function product(P::ProjMPO,
-                 v::ITensor)::ITensor
+                 v::ITensor{N})::ITensor{N} where N
   Hv = v
   if isnothing(lproj(P))
     if !isnothing(rproj(P))
--- a/src/mps/projmpo.jl
+++ b/src/mps/projmpo.jl
@@ -54,6 +54,28 @@ function rproj(P::ProjMPO)
   return P.LR[P.rpos]
 end
 
+function NDTensors.inds(P::ProjMPO)
+  Pinds = []
+  if isnothing(lproj(P))
+    if !isnothing(rproj(P))
+      push!(Pinds, inds(rproj(P))...)
+    else
+      throw("ProjMPO not initialized, call `position!` first.")
+    end
+  else
+    push!(Pinds,inds(lproj(P))...)
+    if !isnothing(rproj(P))
+      push!(Pinds,inds(rproj(P))...)
+    end
+  end
+  for j in P.rpos-1:-1:P.lpos+1
+    push!(Pinds,filter(x->hastags(x,"Site"), inds(P.H[j]))...)
+  end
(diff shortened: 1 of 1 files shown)


mtfishman:
Hey Ori,

Looks like a good check to add.

We could also add a compile-time check, the advantage would be that it would make this function more type-stable (the order of the output should be inferable). I think we could just make the function signature:
```julia
function product(P::ProjMPO,
                 v::ITensor{N})::ITensor{N} where {N}
   [...]
end
```
The runtime check you added should get called as well and is useful to have, so I think we should have both this extra type information and the check and error message you added.

Having an `inds` function for `ProjMPO` is an interesting idea. The main reason I'm kind of hesitant about it is that I think we would want to make the code for all of the "projected operator linear maps" (ProjMPO, ProjMPO_MPS, ProjMPOSum) as simple and generic as possible, so it is easy for other people to implement their own. People might get confused about why `inds(::ProjMPO)` is there and think they have to implement it as well, which is not so trivial in general. Couldn't we just print `inds(v)` and `ind(Hv)` in the error message?

Also it is a good idea to have guidelines for contributing to ITensors.jl. I've started an "advanced usage" section of the docs here: https://github.com/ITensor/ITensors.jl/pull/387, I can start adding that there. I think with multi-line calls you are referring to here, it may be better to split it into multiple lines with temporary variables.

-Matt

orialb:
>Having an inds function for ProjMPO is an interesting idea. The main reason I'm kind of hesitant about it is that I think we would want to make the code for all of the "projected operator linear maps" (ProjMPO, ProjMPO_MPS, ProjMPOSum) as simple and generic as possible, so it is easy for other people to implement their own. People might get confused about why inds(::ProjMPO) is there and think they have to implement it as well, which is not so trivial in general. Couldn't we just print inds(v) and inds(Hv) in the error message?

We could print just `inds(Hv)`, I just thought having the indices of the `ProjMPO` would be slightly more useful for debugging (because it tells the user exactly what kind of tensor it expects).
How about giving that function another name, or clarifying in the documentation (could be in a docstring for `inds(ProjMPO)`) that the only thing a `ProjMPO` like object needs to implement is `product` ? 
But if it's too much of a hassle then I can remove it.

I'll add the compile time check.

mtfishman:
I think we should remove it for now, since `inds(v)` and `inds(Hv)` should give enough information.

orialb:
updated the PR with the type annotation and removed `inds(::ProjMPO)`.

mtfishman:
Looks great, thanks!

PR #391, opened 2020-05-23
small fix for productMPS with ComplexF64 eltype 

orialb:
tiny fix to `productMPS(::T,args...)` to make sure that the first tensor has the requested `eltype`.
I also added a small test.


emstoudenmire:
Good catch!

Issue #395, opened 2020-05-26
possible bug in combiner


orialb:
The following code:
```julia
i1 = Index([QN(0)=>1,QN(1)=>2],"i1")
i2 = Index([QN(0)=>1,QN(1)=>2],"i2")
A = randomITensor(QN(),i1,i2,dag(i1)', dag(i2)')
CL = combiner(i1,i2)
CR = combiner(dag(i1)',dag(i2)')
AC = A *CR *CL

AC*dag(CR)*dag(CL)
```

results in an error:
```
ERROR: BoundsError: attempt to access Dim 1: 1
Dim 2: 1
Dim 3: 4
NDTensors.Dense{Float64,SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}}
 1×1×4
 at index [1, 2, 3, 4, 5, 6, 7, 8]
Stacktrace:
 [1] copyto!(::IndexLinear, ::NDTensors.Tensor{Float64,3,NDTensors.Dense{Float64,SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true}},Tuple{Int
64,Int64,Int64}}, ::IndexCartesian, ::SubArray{Float64,2,Base.ReshapedArray{Float64,2,SubArray{Float64,1,Array{Float64,1},Tuple{UnitRange{Int64}},true},Tuple
{}},Tuple{UnitRange{Int64},UnitRange{Int64}},false}) at ./abstractarray.jl:813
 [2] copyto! at ./abstractarray.jl:798 [inlined]
 [3] uncombine(::NDTensors.Tensor{Float64,2,NDTensors.BlockSparse{Float64,Array{Float64,1},2},IndexSet{2,Index{Array{Pair{QN,Int64},1}},Tuple{Index{Array{Pai
r{QN,Int64},1}},Index{Array{Pair{QN,Int64},1}}}}}, ::IndexSet{3,Index{Array{Pair{QN,Int64},1}},Tuple{Index{Array{Pair{QN,Int64},1}},Index{Array{Pair{QN,Int64
},1}},Index{Array{Pair{QN,Int64},1}}}}, ::Int64, ::Array{Int64,1}, ::Array{Int64,1}) at /Users/oria/src/NDTensors.jl/src/blocksparse/blocksparsetensor.jl:621
[...]

PR #401, opened 2020-05-28
Fix uncombining combiner bug when combined Index is not first index

mtfishman:
This bumps the version of NDTensors, which includes a bug fix for a certain case of uncombining when the index to be uncombined is not the first index in the IndexSet.


mtfishman:
This fixes #395. Need to wait for the new NDTensors version to get registered for the tests to be able to run.

PR #402, opened 2020-05-28
implement exponentiation for ITensors with QNs

orialb:
One thing about this implementation that I'm not sure about- 
currently it will not support the scenario :
``` julia 
i1 = Index([QN(0)=>1,QN(1)=>2],"i1")
i2 = Index([QN(0)=>1,QN(1)=>2],"i2")
A = randomITensor(QN(),i1,i2,dag(i1)', dag(i2)')
Lis = IndexSet(i1,dag(i1)')
Ris = IndexSet(i2,dag(i2)')
exp(A,Lis,Ris )
```
Because the order 2 block-sparse tensor with combined indices `AC= A*combiner(Lis...)*combiner(Ris...)` will not be block diagonal. 

But since `AC` still has zero flux it is my understanding (which might be incorrect) that it should be possible to exponentiate it block by block, preserving the block structure, and the reason it is not represented as a block diagonal tensor is because both indices  of `AC` have an  (arbitrary?) `OUT` direction.

It is very possible that this scenario has no use-case, but it just made me wonder whether I made some mistake in the implementation and it should be also supported.

### DIFF
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1072,26 +1072,67 @@ end
 LinearAlgebra.dot(A::ITensor, B::ITensor) = (dag(A)*B)[]
 
 """
-    exp(A::ITensor, Linds, Rinds = prime(dag(IndexSet(Linds))); ishermitian = false)
+    exp(A::ITensor, Linds=Rinds', Rinds=inds(A,plev=0); ishermitian = false)
 
 Compute the exponential of the tensor `A` by treating it as a matrix ``A_{lr}`` with
 the left index `l` running over all indices in `Linds` and `r` running over all
-indices in `Rinds`. Must have `dim(Linds) == dim(inds(A))/dim(Linds)` for the exponentiation to
-be defined.
+indices in `Rinds`.
+
+Only accepts index lists `Linds`,`Rinds` such that: (1) `length(Linds) +
+length(Rinds) == length(inds(A))` (2) `length(Linds) == length(Rinds)` (3) For
+each pair of indices `(Linds[n],Rinds[n])`, `Linds[n]` and `Rinds[n]` represent
+the same Hilbert space (the same QN structure in the QN case, or just the same
+length in the dense case), and appear in `A` with opposite directions.
+
 When `ishermitian=true` the exponential of `Hermitian(A_{lr})` is
 computed internally.
 """
-function LinearAlgebra.exp(A::ITensor,
-                           Linds,
-                           Rinds = prime(dag(IndexSet(Linds)));
-                           ishermitian = false)
-  CL = combiner(Linds...)
-  CR = combiner(Rinds...)
+function LinearAlgebra.exp(A::ITensor{N}, Linds, Rinds; kwargs...) where N
+  ishermitian=get(kwargs,:ishermitian,false)
+
+  @debug begin
+    if hasqns(A)
+      @assert flux(A) == QN()
+    end
+  end
+
+  NL = length(Linds)
+  NR = length(Rinds)
+  NL != NR && error("Must have equal number of left and right indices")
+  N != NL + NR && error("Number of left and right indices must add up to total number of indices")
+
+  # Linds, Rinds may not have the correct directions
+  Lis = IndexSet(Linds...)
+  Ris = IndexSet(Rinds...)
+
+  Lis = setdirs(Lis, dirs(A, Lis))
+  Ris = setdirs(Ris, dirs(A, Ris))
+
+  for (l, r) in zip(Lis, Ris)
+    if space(l) != space(r)
+      error("In exp, indices must come in pairs with equal spaces.")
+    end
+    if hasqns(A)
+      if dir(l) == dir(r)
+        error("In exp, indices must come in pairs with opposite directions")
+      end
+    end
+  end
+
+  CL = combiner(Lis...)
+  CR = combiner(Ris...)
   AC = A * CR * CL
   expAT = ishermitian ? exp(Hermitian(tensor(AC))) : exp(tensor(AC))
   return itensor(expAT)*dag(CR)*dag(CL)
 end
 
+function LinearAlgebra.exp(A::ITensor;
+                             kwargs...)
+  Ris = inds(A; plev = 0)
+  Lis = Ris'
+
+  return exp(A, Lis, Ris; kwargs...)
+end
 
 """
     product(A::ITensor, B::ITensor)
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1072,24 +1072,26 @@ end
 LinearAlgebra.dot(A::ITensor, B::ITensor) = (dag(A)*B)[]
 
 """
-    exp(A::ITensor, Lis; ishermitian = false)
+    exp(A::ITensor, Linds, Rinds = prime(dag(IndexSet(Linds))); ishermitian = false)
 
 Compute the exponential of the tensor `A` by treating it as a matrix ``A_{lr}`` with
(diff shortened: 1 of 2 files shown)


mtfishman:
Hi Ori,

Thanks for the contribution, this will be useful to have.

The case you bring up is a good one, and I realized that it is also a problem for the `eigen` function right now:
```julia
i1 = Index([QN(0)=>1,QN(1)=>2],"i1")
i2 = Index([QN(0)=>1,QN(1)=>2],"i2")
A = randomITensor(i1, dag(i1)', dag(i2), i2')
eigen(A, (i1,i1'), (i2,i2'))
```
which leads to the same kind of error about requiring block diagonal matrices. This is a very important case to handle for both `eigen` and `exp`, since operators like this with mixed arrows show up naturally in MPS transfer matrices:
```
i_1  ---<--------<----- i_2
          |  T |
i_1' --->-------->----- i_2'
```
(for an infinite MPS with a one-site unit cell `A` and indices `(i_1, s, dag(i_2))`, the transfer matrix `T` is `A * dag(prime(A, not(s)))`).

There are a few things here:

1. Right now, the combiner is a bit arbitrary about the direction it gives to the combined Index. If the inputs to `combiner` are all `Out` or all `In`, the combined Index will be `Out` or `In` respectively. If the arrows of the input indices are mixed, then the combined Index will arbitrarily be `Out`. For applications like this, I can add a keyword argument to `combiner` that lets the user choose the direction of the combined Index. That should fix the problem for both `eigen` and `exp`, since they can specify the combiners should have opposite directions and then the matrices will be block diagonal. (Possibly combiner could pick the direction based on the first input Index, that could also solve this particular problem.)
2. Like in `eigen`, we should check that pairs of indices (i.e. `(Linds[n],Rinds[n])`) have the same spaces and opposite directions. This would define them as proper maps between the same spaces. I believe this restriction makes sense given the definition of the matrix exponential in terms of a power series. See [here](https://github.com/ITensor/ITensors.jl/blob/master/src/decomp.jl#L185) for checks we should add.
3. For functions like this, we are allowing users to pass indices with any directions (many ITensor functions don't require input index directions to be correct, like `eigen`, `svd`, `replaceinds`, etc.). In `eigen`, I explicitly just fix the directions to the correct ones [here](https://github.com/ITensor/ITensors.jl/blob/master/src/decomp.jl#L182).
4. I think it would be nice to design this function like `eigen`, in two layers. The inner layer [here](https://github.com/ITensor/ITensors.jl/blob/master/src/decomp.jl#L149) explicitly accepts two collections of indices. Then, the outer layer [here](https://github.com/ITensor/ITensors.jl/blob/master/src/decomp.jl#L251) wraps the inner layer, and assumes the left set of indices are primed and the right set of indices are unprimed. That convention is used there since the ITensor is seen as a map from the unprimed right indices to the primed left indices, and that function is finding right eigenvectors. I think it would be nice to use the same convention for `exp` (especially considering that the matrix exponential is very related to an eigendecomposition, since it can be computed using an eigendecomposition).

So for now, could you please address (2), (3) and (4)? I can fix (1) for both `eigen` and `exp` in a different PR.

-Matt

mtfishman:
Yes, that is true, there are definitely more general inputs that would allow for proper maps between isomorphic spaces, and we could consider allowing more general inputs in the future. Part of the motivation is the Hermitian case of `exp` (and `eigen`), where not only the size of the space matters but also the ordering of the local Hilbert spaces. For example, for an ITensor with indices `(i', j', dag(i), dag(j))`, it may only be Hermitian if it is considered as a map from `(i, j) -> (i', j')`, but not `(i, j) -> (j', i')`. That becomes more difficult to reason about (for the user and for us) if the function doesn't restrict to pairs of indices of equal spaces and opposite arrows. Just to illustrate why it gets subtle:
```julia
julia> i = Index(2, "i");

julia> j = Index(2, "j");

julia> A = randomITensor(i', j', dag(i), dag(j));

julia> A += swapprime(dag(A), 0, 1);

julia> Cij = combiner(i, j);

julia> Cji = combiner(j, i);

julia> @show A*Cij*Cij';
A * Cij * Cij' = ITensor ord=2
Dim 1: (dim=4|id=457|"CMB,Link")
Dim 2: (dim=4|id=457|"CMB,Link")'
 4×4
 -1.0625949643764634   1.4012742994320175  -1.0021983667374323   -2.4843572644996805
  1.4012742994320175  -1.280522473293159    2.9234750385583164   -0.5344858357713442
 -1.0021983667374323   2.9234750385583164  -0.11195568258709207  -1.3774283944554442
 -2.4843572644996805  -0.5344858357713442  -1.3774283944554442    0.9449780471267247

julia> @show A*Cji*Cji';
A * Cji * Cji' = ITensor ord=2
Dim 1: (dim=4|id=828|"CMB,Link")
Dim 2: (dim=4|id=828|"CMB,Link")'
 4×4
 -1.0625949643764634  -1.0021983667374323    1.4012742994320175  -2.4843572644996805
 -1.0021983667374323  -0.11195568258709207   2.9234750385583164  -1.3774283944554442
  1.4012742994320175   2.9234750385583164   -1.280522473293159   -0.5344858357713442
 -2.4843572644996805  -1.3774283944554442   -0.5344858357713442   0.9449780471267247

julia> @show A*Cji*Cij';
A * Cji * Cij' = ITensor ord=2
Dim 1: (dim=4|id=828|"CMB,Link")
Dim 2: (dim=4|id=457|"CMB,Link")'
 4×4
 -1.0625949643764634   1.4012742994320175  -1.0021983667374323   -2.4843572644996805
 -1.0021983667374323   2.9234750385583164  -0.11195568258709207  -1.3774283944554442
  1.4012742994320175  -1.280522473293159    2.9234750385583164   -0.5344858357713442
 -2.4843572644996805  -0.5344858357713442  -1.3774283944554442    0.9449780471267247
```
Additionally, at least for `eigen`, it is helpful to limit the accepted input to pairs of indices because when the eigenvector is applied to the ITensor operator on the right, it will result in a new vector scaled by the eigenvalue with the same set of indices (though maybe primed, or however else the Index labels might get modified). One could argue that a similar reasoning should apply to `exp`, where the ITensor operator would be used in a Taylor series expansion and must be multiplied by itself over and over.

More practically, if we allowed more general inputs (like maps from 2 indices to 1), for QN ITensors it is not guaranteed that when the two indices are combined into 1 that the combiner would combine them such that the QNs are ordered in the same way as the other Index. For example, for an ITensor mapping `(i,j) -> k`, when `(i,j)` are combined into `c` in general the QNs may be ordered differently than the QNs of `k`. We could try to fix cases like that, but it starts to lead to a rabbithole where allowing more general inputs can lead to much more complicated code, with possibly little payoff.

mtfishman:
Also, to give some context, the C++ version of ITensor only allows pairs of primed and unprimed indices for functions like this, and that had been sufficient for nearly every use case. These versions in the Julia version are already significantly more general by allowing any pairs of indices, not just primed and unprimed pairs.

orialb:
Thanks for taking the time to clarify Matt. 
This makes sense (I was mainly just wondering if I'm not understanding something mathematically about QN tensors or about how they are implemented when asking this), and it's definitely good enough for my use case where I just want to exponentiate few-site operators :)

orialb:
I updated the PR adding the checks similar to `eigen` + adding the method allowing exponentiation without specifying the indices. 
I tried to clarify the restrictions on the indices in the docstring, not sure it is clear enough.

mtfishman:
Looks great, thanks again for your work on this! I'll start working on the new combiner feature so we can fix the issue you brought up in your first comment.

Issue #400, opened 2020-05-28
Define `TagType(::AbstractString)`


mtfishman:
We could consider defining `TagType(s::AbstractString) = TagType{Tag(s)}()` as a nicer syntax for making an instance of a TagType, so instead of `TagType"S=1/2"()` you can write `TagType("S=1/2")`.

This is similar to how Julia defines `Val(x) = Val{x}()`.

emstoudenmire:
Now supported for SiteType and similar types since version 0.1.13

PR #406, opened 2020-05-31
Extensible op function system

emstoudenmire:
This PR redesigns the `op` function system in a backwards compatible
way, where if function overloads with the signature 

    op!(::TagType"TagName",::OpName"Name",Op::ITensor,s::Index)

are present, then these are passed an `emptyITensor(s',dag(s))` to
have its elements written to create the operator named `"Name"`.
The advantages of this approach are:
* extensible by the user just by overloading `ITensors.op!` outside
  of the library code
* automatically generated error message if the opname isn't recognized,
  which the user doesn't have to write or maintain themself
* no more long `if...elseif...end` blocks, and the user facing code
  is very compact
* user doesn't have to construct the ITensor themself and this approach ensures it gets done correctly regarding Index arrows, use of `emptyITensor`, etc.

The only disadvantage so far I can see is that arbitrary strings can't be used as operator names right now, ruling out unicode names. Maybe we can come up with a fix for this but it doesn't seem like a crucial issue.

This PR also includes a function `complex!(::ITensor)` for complexifying
ITensors in-place, which is useful inside other in-place modifying functions.

Only the S=1/2 and S=1 tags are upgraded currently; I will upgrade the rest after this PR is reviewed, since it's a time-consuming process.

Descriptions of commits included:
- Fix Sy definition
- Upgrade spinone.jl to new system
- Update unit tests
- Add backwards compatibility layer for op system
- Implement `complex!(::ITensor)` for use in modifying functions
- Prefer `TagType"S=1/2"`
- Comment newly added `val` functions for now
- Reimplement S=1/2 operators in terms of `op!`
- Initial redesign of `op` function around `op!` overloads
- Generalize SmallString constructor to AbstractString


src/exports.jl
```diff
@@ -239,6 +240,8 @@ export
   TagType,
   @TagType_str,
   op,
+  op!,
```
mtfishman, line 243: Do we really need to export this? Since it is mostly if not entirely for overloading, someone would need to import it or define their overload with:
```
function ITensors.op!(...)
  [...]
end
```
anyway. Just trying to be conservative about exports for these earlier versions of the code.

src/physics/tag_types.jl
```diff
@@ -1,4 +1,6 @@
 
+abstract type AbstractTagType end
```
mtfishman, line 2: Is the abstract type really needed? It doesn't seem to be used anywhere (and `TagType` leaving out the type parameter would probably suffice in any use case).

src/physics/tag_types.jl
```diff
@@ -9,16 +11,47 @@ with certain tags such as "S=1/2".
...
 end
 
-function _call_op(s::Index,
-                  opname::AbstractString;
-                  kwargs...)
+#function val(::TagType{T})::Tag where {T} 
+#  return T
+#end
+
+abstract type AbstractOpName end
```
mtfishman, line 26: Same question about whether or not the abstract type is necessary.

src/physics/site_types/spinhalf.jl
```diff
@@ -22,47 +22,56 @@ function state(::SpinHalfSite,
...
+             s::Index)
+  complex!(Op)
+  Op[s'=>1, s=>2] = -0.5im
+  Op[s'=>2, s=>1] = 0.5im
+end
+
+op!(::TagType"SpinHalf",
+    o::AbstractOpName,
+    Op::ITensor,
+    s::Index) = op!(TagType"S=1/2"(),o,Op,s)
```
mtfishman, line 76: Maybe would be nice to introduce #400 so that we can just use `TagType("S=1/2")` to make the instance of a `TagType` directly.

src/physics/site_types/spinhalf.jl
```diff
@@ -22,47 +22,56 @@ function state(::SpinHalfSite,
...
+
+function op!(::TagType"S=1/2",
+             ::Union{OpName"S-",OpName"Sminus"},
+             Op::ITensor,
+             s::Index)
+  Op[s'=>2, s=>1] = 1.0
+end
+
+function op!(::TagType"S=1/2",
+             ::Union{OpName"Sx"},
```
mtfishman, line 49: Is the `Union` supposed to be here?

src/physics/site_types/spinhalf.jl
```diff
@@ -22,47 +22,56 @@ function state(::SpinHalfSite,
...
+             ::Union{OpName"Sy"},
+             Op::ITensor,
+             s::Index)
+  complex!(Op)
+  Op[s'=>1, s=>2] = -0.5im
+  Op[s'=>2, s=>1] = 0.5im
+end
+
+op!(::TagType"SpinHalf",
+    o::AbstractOpName,
```
mtfishman, line 74: I think it should work to just use `o::OpName` here, since that is interpreted as `OpName{T} where {T}` so should accept any `OpName`.

src/physics/site_types/spinhalf.jl
```diff
@@ -22,47 +22,56 @@ function state(::SpinHalfSite,
...
+
+function op!(::TagType"S=1/2",
+             ::Union{OpName"S+",OpName"Splus"},
+             Op::ITensor,
+             s::Index)
+  Op[s'=>1, s=>2] = 1.0
+end
+
+function op!(::TagType"S=1/2",
+             ::Union{OpName"S-",OpName"Sminus"},
```
mtfishman, line 42: I think I would prefer using the strategy you use down below of having two functions (one which calls the other), instead of using `Union`, but I don't care too much either way. It seems a bit more "extensible" that way, like it makes it a bit clearer how someone would give their own name for an operator.

src/physics/site_types/spinhalf.jl
```diff
@@ -22,47 +22,56 @@ function state(::SpinHalfSite,
...
+             s::Index)
+  Op[s'=>1, s=>2] = +0.5
+  Op[s'=>2, s=>1] = -0.5
+end
+
+function op!(::TagType"S=1/2",
+             ::Union{OpName"Sy"},
+             Op::ITensor,
+             s::Index)
+  complex!(Op)
```
mtfishman, line 68: I meant to make it so that for an `Empty` storage you could set either Real or Complex elements and it would make the correct storage, but haven't done that yet. Would be nice to avoid having to do this.

emstoudenmire:
Very useful comments thanks. I agree with all the changes you suggest and will implement those. I had forgotten the trick of `OpName` already being an abstract parent type of `OpName{T}` which is much cleaner. (I use the abstract type in one place, which is implementing a generic error message for any missing op name overload so users no longer have to make an error condition themselves.)

I had also wondered if emptyITensor would allow setting complex elements or not. I see that right now it has an element type, but since it models the concept of a very generic / overwriteable ITensor and zero is the same (conceptually) for real or complex I agree it would be great to relax the current constraint of needing to do something like `complex!(T)` before setting elements.

src/physics/site_types/spinhalf.jl
emstoudenmire: Agree about removing this in the future in favor of a more flexible Empty storage type

src/physics/site_types/spinhalf.jl
emstoudenmire: Agreed this should be removed - thanks!

emstoudenmire:
Note that I'm planning to upgrade some other functions in the TagType system, including the `state` and `siteinds` functions in a separate PR. The planned upgrades include implementing `siteinds` in terms of a single-site version of the form `siteind(::TagType,n::Int)`. I may also have `state` follow an extensible pattern like `op` does and we could consider support for turning `Pair{Index,String}` into `IndexVal` using the `state` functions.

mtfishman:
Looks like a good design over all, thanks for tackling this!

Regarding the unicode issue, I believe it would mostly be a matter of expanding the storage size of `SmallString`, for example [here](https://github.com/ITensor/ITensors.jl/blob/master/src/smallstring.jl#L2) making `IntChar` `UInt16` and `IntSmallString` `UInt128`. For example:
```julia
julia> UInt8('a')
0x61

julia> UInt8('α')
ERROR: InexactError: trunc(UInt8, 945)
Stacktrace:
 [1] throw_inexacterror(::Symbol, ::Type{UInt8}, ::UInt32) at ./boot.jl:557
 [2] checked_trunc_uint at ./boot.jl:587 [inlined]
 [3] toUInt8 at ./boot.jl:653 [inlined]
 [4] UInt8 at ./boot.jl:709 [inlined]
 [5] UInt8(::Char) at ./char.jl:171
 [6] top-level scope at REPL[27]:1
 [7] eval(::Module, ::Any) at ./boot.jl:331
 [8] eval_user_input(::Any, ::REPL.REPLBackend) at /home/mfishman/software/julia-1.4.0/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86
 [9] run_backend(::REPL.REPLBackend) at /home/mfishman/.julia/packages/Revise/XFtoQ/src/Revise.jl:1162
 [10] top-level scope at none:0

julia> UInt16('α')
0x03b1
```
I'll raise an issue about it, I think it would be nice in general to allow tags with unicode. I only limited it to `UInt8` to make sure it was as efficient as possible at first, but we can do some benchmarking after we change it.

The reason why `Empty` has an element type is that it actually wraps a full storage type, as you can see [here](https://github.com/ITensor/NDTensors.jl/blob/master/src/empty.jl#L6). That is helpful so that when you do something like set an element, `NDTensors` knows what kind of storage to allocate (for example, when `emptyITensor` accepts QN Indices, it make an `Empty{Float64, <: BlockSparse}` storage so it knows to make a `BlockSparse` storage when an element is set). But yeah, it would be no problem to convert `Empty{Float64} -> `Empty{ComplexF64}` if the element being set is complex, it just using `Float64` by default. 

On a side note about the `Empty` type, I recently noticed the package [BangBang.jl](https://github.com/JuliaFolds/BangBang.jl) (which inspired the `contract!!` syntax) actually also has an `Empty` type, as you can see [here](https://github.com/JuliaFolds/BangBang.jl/blob/b5dedefd31d8de5a4974779fb6fe2b2666efe3d4/src/NoBang/emptycontainers.jl#L65). It plays essentially the same role as ours, as a way to generically represent an empty storage that needs to be allocated after doing something like `setindex` or `push`. Interesting that we came across the same design!

emstoudenmire:
That is interesting about the BangBang.jl library and their design with Empty too!

Regarding Unicode, it would definitely be worth investigating that for Tags if it doesn't cause any measureable slowdown of the code (I sort of doubt it will).

src/physics/sitetype.jl
```diff
@@ -0,0 +1,211 @@
...
+  return Op
+end
+
+function op(opname::AbstractString,
+            s::Index;
+            kwargs...)::ITensor
+
+  opname = strip(opname)
+
+  if opname == "Id"
```
mtfishman, line 112: Can't this also be replaced by an overload of `op!`?

emstoudenmire: Yes. I was originally thinking that we would eventually deprecate the `op` back-end / implementation functions entirely in favor of the `op!` ones. So I was just leaving these `op` function in place mostly for backwards compatibility and other reasons. But I like your idea of having this system be overloadable at two levels, the `op` level and the `op!` level with the `op` level calling down to the `op!` level and having a generic implementation if it's not overloaded.

So I'm going to redesign the order in which these functions get called, first going through one or two functions called `op` then drilling down into the `op!` overloads.

emstoudenmire:
Thanks for the careful second review! You can see from my comment above that I like the idea of having two different customization points, either `op!` for the more simple cases of `op` in case one wants to control things even more. Having the generic `op` implementation taking any `SiteType` and `OpName` will also make the code nicer while keeping backward compatibility to the previously-advertised way of doing custom overloads.

emstoudenmire:
@mtfishman ok so this latest version maintains the ability to make custom op function overloads at the level of `ITensors.op!` or `ITensors.op`. Going forward, the second version will mainly be for advanced users to customize the construction of the operator ITensor, as you suggested, but also keeps the code backward compatible with the 0.1.x series of ITensors.jl as a bonus.

The unit test file "sitetype.jl" tests both ways of customizing the `op` function implementation.

I didn't end up managing to use the generic `op(::SiteType,opname,s)` function approach you suggested even though it was a nice suggestion. I thought quite a bit about it, but I think it always introduces a problem that incidental tags on the Index, such as "n=3" end up satisfying that generic implementation, so that the loop which calls `hasmethod(op,...)` ends up thinking that there are multiple, ambiguous implementations of `op` function overloads and doesn't know how to select which one to use.

emstoudenmire:
This update now includes the changes checked off in the list above.

Unfortunately the idea of supporting the "Id" through a generic `op` function is not quite possible in the way the current system works, because it rejects overloads that work for more than one Index tag as ambiguous. Changing the system to just use the first tag for which an `op` method succeeds could allow the "Id" generic overload to work, but then it could lead to cases where an Index has two tags which have overloads defined on them and it would then just pick the first matching tag in an arbitrary way. That might be ok and we can just warn people not to use two special tags on the same Index. That or the system would have to know to act differently for the "Id" op name as it currently does. So hard to think of the best solution but happy to discuss more.

emstoudenmire:
I went ahead and made the move of allowing generic `SiteType` overloads for the following reasons (in order of importance):
- useful and cleaner to have generic cases such as for "Id" handled by separate functions
- loss of the ambiguous-case error message is not so bad: for one thing if two tags have a function, in many cases both functions may do the same thing or a sensible thing; also the behavior is deterministic since tags are sorted
- code is a bit cleaner and shorter now
- code may be slightly faster in some cases

Again, the main drawback of the present version is that if someone made an Index with two tags that both have an `op` overload for the same opname, the current code will silently call the version for whichever tag comes first without any warning.

mtfishman:
I was trying this out and noticed the following:
```julia
julia> @show op("Id", Index(2));
ERROR: ArgumentError: Overload of "op" or "op!" functions not found for operator name "Id" and Index tags ""
```
I think there needs to be a special case when there are no tags (not the most important case to cover, since that won't be very common, but I could see using this as a quick way to make an identity ITensor).

I think there are a lot of advantages to just picking the first tag that is found. If the only disadvantage is not having an error for cases when there are multiple valid SiteTypes, I don't think that will come up very often and we can just document the behavior.

Also, is there a particular reason why `SiteType"x"` is exported but not `SiteType("x")` (and same for `OpName`)?

There may be a performance problem with the current design:
```julia
julia> s = Index(2, "S=1/2");

julia> @btime op("Sx", $s);
  17.779 μs (57 allocations: 3.72 KiB)
```
and with more tags:
```julia
julia> s = Index(2, "A,B,C,S=1/2");

julia> @btime op("Sx", $s);
  42.979 μs (126 allocations: 7.05 KiB)
```
Most of this time is just creating `OpName`s and `SiteType`s:
```julia
julia> @btime ITensors.OpName("Sx");
  4.221 μs (2 allocations: 64 bytes)

julia> @btime ITensors.SiteType("S=1/2");
  4.182 μs (2 allocations: 64 bytes)
```
I believe they are just inherently slow since they are type unstable, since they are making types from runtime values. I have seen similar issues using `Val(x)`.

The `SiteType` constructor dominates when there are many tags to search through. Also, the design of first searching for `op` and then `op!` is problematic, since all of the tags have to get searched twice if only `op!` is defined.

As a small optimizaiton, we may be able to precompute the SiteTypes of the tags (and the `OpName`) so they don't have to be computed twice. 

One design I could think of to mitigate this is having a special marker at the start of the tag you want to be used as a SiteType, for example `"st: S=1/2"`. Then we could search for that special tag at the beginning, and could avoid the loops that involve creating SiteTypes. This would also avoid having to use `hasmethod`, which also takes up a nontrival amount of time, and can allow us to be more flexible in the design of the `op` function (like it would allow us to have a generic `op(::SiteType, ::OpName, ::Index)` function that wraps `op!(::ITensor, ::SiteType, ::OpName, ::Index)`). It doesn't look very nice and shortens the allowed names for SiteType tags, however.

emstoudenmire:
Thanks for taking another detailed look at this. I agree that "Id" case with no tags should work, so I'll see if there's an easy way to handle that case.

The implementation and corner cases of this system are probably something we'll have to refine over time, since it's necessarily doing some fancier 'language reflection' type things so using rather slower parts of the language (making type unstable code, calling things like `hasmethod`). But I agree one of my biggest concerns about this whole approach is the performance, and I've also noticed it has a lot of overhead.

So I can imagine some other ways of doing it that could be faster (though it would have to be tried) like having a generic case that just returns either `nothing` or an empty ITensor, and then would 'signal' that there's no overload for that tag and to try the next tag. But one would have to try it to see that it still meets the design goals and is actually faster. Another approach could just be to try calling the overloads and use a try..catch block to handle the cases where there is no overload to try the next one. But that might be a lot slower even than `hasmethod`; not sure.

Using a special tag convention is something to consider. We might also be able to cache results of these calls, either by using the Memoize library or doing a custom global cache within ITensor.

Finally, I'm just not sure how to handle both the `op` and `op!` overload cases without looping over both, but open to any suggestions.

If I get the no-tag case solved for the generic operators (just "Id" right now) what do you think about merging this then treating the performance as something to be investigated separately? I think for most use cases of AutoMPO right now, it just happens once or twice within most user codes and is still rather fast for such applications. (I just timed AutoMPO within this PR and it takes about 0.02 seconds to make the Heisenberg Hamiltonian for N=100, and 0.2 s for N=1000, which is slow but certainly not a bottleneck of a DMRG code.)

mtfishman:
The `hasmethod` function is actually not the slowest part, from what I saw with profiling. The slowest parts were the constructor calls `SiteType(_)` and `OpName(_)`. So my main concern would be those, which from what I can tell can't be sped up, so we have to minimize calling them. Avoiding the `hasmethod` call could be helpful as well, but I think it is secondary.

I'm not sure there is a way to avoid looping over the tags multiple times to allow overloading `op` and `op!` besides specifying which tag is the SiteType. 

There are cases I am working on with Giacomo, like learning a wavefunction from data, where many operators have to be created on the fly (like from a long list of measurements that were made, so every sweep of the MPS has different operators applied). For small bond dimensions, if not done carefully, just creating the operators dominates the computation time. It would be unfortunate if we ended up with a design that couldn't be optimized enough for those situations. I agree this is less of a concern for AutoMPO with DMRG where a single MPO is made all at once at the beginning of the calculation.

mtfishman:
Memoization or caching is an interesting idea, perhaps that could help in cases that I have in mind (we would have to think about how it would work with indices, since their ids might change, but I guess that should be solvable).

emstoudenmire:
Latest commits include the trick to make construction of Val-like types fast. Timings for calls like `op("Sz",s)` or `op("Id",s)` are now < 5 microseconds on my machine, including cases where the special Index tag is the last of the four tags.

PR #420, opened 2020-06-05
Add swapsites keyword argument to replacebond

mtfishman:
This adds a `swapsites` keyword argument to `replacebond!`. When this is turned on, it would replace bond 3 of an MPS:
```
   s3  s4
   |   |
--A3---A4--
```
with the the following:
```
   s3  s4            s4   s3
   |   |             |    |
--( phi ) --   ->  --A3'--A4'--
```
so it acts like a swap gate, but just does it by inputting the appropriate indices into `factorize`.

The motivation is that this makes defining an operation for swapping sites of an MPS very simple (just a wrapper around `replacebond!` with `swapsites=true`, where `phi = A3 * A4`).

I also added a `replacebond` function that wraps `replacebond!` but returns a new MPS.

@emstoudenmire, I don't recall the C++ having this option, let me know if this kind of interface makes sense to you. We could also consider adding a function `swapbondsites(M::MPS, b::Int)`:
```julia
function swapbondsites(M::MPS, b::Int;
                       ortho::String = "left",
                       cutoff::Real = 1e-15)
  return replacebond(M, b, M[b] * M[b+1];
                     ortho = ortho,
                     swapsites = true,
                     cutoff = cutoff)
end
```
since that is quite useful for gate evolution applications. We would have to distinguish that functionality from something like `swapsiteinds(M::MPS, n1::Int, n2::Int)` (which doesn't exist right now, but might) which instead of doing a swap gate, actually did a `swapinds`.

### DIFF
--- a/src/exports.jl
+++ b/src/exports.jl
@@ -184,6 +184,7 @@ export
   linkind,
   productMPS,
   randomMPS,
+  replacebond,
   replacebond!,
   sample,
   sample!,
--- a/src/exports.jl
+++ b/src/exports.jl
@@ -190,6 +190,7 @@ export
   sample!,
   siteind,
   siteinds,
+  swapbondsites,
   totalqn,
 
 # mps/observer.jl
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -393,6 +393,16 @@ function replacebond(M0::MPS, b::Int, phi::ITensor;
 end
 
 """
+    swapbondsites(ψ::MPS, b::Int; kwargs...)
+
+Swap the sites `b` and `b+1`.
+"""
+function swapbondsites(ψ::MPS, b::Int; kwargs...)
+  kwargs = setindex(values(kwargs), true, :swapsites)
+  return replacebond(ψ, b, ψ[b] * ψ[b+1]; kwargs...)
+end
+
+"""
     sample!(m::MPS)
 
 Given a normalized MPS m, returns a `Vector{Int}`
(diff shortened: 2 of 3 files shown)


emstoudenmire:
This is a great interface to offer - I hadn’t thought about doing it this way, but it makes a nice base layer to implement things like the time evolution codes we talked about on top.

My main review comment would be: let’s make sure this would also work for different kinds of sites, such as alternating S=1/2 and S=1 sites. I think a good way to do this is not to use swapinds internally, but instead just change which indices get passed to factorize as being the ones that go onto the unitary afterward.

mtfishman:
Yes, I was thinking about the fermion case as well. I think for now people just have to make their own fermionic swap gates...

Do you think the `swapbondsites` is worth including? It is quite easy to make anyway, so we could always add it later.

emstoudenmire:
Yes I'm supportive of adding the `swapbondsites` function you've defined there. Let's think about if there's any need for `swapbondsites!` (an in-place version) but probably the view semantics of copying most of the MPS makes that not a big deal from an efficiency perspective. So we can always add it later just if it's more convenient.

I think there shouldn't be a confusion with the other function you discussed, `swapsiteinds`, since its name ends in "inds".

mtfishman:
Generally I would prefer out-of-place semantics for high level MPS functions like this. There isn't any efficiency gain to using in-place versions, since a shallow copy of an MPS is fast (I think it is just setting new pointers):
```julia
julia> using ITensors

julia> s = siteinds("S=1/2", 100);

julia> M = randomMPS(s, 50);

julia> @btime copy($M);
  80.958 ns (2 allocations: 928 bytes)

julia> A = 2*M[50];

julia> @btime $M[50] = $A;
  3.770 ns (0 allocations: 0 bytes)
```
I also think it starts to get confusing, since it becomes unclear if the ITensors themselves might be modified in-place (a "deep in-place operation"), or just that the ITensors are being replaced in the Vector storage. That could get particularly confusing for functions like `scale!(::MPS, ::Number)` or `normalize!(::MPS, ::Number)`.

PR #423, opened 2020-06-09
Axes definition for ITensors

mtfishman:
This defines `axes` for ITensors, which is basically `(1:dim(A, 1), 1:dim(A, 2), ...)`. Broadcasting calls `axes`, but was using a general fallback method which was slow and added some overhead to all ITensor broadcasting calls (about 1 microsecond).

This also bumps the version of NDTensors to v0.1.7, which introduces a lot of optimizations for the contraction logic to minimize overhead for small ITensor contractions. I'll add some benchmarks, but initial tests show that for even very small tensor contractions (for example `(30 x 30) * (30 x 30)`), the overhead is now pretty small, something around 20%. Before this, the overhead was at least 100%.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ITensors"
 uuid = "9136182c-28ba-11e9-034c-db9fb085ebd5"
 authors = ["mfishman <mfishman@caltech.edu>"]
-version = "0.1.6"
+version = "0.1.7"
 
 [deps]
 HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
@@ -17,7 +17,7 @@ TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 julia = "1.4"
 HDF5 = "0.13.1"
 KrylovKit = "0.4.2"
-NDTensors = "0.1.6"
+NDTensors = "0.1.7"
 StaticArrays = "0.12.3"
 TimerOutputs = "0.5.5"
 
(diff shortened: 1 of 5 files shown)


mtfishman:
Here are some initial benchmarks. Running the following:
```julia
using ITensors

function main(n::Int)
  i = Index(n)
  A = randomITensor(i',i)
  B = randomITensor(i'',i')
  C = randomITensor(i'',i)

  a = randn(n, n)
  b = randn(n, n)
  c = randn(n, n)

  @btime $C .= $B .* $A
  @btime mul!($c, $b, $a)
end
```
I get:
```julia
julia> main(20);
  1.039 μs (26 allocations: 1.61 KiB)
  571.359 ns (0 allocations: 0 bytes)

julia> main(40);
  2.787 μs (26 allocations: 1.61 KiB)
  2.321 μs (0 allocations: 0 bytes)

julia> main(60);
  3.780 μs (26 allocations: 1.61 KiB)
  3.098 μs (0 allocations: 0 bytes)

julia> main(80);
  5.948 μs (26 allocations: 1.61 KiB)
  5.409 μs (0 allocations: 0 bytes)

julia> main(100);
  9.936 μs (26 allocations: 1.61 KiB)
  9.295 μs (0 allocations: 0 bytes)
```
The first is an ITensors.jl in-place contraction for a simple matrix multiplication, called using broadcasting.  The second is an in-place matrix multiplication in native Julia of the same size, so the ratio is the ITensor overhead. The ratios are:
```julia
 1.8184713988928152
 1.2007755277897456
 1.2201420271142673
 1.0996487335921614
 1.0689618074233458
```
which is quite good. I'll add some timings using NDTensors v0.1.6 and ITensors v0.1.6 to get an idea for the speedup.

Issue #220, opened 2020-03-08
Restore QR in orthogonalize!


emstoudenmire:
Recently switched `orthogonalize!` (in mps/mpo.jl) to use SVD instead of QR so it could be used for QN conserving MPS and MPO. Need to switch this back once QN conserving QR is implemented.

orialb:
Is someone working on this? Otherwise I'll go on and implement this as the lack of QR in `orthogonalize!` makes my TEBD code 2X slower.

emstoudenmire:
Hi Ori, do you mean implement the lack of a QN-conserving `qr` function or just turning on QR again in `orthogonalize!` ? As you may know, currently the QR decomposition is not implemented for QN-conserving, block-sparse ITensors (QNITensors).

orialb:
Hi Miles, I meant that I could try to implement everything needed to get `orthogonalize!` working with QR for both dense and block-sparse ITensors. Unless of course this is something you want to work on yourselves or there is some reason to wait with this.

orialb:
p.s. If you prefer to wait with block-sparse QR until you have time to implement it yourselves, maybe QR could be reactivated for dense tensors in `orhotogonalize!` in the meantime? 
The code will be less elegant but it will be much faster (I guess this is also relevant for performing measurements on MPS and not just my TEBD code that probably no one uses currently).

emstoudenmire:
Hi Ori, that’s a good suggestion. Actually the situation is kind of a bit nicer than that I just realized. The `orthogonalize!` function calls a general-purpose routine Matt created called `factorize` which is intended as kind of a smart, automatic factorization method that selects the best method depending on the situation. Right now it only has SVD and density matrix decomposition (not sure what a math person would call this) backends. QR is planned to be one of the future backends and could be the default for cases when no truncation is requested. 

So if we just plug in a QR backend for the case of (1) no truncation and (2) non-QN-conserving ITensors for now, then `orthogonalize!` will start using it automatically when it is called without truncation parameters.

@mtfishman does all of the above sound correct to you?

mtfishman:
Yes, that sounds like a good plan. In the `factorize` function, the plan all along was that if you asked for left or right orthogonality, and didn't ask for any truncation, it would automatically use a qr decomposition. I only didn't put it in because we don't have a block-sparse qr decomposition yet and I was still working out the internal organization of the function, but it would be fine to put it in for the dense case now.

mtfishman:
[...]

PR #427, opened 2020-06-13
add QR option in factorize (for dense ITensors)

orialb:
The main goal of this PR is to re-enable the usage of QR in `orthogonalize!` for dense ITensors, as discussed in https://github.com/ITensor/ITensors.jl/issues/220. 

One thing I was not sure about is what to do with the `Spectrum` object in the case of QR. Currently I set it to `nothing` but maybe one would like to actually return an empty spectrum with zero truncation error.


mtfishman:
Thanks for the contribution Ori, I'll take a closer look in the next few days.

I think returning a Spectrum object with zero truncation is better (since this is supposed to be a black box function, where QR is just one special way to do a decomposition which happens to not truncate).

mtfishman:
Also one thing I was wondering about this functionality was whether `ortho = "right"` should do something like an LQ decomposition. But maybe `"LQ"` could just be a separate option (and perhaps the way we use ITensor indices, it already is something like an LQ decomposition). I think there could be situations where you specifically want either upper triangular or lower triangular matrices, like in more advanced methods for gauging MPOs that are sums of local terms (https://arxiv.org/abs/1909.06341).

emstoudenmire:
It’s not a totally obvious thing, but I’d also vote for an empty `Spectrum` object (making sure that it reports a 0.0 truncation error) over returning `nothing`. The argument I’d make is that it will work more smoothly in generic code that sometimes truncates and sometimes doesn’t.

orialb:
I also think it's better to return something. My main problem with returning a `Spectrum` is that it's not really a spectrum, just truncation information, so it might lead to some problems if one tries to compute entanglement-entropy with `entropy(spec)`. I guess we can make `entropy` throw an exception in that case. 
Another option would be to rethink the name of the `Spectrum` object or maybe adding a type parameter to distinguish the different cases.

>Also one thing I was wondering about this functionality was whether ortho = "right" should do something like an LQ decomposition.

How I implemented it now, if `ortho=="right"` it amounts to calling `qr(A,Rinds)`. I guess this is equivalent to doing QR decomp. of A_{lr}^T and then returning  R^T , Q^T  which is in fact an LQ decomposition ?

mtfishman:
> I also think it's better to return something. My main problem with returning a `Spectrum` is that it's not really a spectrum, just truncation information, so it might lead to some problems if one tries to compute entanglement-entropy with `entropy(spec)`. I guess we can make `entropy` throw an exception in that case.
> Another option would be to rethink the name of the `Spectrum` object or maybe adding a type parameter to distinguish the different cases.
> 

Good point about the `entropy` function (and also the same would apply for the `eigs` function [here](https://github.com/ITensor/NDTensors.jl/blob/90f7e0493606c48abc9468871a61bb0a6dbdaa8e/src/linearalgebra.jl#L49)). What we could do is extend the parametrization of `Spectrum` [here](https://github.com/ITensor/NDTensors.jl/blob/90f7e0493606c48abc9468871a61bb0a6dbdaa8e/src/linearalgebra.jl#L44) to `Union{<:AbstractVector, Nothing}`. Then, for cases like this where it is not natural to compute the eigenvalues, we can make the `eigs` field `nothing`. (Also a side point, it might be nice to rename `eigs` to `eigavals` to be in line with the Julia LinearAlgebra `eigvals` function).

> > Also one thing I was wondering about this functionality was whether ortho = "right" should do something like an LQ decomposition.
> 
> How I implemented it now, if `ortho=="right"` it amounts to calling `qr(A,Rinds)`. I guess this is equivalent to doing QR decomp. of A_{lr}^T and then returning R^T , Q^T which is in fact an LQ decomposition ?

Indeed, it looks like in your current implementation, `ortho = "left"` returns a QR decomposition (`A_lr = Q_lq R_qr`) and `ortho = "right"` returns an LQ decomposition (`A_lr = L_lq Q_qr`). Perhaps we want `ortho = "right"` to return an RQ decomposition. Then we can have a separate `which_decomp = "ql"` option (which we can add later) that either returns a QL or LQ decomposition depending on the `ortho` keyword. I guess I would expect if the keyword is `which_decomp = "qr"`, it would always return an upper triangular matrix for the non-orthogonal factor. @emstoudenmire, do you have an opinion on this choice?

Taking a step back from this, I would say that the only guarantee of the function `factorize` is that if you use `ortho = "left"` or `ortho = "right"`, the left or right factors are orthogonal bases. Then, `which_decomp` gives more specific hints about how it should do the decomposition. If you want more guarantees about the output, then you should use more specific decomposition functions, like `qr(...)`, and we can offer other options like `rq(...)`, `ql(...)`, and `lq(...)`. So I would consider this choice to be non-breaking (in the sense that people should be aware specific outputs can change, like in this PR if no truncation is requested), and we can decide later if one or the other is more useful.

mtfishman:
In the latest NDTensors and ITensors, `Spectrum` now allows the `eigs` field to be `nothing`, as you can see in these tests: https://github.com/ITensor/ITensors.jl/blob/master/test/decomp.jl#L86. So we can have the QR version of factorize return that type of Spectrum object.

I think also for now, we can just keep the `ortho = "right"` case as it is, and decide later about whether we want it to do an RQ or LQ decomposition.

orialb:
Thanks Matt I'll update the PR.

>I think also for now, we can just keep the ortho = "right" case as it is, and decide later about whether we want it to do an RQ or LQ decomposition.

I was just now looking at this, it seems that there is no `rq` wrapper currently implemented in `LinearAlgebra` (one can access the LAPACK functions directly though), so this gives me the feeling that it would be cleaner to use LQ decomposition at the moment.

mtfishman:
Yes, I also noticed there is no `rq` wrapper in `LinearAlgebra`. We may want that eventually, but not a high priority right now.

orialb:
I updated the PR to return an empty Spectrum. I also tried to update the docstring a bit, not sure if it is clear enough.

src/decomp.jl
```diff
@@ -394,14 +408,27 @@ function LinearAlgebra.factorize(A::ITensor,
   # Determines when to use eigen vs. svd (eigen is less precise,
   # so eigen should only be used if a larger cutoff is requested)
   automatic_cutoff = 1e-12
-  if which_decomp == "svd" || 
-     (isnothing(which_decomp) && cutoff ≤ automatic_cutoff)
+
+  if isnothing(which_decomp)
+    if cutoff==0.0 && !hasqns(A) && ortho != "none"
```
mtfishman, line 413: We should also add a check for `maxdim` here. Possibly we could check if `maxdim <= min(dim(Lis), dim(Ris))` to see if `maxdim` would lead to a truncation.

A subtle point is that with `cutoff == 0.0`, the spectrum might have zero values, which we would want to truncate. A better code pattern would be to have the `cutoff` default to `nothing`, and then check if it has been set using `isnothing(cutoff)`, and if it has then we don't call `factorize_qr`.

We could turn this into a function like `might_truncate(m::Int, n::Int; cutoff, maxdim, ...) -> Bool` which takes keyword arguments that are used for possibly truncating and return `true` if a truncation might occur (`m` and `n` would be the matrix dimensions). Whether or not a truncation might happen is a common concept that can arise throughout the library.

In C++ ITensor we also have a keyword argument, I think called `truncate`, which you can explicitly set to `true` or `false`. If it is false, it ignores all of the keyword arguments related to truncation and does not truncate, while if it is true then it would use `cutoff = 0.0` if `cutoff` wasn't set. The logic starts to get kind of complicated, I can do another PR adding a `might_truncate(...)` function and a `truncate` keyword argument if you can make the fixes for `maxdim` and `cutoff == 0.0` here.

orialb: I added a check for `maxdim` and changed the default value of `cutoff` to `nothing` in the latest commit.

mtfishman:
Looks good Ori, thanks for the contribution! I started working on the QN version of the QR decomposition, I'll try to finish that soon.

emstoudenmire:
Thanks Ori! Appreciate the high-quality code as usual

PR #434, opened 2020-06-23
Generalize to n-site DMRG for any integer n >= 2

sujay-kazi:
The current implementation of DMRG in ITensors.jl is restricted to 2-site DMRG, meaning that each optimization step handles two adjacent sites. This commit generalizes DMRG to n-site DMRG for any integer n >= 2. The sweeps are still conducted in the same way as before, i.e. progressing forward one site at a time and then backward one site at a time. For example, if one is carrying out n-site DMRG on an N-site chain, then the first optimization covers sites 1 through n, the second optimization covers sites 2 through n+1, and so on, until the (N-n+1)th optimization covers sites n-N+1 through N. Then you do all of those again in reverse order, and that completes one sweep.

### DIFF
--- a/src/mps/abstractmps.jl
+++ b/src/mps/abstractmps.jl
@@ -694,7 +694,7 @@ the tensors in the network. The name
 `totalqn` is an alias for `flux`.
 """
 function flux(M::AbstractMPS)::QN
-  hasqns(M) || error("MPS or MPO does not conserve QNs")
+  hasqns(M) || return nothing
   q = QN()
   for j=M.llim+1:M.rlim-1
     q += flux(M[j])
--- a/src/mps/abstractmps.jl
+++ b/src/mps/abstractmps.jl
@@ -693,7 +693,7 @@ the flux is the sum of fluxes of each of
 the tensors in the network. The name
 `totalqn` is an alias for `flux`.
 """
-function flux(M::AbstractMPS)::QN
+function flux(M::AbstractMPS)
   hasqns(M) || return nothing
   q = QN()
   for j=M.llim+1:M.rlim-1
--- a/src/mps/abstractmps.jl
+++ b/src/mps/abstractmps.jl
@@ -693,8 +693,8 @@ the flux is the sum of fluxes of each of
 the tensors in the network. The name
 `totalqn` is an alias for `flux`.
 """
-function flux(M::AbstractMPS)
-  hasqns(M) || return nothing
+function flux(M::AbstractMPS)::QN
+  hasqns(M) || error("MPS or MPO does not conserve QNs")
   q = QN()
   for j=M.llim+1:M.rlim-1
     q += flux(M[j])
--- a/src/mps/dmrg.jl
+++ b/src/mps/dmrg.jl
@@ -18,7 +18,8 @@ function dmrg(H::MPO,
               psi0::MPS,
               sweeps::Sweeps;
               kwargs...)
-  PH = ProjMPO(H)
+  ncenter::Int = get(kwargs, :ncenter, 2)
+  PH = ProjMPO(H, ncenter)
   return dmrg(PH,psi0,sweeps;kwargs...)
 end
 
@@ -47,6 +48,7 @@ function dmrg(Hs::Vector{MPO},
               psi0::MPS,
               sweeps::Sweeps;
               kwargs...)
+  ncenter::Int = get(kwargs, :ncenter, 2)
   PHS = ProjMPOSum(Hs)
   return dmrg(PHS,psi0,sweeps;kwargs...)
 end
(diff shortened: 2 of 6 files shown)


sujay-kazi:
Two things I should mention:

(1) I have made replacebond! (in mps.jl) throw an error if swapsites is true and ncenter is anything other than 2. I did this because, to my understanding, one should only use swapsites on 2-site DMRG. Please correct me if I am wrong about this; I will admit that I don't understand the purpose or functioning of swapsites all that much.

(2) You should add some tests for finding excited states as well. My code passed all of the tests here, but I know on my end that my code still throws errors on the dmrg function that accepts a vector with MPS's to project out. (I can still promise that my code correctly finds the ground state when running any of the dmrg functions that does not take a vector of MPS's. Feel free to merge this pull request now if you wish, and I can always come back and put an update once I fix this issue.)

sujay-kazi:
It took me a while to make this work, but I was able to update my generalized n-site DMRG code to find excited states as well (using the DMRG function that takes a list of MPSs to project out). However, it is worth pointing out that the current test folder in ITensors.jl still doesn't have any tests for the use of DMRG to find excited states. Should I try to add some? And is there anything else I should do before this pull request can be merged?

mtfishman:
Hi Sujay,

Thanks for all your work on this and sorry for the slow response.

Over all the PR looks good. My main concern is that `replacebond!` is not the best name for the more general n-site case since it is no long just a bond being replaced. As you point out, parts of `replacebond!` functionality like `swapsites` only make sense for replacing 2 sites of the MPS. `swapsites` is useful in the context of doing time evolution of an MPS using gates, where if the gates are nonlocal you need to swap the sites of the MPS to be next to each other. You can do this by explicitly applying a "swap gate", however it is more efficient to do that operation at the level of `replacebond!`.

I'm currently working on adding functionality for doing general gate evolution of an MPS/MPO with nonlocal n-site gates. In the process of doing that, I wrote a function `replacesites(::MPS, ::ITensors; firstsite = n1, lastsite = n2)` where you can specify a contiguous range of sites you want to replace with the keyword arguments `firstsite` and `lastsite`. In addition, it generalizes the `swapsites` functionality with a keyword argument to specify a permutation of the sites. The idea would be that `replacebond(psi, phi, b)` would call `replacesites(psi, phi; firstsite = b, lastsite = b+1)`. `replacesites(...)` would be used in the generalized n-site DMRG code you have written in this PR.

I hope to make a PR for the `replacesites` function soon, in which case you could then use that in this PR. Sorry we ended up with some overlapping work, but it sounded like you needed this functionality for a project anyway and I wasn't sure how long it would take for functionality like `replacesites` to get into ITensors officially (it is in a separate project right now and relies on a lot of small helper functions that need to get included as well).

And yes, that would be great to add tests for excited state DMRG, thanks!

-Matt

sujay-kazi:
Actually, I was wrong about there not being any tests for excited-state DMRG. There is a single test for excited-state DMRG in the test/dmrg.jl file, which looks like this:

```
@testset "Excited-state DMRG" begin
    N = 10
    weight = 15.0

    sites = siteinds("S=1",N)
    sites[1] = Index(2,"S=1/2,n=1,Site")
    sites[N] = Index(2,"S=1/2,n=$N,Site")

    ampo = AutoMPO()
    for j=1:N-1
      ampo += "Sz",j,"Sz",j+1
      ampo += 0.5,"S+",j,"S-",j+1
      ampo += 0.5,"S-",j,"S+",j+1
    end
    H = MPO(ampo,sites)

    psi0i = randomMPS(sites,10)

    sweeps = Sweeps(4)
    maxdim!(sweeps, 10,20,100,100)
    cutoff!(sweeps, 1E-11)
    noise!(sweeps,1E-10)

    energy0, psi0 = dmrg(H,psi0i, sweeps; outputlevel=0)
    @test energy0 < -11.5

    psi1i = randomMPS(sites,10)
    energy1,psi1 = dmrg(H,[psi0],psi1i,sweeps;outputlevel=0,weight=weight)

    @test energy1 > energy0
    @test energy1 < -11.1

    @test inner(psi1,psi0) < 1E-6
  end
```

The reason I didn't notice this earlier is that, if I call dmrg() without specifying an ncenter value, it defaults to 2, and of course that code works properly, so at that point it was no longer testing my n-site DMRG code.

Do you want me to add any test cases for general n-site DMRG? I can basically just have a few of the same tests run but with a higher specified value of ncenter, and if you want, I can also add one test case for something that works for 3-site DMRG that doesn't work for 2-site DMRG (it's related to the Schwinger model, which I have asked a number of questions about on the ITensor Support Q&A).

mtfishman:
Hi Sujay,

Copying the existing tests and adding `nsites = 3` should be sufficient, thanks. The Schwinger model would be a great example to add to the `examples/` directory.

Yes, the `replacesites!` function is available, you can see it here: https://github.com/ITensor/ITensors.jl/blob/master/src/mps/abstractmps.jl#L908 (for this function, `setindex!` is just an alias for `replacesites!`). Let me know if you have questions about how it is used. Currently it doesn't return a "Spectrum" object and doesn't normalize by default. We could just normalize from within DMRG and calculate the Spectrum in dmrg for now, and I can work on adding those features to `replacesites!`.

-Matt

PR #436, opened 2020-06-23
Add ITensors.compile() function

mtfishman:
This adds the function `ITensors.compile()`, which compiles ITensors using PackageCompiler. By default, it creates a system image `sys_itensors.so` in the directory `~/.julia/sysimages`. After compilation, it prints instructions on how to use the system image to run Julia with the ITensors system image. It also updates the Advanced Usage documentation with the simpler process of using this function.

I'm not sure it is great practice to add our own directories to `~/.julia`, but I don't see the harm here since it is unlikely that someone already has the file `sys_itensors.so` in that location. Also, the function has keyword arguments to choose both the path and file name.


mtfishman:
We should also add this to the installation instructions (perhaps once it is tested a bit in the wild). Also, the script with commands that get run and compiled, `src/packagecompile/precompile_itensors.jl`, should have more commands added to it to make the compilation more comprehensive.

emstoudenmire:
Yes one thing it could be good to do is to turn on QN conservation in the precompile_itensors.jl file. I’m not sure if the code there currently touches anything in the QNITensor system but maybe it does.

emstoudenmire:
Let’s merge this when you think it’s ready and I agree about testing it in the wild a little bit before recommending it as a standard step. But we may conclude pretty quickly that it’s a good thing for most users to do and that this approach works well.

mtfishman:
Yes, it doesn't cover QN conservation code right now. I think it is ready, we can try to advertise it as a more "experimental" feature.

It looks like we could use the tests to generate precompile statements: https://julialang.github.io/PackageCompiler.jl/dev/sysimages/#Using-a-package's-test-suite-to-generate-precompile-statements-1. My main concern would be that it would take a long time, we'll have to test it out.

PR #442, opened 2020-06-25
Compile QN DMRG

mtfishman:
With this added, on my laptop, compilation took about 3 minutes, so not too bad.

This also improves the docstring and output message of the `ITensors.compile()` command.

I tried using the tests to generate precompilation commands for PackageCompiler, but the compilation time increased to over 10 minutes, so that is not a good idea right now. I left the command to do that in `precompile_itensors.jl` and we can revisit it later.

### DIFF
--- a/src/packagecompile/compile.jl
+++ b/src/packagecompile/compile.jl
@@ -1,59 +1,61 @@
 
-"""
-    ITensors.compile(; path = "~/.julia/sysimages",
-                       filename = "sys_itensors.so")
+const default_compile_dir = joinpath(ENV["HOME"],
+                                     ".julia",
+                                     "sysimages")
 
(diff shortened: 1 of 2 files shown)


mtfishman:
@emstoudenmire, could you test this out locally to make sure it works correctly? Also take a look at the docstring and command output, I tried to make it so that it prints the local location of the compilation for the user (i.e. the full path to `~/.julia/sysimages`). I'm curious if it works correctly for you.

emstoudenmire:
@mtfishman I checked and it worked on my machine - so looks good to go.

emstoudenmire:
Also it took a little under four minutes on my machine

mtfishman:
Great, thanks for checking. I'll merge it.

PR #444, opened 2020-06-26
Add multi-site op support

mtfishman:
This adds support for adding ops that have more than one site.

The current implementation is a bit too simplistic, because it just searches through the tags of the first input Index for an `op` overload. What we could do is first make a TagSet of the tags that are common between all of the input indices (through something like a `commontags(::Index...) -> TagSet` function), and then search for `op` overloads of just the common tags. This would catch issues like trying to make a multi-site `op` of mixed site types, which for now is not officially support.

### DIFF
--- a/src/index.jl
+++ b/src/index.jl
@@ -125,6 +125,8 @@ Obtain the TagSet of an Index.
 """
 tags(i::Index) = i.tags
 
+commontags(is::Index...) = commontags(tags.(is)...)
+
 """
     plev(i::Index)
 
--- a/src/physics/autompo.jl
+++ b/src/physics/autompo.jl
@@ -59,7 +59,7 @@ mult(t1::OpTerm,t2::OpTerm) = isempty(t2) ? t1 : vcat(t1,t2)
 function isfermionic(t::OpTerm,sites)::Bool
   p = +1
   for op in t
-    if has_fermion_string(sites[site(op)],name(op))
+    if has_fermion_string(name(op), sites[site(op)])
       p *= -1
     end
   end
@@ -879,7 +879,8 @@ function sorteachterm!(ampo::AutoMPO, sites)
     # and inserting string "F" operators
     parity = +1
     for n=Nt:-1:1
-      fermionic = has_fermion_string(sites[site(t.ops[n])],name(t.ops[n]))
+      fermionic = has_fermion_string(name(t.ops[n]),
+                                     sites[site(t.ops[n])])
       if parity == -1
         # Put Jordan-Wigner string emanating
         # from fermionic operators to the right
--- a/src/physics/site_types/electron.jl
+++ b/src/physics/site_types/electron.jl
@@ -141,7 +141,7 @@ function op(::SiteType"Electron",
   return Op
 end
 
-has_fermion_string(::SiteType"Electron",::OpName"Cup") = true
-has_fermion_string(::SiteType"Electron",::OpName"Cdagup") = true
-has_fermion_string(::SiteType"Electron",::OpName"Cdn") = true
-has_fermion_string(::SiteType"Electron",::OpName"Cdagdn") = true
+has_fermion_string(::OpName"Cup", ::SiteType"Electron") = true
+has_fermion_string(::OpName"Cdagup", ::SiteType"Electron") = true
+has_fermion_string(::OpName"Cdn", ::SiteType"Electron") = true
+has_fermion_string(::OpName"Cdagdn", ::SiteType"Electron") = true
--- a/src/physics/site_types/fermion.jl
+++ b/src/physics/site_types/fermion.jl
@@ -54,5 +54,9 @@ function op(::SiteType"Fermion",
 end
 
 
-has_fermion_string(::SiteType"Fermion",::OpName"C") = true
-has_fermion_string(::SiteType"Fermion",::OpName"Cdag") = true
+has_fermion_string(::OpName"C",
+                   ::SiteType"Fermion") = true
+
+has_fermion_string(::OpName"Cdag",
+                   ::SiteType"Fermion") = true
+
--- a/src/physics/site_types/generic_sites.jl
+++ b/src/physics/site_types/generic_sites.jl
@@ -1,8 +1,8 @@
 
-function op(::SiteType,
-            ::OpName"Id",
+function op(::OpName"Id",
+            ::SiteType,
             s::Index)
-  Op = emptyITensor(s',dag(s))
+  Op = emptyITensor(s', dag(s))
   for n=1:dim(s)
     Op[n,n] = 1.0
   end
--- a/src/physics/site_types/spinhalf.jl
+++ b/src/physics/site_types/spinhalf.jl
@@ -1,107 +1,131 @@
 
-function space(::SiteType"S=1/2"; kwargs...)
-  conserve_qns = get(kwargs,:conserve_qns,false)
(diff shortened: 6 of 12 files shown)


mtfishman:
We would have to think if it is worth it to support multi-site operators that mix site types. With this design, one way to do that would be to define the indices with a common tag, for example: `Index(2, "Site, S=1/2, n=1, S=1/2_El"), Index(4, "Site, Electron, n=1, S=1/2_El")` and then define `op` overloads of `SiteType"S=1/2_El"`. The problem with that is if you wanted to mix more SiteTypes, you would run out of tags. Also, for certain multi-site operators, you might need to know which Index is which. That might require overloading something like `op(::SiteType"S=1/2", ::SiteType"Electron", OpName"MultiOp", ::Index, ::Index)`, which would seem to get complicated and could make the search for op overloads take a long time. Possibly that could be done within the `op(::SiteType"S=1/2_El", ...)` overload, so we wouldn't have to complicate the general `op` function.

mtfishman:
@emstoudenmire, do you have any opinion on how the multe-site `op` with mixed site types might work or at what level we should support it? It could effect the interface of this PR, depending on how sophisticated our support would be.

I would lean towards allowing people do it through an extra common site type (or some overload trick, like a union of two site types), which would not effect this implementation.

emstoudenmire:
Good questions about this system - it does get more complicated to think about the two-site case.

Personally I'd be for keeping the features on the simple side at first, as long as the design is such that we don't paint ourselves into too much of a corner in terms of expanded what's allowed later. So basically the design you are currently leaning toward. I agree we can expand later in two ways:
1. using tricks to fit other designs into this simpler system, such as overloads with Union type arguments
2. modifying the internals in a way that doesn't break existing cases, but quietly expands the allowed cases

Also the mixed-site case is kind of dicey for a number of reasons, and might be pushing the limits of what this system is intended for a little bit. The reason I say this is that usually there aren't many "textbook" operators I can think of that are (1) defined on mixed site types but (2) which aren't actually just sums of products of single-site operators. The cases I can think of where there are "true" two-site operators are mainly things like special gates for quantum circuits. Other cases people might define, such as defining things like the "S.S" op (a single term of the Heisenberg model) to me are a bit of a stretch of what this system is really for. But anyway we could still move toward that later as we allow more inputs.

One little thing we could support later is allowing inputs like "Sz⊗Sz" which would be defined to always give the same output as doing op("Sz",s1)*op("Sz",s2) but which might be useful in generic code that always deals with two-site ops. (Thinking again here of quantum circuits.)

Another thing we can add are some generic two-site ops such as the swap operator. (Only downside being it would not automatically work correctly for the fermionic case, so a little caution needed there, or we can make sure to always do a special overload for fermionic site types.)

mtfishman:
I agree. I guess technically all multi-site operators can be defined from sums of products of single site operators (using a complete operator basis), so this is mostly a convenience feature and allows for optimizations when creating more complicate operators. 

I think the current system can handle mixed site types naturally. For example, say I define some sort of swap operator of a spin half and electron site. I could define it as:
```julia
function ITensors.op(::Union{SiteType"S=1/2", SiteType"Electron"}, OpName"Swap", s1::Index, s2::Index)
  [...]
end
```
Then, if I define the following sites:
```julia
s1 = Index(2, "S=1/2")
s2 = Index(4, "Electron")
op("Swap", s1, s2)
op("Swap", s2, s1)
```
then it will call that version of `op` (which works because right now it just searches the tags of the first input Index for a proper `SiteType`). The problem with that strategy is that it doesn't catch cases where Index `s1` has the tag `"S=1/2"` or `"Electron"`, but the second Index doesn't have either of those. That could be handled internally in the above `op` overload (for example by checking the tags of Index `s2`). So I think the current design is both simple and can handle these more complicated cases, if someone wants to put in the effort.

That's a nice idea to have inputs like `"Sz⊗Sz"` interpreted as the equivalent of `op("Sz",s1)*op("Sz",s2)`. We could use inputs like that in AutoMPO, since then people could input `"Sx⊗Sx"` and we could see that is has a flux of zero instead of having to use the `"S+"`/`"S-"` trick.

mtfishman:
I think the main question for the current PR is if the strategy of only looking through the tags of the first input Index for a `SiteType` is a good one. I think based on the example I give above, it seems pretty good and allows for a lot of flexibility. Let me know if you foresee any pitfalls of that design, but otherwise this is ready to merge.

mtfishman:
The latest version allows more fine-grained overloading of multi-site operators, for example:

```julia
    using ITensors
    using Test

    function ITensors.op(::SiteType"_Custom1",
                         ::OpName"α",
                         s1::Index,
                         s2::Index)
      Op = emptyITensor(s1', s2',
                        dag(s1), dag(s2))
      Op[s1'=>1, s2'=>2, s1=>1, s2=>2] = +3/2
      Op[s1'=>2, s2'=>1, s1=>2, s2=>2] = +1/2
      Op[s1'=>3, s2'=>3, s1=>3, s2=>4] = -1/2
      Op[s1'=>4, s2'=>1, s1=>4, s2=>2] = -3/2
      return Op
    end

    function ITensors.op(::SiteType"_Custom1",
                         ::SiteType"_Custom2",
                         ::OpName"β",
                         s1::Index,
                         s2::Index)
      Op = emptyITensor(s1', s2',
                        dag(s1), dag(s2))
      Op[s1'=>1, s2'=>2, s1=>1, s2=>2] = +5/2
      Op[s1'=>2, s2'=>1, s1=>2, s2=>2] = +3/2
      Op[s1'=>3, s2'=>3, s1=>3, s2=>4] = -3/2
      Op[s1'=>4, s2'=>1, s1=>4, s2=>2] = -5/2
      return Op
    end

    s = Index(4, "_Custom1, __x")
    t = Index(4, "_Custom1, __x")
    α = op("α", s, t)
    @test α[s'=>1, t'=>2, s=>1, t=>2] ≈ +3/2
    @test α[s'=>2, t'=>1, s=>2, t=>2] ≈ +1/2
    @test α[s'=>3, t'=>3, s=>3, t=>4] ≈ -1/2
    @test α[s'=>4, t'=>1, s=>4, t=>2] ≈ -3/2

    s2 = Index(4, "_Custom2, __x")

    β = op("β", s, s2)
    @test β[s'=>1, s2'=>2, s=>1, s2=>2] ≈ +5/2
    @test β[s'=>2, s2'=>1, s=>2, s2=>2] ≈ +3/2
    @test β[s'=>3, s2'=>3, s=>3, s2=>4] ≈ -3/2
    @test β[s'=>4, s2'=>1, s=>4, s2=>2] ≈ -5/2

    # This fails since the order matters in the definition of `op`
    @test_throws ErrorException op("β", s2, s)
```

The performance of the single-site operator case should be unaffected, and the multi-site case with a uniform site type only adds on the step of first getting the tags that are common between all of the sites. Then, the case of multiple different site types is checked (so that has to go through all of the other cases, and is a bit slower, but that is not such a concern since I don't think it will be very common anyway).

mtfishman:
Note that if you also defined the following:
```julia
    function ITensors.op(::SiteType"_Custom1",
                         ::SiteType"_Custom1",
                         ::OpName"α",
                         s1::Index,
                         s2::Index)
      [...]
    end
```
the version above would take precedence, which I think is reasonable behavior.

mtfishman:
In the latest commit, I changed the argument ordering so that `OpName` comes before the `SiteType` inputs (so `op(::OpName, ::SiteType, ::Index)`, for example).

PR #446, opened 2020-06-29
Customization of siteinds and siteind

emstoudenmire:
This PR expands the customizability of the `siteinds` function in a  
backwards compatible way, by reimplementing it in terms of functions named
`siteind` which just make one site index, which itself calls by default another
function called `space` that just returns the space of a site index.

One advantage of having the function `siteind` is it makes it easy to make
site arrays which are mixtures of multiple site types.

This design gives the user multiple points of customization of how site indices 
get created. From lowest to highest level:
- create an overload of `space(::SiteType"Tag"; kwargs...)`
- create an overload of `siteind(::SiteType"Tag"; kwargs...)`
- create an overload of `siteind(::SiteType"Tag", n::Int; kwargs...)`
- create an overload of `siteinds(::SiteType"Tag", N::Int; kwargs...)`

This system works well, and the only drawbacks I can see are:
1. It might be a tad slower than before, especially in the QN case since the QN spaces get constructed N times rather than just once. But I don't think `siteinds` is a bottleneck in any case we know of.
2. If a user overloads `siteind(::SiteType"Tag",n)` but not `siteind(::SiteType"Tag")` or `space(::SiteType"Tag")`, then calling `siteind("Tag")` will result in an error. The solution is just to define one of those other functions but it could be a bit confusing to someone encountering this situation.

Commits included:
- Add functions for retrieving tag and name back from SiteType and OpName instances
- Initial draft of new siteinds and siteind system
- Unit tests for custom siteind and siteinds
- Compress code a bit and remove int argument from lowest-level siteind function
- Fix error message
- Upgrade spinhalf.jl and spinone.jl to use space
- Slight rewrite of test code
- Update all site types to use space function
- Add nblocks and qn to exports
- Test siteind function for "S=1/2" tag

### DIFF
--- a/src/physics/sitetype.jl
+++ b/src/physics/sitetype.jl
@@ -239,20 +239,19 @@ end
 #
 #---------------------------------------
 
-function siteind(st::SiteType,n;kwargs...) 
-  t = String(tag(st))
+function siteind(st::SiteType; kwargs...) 
   if applicable(space,st)
     sp = space(st;kwargs...)
-    return Index(sp,"Site,$t,n=$n")
+    return Index(sp,"Site,$(tag(st))")
   end
   throw(ArgumentError("Overload of \"siteind\" or \"space\" functions not found for Index tag: $t"))
 end
 
-function siteind(tag::String,
-                 n::Integer; kwargs...)
-  st = SiteType(tag)
-  return siteind(st,n;kwargs...)
-end
+siteind(st::SiteType, n; kwargs...) = addtags(siteind(st;kwargs...),"n=$n")
+
+siteind(tag::String; kwargs...) = siteind(SiteType(tag);kwargs...)
+
+siteind(tag::String,n; kwargs...) = siteind(SiteType(tag),n;kwargs...)
 
 # Special case of `siteind` where integer (dim) provided
 # instead of a tag string
--- a/src/physics/sitetype.jl
+++ b/src/physics/sitetype.jl
@@ -27,6 +27,7 @@ end
 
 SiteType(s::AbstractString) = SiteType{Tag(s)}()
 SiteType(t::Tag) = SiteType{t}()
+tag(::SiteType{T}) where {T} = T
 
 macro SiteType_str(s)
   SiteType{Tag(s)}
@@ -67,6 +68,7 @@ end
 
 OpName(s::AbstractString) = OpName{SmallString(s)}()
 OpName(s::SmallString) = OpName{s}()
+name(::OpName{N}) where {N} = N
 
 macro OpName_str(s)
   OpName{SmallString(s)}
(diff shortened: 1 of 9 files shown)


src/physics/sitetype.jl
```diff
@@ -27,6 +27,7 @@ end
 
 SiteType(s::AbstractString) = SiteType{Tag(s)}()
 SiteType(t::Tag) = SiteType{t}()
+tag(::SiteType{T}) where {T} = T
```
mtfishman, line 30: Glad you added this, I was thinking it would be useful.

src/physics/sitetype.jl
```diff
@@ -220,20 +233,64 @@ function state(sset::Vector{<:Index},
...
-                  N::Integer; kwargs...)
-  return [Index(d,"Site,n=$n") for n=1:N]
+#---------------------------------------
+#
+# siteind system
+#
+#---------------------------------------
+
+function siteind(st::SiteType; kwargs...) 
+  if applicable(space,st)
```
mtfishman, line 243: I think in general people should overload `space`, but for the sake of making `siteind` easier to overload, is there a way we could avoid this? For example, maybe we could have a generic `space(::SiteType) = error("...")` function, which Julia will fall back to if no overload is made for a particular `SiteType`.

mtfishman:
> This system works well, and the only drawbacks I can see are:
> 
>     1. It might be a tad slower than before, especially in the QN case since the QN spaces get constructed N times rather than just once. But I don't think `siteinds` is a bottleneck in any case we know of.

I agree, I'm not so worried about that. Unlike `op`, I can't imagine situations where people are making site indices on the fly in a calculation, or that it would ever be a bottleneck even if they were.

> 
>     2. If a user overloads `siteind(::SiteType"Tag",n)` but not `siteind(::SiteType"Tag")` or `space(::SiteType"Tag")`, then calling `siteind("Tag")` will result in an error. The solution is just to define one of those other functions but it could be a bit confusing to someone encountering this situation.

That doesn't seem so concerning. I think we should emphasize overloading `space(::SiteType"Tag")` anyway, I don't see a reason why someone would want to overload `siteind(::SiteType"Tag", n)` in particular (unless they wanted to change the tag `"n=$n"` to something else).

src/physics/sitetype.jl
```diff
@@ -220,20 +233,67 @@ function state(sset::Vector{<:Index},
...
-  return [Index(d,"Site,n=$n") for n=1:N]
+#---------------------------------------
+#
+# siteind system
+#
+#---------------------------------------
+
+space(st::SiteType; kwargs...) = throw(MethodError("Overload of \"space\",\"siteind\", or \"siteinds\" functions not found for Index tag: $(tag(st))"))
+
+function siteind(st::SiteType; add_tags="", kwargs...)
```
mtfishman, line 244: Could we just call it `addtags`, since that is the name of the function?

emstoudenmire: Changed in latest commits

PR #449, opened 2020-07-01
Update state system to use StateName

emstoudenmire:
This PR updates the `state(::SiteType,...)` system to use a new type `StateName"Name"`
to allow the system to be user-extensible and offer automatic error handling, etc. 
It falls back to the older system for backwards compatibility.

This PR also updates the `has_fermion_string` system used by AutoMPO to use `OpName`
for similar reasons.

- Redesign of `state` system to use StateName type
- Update physics site types to use StateName overloads
- Update has_fermion_string system to use OpName

### DIFF
--- a/src/physics/site_types/electron.jl
+++ b/src/physics/site_types/electron.jl
@@ -35,10 +35,10 @@ state(::SiteType"Electron",::StateName"Emp")  = 1
 state(::SiteType"Electron",::StateName"Up")   = 2
 state(::SiteType"Electron",::StateName"Dn")   = 3
 state(::SiteType"Electron",::StateName"UpDn") = 4
-state(::SiteType"Electron",::StateName"0")    = 1
-state(::SiteType"Electron",::StateName"↑")    = 2
-state(::SiteType"Electron",::StateName"↓")    = 3
-state(::SiteType"Electron",::StateName"↑↓")   = 4
+state(st::SiteType"Electron",::StateName"0")    = state(st,StateName("Emp"))
+state(st::SiteType"Electron",::StateName"↑")    = state(st,StateName("Up"))
+state(st::SiteType"Electron",::StateName"↓")    = state(st,StateName("Dn"))
+state(st::SiteType"Electron",::StateName"↑↓")   = state(st,StateName("UpDn"))
 
 function op(::SiteType"Electron",
             s::Index,
--- a/src/physics/site_types/electron.jl
+++ b/src/physics/site_types/electron.jl
@@ -31,20 +31,14 @@ function space(::SiteType"Electron"; kwargs...)
   return 4
 end
 
-function state(::SiteType"Electron",
-               st::AbstractString)
-  if st == "Emp" || st == "0"
-    return 1
-  elseif st == "Up" || st == "↑"
-    return 2
-  elseif st == "Dn" || st == "↓"
-    return 3
-  elseif st == "UpDn" || st == "↑↓"
-    return 4
-  end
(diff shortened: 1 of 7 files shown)


src/physics/site_types/electron.jl
```diff
@@ -31,20 +31,14 @@ function space(::SiteType"Electron"; kwargs...)
...
-    return 4
-  end
-  throw(ArgumentError("State string \"$st\" not recognized for Electron site"))
-  return 0
-end
+state(::SiteType"Electron",::StateName"Emp")  = 1
+state(::SiteType"Electron",::StateName"Up")   = 2
+state(::SiteType"Electron",::StateName"Dn")   = 3
+state(::SiteType"Electron",::StateName"UpDn") = 4
+state(::SiteType"Electron",::StateName"0")    = 1
```
mtfishman, line 38: Should we define these aliases as `state(::SiteType"Electron",::StateName"0") = state(::SiteType"Electron",::StateName"Emp")`? This would mostly be to show people how they might make their own aliases (since then aliases don't rely on lower level details).

src/physics/site_types/spinhalf.jl
```diff
@@ -104,6 +94,12 @@ end
 op!(Op::ITensor,t::SiteType"S=1/2",
     ::OpName"S²",s::Index) = op!(Op,t,OpName("S2"),s)
 
+
+# Support the tag "SpinHalf" as equivalent to "S=1/2"
+
+space(::SiteType"SpinHalf"; kwargs...) = space(SiteType("S=1/2");kwargs...)
+state(::SiteType"SpinHalf",n::StateName) = state(SiteType("S=1/2"),n)
```
mtfishman, line 101: This reminded me that we could also support `"S=½"`.

src/physics/sitetype.jl
```diff
@@ -204,35 +204,48 @@ op(s::Vector{<:Index},
...
+    res = state(st,sname)
+    !isnothing(res) && return s(res)
   end
-  if nfound == 0
-    throw(ArgumentError("Overload of \"state\" function not found for Index tags $(tags(s))"))
-  elseif nfound > 1
-    throw(ArgumentError("Multiple tags from $(tags(s)) overload the function \"state\""))
+
+  # Try calling state(::SiteType"Tag","Name")
+  for st in stypes
```
mtfishman, line 237: Is this version for backwards compatibility, in case someone had already defined their own `state` definitions?

mtfishman:
Looks good! Nice to get rid of all of those if-statements, the code looks much cleaner now.

src/physics/site_types/electron.jl
emstoudenmire: That's a good suggestion. I just did it this way because the code was simpler but I like the pedagogical idea you suggest.

src/physics/site_types/spinhalf.jl
emstoudenmire: I'd be ok with adding that. I'd have to learn how to type that though!

src/physics/sitetype.jl
emstoudenmire: Yes, exactly. Though I doubt anyone has. Of course it will only run in either that case or when an error is about to occur anyway, so no loss of speed except when using the older function signature.

emstoudenmire:
Thanks for the helpful review

src/physics/site_types/spinhalf.jl
mtfishman:
```julia
help?> ½
"½" can be typed by \1/2<tab>
```

PR #452, opened 2020-07-08
Fix some issues with precompile_itensors.jl code

emstoudenmire:
The ITensors.compile() command was failing due to some incorrect code usage in the QN part of the precompile_itensors.jl file. This PR updates that code to restore ITensors.compile() to working.

### DIFF
--- a/src/packagecompile/precompile_itensors.jl
+++ b/src/packagecompile/precompile_itensors.jl
@@ -35,8 +35,7 @@ for j=1:N-1
     ampo .+= (0.5,"S-",j,"S+",j+1)
 end
 H = MPO(ampo,sites)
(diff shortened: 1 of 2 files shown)


mtfishman:
Looks good, feel free to merge when you're ready.

It would be nice to have automated tests for this functionality. I was thinking we could just add `ITensors.compile()` in the tests somewhere, but that would add a lot of time to the tests. Of course that wouldn't catch a mistake like this anyway.

emstoudenmire:
Your comment got me thinking, and I remembered we have the examples.jl unit test file. So I added an include of the  precompile_itensors.jl code to that and now it will get automatically tested!

PR #453, opened 2020-07-09
Update site definitions

emstoudenmire:
- Update keyword args style of Electron space function and test
- Update keyword args of Fermion space function and test
- Update Fermion op function overloads to use op!
- Update Electron op overloads to use op!
- Update generic "Id" op overload to use op!
- Update keyword argument style for tJ space function
- Update tJ op overloads to use op! and fix Sx defn


mtfishman:
Great, thanks for tackling this!

PR #457, opened 2020-07-10
ITensor iteration and other miscellaneous ITensor and MPS/MPO functionality

mtfishman:
This adds quite a number of small functions I have been accumulating in separate projects for MPS/MPO gate evolution and other MPS operations. Here is a list of the new functions:

- Add `eachindex(T::ITensor)` to return an iterator over each cartesian index of an ITensor (i.e. for an `d x d` ITensor, either `1:d^2` or `(1,1), (1,2), ..., (d, d)`). For sparse ITensors, this includes structurally zero and nonzero entries.
- Add `iterate(A::ITensor, args...)`, which allows using `for a in A @show a end` to print all elements (zero and nonzero, for sparse tensors).
- Add `setindex!(T::ITensor, x::Number, I::CartesianIndex)` to allow indexing with a `CartesianIndex`, which is naturally returned by functions like `eachindex`.
- Add `hasplev(pl::Int)` that returns a function `x -> hasplev(x, pl)` (useful in functions like `map`).
- Add `hasind[s](i::Index)` that returns a function `x -> hasind[s](x, i)` (useful in functions like `map`).
- Add `hascommoninds(A, B; kwargs...)` which returns true if `A` and `B` have common indices.
- Add `findfirstsiteind(M::MPS/MPO, s::Index)` that returns which site of the MPS/MPO has the site index `s`.
- Add `findfirstsiteinds(M::MPS/MPO, is)` that returns which site of the MPS/MPO has the site indices `is`.
- Add `linkinds(::MPS/MPO)` that returns a vector of the link indices.
- Add `linkdim(::MPS/MPO, ::Int)` that returns the dimension of the specified link, and nothing if there is no link found.
- Add `linkdims(::MPS/MPO)` that returns a vector of the link dimensions.
- Fix a bug in `+(::MPST, ::MPST)` that the inputs were getting modified (the inputs were getting orthogonalized and the prime levels were being modified).
- Add `productMPS(sites, state::Union{String, Int})` to create an MPS with the same `state` at each site (for example, `productMPS(sites, "Up")` makes a state with all Up spins).

There are more general-interest functions I plan to add, but they are more involved.


mtfishman:
@emstoudenmire, I don't think anything here is too controversial, let me know if anything on the list of changes looks objectionable.

I would say the most subtle part is ITensor iteration, and how the functions `iterate` and `eachindex` should work for sparse ITensors. Here I take the approach that Julia's SparseArrays uses that they iterate through all cartesian values (structural nonzeros and zeros), and then we can provide specialized iteration tools in the future for iterating through only nonzero/stored values.

emstoudenmire:
These all look good to me, so I'm happy if you merge it. Nice to see how minimal the extra code is that's needed to add these. 

It seems like the only tricky thing here is what you said about iterating over entries which are known to always be zero for sparse tensor storage types. I think there we have reasons to want each type of behavior, so it makes sense that if these functions would show every element for a dense tensor then they would show the zeros of a sparse tensor too. 

As you said, we can have other functions in the future which behave differently, or we could add various keyword arguments to these functions. The keyword args I have in mind could include a boolean one which toggles the behavior about including zeros, or a threshold Float64 argument which skips over values whose magnitude is below that threshold. (The threshold one could be useful for printing even for dense tensors.)

mtfishman:
Thanks for taking a look.

For just iterating over the nonzero values, one can already iterate over `ITensors.data(A)`. Julia's SparseArrays offers this with the function `nonzeros` (https://docs.julialang.org/en/v1/stdlib/SparseArrays/index.html#SparseArrays.nonzeros) so we could consider that as an exported name. Having a threshold is an interesting idea.

I think the more subtle one is the nonzero version of `eachindex`, but I think we can have a devoted function/iterator type for that.

I'll go ahead and merge and tag a new version.

Issue #450, opened 2020-07-01
Helper functions for warn tensor order


mtfishman:
It might be nice to have some helper functions for the "WarnTensorOrder" global parameter, to make it easier to turn on and off temporarily (for example if you know that a certain operation may require more indices, but want it on generally). For example:
```julia
C = A * B  # I want the standard check

set_warn_itensor_order!(18)  # ITensors.GLOBAL_PARAMS["WarnTensorOrder"] = 18
E = C * D  # I know this will go over the default order, and I'm ok with that
reset_warn_itensor_order!()  # Sets back to default
```
We could also have a function `disable_warn_itensor_order!()` which just sets it to something very large.

If we want to be very fancy, we could turn some of these into macros which sets or disables the warning in a local block of code, like:
```julia
@set_warn_itensor_order 18 begin
  E = C*D
end
# GLOBAL_PARAMS["WarnTensorOrder"] is set back to the default
```
or
```julia
@disable_warn_itensor_order begin
  E = C*D
end
# GLOBAL_PARAMS["WarnTensorOrder"] is set back to the default
```

This came up when I was comparing against ED results for open state evolution, and some temporary tensors were of order 17 or 18. Many warnings showed up, but I only wanted to turn off the checks locally in the ED comparison, but not in the rest of the code.

Also, we might consider renaming it to `"WarnITensorOrder"` (to make it clear it is used for ITensors and not the Tensors of NDTensors).

mtfishman:
Closed by #458.

PR #458, opened 2020-07-10
Warn ITensor order helper functions

mtfishman:
Implements #450 (everything except for the macros, which I don't think are necessary right now).

Basically the syntax is something like:
```julia
# Do some contractions with standard checks
C = A * B
D = B * C

# Disable temporarily
disable_warn_itensor_order!()
E = D*F

# Reset to the default value
reset_warn_itensor_order!()
```
You can also set it to something different with `set_warn_itensor_order!(16)` and get the current value with `get_warn_itensor_order()`.

Implementation-wise, I also changed over to storing the actual global variable into a const `Ref`, i.e. I define something like:
```julia
const warn_itensor_order = Ref{Union{Int, Nothing}}(14)
```
where it is set to nothing when it is disabled. Getting the value is the notation `warn_itensor_order[]`, and setting it is the notation `warn_itensor_order[] = 4`. I like this approach better than using a dictionary (it is a bit simpler), and I think functions should be the "official" interface for controlling this anyway so the implementation shouldn't matter.


mtfishman:
@emstoudenmire, let me know if the interface looks good to you, otherwise it is ready to merge.

emstoudenmire:
This looks good to go. I think the lengthy interface here is ok since it communicates that this is an operation more aimed at configuring the internals of the library. It would definitely be good later to add a macro with a shorter name for disabling or raising the warning order for a block of code.

mtfishman:
Yes, it is definitely a bit clunky right now, a macro would be nice but this does the trick for now (I've been using it in tests that compare to exact diagonalization, where it has been very helpful to turn it on and off locally).

Issue #451, opened 2020-07-07
Use Julia's replace syntax for replaceinds


mtfishman:
Julia's `replace` function has syntax like the following:
```julia
julia> replace([1, 2, 3, 4], 4=>3, 3=>4)
4-element Array{Int64,1}:
 1
 2
 4
 3
```
We could consider the same kind of syntax for functions like `replaceinds`, for example:
```julia
julia> i = Index(2, "i")
(dim=2|id=344|"i")

julia> j = Index(2, "j")
(dim=2|id=664|"j")

julia> k = Index(2, "k")
(dim=2|id=563|"k")

julia> A = randomITensor(i, j)
ITensor ord=2 (dim=2|id=344|"i") (dim=2|id=664|"j")
NDTensors.Dense{Float64,Array{Float64,1}}

julia> replaceinds(A, (j,), (k,))  # Current syntax
ITensor ord=2 (dim=2|id=344|"i") (dim=2|id=563|"k")
NDTensors.Dense{Float64,Array{Float64,1}}

julia> replaceinds(A, j => k)  # New syntax
```
We could even just use the function name `replace` as an alias for `replaceinds`, when the input is an IndexSet.

We could also consider something similar for `replacetags`, such as `replacetags(i, "x" => "y")` to replace the tag `"x"` with `"y"`.

mtfishman:
Closed by  #459.

PR #459, opened 2020-07-13
Add pair syntax to mapprime, replacetags, and replaceinds

mtfishman:
This implements #451. It adds `Pair` syntax to `mapprime`, `replacetags`, and `replaceinds`. For example:
```julia
i = Index(2, "i")
j = Index(2, "j")
A = randomITensor(i, j')
mapprime(A, 0 => 1)
replacetags(A, "i" => "k")
replaceinds(A, i => i')
```
What is cool about this syntax (and adding support for multiple inputs) is that it adds an alternative syntax for `swapprime`, `swaptags`, and `swapinds`, i.e.:
```julia
mapprime(A, 0 => 1, 1 => 0) == swapprime(A, 0, 1)
replacetags(A, "i" => "j", "j" => "i") == swaptags(A, "i", "j")
replaceinds(A, i => j', j' => i) == swapinds(A, i, j')
```

### DIFF
--- a/src/index.jl
+++ b/src/index.jl
@@ -306,7 +306,8 @@ specified tags added to the existing ones.
 The `ts` argument can be a comma-separated 
 string of tags or a TagSet.
 """
-addtags(i::Index, ts) = settags(i, addtags(tags(i), ts))
+addtags(i::Index, ts) =
+  settags(i, addtags(tags(i), ts))
 
 """
     removetags(i::Index, ts)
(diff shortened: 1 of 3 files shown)


emstoudenmire:
Good use of Pair syntax, I agree. It does result in very readable code, though I do think we should remind people that `a=>b` makes a Pair in Julia in the documentation whenever we get a chance, since it's a much more language-specific thing compared to much of the rest of the ITensor interface. (E.g. we otherwise avoid a lot of Julia specific things like `:symbols` and such.)

mtfishman:
Good point about explaining the syntax in the docs. This is ready to go, so I'll merge.

PR #462, opened 2020-07-14
Extend  to support multiple IndexVals

emstoudenmire:


emstoudenmire:
@mtfishman Please let me know if you see any issues, like if you know a better way to extend this function or if the approach here might not work in some case I didn't think of. Otherwise I think it's a useful behavior to support.

mtfishman:
Looks good, definitely useful to have. Mostly I'm just hoping that eventually it is not necessary, and we can replace it with a true sparse tensor type and slicing!

PR #463, opened 2020-07-14
Add setindex!(::MPS, _, ::Colon)

mtfishman:
This adds functionality like the following:
```julia
N = 4
s = siteinds("S=½", N)
ψ = randomMPS(s)
ϕ = productMPS(s, "↑")
orthogonalize!(ϕ, 1) 
ψ[:] = ϕ
@test ITensors.orthocenter(ψ) == 1
```
so the MPS gets the orthogonality of the input MPS. One can also set an MPS from a Vector of ITensors (in which case the orthogonality can't be determined, and it is set to the boundaries).

### DIFF
--- a/src/mps/abstractmps.jl
+++ b/src/mps/abstractmps.jl
@@ -61,19 +61,15 @@ function Base.setindex!(M::AbstractMPS,
 end
 
 function Base.setindex!(M::MPST, v::MPST,
-                        ::Colon;
-                        set_limits::Bool = true) where {MPST <: AbstractMPS}
-  if set_limits
-    setleftlim!(M, -1)
-    setrightlim!(M, length(M)+1)
-  end
+                        ::Colon) where {MPST <: AbstractMPS}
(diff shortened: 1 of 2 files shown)


mtfishman:
As you might guess, I plan to add something like `ψ[4:7] = ϕ` where `length(ϕ) == 4` next, to set a subset of MPS tensors. This is useful for things like n-site DMRG and n-site gate evolution. In that case, we can determine the orthogonality of the modified MPS from the orthogonality of the original MPS and the input MPS.

emstoudenmire:
Looks like it's leading up to some nice features. I was wondering for this case why it's not enough to just do `psi = copy(phi)` but I see now that it's more toward taking a subset of the tensors.

mtfishman:
Also it can be helpful to have in-place semantics (but yeah generally I prefer out-of-place semantics for high level things like MPS, since the way we are doing everything with shallow copies there is rarely a performance benefit for in-place operations).

PR #464, opened 2020-07-15
Add QN ITensor from Array constructor

mtfishman:
This adds a constructor for QN ITensors from an Array, where the block structure and flux is determined from the nonzero values of the Array. It also provides a keyword argument `tol` to choose a tolerance for when to drop values. For example:
```julia
julia> i = Index([QN(0)=>1, QN(1)=>2], "i");

julia> A = [1e-9 0.0 0.0;
            0.0 2.0 3.0;
            0.0 1e-10 4.0];

julia> @show ITensor(A, i', dag(i); tol = 1e-8);
ITensor(A, i', dag(i); tol = 1.0e-8) = ITensor ord=2
Dim 1: (dim=3|id=468|"i")' <Out>
 1: QN(0) => 1
 2: QN(1) => 2
Dim 2: (dim=3|id=468|"i") <In>
 1: QN(0) => 1
 2: QN(1) => 2
NDTensors.BlockSparse{Float64,Array{Float64,1},2}
 3×3
Block: (2, 2)
 [2:3, 2:3]
 2.0  3.0
 0.0  4.0
```

### DIFF
--- a/docs/src/ITensorType.md
+++ b/docs/src/ITensorType.md
@@ -19,6 +19,7 @@ setelt(::IndexVal)
 
 ```@docs
 ITensor(::Type{<:Number}, ::QN, ::ITensors.Indices)
+ITensor(A::Array, inds::ITensors.QNIndexSet)
 ```
 
 ## Empty Constructors
--- a/src/qn/qnitensor.jl
+++ b/src/qn/qnitensor.jl
@@ -106,6 +106,22 @@ ITensor(A::Array,
         tol = 0) =
   _ITensor(A, IndexSet(inds...); tol = tol)
 
+itensor(A::Array,
+        inds::QNIndexSet;
+        tol = 0) =
+  ITensor(A, inds; tol = tol)
+
+itensor(A::Array,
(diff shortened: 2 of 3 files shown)


mtfishman:
@emstoudenmire, could you look over the interface? I think it is pretty straightforward.

I would like to eventually have more efficient ways of doing this, for example once a block is known to exist, set the elements of the entire block at once as a big chunk. But this is supposed to be a convenience function for helping make more generic code between Dense and BlockSparse to make small ITensors, so this is fine for now.

emstoudenmire:
Looks good! I guess the checking that all blocks have the same flux gets done by the `setindex!` function calls of the form `T[vs]` ?

emstoudenmire:
Agreed there are more efficient ways to read in the values but this should be good as a starting point.

mtfishman:
Yes, bad fluxes are checked in `setindex!`. It makes me think that we could add a keyword argument to `setindex!` to avoid checking the flux, for cases when you know the flux is fine.

PR #467, opened 2020-07-16
Add parity conservation to S=1/2 sitetype

mtfishman:
* Add option for parity convervation to S=1/2 site type.

* Add "ProjUp" and "ProjDn" operator definitions to S=1/2 site type.


mtfishman:
I'm not sure if there is a use case for conserving both total spin and spin parity, since total spin inherently conserves spin parity, but I don't see the harm in allowing it.

emstoudenmire:
I agree that it doesn't hurt to allow it, because someone might like to see the extra information.

mtfishman:
Is "Parity" a good name here? Should it have "Sz" in it, like "ParitySz"?

emstoudenmire:
You know, that's a good point: it should have Sz in the name. The reason is that one might want to mix this with fermion parity which I think we just call "Parity" right now too.

mtfishman:
I think fermion parity is "Pf". It would be kind of nice if the names were consistent, but "PSz" looks a bit terse.

emstoudenmire:
Yes you're right it's "Pf". With the benefit of hindsight, a better name might be "FParity" and we could use "SzParity" for this one. Technically it would be a breaking change to change "Pf" to something else but it may not affect any users currently.

mtfishman:
This makes me think, should we give keyword arguments to allow renaming the QN names? So that people can decide how they can mix? Like:
```julia
function space(::SiteType"S=1/2";
               conserve_qns = false,
               conserve_sz = conserve_qns,
               conserve_psz = false,
               qnname_sz = "Sz",
               qnname_psz = "PSz")
  if conserve_sz
    return [QN(qnname_sz,+1)=>1,QN(qnname_sz,-1)=>1]
               conserve_sz = conserve_qns,
               conserve_parity = false)
  if conserve_sz && conserve_parity
    return [QN((qnname_sz, +1), (qnname_psz, 1, 2)) => 1,
            QN((qnname_sz, -1), (qnname_psz, 0, 2)) => 1]
  elseif conserve_sz
    return [QN(qnname_sz, +1) => 1,
            QN(qnname_sz, -1) => 1]
  elseif conserve_parity
    return [QN(qnname_psz, 1, 2) => 1,
            QN(qnname_psz, 0, 2) => 1]
  end
  return 2
end
```

mtfishman:
Yeah, "FParity" would be good. My main concern is if we go with "*Parity" as the convention, then the names might get too long for certain symmetries.

PR #469, opened 2020-07-17
Add generic version of "F" operator for non-fermion sites in AutoMPO

emstoudenmire:


src/physics/site_types/generic_sites.jl
```diff
@@ -1,10 +1,15 @@
 
 function op!(Op::ITensor,
              ::OpName"Id",
-             ::SiteType,
+             ::SiteType"_Generic",
```
mtfishman, line 4: Why is this change needed?

emstoudenmire: Yes, I was going to get your feedback on this change, thanks. So the problem I ran into is that AutoMPO need to be able to construct an "F" operator for all site types, even ones that don't define it explicitly. (I could put a try...catch block in AutoMPO to supply a default F there if one doesn't exist but thought that was less nice.)

But the problem I found is that if I make a generic "F" in the approach done before this PR, then for site types that *do* define "F", sometimes the generic version still gets called depending on the ordering of the tags in the Index. This happens when the special tag isn't the first tag, so that the first tag triggers this generic "F" overload and that preempts the custom "F" overload.

So my solution here is to always put a special SiteType"_Generic" at the very end of the vector of SiteTypes internally used by the`op` function. Then it is guaranteed to only call these generic versions as a last option.

Please let me know if you can think of a different fix - I'd be happy to have one.

mtfishman: Ah right, I didn't realize that would be a problem with the generic "Id" fallback. Indeed, it seems like we need a better system for that.

I guess the minimal example of this problem is:
```julia
julia> s = Index(2, "A,B");

julia> ITensors.op!(op::ITensor, ::OpName"Id", ::SiteType"B", s::Index) = (op[1, 1] = 2);

julia> @show op(s, "Id");
op(s, "Id") = ITensor ord=2
Dim 1: (dim=2|id=321|"A,B")'
Dim 2: (dim=2|id=321|"A,B")
 2×2
 1.0  0.0
 0.0  1.0
```
Interestingly, if you overload the `op` version it works fine:
```julia
julia> ITensors.op(::OpName"Id", ::SiteType"B", s::Index) = randomITensor(s', s);

julia> @show op(s, "Id");
op(s, "Id") = ITensor ord=2
Dim 1: (dim=2|id=928|"A,B")'
Dim 2: (dim=2|id=928|"A,B")
 2×2
 -1.5652147535991363  -0.6779199201129691
  0.9558799258795723   0.4668127333439197
```
since there is not generic `op(::OpName"Id", ...)` overload.

Anyway, this approach seems like a good way to go (can't think of anything better). I guess some users might want to define their own SiteType"_Generic" overloads (for example, a fermionic swap gate). Overloading `op!(::ITensor, ::OpName, ::SiteType"_Generic", ::Index...)` seems reasonable.

Issue #470, opened 2020-07-20
Use ! function for not logic


mtfishman:
As an alias for the `not` function, we can use the function `!` (I just realized it was available as a function).

For example, to prime all indices in ITensor `A` besides `i`, we could use the syntax: `prime(A; inds = !i)`. Right now this gives:
```julia
julia> i = Index(2, "i"); j = Index(2, "j");

julia> A = randomITensor(i, j);

julia> prime(A, inds = !i)
ERROR: MethodError: no method matching !(::Index{Int64})
[...]
```
We can also use the same syntax for not priming indices with a certain tag, such as:
```julia
julia> prime(A, tags = !"i")
ERROR: MethodError: no method matching !(::String)
[...]
```
but we should be careful about that, since both the `!` function and the `String` type are not defined by ITensors.jl. Therefore, it is type piracy, and other packages may have a definition for `!(::String)`.

mtfishman:
Another interesting point I hadn't noticed is that Julia automatically defines the operation `!(::Function)` to negate the boolean result of the function. So that means the following already works:
```julia
julia> prime(!hastags("i"), A)
ITensor ord=2 (dim=2|id=41|"i") (dim=2|id=812|"j")'
NDTensors.Dense{Float64,Array{Float64,1}}
```
which could make the implementation of the `not` logic in ITensors.jl simpler.

mtfishman:
Closed by #471.

PR #471, opened 2020-07-20
Add ! as not syntax

mtfishman:
This adds `!` as an alternative syntax for the `not` function, as suggested in #470.

The main concern is allowing syntax like `prime(A, plev = !0)` and `prime(A, tags = !"i")`. Right now, this is done by defining:
```julia
!(n::Int) = Not{Int}(n)
!(s::String) = Not{String}(s)
```
where `Not` is just a lossless wrapper type. I think this is not too harmful. I'm not sure how many packages would be defining `!(::Int)` and `!(::String)`, and if they are either they can play nicely with our `Not` wrapper type or just overwrite our definition.


emstoudenmire:
It's a nice notation for indices, but I am definitely concerned about redefining or defining "!" for Int and String. I could imagine some very surprising behavior when combining ITensors with other libraries if we export these, especially the Int case (I'm thinking of binary logic use cases).

mtfishman:
Yeah, the more I think about it the more I think it is a bad idea to define `!(::Int)` and `!(::String)`. We could just define it for types that are "owned" by ITensors, like Index, IndexSet, and TagSet. Perhaps we could make a simpler TagSet constructor, like `ts"mytag"`, so then we can still use `!ts"mytag"`.

mtfishman:
In the latest, I removed the overloads for `!(::Int)` and `!(::String)`. I also add the macro `@ts_str` to construct a TagSet, which I think is helpful (it is pretty verbose to type `TagSet("Site,n=2")`).

So now you can do:
```julia
julia> i = Index(2, "i");

julia> j = Index(2, "j");

julia> A = randomITensor(i,j)
ITensor ord=2 (dim=2|id=461|"i") (dim=2|id=857|"j")

julia> prime(A, tags = !ts"i")
ITensor ord=2 (dim=2|id=461|"i") (dim=2|id=857|"j")'
```

I noticed that there is a package [InvertedIndices.jl](https://github.com/mbauman/InvertedIndices.jl) which implements something similar to our `Not` type. It is used to get slices of a Julia Array, i.e.
```julia
julia> using InvertedIndices

julia> A = randn(2,2)
2×2 Array{Float64,2}:
 0.353677   0.0622453
 0.485538  -0.0994356

julia> A[Not(1), :]
1×2 Array{Float64,2}:
 0.485538  -0.0994356
```
It is used by [DataFrames.jl](https://github.com/JuliaData/DataFrames.jl), for example. We could consider that kind of usage for ITensor slicing as well.

PR #472, opened 2020-07-20
Add Sweeps constructor from matrix of parameters

mtfishman:
This adds the following Sweeps constructor:
```julia
  julia> Sweeps([
           "maxdim" "mindim" "cutoff" "noise"
            50       10       1e-12    1E-7
            100      20       1e-12    1E-8
            200      20       1e-12    1E-10
            400      20       1e-12    0
            800      20       1e-12    1E-11
            800      20       1e-12    0
           ])
  Sweeps
  1 cutoff=1.0E-12, maxdim=50, mindim=10, noise=1.0E-07
  2 cutoff=1.0E-12, maxdim=100, mindim=20, noise=1.0E-08
  3 cutoff=1.0E-12, maxdim=200, mindim=20, noise=1.0E-10
  4 cutoff=1.0E-12, maxdim=400, mindim=20, noise=0.0E+00
  5 cutoff=1.0E-12, maxdim=800, mindim=20, noise=1.0E-11
  6 cutoff=1.0E-12, maxdim=800, mindim=20, noise=0.0E+00
```
This is useful for input files for complicated DMRG runs.

In the future, we could consider adding a macro `@sweeps` that simplifies the syntax a little bit (like not requiring quotation marks around the Strings and not requiring brackets). We could also add a syntax for missing values for indicating that a value gets repeated.


emstoudenmire:
Great - I really like this idea and syntax. (I didn't realize you could leave out commas entirely for matrices, I thought you still needed them at the end of each row, so that's nice to learn.)

My only suggestion was going to be something I see you've already implemented, which is to allow specifying only the columns you want in the order you want to by looking at the strings provided.

Nice how simple Julia's slicing makes the implementation too.

PR #473, opened 2020-07-20
Add examples of input files

mtfishman:
* Add examples of using script input file and ArgParse.

* Allow specifying number of sweeps in matrx Sweeps constructor.


mtfishman:
In the latest, I added the `argsdict()` function to ITensors, which does custom command line parsing. The idea is to use it in conjunction with an input file, to overwrite default values from the command line. For example with the files `input` and `dmrg.jl`:
```julia
# input
N = 100
```
and the DMRG file:
```julia
# dmrg.jl
using ITensors
args = argsdict(; prefix = "file")
include(args["file1"])
for (k, v) in args
  @show k, v
end
N = get(args, "N", N)
@show N
```
which is run as follows:
```
$ julia dmrg.jl input N=20                
(k, v) = ("file1", "input")
(k, v) = ("N", 20)
N = 20
```

I also expanded on the input file examples. Now there is one simple one with just an input file script, one example of using ITensor's new `argsdict()` function, and one example of using the package `ArgParse.jl`.

PR #480, opened 2020-08-04
Add MPS/MPO circuit evolution

mtfishman:
This adds circuit evolution functionality for MPS and MPO. The basic interface is the following:
```julia
using ITensors

N = 3

ITensors.op(::OpName"σx", ::SiteType"S=1/2", s::Index) =
  2*op("Sx", s)

ITensors.op(::OpName"σz", ::SiteType"S=1/2", s::Index) =
  2*op("Sz", s)

# Make the operator list.
os = [("σx", n) for n in 1:N]
append!(os, [("σz", n) for n in 1:N])

@show os

s = siteinds("S=1/2", N)
gates = ops(os, s)

# Starting state |↑↑↑⟩
ψ0 = productMPS(s, "↑")

# Apply the gates.
ψ = apply(gates, ψ0; cutoff = 1e-15)

# Test against exact (full) wavefunction
prodψ = apply(gates, prod(ψ0))
@show prod(ψ) ≈ prodψ

# The result is:
# σz₃ σz₂ σz₁ σx₃ σx₂ σx₁ |↑↑↑⟩ = -|↓↓↓⟩
@show inner(ψ, productMPS(s, "↓")) == -1
```
(updated for new, simpler notation).

This supports general n-site gates applied to arbitrary non-local site, where swaps are performed to make the sites contiguous to apply the gate. This also supports applying gates either to one side of an MPO, or applying the same set of gates to both sides of an MPO.

A big caveat is that fermionic swaps are not currently implemented, so for fermionic site types the answers will be wrong. This will be addressed in future work.


mtfishman:
Here is an example with nonlocal gates:
```julia
# 2-site gate
function ITensors.op(::OpName"CX", ::SiteType"S=1/2", s1::Index, s2::Index)
  mat = [1 0 0 0
         0 1 0 0
         0 0 0 1
         0 0 1 0]
  return itensor(mat, s2', s1', s2, s1)
end

os = ProductOps()
os >>= "CX", 1, 3
os >>= "σz", 3
gate = ops(os, s)

# Start with the state |↓↑↑⟩
ψ0 = productMPS(s, n -> n == 1 ? "↓" : "↑")

# The result is:
# σz₃ CX₁₃ |↓↑↑⟩ = -|↓↑↓⟩
ψ = apply(ops(os, s)..., ψ0; cutoff = 1e-15)
@show inner(ψ, productMPS(s, n -> n == 1 || n == 3 ? "↓" : "↑")) == -1
```

mtfishman:
Using the interface from this PR, here is a very minimal example of a TEBD-like time evolution of the Heisenberg model:
```julia
# Define the nearest neighbor term `S⋅S` for the Heisenberg model
function ITensors.op(::OpName"S⋅S", ::SiteType"S=1/2", s1::Index, s2::Index; f::Function = identity)
  return f(0.5 * op("S+", s1) * op("S-", s2) +
           0.5 * op("S-", s1) * op("S+", s2) +
           op("Sz", s1) * op("Sz", s2))
end

# Define a function to scale and exponentiate each term
τ = -0.1im
expτh = o -> exp(τ*o; ishermitian = true)

# Define the gates
os = ProductOps()
os >>= "S⋅S", (1, 2), (f = expτh,)
os >>= "S⋅S", (2, 3), (f = expτh,)
expτH = ops(os, s)

# Make the starting state and perform the evolution
ψ0 = productMPS(s, n -> n == 1 ? "↓" : "↑")
ψτ = apply(expτH..., ψ0)
```

An interesting question would be if we can make the `ProductOps` syntax more sophisticated by allowing sums of local operators, so the user doesn't have to define the `op` overload. However, I think the approach shown above is already quite flexible.

Another approach would be to make the list of gates for TEBD directly from an `AutoMPO`, which would be a nice "black box" solution.

mtfishman:
In the latest version of this PR, I've removed the `ProductOps` type. Instead, operators are just defined in terms of a vector of tuples that gate the gate name and positions, for example:
```julia
os = [("Sx", 1), ("Sx", 2), ("Sx", 3)]
```
Then, the gates are applied as follows:
```julia
s = siteinds("S=1/2", 3)
product(ops(os, s), productMPS(s, "↑"); cutoff = 1e-15)
```
where the gates are applied in the order they appear in the vector `os`. The examples in the previous comments are updated with the new syntax.

I'm working on getting gate evolution working properly for fermionic models (where the fermion sign is handled automatically when swapping sites). That is the last thing that is needed before this PR is ready to merge.

src/exports.jl
```diff
@@ -280,6 +282,10 @@ export
   square_lattice,
   triangular_lattice,
 
+# physics/productops.jl
+  ProductOps,
```
emstoudenmire, line 286: ProductOps is still in exports here

test/qubit.jl
```diff
@@ -0,0 +1,270 @@
+using ITensors
+
+#
+# qubit
+#
+
+ITensors.space(::SiteType"qubit") = 2
```
emstoudenmire, line 7: Should we consider moving the qubit site type into the library? Seems like it might get used a lot and we could publicize it as an out-of-the-box feature of ITensors. If we do, let's name it "Qubit" for better consistency with "Electron" and "Fermion".

emstoudenmire:
Looks really good. Was the move away from ProductOps because of the issues around the `*` and `*=` notation, and possible differences from conventions like in quantum computing that we had discussed? Or a different reason?

test/qubit.jl
mtfishman: We can consider it, let's discuss it over Slack. Also agreed about changing it to "Qubit".

mtfishman:
> Looks really good. Was the move away from ProductOps because of the issues around the `*` and `*=` notation, and possible differences from conventions like in quantum computing that we had discussed? Or a different reason?

Yes, I couldn't decide on the notation and conventions for `ProductOps`, we can always add that on later as we see how this gets used. The current design is pretty unambiguous and straightforward, so it is a good starting point.

test/qubit.jl
mtfishman: Basically, at least at first I would lean towards providing it as an "example" rather than as an officially supported site type, since there are so many different gates that could be added which have different naming conventions.

src/exports.jl
mtfishman: Thanks, just fixed this.

test/qubit.jl
mtfishman: Changed to "Qubit", and I've moved `qubit.jl` to `examples/gate_evolution/` (and will hold off on making it an "official" site type for a little while once this gets tested a bit more).

mtfishman:
I fixed a few bugs in the fermionic swap logic, and now evolution with fermionic gates looks like it is working (there are tests for computing non-local 2-point and 4-point correlation functions for the Fermion and Electron site type).

Over all, once tests pass, this is ready to go from my end. @emstoudenmire, please let me know if you see anything objectionable about the interface.

emstoudenmire:
I think the interface looks great, judging from the example code. There are some strong positives about using plain arrays over ProductOps, such as being able to use array comprehension notation or the `append!` function, and just the simplicity of it. (Maybe we should make the AutoMPO system work in a similar way, where you can instead just pass an array of tuples to the MPO constructor, and reduce AutoMPO to just a tool to generate such an array with convenient math notation.)

PR #481, opened 2020-08-07
add docstrings for siteinds methods

orialb:


emstoudenmire:
Great! The only improvements I’d suggest are to add a brief sentence for the second one clarifying that `f(i)` should return a string. Also I’d suggest using a different variable than `i` because to me here `i` kind of evokes an Index object. So `n` would likely clarify that it’s an integer argument that gets put into f.

orialb:
Thanks for the improvement suggestions Miles, I updated the PR.

mtfishman:
Looks like the failing test is just a floating point check that is too strict. Thanks for the contribution!

PR #483, opened 2020-08-25
Add dense function for MPS and MPO

emstoudenmire:


src/itensor.jl
```diff
@@ -419,8 +419,8 @@ avoiding allocating new data if possible.
 For example, an ITensor with Diag storage will become Dense storage,
 filled with zeros except for the diagonal values.
 """
-function NDTensors.dense(A::ITensor)
-  T = dense(tensor(A))
+function dense(A::ITensor)
```
mtfishman, line 422: What was the reason for this change? I think we want the original version. The new version would mean that both ITensors and NDTensors would have their own version of `dense`, instead of a single version in NDTensors that is overloaded by ITensors.

emstoudenmire: Thanks for noticing - the reason for this change was that I was getting a lot of strange errors about `dense` not being defined. A lot of things I tried didn't work. Let's discuss on Slack because I'm probably thinking about it the wrong way.

mtfishman: Ah, I think you just want to make sure that you overload this version and the `AbstractMPS` version with `NDTensors.dense`.

I believe the problem was that since you defined the `AbstractMPS` version as `dense(M::AbstractMPS)` instead of `NDTensors.dense(M::AbstractMPS)`, it creates a new local version of the `dense` function within the ITensors module, so then Julia can't "see" the NDTensors version of the `dense` function without calling it as `NDTensors.dense`.

src/mps/abstractmps.jl
```diff
@@ -88,6 +88,20 @@ Base.iterate(M::AbstractMPS) = iterate(data(M))
...
+"""
+    dense(M::MPS)
+    dense(M::MPO)
+
+Given an MPS (or MPO), return a new MPS (or MPO) 
+having called `dense` on each ITensor to convert each
+tensor to use dense storage and remove any QN or other
+sparse structure information, if it is not dense already.
+"""
+function dense(M::AbstractMPS)
```
mtfishman, line 100: Based on my comment above, please try defining this with `NDTensors.dense`.

mtfishman:
Thanks for the addition, this is good to have. It also reminds me that it would be nice to implement a proper `removeqns` function...

Also, it would be nice if we could have an alternative to this that is implemented through broadcasting, i.e.:
```julia
julia> using ITensors

julia> s = siteinds("S=1/2", 3, conserve_qns = true);

julia> ψ = productMPS(s, "↑");

julia> dense.(ψ)
3-element Array{ITensor,1}:
 ITensor ord=2
Dim 1: (dim=2|id=810|"S=1/2,Site,n=1")
Dim 2: (dim=1|id=121|"Link,l=1")
[...]
```
Unfortunately right now the broadcasting returns a `Vector{ITensor}`, since we need to tell the broadcasting system in some way that it should return an MPS with the correct orthogonality center (it just default to `Vector` storage).

mtfishman:
Looks good to me, is this ready to merge?

emstoudenmire:
Yes ready to merge. I'll go ahead!

Issue #266, opened 2020-04-10
Optimizations for block sparse operations


mtfishman:
Here is an issue listing potential improvements to block sparse tensor operations.

- [ ] Add multithreading to contraction and combining.
- [ ] Use a Dict instead of a sorted Vector for the block offsets list (like https://github.com/ITensor/ITensor/pull/336).
- [ ] In contraction, try combining indices first and then contracting with a matrix-matrix multiplication.
- [x] Try Jutho's [Strided.jl](https://github.com/Jutho/Strided.jl) package for faster permutations.

mtfishman:
I was curious about how the strategy of combining and then contracting would work right now. Here is the function for testing that:
```julia
function contract_combine(A::ITensor, B::ITensor)
  A_uniqueinds = uniqueinds(A, B)
  B_uniqueinds = uniqueinds(B, A)
  AB_commoninds = commoninds(A, B)
  CA = combiner(A_uniqueinds)
  CB = combiner(B_uniqueinds)
  CAB = combiner(AB_commoninds)
  AC = A * CA * CAB
  BC = B * CB * dag(CAB)
  CC = AC * BC
  C = CC * dag(CA) * dag(CB)
  return C
end
```
Running the following code:
```julia
function main()
  Nmax = 8
  for N in 1:Nmax
    println("order = ", 2 * N)
    d = 1
    i = Index(QN(0,2) => d, QN(1,2) => d)
    is = IndexSet(ntuple(n -> settags(i, "i$n"), Val(N)))
    A = randomITensor(is'..., dag(is)...)
    B = randomITensor(is'..., dag(is)...)
    println("Contract:              ", @elapsed A' * B)
    println("Combine then contract: ", @elapsed contract_combine(A', B))
    println()
  end
end
```
I get:
```
order = 2
Contract:              7.1355e-5
Combine then contract: 0.004867469

order = 4
Contract:              0.000149889
Combine then contract: 0.006314275

order = 6
Contract:              0.000469303
Combine then contract: 0.006801207

order = 8
Contract:              0.00263369
Combine then contract: 0.007549091

order = 10
Contract:              0.027489216
Combine then contract: 0.01168987

order = 12
Contract:              0.525102576
Combine then contract: 0.034727689

order = 14
Contract:              4.987328203
Combine then contract: 0.146389961

order = 16
Contract:              49.175892434
Combine then contract: 2.307021707
```
[...]

PR #536, opened 2020-12-04
Add global flag for combining before contracting QN ITensors

mtfishman:
This adds a global flag for reshaping QN ITensors into matrices before contracting, which can improve the efficiency for contracting high order block sparse tensors. It is basically the technique described in this comment: https://github.com/ITensor/ITensors.jl/issues/266#issuecomment-629369474

Here is a benchmark using the technique, which you can turn on and off with the functions `enable_combine_contract!()` and `disable_combine_contract!()`:
```julia
using BenchmarkTools
using ITensors

function main(; Nmax = 8)
  # Don't warn about large tensor orders
  disable_warn_order!()

  for N in 1:Nmax
    println("#################################################")
    println("# order = ", 2 * N)
    println("#################################################")
    println()

    d = 1
    i = Index(QN(0,2) => d, QN(1,2) => d)
    is = IndexSet(ntuple(n -> settags(i, "i$n"), Val(N)))
    A = randomITensor(is'..., dag(is)...)
    B = randomITensor(is'..., dag(is)...)

    # Use standard algorithm, without combining first
    println("Contract:")
    C_contract = @btime $A' * $B samples = 5
    println()

    # Reshape the ITensors into matrices before contracting
    println("Combine then contract:")
    enable_combine_contract!()
    C_combine_contract = @btime $A' * $B samples = 5
    disable_combine_contract!()
    println()

    @show C_contract ≈ C_combine_contract
    println()
  end
end
```
which gives the results (using NDTensors#master which has some block sparse optimizations):
```julia
julia> main(Nmax = 8)
#################################################
# order = 2
#################################################

Contract:
  2.410 μs (76 allocations: 5.70 KiB)

Combine then contract:
  48.111 μs (899 allocations: 92.98 KiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 4
#################################################

Contract:
  11.329 μs (311 allocations: 24.73 KiB)

Combine then contract:
  111.227 μs (1628 allocations: 182.69 KiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 6
#################################################

Contract:
  119.332 μs (2125 allocations: 200.11 KiB)

Combine then contract:
  284.701 μs (4019 allocations: 470.45 KiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 8
#################################################

Contract:
  1.291 ms (16988 allocations: 1.75 MiB)

Combine then contract:
  1.250 ms (18801 allocations: 1.75 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 10
#################################################

Contract:
  13.555 ms (138908 allocations: 16.22 MiB)

Combine then contract:
  6.233 ms (103211 allocations: 8.28 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 12
#################################################

Contract:
  165.222 ms (1261274 allocations: 141.49 MiB)

Combine then contract:
  37.353 ms (492436 allocations: 41.43 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 14
#################################################

Contract:
  2.256 s (10387274 allocations: 1.24 GiB)

Combine then contract:
  187.437 ms (2158837 allocations: 187.26 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 16
#################################################

Contract:
  28.637 s (104661017 allocations: 14.86 GiB)

Combine then contract:
  1.416 s (17813407 allocations: 1.32 GiB)

C_contract ≈ C_combine_contract = true

```
You can see the crossover is now around order 6-8 ITensors, which is a bit better than the last time I checked.

An immediate way to make this work better, besides low-level optimizations of the combiner code, would be to fuse the two combiners/uncombiners into one operation. Currently, the QN ITensors are turned into matrix-like (order 2) ITensors by contracting two combiners sequentially. We could fuse that operation into a single operation and minimize the amount of work the combiner step has to take (minimize the number of permutations needed, for example). My guess is that, at least for this example, that could make this combining technique match the speed of the standard contraction method for all of the ITensor orders. Note that this example of parity symmetry is particularly suited for this technique, since the order 2 ITensors end up with only 2 dense blocks to contract, so for other symmetries like U(1) this may not help as much (for example I see a crossover at order 8-10 for this same example when I switch to U(1) symmetry).

### DIFF
--- a/examples/combine_contract/README.md
+++ b/examples/combine_contract/README.md
@@ -1,0 +1,114 @@
+For QN ITensors (ITensors with block sparse structure), ITensors.jl
+provides two "modes" for contracting them. The default mode is treating
+them as block sparse multi-dimensional arrays and contracting them
+"as is".
+
+In the second mode, the ITensors are reshaped into block
+sparse matrices before contracting. Making use of the QN information
+in the tensors, the second mode can lead to fewer blocks being contracted,
+which can lead to a speedup depending on the number of blocks, the sizes
+of the blocks, the order of the ITensors being contracted, etc.
+
+Run the example using parity symmetry as follows:
+```julia
+julia> include("main.jl")
+
+julia> main(Nmax = 8)
+#################################################
+# order = 2
+#################################################
+
+Contract:
+  2.410 μs (76 allocations: 5.70 KiB)
+
+Combine then contract:
+  48.111 μs (899 allocations: 92.98 KiB)
+
+C_contract ≈ C_combine_contract = true
+
+#################################################
+# order = 4
+#################################################
+
+Contract:
+  11.329 μs (311 allocations: 24.73 KiB)
+
+Combine then contract:
+  111.227 μs (1628 allocations: 182.69 KiB)
+
+C_contract ≈ C_combine_contract = true
+
+#################################################
+# order = 6
+#################################################
+
+Contract:
+  119.332 μs (2125 allocations: 200.11 KiB)
+
+Combine then contract:
+  284.701 μs (4019 allocations: 470.45 KiB)
+
+C_contract ≈ C_combine_contract = true
+
+#################################################
+# order = 8
+#################################################
+
+Contract:
+  1.291 ms (16988 allocations: 1.75 MiB)
+
+Combine then contract:
+  1.250 ms (18801 allocations: 1.75 MiB)
+
+C_contract ≈ C_combine_contract = true
+
+#################################################
+# order = 10
+#################################################
+
+Contract:
+  13.555 ms (138908 allocations: 16.22 MiB)
+
+Combine then contract:
+  6.233 ms (103211 allocations: 8.28 MiB)
+
+C_contract ≈ C_combine_contract = true
+
+#################################################
+# order = 12
+#################################################
+
+Contract:
+  165.222 ms (1261274 allocations: 141.49 MiB)
+
+Combine then contract:
+  37.353 ms (492436 allocations: 41.43 MiB)
+
+C_contract ≈ C_combine_contract = true
+
+#################################################
+# order = 14
+#################################################
+
+Contract:
+  2.256 s (10387274 allocations: 1.24 GiB)
+
+Combine then contract:
+  187.437 ms (2158837 allocations: 187.26 MiB)
+
+C_contract ≈ C_combine_contract = true
+
+#################################################
+# order = 16
+#################################################
+
+Contract:
+  28.637 s (104661017 allocations: 14.86 GiB)
+
+Combine then contract:
+  1.416 s (17813407 allocations: 1.32 GiB)
+
+C_contract ≈ C_combine_contract = true
+
+```
+
--- a/examples/combine_contract/main.jl
+++ b/examples/combine_contract/main.jl
@@ -1,0 +1,36 @@
+using BenchmarkTools
+using ITensors
+
+function main(; Nmax = 8)
+  # Don't warn about large tensor orders
+  disable_warn_order!()
+
+  for N in 1:Nmax
+    println("#################################################")
+    println("# order = ", 2 * N)
+    println("#################################################")
+    println()
+
+    d = 1
+    i = Index(QN(0) => d, QN(1) => d)
+    is = IndexSet(ntuple(n -> settags(i, "i$n"), Val(N)))
+    A = randomITensor(is'..., dag(is)...)
+    B = randomITensor(is'..., dag(is)...)
+
+    # Use standard algorithm, without combining first
+    println("Contract:")
+    C_contract = @btime $A' * $B samples = 5
+    println()
+
+    # Reshape the ITensors into matrices before contracting
+    println("Combine then contract:")
+    enable_combine_contract!()
+    C_combine_contract = @btime $A' * $B samples = 5
+    disable_combine_contract!()
+    println()
+
+    @show C_contract ≈ C_combine_contract
+    println()
+  end
+end
+
--- a/src/ITensors.jl
+++ b/src/ITensors.jl
@@ -22,24 +22,6 @@ using StaticArrays
 using TimerOutputs
 
 #####################################
-# Global flags
-#
-const _use_combine_contract = Ref(false)
-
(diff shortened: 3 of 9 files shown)


mtfishman:
(Also credit to Glen Evenbly for pointing out this alternative QN contraction technique.)

emstoudenmire:
Great, this will be useful for testing at the very least, or as an advanced feature. Let's think in the future about whether we can toggle this option automatically for people using some heuristic (like the crudest one would be counting the number of indices involved in the contraction, whether contracted, uncontracted, or both; a better one might involve estimating or fully calculating the sizes and number of blocks before versus after combining, though that could be too expensive or tricky). It might be a difficult thing to get right or too much of a research frontier, though.

We should keep this in mind, by the way, for having discussions with the broader tensor computation community in the sense of showing that there are a lot of non-trivial things to think about regarding performance of block-sparse tensor contraction and it would be good for someone to make a common low-level library that incorporates tricks like these internally one day.

mtfishman:
Right, it would be nice to have a heuristic, but I think it could be quite complicated (since it would depend on the order of the tensor, the number of QN sectors within each index, the number of nonzero blocks in each ITensor before and after combining, the type of symmetry, etc.). I was picturing that for now, a user would keep this off for something like DMRG, but then turn it on for algorithms that involve higher order tensors like MERA, TTN, and PEPS.

My hope would be that we could just make it so that it is always as fast or faster than not combining (Glen said he always uses this technique in his code, but he is often working with very high order tensors).

A complication with making this technique more general is that, at least right now, this relies on information about the QNs, which tells you what blocks can get combined. However perhaps this could be made blind to the QNs in some way.

mtfishman:
Here are some updated benchmarks using the improvements in https://github.com/ITensor/NDTensors.jl/pull/53:
```julia
julia> main()
#################################################
# order = 2
#################################################

Contract:
  2.407 μs (64 allocations: 5.42 KiB)

Combine then contract:
  47.069 μs (887 allocations: 92.70 KiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 4
#################################################

Contract:
  10.741 μs (215 allocations: 22.23 KiB)

Combine then contract:
  108.055 μs (1616 allocations: 182.41 KiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 6
#################################################

Contract:
  101.901 μs (1357 allocations: 188.11 KiB)

Combine then contract:
  284.599 μs (3991 allocations: 467.80 KiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 8
#################################################

Contract:
  1.074 ms (10844 allocations: 1.63 MiB)

Combine then contract:
  1.291 ms (18789 allocations: 1.75 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 10
#################################################

Contract:
  9.830 ms (89756 allocations: 14.97 MiB)

Combine then contract:
  6.418 ms (103135 allocations: 8.23 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 12
#################################################

Contract:
  121.839 ms (868058 allocations: 138.49 MiB)

Combine then contract:
  38.362 ms (492424 allocations: 41.24 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 14
#################################################

Contract:
  1.675 s (7241546 allocations: 1.15 GiB)

Combine then contract:
  199.868 ms (2158697 allocations: 186.37 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 16
#################################################

Contract:
  22.563 s (79495193 allocations: 14.61 GiB)

Combine then contract:
  1.493 s (17813395 allocations: 1.32 GiB)

C_contract ≈ C_combine_contract = true

```
That mostly improves the uncombined version, since those optimizations focused on contraction optimizations for very small blocks.

mtfishman:
Here are the same benchmarks after the changes introduced in (https://github.com/ITensor/NDTensors.jl/pull/54):
```julia
julia> main()
#################################################
# order = 2
#################################################

Contract:
  1.580 μs (51 allocations: 4.83 KiB)

Combine then contract:
  50.242 μs (877 allocations: 94.34 KiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 4
#################################################

Contract:
  4.091 μs (60 allocations: 9.66 KiB)

Combine then contract:
  110.874 μs (1620 allocations: 186.27 KiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 6
#################################################

Contract:
  30.831 μs (83 allocations: 55.81 KiB)

Combine then contract:
  290.229 μs (3867 allocations: 464.59 KiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 8
#################################################

Contract:
  290.239 μs (99 allocations: 467.00 KiB)

Combine then contract:
  1.060 ms (13371 allocations: 1.54 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 10
#################################################

Contract:
  3.534 ms (115 allocations: 3.21 MiB)

Combine then contract:
  5.125 ms (76588 allocations: 7.12 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 12
#################################################

Contract:
  39.490 ms (131 allocations: 21.20 MiB)

Combine then contract:
  28.124 ms (369079 allocations: 37.02 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 14
#################################################

Contract:
  726.577 ms (150 allocations: 184.92 MiB)

Combine then contract:
  156.128 ms (1576503 allocations: 165.32 MiB)

C_contract ≈ C_combine_contract = true

#################################################
# order = 16
#################################################

Contract:
  13.567 s (12583081 allocations: 4.12 GiB)

Combine then contract:
  1.229 s (14995579 allocations: 1.22 GiB)

C_contract ≈ C_combine_contract = true

```
Note that only part of the speedup is strictly because of the new Dictionary design (maybe about half of the speedup). The rest of the speedup is from optimizations particular to blocks of size 1 which we have in this case, for example now there is a special code path when contracting blocks of size 1 to avoid all of the contraction logic and simply do arithmetic on the single elements of the block.

Also, again the optimizations were primarily focused on optimizing block sparse contractions, not combining, so the combiner could could be optimized quite a bit.

PR #539, opened 2020-12-10
Add Ising YY and ZZ gates

anupam-mitra:


mtfishman:
Thanks for the contribution.

On Wikipedia, it looks like there is a different convention for the angle in the Ising ZZ gate: https://en.wikipedia.org/wiki/Quantum_logic_gate#Ising_(ZZ)_coupling_gate (both a minus sign and a factor of 2).

I also see the same definition here: https://arxiv.org/pdf/quant-ph/0209049.pdf

I'm not familiar enough with the quantum computing literature to know which convention is more common.

anupam-mitra:
Thank you for developing ITensors.jl, ITensor, and several other associated libraries, all of which are invaluable to the community.

The convention of the angle parameter in the Ising gates XX(ϕ), YY(ϕ), ZZ(ϕ) is not universal. Based on how the XX(ϕ) gate was defined in "examples/gate_evolution/qubit.jl" , that is XX(ϕ) = exp(-im *ϕ* X⊗X), I defined the YY(ϕ) gate as exp(-im *ϕ* Y⊗Y) and ZZ(ϕ) gate as exp(-im *ϕ* Z⊗Z). I think the Wikipedia article and the article by Jonathan Jones define ZZ(ϕ) as  exp(+im *ϕ/2 * Z⊗Z), while the Wikipedia article defines XX(ϕ) = exp(-im *ϕ* X⊗X) and YY(ϕ) = exp(-im *ϕ* Y⊗Y).

All these gates can be written as
 XX(ϕ) = exp(-im *ϕ* X⊗X) = cos(ϕ) I⊗I - im sin(ϕ) X⊗X
 YY(ϕ) = exp(-im *ϕ* X⊗X) = cos(ϕ) I⊗I - im sin(ϕ) Y⊗Y
 ZZ(ϕ) = exp(-im *ϕ* X⊗X) = cos(ϕ) I⊗I - im sin(ϕ) Z⊗Z

The Wikipedia article shows this for  XX(ϕ). Plugging in the matrix representations of  I⊗I, X⊗X, Y⊗Y and Z⊗Z, one would obtain the representations for the Ising gates.

I think much of the literature discussing XX(ϕ) and YY(ϕ) gates use the convention of exp(-im *ϕ* X⊗X) and exp(-im *ϕ* Y⊗Y), while much of the literature discussing  ZZ(ϕ) uses the convention of  exp(+im *ϕ/2 * Z⊗Z).

mtfishman:
Thanks for the explanation. Generally I would say to go with the convention that is more common, but the consistency with the `XX` and `YY` definitions makes sense. I'll merge it.

anupam-mitra:
Thank you.

PR #545, opened 2020-12-17
Improve contractions involving scalar-like ITensors or blocks

mtfishman:
This bumps NDTensors to v0.1.20, which has optimizations for contractions involving scalar-like tensors (https://github.com/ITensor/NDTensors.jl/pull/57). This particularly improves the efficiency of block sparse contractions that involve many blocks of size 1 (which is common for MPO tensors of local Hamiltonians).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ITensors"
 uuid = "9136182c-28ba-11e9-034c-db9fb085ebd5"
 authors = ["mfishman <mfishman@caltech.edu>"]
-version = "0.1.30"
+version = "0.1.31"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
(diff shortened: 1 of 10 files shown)


mtfishman:
Here are results for the 1D S=1 Heisenberg model with `chi = 800`, 4 BLAS threads, with and without `splitblocks`, using NDTensor master:
```julia
julia> main(800; use_splitblocks = false)
After sweep 1 energy=-138.940086143531 maxlinkdim=800 time=51.740
 53.246273 seconds (23.64 M allocations: 170.844 GiB, 2.88% gc time)

julia> main(800; use_splitblocks = true)
After sweep 1 energy=-138.940086143531 maxlinkdim=800 time=41.392
 42.461887 seconds (21.85 M allocations: 107.442 GiB, 2.99% gc time)
```
while using NDTensors before the scalar contraction optimizations:
```julia
julia> main(800; use_splitblocks = false)
After sweep 1 energy=-138.940086143531 maxlinkdim=800 time=52.738
 54.213462 seconds (29.86 M allocations: 171.229 GiB, 2.95% gc time)

julia> main(800; use_splitblocks = true)
After sweep 1 energy=-138.940086143531 maxlinkdim=800 time=49.531
 51.053270 seconds (44.21 M allocations: 152.692 GiB, 2.99% gc time)
```
so there is not much change without using `splitblocks`, but a nearly 20% speedup using `splitblocks`.

PR #551, opened 2020-12-23
Make OpName use Symbol instead of SmallString for longer operator names

### LINKED ISSUE #550: Use Symbol instead of SmallString in OpName
If we switch OpName to use a Symbol instead of a SmallString, we could use longer operator names.

mtfishman:
Closes #550.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+ITensors v0.1.34 Release Notes
+==============================
+* Allow operator names in the `op` system that are longer than 8 characters (PR #551).
+
 ITensors v0.1.33 Release Notes
 ==============================
 * Fix bug introduced in v0.1.32 involving inner(::MPS, ::MPS) if the MPS have more than one site Index per tensor (PR #549).
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,7 @@ name = "ITensors"
 uuid = "9136182c-28ba-11e9-034c-db9fb085ebd5"
 authors = ["Matthew Fishman <mfishman@flatironinstitute.org>",
            "Miles Stoudenmire <mstoudenmire@flatironinstitute.org>"]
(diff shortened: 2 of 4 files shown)


mtfishman:
@emstoudenmire, just 2 lines of code needed. I don't really see any disadvantage. The interface doesn't change, since it is still overloaded with the same notation.

emstoudenmire:
Looks good! I guess the only possible downside I can think of then is performance, though it could be that this approach is actually faster.

mtfishman:
With this design, making an OpName from a String is very fast:
```julia
julia> using ITensors

julia> @btime OpName($(Ref("abcdefgh"))[])
  104.089 ns (0 allocations: 0 bytes)
OpName{:abcdefgh}()

julia> @btime OpName($(Ref("abcdefghijklmnopqr"))[])
  102.066 ns (0 allocations: 0 bytes)
OpName{:abcdefghijklmnopqr}()
```
(The use of `Ref` stops the compiler from optimizing away the operation, see here: https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/doc/manual.md#understanding-compiler-optimizations).

On v0.1.33 the time was:
```julia
julia> @btime OpName($(Ref("abcdefgh"))[])
  134.797 ns (3 allocations: 96 bytes)
OpName{abcdefgh}()
```
so a bit slower because of the String to SmallString conversion:
```julia
julia> @btime ITensors.SmallString($(Ref("abcdefgh"))[])
  27.446 ns (1 allocation: 32 bytes)
abcdefgh
```

mtfishman:
I'll merge this once I see tests passing.

emstoudenmire:
Great. Of course we have the benchmarks above but good to see that OpName-specific micro benchmark. I'm confident it's a good change now.

PR #553, opened 2021-01-06
Implement iterate for TagSet

### LINKED ISSUE #542: Should we define iterate for TagSet?
Currently, things like `for tag in tags(ind)` errors, since `iterate(t::TagSet)` is not defined. `TagSet` behaves like an iterable in most respects (e.g. it has `length`, it has linear `getindex`) so it might make sense to define `iterate` for it. I can open a pull request, if this looks like a good idea.

tomohiro-soejima:
Closes #542.

### DIFF
--- a/src/tagset.jl
+++ b/src/tagset.jl
@@ -241,7 +241,7 @@ l
 tag
 ```
 """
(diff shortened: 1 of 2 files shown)


tomohiro-soejima:
Oops, there was a typo in the new test, so I corrected it.

emstoudenmire:
Thank you Tomohiro!

PR #569, opened 2021-02-03
Add generic support for scalar ITensor contraction

### LINKED ISSUE #564: Scalar ITensor for QN or BlockSparse Case
It is a useful code pattern to start a chain of contractions from a scalar ITensor constructed as `ITensor(1.0)`. One example is building 'environment' tensors such as part of an MPO projected into an MPS basis.

Currently it appears that contraction of the form `ITensor(1.0) * T` where T is a QNITensor is not supported. I think this would be a good special case to handle, and the code could be specialized to the case of one of the tensors having zero indices.

mtfishman:
This fixes #564.

@emstoudenmire, the main subtlety here is that in this implementation, it adds a special case when the ITensor has a scalar value of 1 and does nothing at all to it (just returns the non-scalar ITensor). The alternative would be that it performs a copy of the ITensor in that case, since it may be surprising for people that contractions sometimes return views of the input ITensors. Do you have an opinion on this?

There are other cases where we would not want to do a copy in ITensor contraction, for example with delta contractions that are just replacing indices (i.e. `randomITensor(i, j) * delta(j, j')`). In those cases, I would also want to return a view of the ITensor data, so perhaps contraction with `ITensor(1)` can be lumped in with that category. The main subtlety for contraction with `ITensor(1)` is that it is a pretty sneaky behavior change between contracting with `ITensor(1)` and any other scalar value, so maybe someone relies on the return not being a view and then one day they don't realize they are contracting by a value of exactly `1` and the view behavior changes... But maybe we can just make it clear in the docs that scalar-like ITensors have special behavior.

### DIFF
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1268,9 +1268,6 @@ permute(T::ITensor,
                                             IndexSet(inds...); vargs...)
 
 function (T::ITensor{N} * x::Number)::ITensor{N} where {N}
-  if isone(x)
-    return T
-  end
   return itensor(x * tensor(T))
 end
 
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1267,7 +1267,12 @@ permute(T::ITensor,
         inds::Index...; vargs...) = permute(T,
                                             IndexSet(inds...); vargs...)
 
-(T::ITensor * x::Number) = itensor(x * tensor(T))
+function (T::ITensor{N} * x::Number)::ITensor{N} where {N}
+  if isone(x)
+    return T
+  end
+  return itensor(x * tensor(T))
+end
 
 (x::Number * T::ITensor) = T * x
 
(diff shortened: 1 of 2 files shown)


mtfishman:
Thinking about this more, I'm uncomfortable designing this where:
```julia
A = randomITensor(i, j)
B = A * ITensor(1)
```
makes `B` a view of `A` but:
```julia
B = A * ITensor(2)
```
does not. However, I was thinking that if you want the view behavior (i.e. you want to avoid the copy), we could use a `delta` tensor with no indices for that purpose. The proposal would be:
```julia
A = randomITensor(i, j)
B1 = A * ITensor(1) # B1 is a copy of A
B2 = A * delta() # B2 is a view of A
```
I think the interpretation of `delta()` here is pretty unambigous. And as mentioned in the previous comment, the view behavior of contracting with a `delta` would match with how it should work for `delta` contractions that are replacing indices. We could of course introduce a different type for this (like a `Scalar` storage) but I would prefer not to if we can avoid it by using existing types.

mtfishman:
See #618 for a design proposal for tensors that contract and return views instead of copies.

PR #576, opened 2021-02-06
Add Pauli X,Y,Z to S=1/2 site type

emstoudenmire:


mtfishman:
Thanks for adding this, this is something I have thought about adding in the past.

I was thinking about adding the "Qubit" type as an official ITensors.jl site type (https://github.com/ITensor/ITensors.jl/blob/master/examples/gate_evolution/qubit.jl), however maybe we could just define all of those gates for "S=1/2" and make "Qubit" another alias for "S=1/2".

PR #577, opened 2021-02-11
DMRG Improvements

emstoudenmire:


mtfishman:
Looks good, this will be good to have!

emstoudenmire:
Yes based on a forum question I realized that seeing the max trunc err by default is helpful in a lot of cases (and for teaching people what’s important to pay attention to).

PR #591, opened 2021-03-15
Remove size type parameter from ITensor and IndexSet [WIP]

### LINKED ISSUE #584: Remove tensor order as type parameter for `ITensor{N}` and `IndexSet{N}`
I have been thinking about this design a lot, and I think it was bad decision to put the ITensor order as a type parameter in the `ITensor`/`IndexSet` types (I made the decision, so I'll kindly put the blame on myself). The most ubiquitous operation performed with ITensors is contraction, and in the case of contraction the resulting order of the ITensor is not known ahead of time (i.e. for most operations we do it is purely runtime). This makes it so that the `ITensor` order is almost never inferable by the Julia compiler, and puts unnecessary strain on the Julia compiler (it is constantly trying to infer the ITensor order, but most of the time it can't so the order is determined at runtime anyway).

Additionally, the most intensive operations on `IndexSet` objects are set operations (like `uniqueinds` and `commoninds`) which are also size-dynamic operations. That would point towards `IndexSet` naturally having a storage type that is dynamically sized, such as a `Vector{Index}`. In fact, those operations are implemented in terms of forming a Vector of indices and then converting back to an IndexSet (https://github.com/ITensor/ITensors.jl/blob/master/src/indexset.jl#L587-L593), which is not type stable. This of course was the design in C++ ITensor, which looking back I think was the right move.

I think the change won't be too much to make (and I think it could make some code simpler, especially the IndexSet code). However, it will of course be breaking to any users that were relying on those type parameters. I don't expect that is too common, but if it was the case than fixing it will mostly be a matter of removing the type parameter where necessary.

There is a tension with NDTensors that the Tensors in NDTensors do have the order as a type parameter, and I still think that is the correct decision there. I think we should handle that by uniformly converting the IndexSet to a Tuple when a Tensor is constructed, which will also make some NDTensor code simpler since then we can always assume the inds storage is a Tuple (and not rely on some convoluted generic code that had to account for the `inds` possibly being an `IndexSet`.

kshyatt:
Would close #584. Doesn't require any modifications to NDTensors.jl. Happy to change anything if the modifications (especially to `emptyITensor`) don't suit.

### DIFF
--- a/src/decomp.jl
+++ b/src/decomp.jl
@@ -1,11 +1,11 @@
 
 """
-    TruncSVD{N}
+    TruncSVD
 
 ITensor factorization type for a truncated singular-value 
 decomposition, returned by `svd`.
 """
-struct TruncSVD{N1,N2}
+struct TruncSVD
   U::ITensor
   S::ITensor
   V::ITensor
@@ -121,17 +121,17 @@ function svd(A::ITensor, Linds...; kwargs...)
   u = settags(u,utags)
   v = settags(v,vtags)
 
-  return TruncSVD{ndims(U), ndims(V)}(U,S,V,spec,u,v)
+  return TruncSVD(U,S,V,spec,u,v)
 end
 
 
 """
-    TruncEigen{N}
+    TruncEigen
 
 ITensor factorization type for a truncated eigenvalue 
 decomposition, returned by `eigen`.
 """
-struct TruncEigen{N}
+struct TruncEigen
   D::ITensor
   V::ITensor
   Vt::ITensor
@@ -237,7 +237,7 @@ function eigen(A::ITensor, Linds, Rinds; kwargs...)
     end
   end
 
-  return TruncEigen{ndims(V)}(D, V, Vt, spec, l, r)
+  return TruncEigen(D, V, Vt, spec, l, r)
 end
 
 function eigen(A::ITensor; kwargs...)
--- a/src/decomp.jl
+++ b/src/decomp.jl
@@ -6,9 +6,9 @@ ITensor factorization type for a truncated singular-value
 decomposition, returned by `svd`.
 """
 struct TruncSVD{N1,N2}
-  U::ITensor{N1}
-  S::ITensor{2}
-  V::ITensor{N2}
+  U::ITensor
+  S::ITensor
+  V::ITensor
   spec::Spectrum
   u::Index
   v::Index
@@ -121,7 +121,7 @@ function svd(A::ITensor, Linds...; kwargs...)
   u = settags(u,utags)
   v = settags(v,vtags)
 
-  return TruncSVD(U,S,V,spec,u,v)
+  return TruncSVD{ndims(U), ndims(V)}(U,S,V,spec,u,v)
 end
 
 
@@ -132,9 +132,9 @@ ITensor factorization type for a truncated eigenvalue
 decomposition, returned by `eigen`.
 """
 struct TruncEigen{N}
-  D::ITensor{2}
-  V::ITensor{N}
-  Vt::ITensor{N}
+  D::ITensor
+  V::ITensor
+  Vt::ITensor
   spec::Spectrum
   l::Index
   r::Index
@@ -148,17 +148,18 @@ iterate(E::TruncEigen, ::Val{:l}) = (E.l, Val(:r))
 iterate(E::TruncEigen, ::Val{:r}) = (E.r, Val(:done))
 iterate(E::TruncEigen, ::Val{:done}) = nothing
 
-function eigen(A::ITensor{N}, Linds, Rinds; kwargs...) where {N}
+function eigen(A::ITensor, Linds, Rinds; kwargs...)
   @debug_check begin
     if hasqns(A)
       @assert flux(A) == QN()
     end
   end
 
-  NL = length(Linds)
-  NR = length(Rinds)
-  NL != NR && error("Must have equal number of left and right indices")
-  N != NL + NR && error("Number of left and right indices must add up to total number of indices")
+  N   = ndims(A)
+  NL  = length(Linds)
+  NR  = length(Rinds)
+  NL != NR      && error("Must have equal number of left and right indices")
+  N  != NL + NR && error("Number of left and right indices must add up to total number of indices")
 
   ishermitian::Bool = get(kwargs, :ishermitian, false)
 
@@ -236,7 +237,7 @@ function eigen(A::ITensor{N}, Linds, Rinds; kwargs...) where {N}
     end
   end
 
-  return TruncEigen(D, V, Vt, spec, l, r)
+  return TruncEigen{ndims(V)}(D, V, Vt, spec, l, r)
 end
 
 function eigen(A::ITensor; kwargs...)
--- a/src/indexset.jl
+++ b/src/indexset.jl
@@ -1,7 +1,7 @@
 
 # TODO: extend type restriction to `IndexT <: Union{<: Index, <: IndexVal}`
 struct IndexSet{IndexT <: Index}
-  data::Tuple{Vararg{IndexT}}
+  data::Vector{IndexT}
 
   function IndexSet{IndexT}(data) where {IndexT}
     @debug_check begin
@@ -9,55 +9,34 @@ struct IndexSet{IndexT <: Index}
         error("Trying to create IndexSet with collection of indices $data. Indices must be unique.")
       end
     end
-    return new{IndexT}(data)
+    return new{IndexT}([data...])
   end
 
   """
       IndexSet()
 
-  Create a special "empty" IndexSet with data `Tuple{}` and any number of indices.
+  Create a special "empty" `IndexSet` with data `[]` and any number of indices.
 
-  This is used as the IndexSet of an `emptyITensor()`, an ITensor with `NDTensors.Empty` storage and any number of indices.
+  This is used as the `IndexSet` of an `emptyITensor()`, an ITensor with `NDTensors.Empty` storage and any number of indices.
   """
-  IndexSet() = new{Union{}}(())
+  IndexSet() = new{Union{}}([])
 end
 
-#
-# Order value type
-#
-
-# More complicated definition makes Order(Ref(2)[]) faster
 @eval struct Order{N}
-  (OrderT::Type{ <: Order})() = $(Expr(:new, :OrderT))
-end
+      (OrderT::Type{ <: Order})() = $(Expr(:new, :OrderT))
+  end
 
 @doc """
-    Order{N}
-
+   Order{N}
 A value type representing the order of an ITensor.
 """ Order
 
 """
-    Order(N) = Order{N}()
-
+   Order(N) = Order{N}()
 Create an instance of the value type Order representing
 the order of an ITensor.
 """
 Order(N) = Order{N}()
-
-"""
-    IndexSet(::Function, ::Order{N})
-
-Construct an IndexSet of length N from a function that accepts
-an integer between 1:N and returns an Index.
-
-# Examples
-```julia
-IndexSet(n -> Index(1, "i\$n"), Order(4))
-```
-"""
-IndexSet(f::Function, N::Int) =
-  IndexSet(ntuple(f, N))
 
 IndexSet(f::Function, ::Order{N}) where {N} =
   IndexSet(ntuple(f, Val(N)))
@@ -70,8 +49,8 @@ const Indices{N, IndexT} = Union{IndexSet{IndexT},
     IndexSet{IndexT}(inds)
     IndexSet{IndexT}(inds::Index...)
 
-Construct an IndexSet of order N and element type IndexT
-from a collection of indices (any collection that is convertable to a Tuple).
+Construct an `IndexSet` of element type `IndexT`
+from a collection of indices (any collection that is convertable to a `Vector`).
 """
 function IndexSet{IndexT}(inds::Index...) where {IndexT}
   return IndexSet{IndexT}(inds)
@@ -82,18 +61,18 @@ end
     IndexSet(inds::Index...)
 
 Construct an IndexSet from a collection of indices
-(any collection that is convertable to a `Tuple`).
+(any collection that is convertable to a `Vector`).
 """
 IndexSet(inds) = IndexSet{eltype(inds)}(inds)
 
 IndexSet(inds::NTuple{N, IndexT}) where {N, IndexT} =
-  IndexSet{IndexT}(inds)
+    IndexSet{IndexT}(inds)
 
 IndexSet(inds::Index...) = IndexSet(inds)
 
 IndexSet(is::IndexSet) = is
 
-IndexSet(::Tuple{}) = IndexSet{Union{}}(())
+IndexSet(::Tuple{}) = IndexSet{Union{}}([])
 
 """
     convert(::Type{IndexSet}, t)
@@ -160,10 +139,6 @@ function NDTensors._permute(s::T, perm) where {T<:IndexSet}
   return ntuple(i->s[perm[i]], length(s))
 end
 
-Order(is::IndexSet) = Order(length(is))
-
-#Order(::Type{<:IndexSet{N}}) where {N} = Order(N)
-
 # Convert to an Index if there is only one
 # TODO: also define the function `only`
 function Index(is::IndexSet)
@@ -203,7 +178,7 @@ the proper Arrow directions.
 function Base.setindex(is::IndexSet,
                        i::Index,
                        n::Int)
-  return IndexSet(setindex(data(is), i, n))
+  return IndexSet(setindex!(data(is), i, n))
 end
 
 """
@@ -318,6 +293,9 @@ Base.keys(is::IndexSet) = 1:length(is)
 # as IndexSet{4})
 NDTensors.similar_type(::Type{<:IndexSet},
                        ::Val{N}) where {N} = IndexSet
+
+NDTensors.similar_type(::Type{<:Tuple{Vararg{<:Index}}},
+                       ::Type{Val{N}}) where {N} = NTuple{N, Index}
 
 NDTensors.similar_type(::Type{<:IndexSet},
                        ::Type{Val{N}}) where {N} = IndexSet
@@ -485,9 +463,6 @@ function Base.setdiff(f::Function,
   return R
 end
 
-#Base.setdiff(f::Function, is1::IndexSet, iss::IndexSet...) =
-#  setdiff(Order(count(i -> f(i) && all(is -> i ∉ is, iss), is1)), f, is1, iss...)
-
 """
     setdiff(A::IndexSet, Bs::IndexSet...)
 
@@ -496,27 +471,6 @@ the IndexSets `Bs`.
 """
 Base.setdiff(A::IndexSet, Bs::IndexSet...; kwargs...) =
   setdiff(fmatch(; kwargs...), A, Bs...)
-
-Base.setdiff(O::Order, A::IndexSet, Bs::IndexSet...;
-             kwargs...) =
-  setdiff(fmatch(; kwargs...), O, A, Bs...)
-
-"""
-  setdiff(f::Function, ::Order{N}, A::IndexSet, B::IndexSet...)
-
-Output the Vector of Indices in the set difference of `A` and `B`,
-optionally filtering by the function `f`.
-
-Specify the number of output indices with `Order(N)`.
-"""
-function Base.setdiff(f::Function,
-                      ::Order{N},
-                      A::IndexSet{IndexT},
-                      B::IndexSet...) where {N, IndexT}
-  r = mutable_storage(Order{N}, IndexT)
-  setdiff!(f, r, A, B...)
-  return IndexSet{IndexT}(Tuple(r))
-end
 
 function firstsetdiff(f::Function,
                       A::IndexSet,
@@ -546,28 +500,6 @@ function Base.intersect(f::Function, A::IndexSet, B::IndexSet)
   return R
 end
 
-#Base.intersect(f::Function, is1::IndexSet, is2::IndexSet) =
-#  intersect(f, Order(count(i -> f(i) && i ∈ is2, is1)), is1, is2)
-
-mutable_storage(::Type{Order{N}},
-                ::Type{IndexT}) where {N, IndexT <: Index} =
-  MVector{N, IndexT}(undef)
-
-function Base.intersect(f::Function,
-                        ::Order{N},
-                        A::IndexSet{IndexT},
-                        B::IndexSet) where {N, IndexT}
-  R = mutable_storage(Order{N}, IndexT)
-  intersect!(f, R, A, B)
-  return IndexSet{IndexT}(Tuple(R))
-end
-
-Base.intersect(O::Order,
-               A::IndexSet,
-               B::IndexSet;
-               kwargs...) where {N} =
-  intersect(fmatch(; kwargs...), O, A, B)
-
 function Base.intersect!(f::Function,
                          R::AbstractVector,
                          A::IndexSet,
@@ -590,15 +522,9 @@ end
 
     intersect(f::Function, A::IndexSet, B::IndexSet)
 
-    intersect(:Order{N}, A::IndexSet, B::IndexSet; kwargs...)
-
-    intersect(f::Function, ::Order{N}, A::IndexSet, B::IndexSet)
-
 Output the Vector of Indices in the intersection of `A` and `B`,
 optionally filtering with keyword arguments `tags`, `plev`, etc. 
 or by a function `f(::Index) -> Bool`.
-
-Specify the output number of indices `N` with `Order(N)`.
 """
 Base.intersect(A::IndexSet, B::IndexSet; kwargs...) =
   intersect(fmatch(; kwargs...), A, B)
@@ -628,16 +554,11 @@ firstintersect(A::IndexSet, B::IndexSet; kwargs...) =
 """
     filter(f::Function, inds::IndexSet)
 
-    filter(f::Function, ::Order{N}, inds::IndexSet)
-
 Filter the IndexSet by the given function (output a new
 IndexSet with indices `i` for which `f(i)` returns true).
 
 Note that this function is not type stable, since the number
 of output indices is not known at compile time.
-
-To make it type stable, specify the desired order by
-passing an instance of the type `Order`.
 """
 Base.filter(f::Function,
             is::IndexSet) =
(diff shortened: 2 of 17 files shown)


emstoudenmire:
I was thinking of taking a swing at this one too, so much appreciated that you did! I'll look it over but let's definitely see Matt's comments.

src/itensor.jl
```diff
@@ -1277,28 +1266,26 @@ end
...
 -(A::ITensor) = itensor(-tensor(A))
 
-function (A::ITensor{N} + B::ITensor{N}) where {N}
+function (A::ITensor + B::ITensor)
+  if ndims(A) == 0 && ndims(B) > 0
+    return copy(B)
+  end
+  if ndims(B) == 0 && ndims(A) > 0
+    return copy(A)
+  end
```
emstoudenmire, line 1275: Is the above logic to replace the emptyITensor behavior? This may be a good way to go but needs some thought.

emstoudenmire (commented):
Agreed that new behaviors replacing emptyITensor are a key point to discuss. Also important to check whether storage of IndexSet is now what we want, but looks good to me.

src/decomp.jl
```diff
@@ -132,9 +132,9 @@ ITensor factorization type for a truncated eigenvalue
 decomposition, returned by `eigen`.
 """
 struct TruncEigen{N}
```
mtfishman, line 134: We'd want to remove this type parameter as well.

src/itensor.jl
kshyatt: Yes.

src/decomp.jl
kshyatt: And in `TruncSVD` as well presumably.

mtfishman: Yeah, that too.

src/indexset.jl
```diff
@@ -557,11 +511,11 @@ Specify the number of output indices with `Order(N)`.
 """
 function Base.setdiff(f::Function,
                       ::Order{N},
```
mtfishman, line 513: The reason for these functions that could be called with `Order` inputs was to help with type stability when you knew the length of the result. However, with the new design that isn't necessary (other than adding a runtime check for the length of the result, which I don't think is really necessary to have). So versions of functions that accept `Order` probably aren't necessary anymore. I doubt people were using them anyway, but I guess we can deprecate them.

src/indexset.jl
```diff
@@ -1192,8 +1143,8 @@ NDTensors.getindices(is::IndexSet, I...) = getindices(is, I...)
 
 Return a new IndexSet with indices `setdir(is[i], dirs[i])`.
 """
-function setdirs(is::IndexSet{N}, dirs) where {N}
-  return IndexSet(ntuple(i -> setdir(is[i], dirs[i]), Val(N)))
+function setdirs(is::IndexSet, dirs)
+  return IndexSet(ntuple(i -> setdir(is[i], dirs[i]), Val(length(is))))
```
mtfishman, line 1147: Probably just want to replace this with a call to `map` (or maybe broadcast or comprehension, whichever is fastest), once we change to a `Vector{IndexT}` storage.

src/indexset.jl
kshyatt: Sure :)

src/indexset.jl
```diff
@@ -1,25 +1,25 @@
 
 # TODO: extend type restriction to `IndexT <: Union{<: Index, <: IndexVal}`
-struct IndexSet{N, IndexT <: Index, DataT <: Tuple}
-  data::DataT
+struct IndexSet{IndexT <: Index}
+  data::Tuple{Vararg{IndexT}}
```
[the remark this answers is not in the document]
kshyatt: Alright, I'll swap it out

src/itensor.jl
```diff
@@ -67,33 +66,30 @@ NDTensors.Dense{Float64,Array{Float64,1}}
...
-      ITensor{N}(is::IndexSet{N}, st::TensorStorage)
+      ITensor(is::IndexSet, st::TensorStorage)
 
   This is an internal constructor for an ITensor where the ITensor stores a view of the `NDTensors.TensorStorage`.
   """
-  ITensor{N}(is, st::TensorStorage) where {N} = new{N}(st, is)
+  ITensor(is, st::TensorStorage) = new(st, is)
 
-  ITensor{Any}(is, st::Empty) = new{Any}(st, is)
+  ITensor(is, st::Empty) = new(st, is)
```
mtfishman, line 82: This constructor is covered by the one above now is isn't necessary.

The meaning of an ITensor with an order of `Any` was a special case where you wanted an ITensor to an initialize a loop where you are adding onto the ITensor in a loop but didn't know what the order of the ITensor would be:
```julia
A = emptyITensor(Any)
for n in 1:10
  A += X * Y # Some expression where you don't know the order of the result
end
```
I would have to think about the fate of this syntax with the new design. The value `Any` was stored in the type parameter, so the ITensor was of type `ITensor{Any}` and that's how we determined it could be added to an ITensor with any number of indices. This isn't possible with the new design. We could replace this with just:
```julia
A = emptyITensor()
for n in 1:10
  A += X * Y # Some expression where you don't know the order of the result
end
```
and/or
```julia
A = ITensor(0)
for n in 1:10
  A += X * Y # Some expression where you don't know the order of the result
end
```
I think we were hesitant about those versions since it seemed like a slippery slope to allow adding a scalar ITensor to a nonscalar ITensor (it would imply that we should also allow `ITensor(2) + ITensor(i, j)` which was deemed unsavory). @emstoudenmire, any opinion on this? I think the version `A = emptyITensor()` above is pretty unambiguous so we could just use that for this use case.

src/itensor.jl
```diff
@@ -319,7 +311,7 @@ emptyITensor() = emptyITensor(Float64)
 Construct an ITensor with empty storage and `Any` number of indices.
 """
 function emptyITensor(::Type{ElT}, ::Type{Any}) where {ElT <: Number}
```
mtfishman, line 313: We'll probably remove this constructor (see https://github.com/ITensor/ITensors.jl/pull/591/files#r594604222).

src/itensor.jl
mtfishman: It looks like it is to replace the `ITensor{Any}` behavior as I discussed in this comment: https://github.com/ITensor/ITensors.jl/pull/591/files#r594604222. Agreed that we need to think about this. For one, this seems like it would be incorrect if you did `ITensor(2) + ITensor(i, j)`, for example. For now we may only want this behavior if `ndims(A) == 0 && isempty(A)` (where by `isempty(A)` I mean that `store(A) isa Empty`, I'm not sure if that is the current definition of `isempty(::ITensor)`).

src/itensor.jl
```diff
@@ -1277,28 +1266,26 @@ end
...
+  end
+  ndims(A) != ndims(B) && throw(DimensionMismatch("cannot add ITensors with different numbers of indices"))
   C = copy(A)
   C .+= B
   return C
 end
 
-function (A::ITensor{N} - B::ITensor{N}) where {N}
+function (A::ITensor - B::ITensor)
+  ndims(A) != ndims(B) && throw(DimensionMismatch("cannot subtract ITensors with different numbers of indices"))
```
mtfishman, line 1283: We probably want `::ITensor - ::ITensor` to match the behavior of `::ITensor + ::ITensor` in terms of adding to empty order-0 ITensors (or whatever behavior we decide for that as discussed above).

kshyatt:
Thanks for the detailed review so far, @mtfishman! I'll address the comments as I find time, should be soon :)

src/itensor.jl
emstoudenmire: Yes, my vote is that `emptyITensor()` would now replace `emptyITensor(Any)` as the way to get the behavior where you can add any other ITensor to it and it will act as a "universal additive zero". Also we could consider allowing setting the elements of an empty ITensor, and perhaps make the constructor `ITensor()` do the same thing as `emptyITensor()` but those are later considerations.

We definitely don't want `ITensor(x)` for some scalar x to act as an `emptyITensor()`, I think even including the case x=0.

src/itensor.jl
emstoudenmire: Yes

src/itensor.jl
mtfishman: Sounds good. Agreed that we can consider having `ITensor()` act like `emptyITensor()` in the future.

src/itensor.jl
```diff
@@ -221,7 +213,7 @@ itensor(T::Tensor) = itensor(store(T), inds(T))
 Convert the ITensor to a Tensor that shares the same
 storage and indices as the ITensor.
 """
-tensor(A::ITensor) = tensor(store(A),inds(A))
+tensor(A::ITensor) = tensor(store(A), inds(A))
```
mtfishman, line 216: So when we change the storage of `IndexSet` to `Vector`, this conversion becomes subtle (it's something that kept bothering me when I was thinking about this design). There is a tension between `NDTensors` and `ITensors`, where the `IndexSet` acts more like a `Tuple` within `NDTensors` (most of the operations are supposed to be type stable with respect to the tensor order in `NDTensors`) while in `ITensors` the `IndexSet` acts more like an ordered set.

In this new design, if we don't do any conversion of the `IndexSet` here, we are basically passing a `Vector` (i.e. a non-statically sized container) to `NDTensors`. I'm not sure what implications this has for performance within `NDTensors`, since I designed `NDTensors` to be generic with respect to the storage of the indices but had in mind that it would be statically sized, so likely I use that assumption in many places and using a non-statically sized container either doesn't work or leads to type unstable code.

The "simple" solution is to convert the indices to a `Tuple` here, i.e. ` tensor(A::ITensor) = tensor(store(A), Tuple(inds(A)))`, and then the conversion back to an `ITensor` from a `Tensor` would have to convert back to a `Vector`/`IndexSet`. However, this can have a decent amount of overhead if done naively (at least in the `Vector -> Tuple` direction, the `Tuple -> Vector` direction is fast):
```julia
julia> @btime Tuple($([prime(i, n) for n in 0:4]));
  256.702 ns (7 allocations: 1.69 KiB)

julia> @btime Tuple($([prime(i, n) for n in 0:12]));
  569.027 ns (15 allocations: 4.20 KiB)

julia> @btime Tuple($([prime(i, n) for n in 0:16]));
  734.744 ns (19 allocations: 5.61 KiB)

julia> @btime collect($(Tuple([prime(i, n) for n in 0:16])));
  117.304 ns (1 allocation: 1.98 KiB)
```

Interestingly, I can make a faster `Tuple` constructor like this:
```julia
@eval struct Val2{x}
  (f::Type{<:Val2})() = $(Expr(:new, :f))
end

_NTuple(::Val2{N}, v::Vector{T}) where {N, T} = ntuple(n -> v[n], Val(N))
_Tuple(v::Vector{T}) where {T} = _NTuple(Val2{length(v)}(), v)
```
using a faster `Val` as discussed [here](https://github.com/JuliaLang/julia/issues/36384) and I get the following:
```julia
julia> VERSION
v"1.5.3"

julia> @btime _Tuple($([prime(i, n) for n in 0:16]));
  243.172 ns (1 allocation: 1.77 KiB)
```
so we can make this conversion pretty fast. Ultimately I think converting to a `Tuple` before passing it to `NDTensors` would be the right way to go here.

Also note that the performance without the `Val2` hack is better in Julia v1.6:
```julia
julia> VERSION
v"1.6.0-rc2"

julia> @btime Tuple($([prime(i, n) for n in 0:16]));
  585.672 ns (18 allocations: 3.62 KiB)

julia> _NTuple(::Val{N}, v::Vector{T}) where {N, T} = ntuple(n -> v[n], Val(N));

julia> _Tuple(v::Vector{T}) where {T} = _NTuple(Val(length(v)), v);

julia> @btime _Tuple($([prime(i, n) for n in 0:16]));
  308.824 ns (1 allocation: 1.77 KiB)
```
but still better with the `Val2` trick:
```julia
julia> _NTuple(::Val2{N}, v::Vector{T}) where {N, T} = ntuple(n -> v[n], Val(N));

julia> _Tuple(v::Vector{T}) where {T} = _NTuple(Val2{length(v)}(), v);

julia> @btime _Tuple($([prime(i, n) for n in 0:16]));
  222.427 ns (1 allocation: 1.77 KiB)
```

So TLDR: I think we should convert the `IndexSet` to a `Tuple` here, using the specialized `_Tuple` function I wrote above that makes use of `Val2` (or better yet `Order`, which is defined like `Val2` and so is fast when being constructed from runtime values).

mtfishman: It can be easy to lose context about what these timings mean, so for some reference here are benchmarks for small matrix multiplications performed with Julia Arrays and ITensors:
```julia
julia> n = 10; A = randn(n, n); @btime $A * $A;
  281.037 ns (1 allocation: 896 bytes)

julia> n = 20; A = randn(n, n); @btime $A * $A;
  733.406 ns (1 allocation: 3.25 KiB)

julia> n = 30; A = randn(n, n); @btime $A * $A;
  1.794 μs (1 allocation: 7.19 KiB)

julia> n = 40; A = randn(n, n); @btime $A * $A;
  2.958 μs (1 allocation: 12.63 KiB)

julia> n = 10; i = Index(n); A = randomITensor(i', i); @btime $(A') * $A;
  830.237 ns (22 allocations: 2.92 KiB)

julia> n = 20; i = Index(n); A = randomITensor(i', i); @btime $(A') * $A;
  1.364 μs (22 allocations: 5.30 KiB)

julia> n = 30; i = Index(n); A = randomITensor(i', i); @btime $(A') * $A;
  2.480 μs (22 allocations: 9.23 KiB)

julia> n = 40; i = Index(n); A = randomITensor(i', i); @btime $(A') * $A;
  3.235 μs (22 allocations: 14.67 KiB)
```
My goal is always to keep the gap between these down and making sure the conversions between ITensors and Tensors is fast is an important part of that (in fact you can see it can easily dominate the cost of small tensor contractions).

kshyatt: Thanks, this is very helpful.

mtfishman: Another reference benchmark:
```julia
julia> julia> VERSION
v"1.5.3"

julia> using ITensors.NDTensors

julia> i = Index(2, "i");

julia> A = randomITensor(i', i);

julia> @btime tensor($A);
  21.026 ns (1 allocation: 224 bytes)

julia> @btime itensor($(tensor(A)));
  21.543 ns (3 allocations: 272 bytes)
```
so conversion between ITensors and Tensors is very fast right now. Unfortunately with this new design (converting `IndexSet -> Tuple`) this conversion will slow down a bit, but I think the bigger picture is that it will make a lot of `IndexSet` operations faster and more type stable and our focus is more on larger tensor contractions.

kshyatt: Agreed. Anecdotally in code I've worked on I've been slowed down *a lot* by the first run of `contract` or `norm` or even `ndims` because of type inference issues.

mtfishman: I'm surprised by `ndims`, but yes any code that involves contractions and decompositions has very slow compilation right now. I started analyzing it with SnoopCompile and found that the NDTensors `contract`, `svd` and related functions take up a ton of inference time. I think the key things are to make sure the code in NDTensors is type stable and make better precompile statements for those functions with many combinations of tensor orders and storage types.

emstoudenmire:
I just did some timings of the JIT overhead of the master branch versus this branch. Running the 1d_heisenberg.jl DMRG example code, I find that from start to finish (all of the JIT time plus all five sweeps) the master branch takes 2min 20s while this PR branch takes 1min 20s. So a whole minute faster! Of course results will vary depending on the sample code and machine used.

src/itensor.jl
kshyatt: 💯 for SnoopCompile, great package.

mtfishman: Going down a bit of a rabbit hole, but here is an even faster Tuple constructor based on the caching strategy explained [here](https://github.com/JuliaLang/julia/issues/29887):
```julia
const ValCacheLength = 100
const ValCache = [Val(n) for n in 1:ValCacheLength]
_NTuple(::Val{N}, v::Vector{T}) where {N, T} = ntuple(n -> v[n], Val(N))
_Tuple(v::Vector{T}) where {T} = _NTuple(ValCache[length(v)], v)
```
results in:
```julia
julia> @btime _Tuple($([prime(i, n) for n in 0:16]));
  165.160 ns (1 allocation: 1.77 KiB)
```
so this is the fastest version yet. The main downside is that it puts an effective limit on the order of the tensors, but I don't see a downside in making the limit quite large (and an order-100 tensor is already well beyond any we would expect to see).

kshyatt: As long as the limit can be adjusted later (in the unlikely event we do hit such tensors) it seems fine to me.

src/indexset.jl
```diff
@@ -1,64 +1,43 @@
...
+  data::Vector{IndexT}
 
   function IndexSet{IndexT}(data) where {IndexT}
     @debug_check begin
       if !allunique(data)
         error("Trying to create IndexSet with collection of indices $data. Indices must be unique.")
       end
     end
-    return new{IndexT}(data)
+    return new{IndexT}([data...])
```
mtfishman, line 12: I guess this is something to think about. This would copy every time an `IndexSet` gets constructed, i.e. `IndexSet([i, i', i''])` would copy. I'm not sure this is necessary, since the goal is for `IndexSet` to have immutable semantics anyway. I always go back and forth about whether or not constructors should make views or copies, but I think we could think about `IndexSet` as a light wrapper around a `Vector`. Also it will be very common to build an `IndexSet` by first building a `Vector{Index}` so it would be unfortunate if it copied every time...

mtfishman: Presumably you added this in order to handle more generic inputs besides `Vector`? In that case we could have an inner constructor that explicitly accepts `Vector` (without copying) and then an outer constructor that, if the input isn't a `Vector`, converts to `Vector` with `collect`.

src/itensor.jl
mtfishman: I'm confused about where my comments are going (maybe I commented on your branch instead of this PR or something), but this version and the `+` version should also check that in the `ndims(A) == 0` case, you also need that `store(A) isa NDTensors.Empty` in order to get the copying behavior.

kshyatt:
A bunch of these got detached from whatever they're meant to refer to...

mtfishman:
> A bunch of these got detached from whatever they're meant to refer to...

Ugh, pretty annoying. I added notes on which parts of the code the comments were referring to.

kshyatt:
> 
> 
> Why was this definition removed?
> 
> EDIT: referring to removing `IndexSet(f::Function, N::Int)`.

Accidentally 😰

kshyatt:
Ugh, forgot that I'd added a `similar` method to NDTensors. I'll move it over to `ITensors` and check locally that everything works then push.

kshyatt:
Alright, fixed, and added back in the accidentally deleted method.

src/indexset.jl
```diff
@@ -63,108 +60,65 @@ IndexSet(f::Function, ::Order{N}) where {N} =
...
-const Indices{N, IndexT} = Union{IndexSet{N,
-                                          IndexT,
-                                          NTuple{N, IndexT}},
-                                 NTuple{N, IndexT}}
-
-function (IndexSet{N, IndexT}(inds)::IndexSet{N, IndexT, NTuple{N, IndexT}}) where {N, IndexT}
-  data = NTuple{N, IndexT}(inds)
-  return IndexSet{N, IndexT, NTuple{N, IndexT}}(data)
-end
+const Indices{N, IndexT} = Union{IndexSet{IndexT},
```
mtfishman, line 63: This shouldn't need the type parameter `N` anymore. I forget if this type is even being used anywhere...

src/indexset.jl
```diff
@@ -63,108 +60,65 @@ IndexSet(f::Function, ::Order{N}) where {N} =
...
     IndexSet(inds::Vector{<:Index})
 
-Convert a Vector of indices to an IndexSet.
-
-Warning: this is not type stable, since a Vector
-is dynamically sized and an IndexSet is statically sized.
-Consider using the constructor `IndexSet{N}(inds::Vector)`.
+Convert a `Vector` of indices to an `IndexSet`.
 """
 IndexSet(inds::Vector{IndexT}) where {IndexT} = IndexSet(tuple(inds...))
```
mtfishman, line 120: This doesn't need the `tuple` conversion anymore.

kshyatt:
I'll update to address latest comments and fix the benchmarks as well

src/itensor.jl
```diff
@@ -1244,10 +1244,22 @@ end
 
 function (A::ITensor + B::ITensor)
   if ndims(A) == 0 && ndims(B) > 0
-    return copy(B)
+    if store(A) isa NDTensors.Empty
+      return copy(B)
+    else # A is a scalar
+      C = copy(B)
+      C .+= A[]
```
mtfishman, line 1238: We decided that we don't want to allow people to do `ITensor(2) + ITensor(i, j)`, so this case should error (that case should get caught by the `ndims(A) != ndims(B)` check below).

src/itensor.jl
```diff
@@ -1257,10 +1269,22 @@ end
 
 function (A::ITensor - B::ITensor)
   if ndims(A) == 0 && ndims(B) > 0
-    return -copy(B)
+    if store(A) isa NDTensors.Empty
+      return -copy(B)
+    else # A is a scalar
+      C = -copy(B)
+      C .+= A[]
```
mtfishman, line 1263: Same as above, we don't want to allow calls like `ITensor(2) - ITensor(i, j)`.

mtfishman:
Thanks for all the work on this Katie, this is a pretty involved code redesign. It already looks like it simplifies the code a decent amount (and hopefully in the end improves the JIT compilation time). It looks like from the benchmark, small tensor contractions have slowed down. My guess is that it is in the `compute_contraction_labels(inds(A), inds(B))` function call, since that was written assuming that `IndexSet` had static size information available (so uses things like `MVector` and `Tuple`). Once the remaining design issues are hashed out and tests pass, we can merge this and I can make a follow-up PR optimizing things like `compute_contraction_labels` before we include it in a new release.

kshyatt:
Ah alright, I'll remove these then.

kshyatt:
Thanks for detailed review! It seems the DMRH example has sped up a bit. We should also do a follow up with new snoop compile directives for the precompile file.

src/indexset.jl
kshyatt: It is used in `src/itensor.jl`.

kshyatt:
Alright, comments addressed. The default storage type of `ITensor()` was NOT `Empty` so I changed things so `ITensor() = emptyITensor()`. Tests passed locally.

kshyatt:
Or not ... hmmm. I guess we'll need to add a method to write an `emptyITensor` to HDF5?

emstoudenmire:
I'll echo Matt's thanks for the hard work on this and that some optimizations can be done in later PR's. 

Let's not consider the writing to disk of emptyITensor() to be a blocker, partly because the HDF5 writing system needs some attention separately from me anyhow.

kshyatt:
It actually turned out to be an easy fix (it seems), as long as my solution is acceptable as a stopgap.

emstoudenmire:
Good! Yes I think that's a reasonable way to handle it, and I think it's fairly similar to how we handle that case in the C++ version.

kshyatt:
OK, if this is otherwise good to go I think the only thing left to be dealt with is the doc build.

kshyatt:
Which I don't 100% understand, by the way, since it seems like the filtering mechanism isn't working?

emstoudenmire:
Yes, the output is not the most helpful there. What was going on is a minor change in the output of DMRG, so the filter had to be corrected. It was from a different PR where I think I merged it without fixing that doctest! 

But there was a second set of doctest errors too regarding doctests that used the ITensor{N} notation, so those had to be fixed too. Let's see if it passes now! It did locally for me.

kshyatt:
Incredible, thank you! Is this ready to go if the lights turn green?

emstoudenmire:
It has my approval. Lets also get @mtfishman 's final signoff before merging, since he mentioned "remaining design issues" in his last comment.

src/itensor.jl
```diff
@@ -245,7 +235,7 @@ ITensor(is::Indices) = ITensor(Float64, is)
 ITensor(inds::Index...) = ITensor(Float64, IndexSet(inds...))
 
 # To fix ambiguity with QN Index version
-ITensor() = ITensor(Float64, IndexSet())
+ITensor() = emptyITensor()
```
emstoudenmire, line 238: I discussed with Matt and we agreed to hold off on this change for now, just to make users have to be explicit about getting the emptyITensor behavior. This may be changed in a future PR though.

So could you please revert this line to `ITensor() = ITensor(Float64, IndexSet())` ? (Assuming that code still works within this PR.)

kshyatt: It doesn't, that's why I changed it as I explained above.

kshyatt: The problem is that doing this breaks the check in the ITensor addition methods since the storage type of the ITensor is *not* `Empty` if you construct it with previous method.

kshyatt: So, we could either undo this change and find a new way to implement the behaviour for `A::ITensor+B::ITensor` (maybe check there is nothing in the storage?) or keep this and the `store(A) isa NDTensors.Empty` behaviour.

mtfishman: For now, we've decided that we are keeping the behavior of `ITensor()` the same, so we still want it to error as it does now:
```julia
julia> using ITensors

julia> i = Index(2)
(dim=2|id=653)

julia> ITensor() + ITensor(i)
ERROR: cannot add ITensors with different numbers of indices
[...]
```

mtfishman: Currently it is defined as `ITensor() = ITensor(0)`:
```julia
julia> ITensor() == ITensor(0)
true
```
so it should act like `ITensor(0)` for the time being.

kshyatt: OK, this has been restored again. It would have been nice to break out the totally speculative discussion of `emptyITensor` to an issue rather than implying I needed to add it here.

kshyatt:
Miles I should hopefully have kept your commit in even with the force push.

src/itensor.jl
mtfishman: You can also check with us before making changes in the future if you're not sure about the desired design.

emstoudenmire:
Hm, looks like the Doc build is failing again. But it's easy to re-fix now that I know how to do it. Only a few lines to change back!

kshyatt:
Let me fix the HDF5 issue as well. Annoying to catch these when the tests take nearly an hour to run.

emstoudenmire:
Yes I do wish the tests ran more quickly. This PR will actually help with that compared to previous versions of the code! But we could certainly lighten some of the tests which have a combinatorial explosion of cases.

I just pushed a commit which restores the ability of adding emptyITensor() to other tensors, using the code you had in the + and - function previously. The case of ITensor() + ITensor(i,j) errors - as it should - because the storage is not Empty and the dimensions don't match.

Definitely good to have the HDF5 code back. I'll check the last commit to see if anything else we wanted got dropped.

emstoudenmire:
On my end it looks like the HDF5 writing fix for emptyITensor is still there.

kshyatt:
I haven't pushed it yet, one minute, just checking that it doesn't conflict with anything.

Get Jim to buy you a 1000 core CI rig and then `pmap` over the test files 😈

kshyatt:
OK, good to go with a squash and merge if this passes?

mtfishman:
Well that was epic. Thanks again for getting this done!

kshyatt:
Thanks again for the reviews. Look forward to the coming feature film about this PR in theaters near you summer of 2021 😎 .

kshyatt:
I'll try to get ITensorsGPU.jl (which I've been neglecting) harmonized with this as well.

PR #592, opened 2021-03-18
Add Qubit sites and operators

emstoudenmire:
Adds a site tag and SiteType named "Qubit", which is mostly
an alias for "S=1/2" though has different keyword arguments
for its `space` function when using QN conservation.

- Add support for op overload returning matrix
- Update spinhalf.jl to use matrix op overloads
- Create "Qubit" site type and operators Qubit can use S=1/2 operators and vice versa
- Minor edit
- Make sure to pass keyword args
- Updated qubit.jl to define in terms of S=1/2
- Some unit tests


test/mps.jl
```diff
@@ -6,7 +6,7 @@ using Test
 Random.seed!(1234)
 
 include("util.jl")
-include(joinpath(pkgdir(ITensors), "examples", "gate_evolution", "qubit.jl"))
+include(joinpath(pkgdir(ITensors), "examples", "gate_evolution"))
```
mtfishman, line 9: Probably this entire line isn't needed anymore (this tries to `include` a directory, which errors).

emstoudenmire: Yes, thanks. Just fixed that one. Also now just fixing up a few unit tests that use some of the more esoteric Qubit operators like the "randn" one, which I think I didn't bring over because I didn't know it was being used anywhere.

mtfishman: It was just used in testing out the gate evolution functionality, it's not necessary.

A better one to add would be a "RandomUnitary" gate like we have in PastaQ (https://github.com/GTorlai/PastaQ.jl/blob/master/src/circuits/gates.jl#L305-L313) which makes a Haar random unitary of any number of indices. That can be left for a future PR though.

emstoudenmire: Just what I needed to know - thanks!

PR #596, opened 2021-03-18
Add support for using `end` in `setindex!` for ITensors

mtfishman:
Also adds some examples, like MPO gate evolution and an example of using `KrylovKit.linsolve`.


emstoudenmire:
Looks good! A welcome addition. Also just have to say the notation for passing keyword args to `ops` is very handy.

PR #604, opened 2021-03-24
optional argument cutoff should be named. Otherwise, the code doesn't run

kunyuan:
optional argument cutoff should be named. Otherwise, the code doesn't run

### DIFF
--- a/examples/gate_evolution/mpo_gate_evolution.jl
+++ b/examples/gate_evolution/mpo_gate_evolution.jl
@@ -46,7 +46,7 @@ function main(; N = 10, cutoff = 1E-8,
   # for Nsteps steps
   psi = psi0
   for step in 1:Nsteps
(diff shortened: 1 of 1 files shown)


mtfishman:
Thanks. As of Julia v1.5 this automatically gets expanded to `cutoff = cutoff`, but ITensors.jl technically supports versions >= v1.3 so best to use backwards compatible syntax. Are you using a Julia version < v1.5?

kunyuan:
Hi Matt,

    Thanks for your reply. I am using v1.4. As you pointed out, I have to manually add the name.

Best,
Kun

emstoudenmire:
Thanks Kun. This is a good fix because as Matt said we do want to continue supporting Julia 1.4 and even 1.3 for a while. Appreciate you pointing this out!

PR #610, opened 2021-03-30
Replacing hardcoded literals with constants

saolof:
Minor cleanup, so that changing consts at the top of the file actually behaves as one would expect.

### DIFF
--- a/src/tagset.jl
+++ b/src/tagset.jl
@@ -9,14 +9,14 @@ const vmaxTags = Val(maxTags)
 const TagSetStorage{T,N} = SVector{N,T}
 const MTagSetStorage{T,N} = MVector{N,T}  # A mutable tag storage
 
-emptytag(::IntTag) = IntTag(0)
+emptytag(::Type{IntTag}) = IntTag(0)
 
 #TODO: decide which functions on TagSet should be made generic.
 struct GenericTagSet{T,N}
   data::TagSetStorage{T,N}
   length::Int
   function GenericTagSet{T,N}() where {T,N}
-    ts = TagSetStorage(ntuple(_ -> emptytag(T) , N))
+    ts = TagSetStorage(ntuple(_ -> emptytag(T) , Val(N)))
     new(ts, 0)
   end
   GenericTagSet{T,N}(tags::TagSetStorage{T,N}, len::Int) where {T,N} = new(tags, len)
@@ -25,7 +25,7 @@ end
 GenericTagSet(ts::GenericTagSet) = ts
 
 function GenericTagSet{T,N}(t::T) where {T,N}
-  ts = MTagSetStorage(ntuple(_ -> emptytag(T) , N))
+  ts = MTagSetStorage(ntuple(_ -> emptytag(T) , Val(N)))
   ts[1] = IntTag(t)
   return GenericTagSet(TagSetStorage(ts), 1)
 end
@@ -95,7 +95,7 @@ function GenericTagSet{T,N}(str::AbstractString) where {T,N}
   # TODO: refactor the Val here.
   current_tag = MTagStorage(ntuple(_ -> IntChar(0),Val(maxTagLength)))
   # Mutable fixed-size vector as temporary TagSet storage
-  ts = MTagSetStorage(ntuple(_ -> emptytag(T),N))
+  ts = MTagSetStorage(ntuple(_ -> emptytag(T),Val(N)))
   nchar = 0
   ntags = 0
   for current_char in str
--- a/src/tagset.jl
+++ b/src/tagset.jl
@@ -5,26 +5,26 @@ const MTagStorage = MSmallStringStorage # A mutable tag storage
 const IntTag = IntSmallString  # An integer that can be cast to a Tag
 const emptyIntTag = IntTag(0)
 const maxTags = 4
-const vmaxTags = Val(maxTags)
 const TagSetStorage{T,N} = SVector{N,T}
 const MTagSetStorage{T,N} = MVector{N,T}  # A mutable tag storage
 
 emptytag(::Type{IntTag}) = IntTag(0)
-empty_tagsetstorage{T,N}() where {T,N} = TagSetStorage(ntuple(_ -> emptytag(T) , Val(N)))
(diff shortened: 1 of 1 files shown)


src/tagset.jl
```diff
@@ -11,7 +12,7 @@ struct TagSet
   data::TagSetStorage
   length::Int
   function TagSet() 
-    ts = TagSetStorage(ntuple(_ -> IntTag(0), Val(4)))
+    ts = TagSetStorage(ntuple(_ -> emptyIntTag, Val(maxTags)))
```
mtfishman, line 15: So I'm betraying my ignorance of the Julia compiler here, but I just realized that Julia is not able to infer the type of `Val(maxTags)` properly. I believe the previous version would get inferred by constant propagation, but in theory the value of `maxTags` could get changed somewhere else and thus change the value here, so in principle the size wouldn't be inferrable. It would be better to define something like `const ValMaxTags = Val(4)` and use that instead.

I see `Val(maxTags)` is used elsewhere so probably I didn't consider this issue before, it should be change everywhere.

I just checked quickly switching to that design (and also using a corresponding `ValMaxTagLength = Val(8)` and I get the following speedup on `TagSet` construction. Current master branch:
```julia
julia> @btime TagSet($s)
  223.941 ns (10 allocations: 416 bytes)
"abcde,fghij,klmnop,qrstuv"
```
New design:
```julia
julia> @btime TagSet($s)
  103.391 ns (1 allocation: 80 bytes)
"abcde,fghij,klmnop,qrstuv"
```
which is pretty great speedup! I wrote the constructor pretty carefully and I remember wondering where all of those allocations were coming from.

I think there is over all a better design here, which would be defining a general parametrized type like `GenericTagSet{MaxNumTags, MaxTagLength}` and then hardcode `TagSet = TagSet{4, 8}`, and then extract `MaxNumTags` and `MaxTagLength` from the type within functions instead of from global variables. But we can start with the `const Val* = Val(*)` definitions for now.

saolof (commented):
Working on making TagSet generic

saolof (commented):
Allocation for Val(N) may come back. Hopefully there's a nicer solution than using a generated function

saolof (commented):
Fixing a method that didn't seem to specialize.

saolof (commented):
Still trying to figure out why the GenericTagSet constructor method that takes an AbstractString isn't specializing to TagSet properly

src/tagset.jl
```diff
@@ -3,28 +3,33 @@ const Tag = SmallString
 const maxTagLength = smallLength
 const MTagStorage = MSmallStringStorage # A mutable tag storage
 const IntTag = IntSmallString  # An integer that can be cast to a Tag
+const emptyIntTag = IntTag(0)
 const maxTags = 4
-const TagSetStorage = SVector{maxTags,IntTag}
-const MTagSetStorage = MVector{maxTags,IntTag}  # A mutable tag storage
+const vmaxTags = Val(maxTags)
```
mtfishman, line 8: I think you have to literally write this as ` const vmaxTags = Val(4)` so then Julia sees the type parameter as a constant value `4`. Otherwise it has the same global variable/type instability problem as before.

src/tagset.jl
```diff
@@ -3,28 +3,33 @@ const Tag = SmallString
 const maxTagLength = smallLength
 const MTagStorage = MSmallStringStorage # A mutable tag storage
 const IntTag = IntSmallString  # An integer that can be cast to a Tag
+const emptyIntTag = IntTag(0)
```
mtfishman, line 6: Perhaps you are already planning this change, but probably best to remove this definition entirely in favor of ` emptytag(::Type{IntTag}) = IntTag(0)` that you defined below to avoid unnecessary global variables.

saolof: Good call, done.

saolof (commented):
Minor reshuffle due to declaration order

saolof (commented):
Added the missing SmallString constructor that caused the tests to fail earlier. Hopefully now everything should pass.

saolof (commented):
Reverted a change made in the last commit, since it may have caused a type instability. Otherwise, this seems more or less done for now. Scope of this PR was minor cleanup and making the TagSet type & constructors generic, while getting rid of allocations where possible.

mtfishman:
Thanks a lot for your work on this! I like the code patterns like `empty_storage` and `emptytag`, it definitely makes the code cleaner.

Would you mind repeating the benchmark I did from the comment above for your new PR vs. v0.1.41 of ITensors.jl? i.e. `s = "abcde,fghij,klmnop,qrstuv"; @btime TagSet($s)`.

saolof:
> 
> 
> Thanks a lot for your work on this! I like the code patterns like `empty_storage` and `emptytag`, it definitely makes the code cleaner.
> 
> Would you mind repeating the benchmark I did from the comment above for your new PR vs. v0.1.41 of ITensors.jl? i.e. `s = "abcde,fghij,klmnop,qrstuv"; @btime TagSet($s)`.

My version:

```
julia> s = "abcde,fghij,klmnop,qrstuv"; @btime TagSet($s)
  156.500 ns (1 allocation: 80 bytes)
"abcde,fghij,klmnop,qrstuv"

julia> s = "abcde,fghij,klmnop,qrstuv"; @benchmark TagSet($s)
BenchmarkTools.Trial: 
  memory estimate:  80 bytes
  allocs estimate:  1
  --------------
  minimum time:     156.258 ns (0.00% GC)
  median time:      159.975 ns (0.00% GC)
  mean time:        183.772 ns (2.14% GC)
  maximum time:     3.238 μs (89.14% GC)
  --------------
  samples:          10000
  evals/sample:     807
```

Public version:
```

julia> s = "abcde,fghij,klmnop,qrstuv"; @btime TagSet($s)
  357.971 ns (9 allocations: 336 bytes)
"abcde,fghij,klmnop,qrstuv"

julia> s = "abcde,fghij,klmnop,qrstuv"; @benchmark TagSet($s)
BenchmarkTools.Trial: 
  memory estimate:  336 bytes
  allocs estimate:  9
  --------------
  minimum time:     364.115 ns (0.00% GC)
  median time:      390.909 ns (0.00% GC)
  mean time:        484.270 ns (6.06% GC)
  maximum time:     26.401 μs (97.06% GC)
  --------------
  samples:          10000
  evals/sample:     209
```

mtfishman:
Great! I'll merge.

PR #615, opened 2021-04-03
Various bug fixes

### LINKED ISSUE #377: Fix random complex MPS
Hi,
I noticed that `randomMPS(ComplexF64,sites)` (which used to give a random complex prodcut-state MPS) is no longer working because it calls `randomCircuitMPS` also when the bond dimension is 1, which is in turn only implemented for `Float64`. 

Should `randomMPS(complexF64,sites)` be implemented ? (also for the case of bond_dim > 1) 
Is the reason that it's not implemented currently because one has to implement `NDTensors.random_unitary` ?

### LINKED ISSUE #580: Use `onehot` as alternative syntax for `setelt`
I noticed that the terminology "one-hot encoding" used in the context of digital circuits and machine learning [1,2,3] is essentially the same as our `setelt` ITensor constructor. It may be nice to use `onehot` as an alternative syntax, since it may be more familiar to people coming from that background (and I've wanted to find an alternative name for it anyway).

[1] https://en.wikipedia.org/wiki/One-hot
[2] https://www.tensorflow.org/api_docs/python/tf/one_hot
[3] https://fluxml.ai/Flux.jl/stable/data/onehot/

### LINKED ISSUE #602: `ops` is too strict about the type
The `ops` function is too strict about the input type:
```julia
julia> s = siteinds("S=1/2", 2)
2-element Array{Index{Int64},1}:
 (dim=2|id=553|"S=1/2,Site,n=1")
 (dim=2|id=354|"S=1/2,Site,n=2")

julia> ops(s, [("Sz", n) for n in 1:length(s)])
2-element Array{ITensor,1}:
 ITensor ord=2
Dim 1: (dim=2|id=553|"S=1/2,Site,n=1")'
Dim 2: (dim=2|id=553|"S=1/2,Site,n=1")
 2×2
 0.5   0.0
 0.0  -0.5
 ITensor ord=2
Dim 1: (dim=2|id=354|"S=1/2,Site,n=2")'
Dim 2: (dim=2|id=354|"S=1/2,Site,n=2")
 2×2
 0.5   0.0
 0.0  -0.5

julia> ops(s, Tuple[("Sz", n) for n in 1:length(s)])
ERROR: MethodError: no method matching ops(::Array{Index{Int64},1}, ::Array{Tuple,1})
Closest candidates are:
  ops(::Array{var"#s441",1} where var"#s441"<:Index, ::AbstractArray{var"#s440",N} where N where var"#s440"<:Tuple{String,Vararg{Any,N} where N}) at /home/mfishman/.julia/dev/ITensors/src/physics/sitetype.jl:368
Stacktrace:
 [1] top-level scope at ./REPL[35]:1
```

### LINKED ISSUE #614: Incorrect definition of Adagdn for Electron sites

emstoudenmire:
- Fix definition of Adagdn for Electron. Fixes #614
- Make ops less strict about input. Fixes #602
- Add onehot as new name for setelt. Fixes #580
- Support randomMPS(ComplexF64,s,chi). Fixes #377
- Lighten up unit test workload

### DIFF
--- a/docs/src/ITensorType.md
+++ b/docs/src/ITensorType.md
@@ -14,7 +14,7 @@ ITensor(::Type{<:Number}, ::UndefInitializer, ::ITensors.Indices)
 ITensor(::Number, ::ITensors.Indices)
 ITensor(::Array{<:AbstractFloat},::ITensors.Indices)
 randomITensor(::Type{<:Number}, ::ITensors.Indices)
-setelt(::IndexVal)
+onehot(::IndexVal)
 ```
 
 ## Dense View Constructors
--- a/src/exports.jl
+++ b/src/exports.jl
@@ -144,6 +144,7 @@ export
   nzblocks,
   nzblock,
   nnz,
+  onehot,
   order,
   permute,
   prime!,
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -484,6 +484,7 @@ delta(is::Index...) = delta(Float64, IndexSet(is...))
 const δ = delta
 
 """
+    onehot(ivs...)
     setelt(ivs...)
 
 Create an ITensor with all zeros except the specified value,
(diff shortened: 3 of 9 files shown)


mtfishman:
Thanks! All good things to fix.

PR #616, opened 2021-04-04
Undef randomITensor

### LINKED ISSUE #399: Use undef in randomITensor
Make sure `randomITensor` is as fast as possible. I think if we just use the `undef` version of the ITensor constructor here:
https://github.com/ITensor/ITensors.jl/blob/b53cacb107a673de91630d12eefa7027f632edcf/src/itensor.jl#L909
it will basically be how Julia defines things like `randn`:
https://github.com/JuliaLang/julia/blob/b1f51df1088b2ab4e1c954537fd8c22b9b5f19ac/stdlib/Random/src/normal.jl#L201
and `rand`:
https://github.com/JuliaLang/julia/blob/b1f51df1088b2ab4e1c954537fd8c22b9b5f19ac/stdlib/Random/src/Random.jl#L288
(i.e. by initializing `undef` data and using `randn!` and `rand!`).

emstoudenmire:
- Change dense randomITensor to use undef constructor
- Add undef QN ITensor constructor
- Change QN randomITensor to use undef constructor
- Improve docs and document undef QN ITensor constructor
- Add docs for Sweeps objects

Fixes #399

### DIFF
--- a/src/qn/qnitensor.jl
+++ b/src/qn/qnitensor.jl
@@ -6,6 +6,25 @@
 Construct an ITensor with BlockSparse storage filled with `zero(ElT)` where the nonzero blocks are determined by `flux`.
 
 If `ElT` is not specified it defaults to `Float64`.
+
+# Examples
+
+```julia
+i = Index([QN(0)=>1, QN(1)=>2], "i")
+
+# QN ITensors with flux of QN(0):
+
+A = ITensor(i',dag(i))
+B = ITensor(QN(0),i',dag(i))
+
+# QN ITensor with flux of QN(1):
+
+C = ITensor(QN(1),i',dag(i))
+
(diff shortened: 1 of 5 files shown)


src/qn/qnitensor.jl
```diff
@@ -32,6 +51,22 @@ ITensor(::Type{ElT}, inds::QNIndex...) where {ElT<:Number} =
 
 ITensor(inds::QNIndex...) = ITensor(Float64, QN(), IndexSet(inds...))
 
+"""
+    ITensor([::Type{ElT} = Float64, ]::UndefInitializer, inds)
```
mtfishman, line 55: Looks like you forgot the `flux::QN` input here.

emstoudenmire: Good catch! Just fixed this in another PR today.

PR #620, opened 2021-04-06
ITensor constructors

### LINKED ISSUE #612: Do not convert element types to float
```julia
julia> ITensor(TropicalF64.(randn(3,3)), i, j)
ERROR: DimensionMismatch("In ITensor(::Array, ::IndexSet), length of Array (9) must match total dimension of IndexSet (1000000)")
Stacktrace:
 [1] itensor(A::Matrix{TropicalF64}, inds::IndexSet{Index{Int64}})
   @ ITensors ~/.julia/dev/ITensors/src/itensor.jl:363
 [2] ITensor(A::Matrix{TropicalF64}, inds::IndexSet{Index{Int64}})
   @ ITensors ~/.julia/dev/ITensors/src/itensor.jl:394
 [3] ITensor(::Matrix{TropicalF64}, ::Index{Int64}, ::Vararg{Index{Int64}, N} where N)
   @ ITensors ~/.julia/dev/ITensors/src/itensor.jl:396
 [4] top-level scope
   @ REPL[1]:1
```

The relevant code

https://github.com/ITensor/ITensors.jl/blob/ae84b80ef55271dca1aa39dfcd4db350c4e864e1/src/itensor.jl#L364

Wondering if it is plausible to remove this element type conversion?

mtfishman:
This fixes #612. It doesn't remove the floating point conversion entirely, but only does it for `Int` and `Complex{Int}`, i.e. `itensor([0 1; 1 0], i, j)` and `itensor([0 1+2im; 1 0], i, j)` still convert to floating point. In addition, it adds an API where the element type can be specified explicitly with `itensor(Int, [0 1; 1 0], i, j)` in the rare cases when integer types are desired.

I'm open to removing the automatic conversion entirely, since at some point (through packages like https://github.com/JuliaLinearAlgebra/Octavian.jl) we should be able to have fast tensor contractions/matrix multiplications over generic element types. In addition, my guess is that a vast majority of the time people are making relatively small ITensors from array data (for example the common case is `op`) in which case it's not a big deal to convert to float before the BLAS call. In addition, I think in cases like `op` it is best practice to define it as a floating point directly with something like `itensor([0. 1.; 1. 0.], i, j)` or `itensor(Float64[0 1; 1 0], i, j)`, which makes the construction faster. In fact we should be doing this internally in our Qubit `op` definitions, for example:
```julia
julia> f() = [0 1; 1 0]

julia> g() = Float64[0 1; 1 0]
g (generic function with 1 method)

julia> h() = float([0 1; 1 0])
h (generic function with 1 method)

julia> @btime f()
  83.924 ns (2 allocations: 160 bytes)
2×2 Matrix{Int64}:
 0  1
 1  0

julia> @btime g()
  26.859 ns (1 allocation: 112 bytes)
2×2 Matrix{Float64}:
 0.0  1.0
 1.0  0.0

julia> @btime h()
  118.722 ns (3 allocations: 272 bytes)
2×2 Matrix{Float64}:
 0.0  1.0
 1.0  0.0
```
The version with the element type specified is also nice because it allows specifying a different element type as well.

I've added warnings in ITensor constructors saying that the conversion behavior may change in the future and shouldn't be relied on.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@ ITensor v0.2.0 Release Notes
 - Contraction sequence optimization (PR #589) (@mtfishman).
 - Make TagSet code cleaner and more generic, and improve constructor from String performance (PR #610) (@saolof).
 - Define some missing methods for AbstractMPS broadcasting (#609) (@kshyatt).
-- ITensor constructors from Array now only convert to floating point for `Array{Int}` and `Array{Complex{Int}}` (IS #612).
+- ITensor constructors from Array now only convert to floating point for `Array{Int}` and `Array{Complex{Int}}`. That same conversion is added for QN ITensor constructors to be consistent with non-QN versions (IS #612).
 - New ITensor constructors like `itensor(Int, [0 1; 1 0], i, j)` to specify the exact element type desired (IS #612).
 - Speed up randomITensor with undef constructor (PR #616) (@emstoudenmire).
 - Fix definition of Adagdn for Electron (PR #615) (@emstoudenmire).
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,8 +5,8 @@ ITensor v0.2.0 Release Notes
 - Contraction sequence optimization (PR #589) (@mtfishman).
 - Make TagSet code cleaner and more generic, and improve constructor from String performance (PR #610) (@saolof).
 - Define some missing methods for AbstractMPS broadcasting (#609) (@kshyatt).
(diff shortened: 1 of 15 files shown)


mtfishman:
Also closes #620.

emstoudenmire:
I agree with the careful approach here. We can't expect our users to always have to put [0. 1; 1 0] to get the behavior most of them would expect (i.e. the resulting ITensor having float storage). So if we do begin to have ITensor([0 1; 1 0],...) result in an Int-valued ITensor in the future, it should be after we are sufficiently convinced that the code can be made fully generic where these tensors can interoperate with float-valued ones. But I'm reasonably confident that in Julia that generic of code is possible! Some good things could come of such a generic move, like better support for lower-precision float types which can give big speedups, especially on GPUs.

PR #622, opened 2021-04-08
Correlator function

emstoudenmire:
Adds a function `correlator(psi::MPS,Op1,Op2;start_site=i,end_site=j)`
that returns a matrix C such that C[i,j]=<psi| Op1i Op2j |psi>.
Includes handling of fermionic operators, using `has_fermion_string` to
detect whether to put in Jordan-Wigner F string operators.

- Add correlator(MPS,A,B) measuring <A_i B_j>
- Unit tests for correlator function
- Add optional start_site, end_site args. Add docs.

### DIFF
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -552,6 +552,8 @@ function correlation_matrix(psi::MPS,
 
   psi = copy(psi)
   orthogonalize!(psi,start_site)
+  psi[start_site] ./= norm(psi[start_site])
+
   s = siteinds(psi)
   onsiteOp = "$Op1*$Op2"
   fermionic2 = has_fermion_string(Op2,s[1])
@@ -646,7 +648,11 @@ function expect(psi::MPS,
 
   site_range::UnitRange{Int} = get(kwargs,:site_range,1:N)
   Ns = length(site_range)
-  offset = first(site_range)-1
+  start_site = first(site_range)
+  offset = start_site-1
+
+  orthogonalize!(psi,start_site)
+  psi[start_site] ./= norm(psi[start_site])
 
   ex = ntuple(n->zeros(ElT,Ns),Nops)
   for j=site_range
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -507,6 +507,85 @@ function sample(m::MPS)
   return result
 end
 
+"""
+    correlator(M::MPS,
+               Op1::AbstractString,
+               Op2::AbstractString;
+               kwargs...)
+
+Given an MPS psi and two strings denoting
+operators (as recognized by the `op` function), 
+computes the two-point correlation function 
+C[i,j] = <psi|Op1_i Op2_j|psi>
+using efficient MPS techniques.
+Returns the matrix C.
+
+For an MPS of length N and typical
+bond dimension m, the scaling
+of this algorithm is N^2*m^3.
+
+# Examples
+```julia
+N = 30
+m = 4
+
+s = siteinds("S=1/2",N)
+psi = randomMPS(s,m)
+Czz = correlator(psi,"Sz","Sz")
+
+s = siteinds("Electron",N; conserve_qns=true)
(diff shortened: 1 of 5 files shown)


mtfishman:
Looks good, thanks!

What if we used a single keyword argument `site_range` with inputs like `site_range = 2:8`? Also, what about naming it `correlation_matrix`? So:
```julia
Cuu = correlation_matrix(psi, "Cdagup", "Cup"; site_range = 2:8)
```
When I first read your description and the docs, I thought that `correlator(psi::MPS,Op1,Op2;start_site=i,end_site=j)` might mean that you are only measuring a single correlator `<psi| Op1i Op2j |psi>` (not the entire matrix within the range of sites `i:j`).

emstoudenmire:
Good notes - I agree with those changes.

I also added a new function `expect` which computes the expected values of one or more operators. Please let me know if it looks ok to you. Maybe the only change to it I can think of would be to also offer `site_range` for it?

mtfishman:
Yeah, `site_range` would also be nice for `expect` as well.

src/mps/mps.jl
```diff
@@ -507,6 +507,155 @@ function sample(m::MPS)
...
+  N = length(psi)
+  site_range::UnitRange{Int} = get(kwargs,:site_range,1:N)
+
+  Ns = length(site_range)
+  offset = first(site_range)-1
+
+  psi = copy(psi)
+  Nops = length(ops)
+  s = siteinds(psi)
+  ex = ntuple(n->zeros(Ns),Nops)
```
mtfishman, line 649: I just realized that this assumes the result will be real. I suppose we are limiting this to Hermitian operators (otherwise an expectation value isn't defined)? Also this will probably error in general for complex cases since the result won't be exactly real, i.e.:
```julia
julia> x = zeros(4)
4-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0

julia> x[1] = 3.0 + 0.0im
3.0 + 0.0im

julia> x[1] = 3.0 + eps()im
ERROR: InexactError: Float64(3.0 + 2.220446049250313e-16im)
Stacktrace:
 [1] Real
   @ ./complex.jl:37 [inlined]
 [2] convert
   @ ./number.jl:7 [inlined]
 [3] setindex!(A::Vector{Float64}, x::ComplexF64, i1::Int64)
   @ Base ./array.jl:839
 [4] top-level scope
   @ REPL[9]:1
```
For now we could take the real part and check that the imaginary part is small, and limit to Hermitian operators.

This reminds me that we should also start to be careful about preserving element types by using generic code patterns like `zeros(real(eltype(psi)), Ns)`, in case people are using things like `Float32`, `BigFloat` or `Tropical` numbers, but maybe that is a story for another day.

emstoudenmire: Good catch! I will try it out with some complex random MPS to make sure the final version works.

Yes, I do think it's reasonable to limit to Hermitian operators, so we can just take the real part.

I'll do some of the generic element type code too, good suggestion, but we will probably have to really test it out in the future to be sure all our code is fully generic.

emstoudenmire:
The `correlation_matrix` function was also not handling complex MPS correctly, so I just fixed that in addition to making `expect` handle that case in a generic way.

src/mps/mps.jl
```diff
@@ -507,6 +507,158 @@ function sample(m::MPS)
...
+  onsiteOp = "$Op1*$Op2"
+  fermionic2 = has_fermion_string(Op2,s[1])
+  if fermionic2
+    Op1 = "$Op1*F"
+  end
+
+  # Nb = size of block of correlation matrix
+  Nb = end_site-start_site+1
+
+  C = zeros(ElT,Nb,Nb)
```
mtfishman, line 565: I think here we want to use `real(ElT)`.

emstoudenmire:
Just realized that if the input MPS isn't normalized, we'd want to normalize it, since expectation values are defined that way i.e. <psi|A|psi>/<psi|psi>. Similar for the correlation_matrix function.

Issue #513, opened 2020-10-19
Allow more general inputs in Sweep functions like `noise!`


mtfishman:
This kind of input currently errors:
```julia
sweeps = Sweeps(8)
noise!(sweeps, 1e-8, 0)
```
with:
```julia
ERROR: MethodError: no method matching noise!(::Sweeps, ::Float64, ::Int64)
[...]
```
We should allow any real number type, and then convert to `Float64`.

Also, I would consider changing the names to things like `setnoise!`.

PR #624, opened 2021-04-09
Simplify implementation of Sweeps setting functions

### LINKED ISSUE #474: Simplify maxdim!, mindim!, etc. implementation with get
In places like this:

https://github.com/ITensor/ITensors.jl/blob/2c4a59fe18d3f1f0d1e5e5b905c75c3a3060d6cf/src/mps/sweeps.jl#L135

we could simplify the implementation with the `get` function, for example:
```julia
for i in 1:nsweep(sw)
  sw.maxdim[i] = get(maxdims, i, maxdims[end])
end
```
The last value is a default if the index `i` is out of range.

We could even just not store all of the sweep values in the field `maxdim`, and define `getindex` through `get`, but that is probably not necessary.

emstoudenmire:
Use `get` and `end` to shorten code for setting parameters of Sweeps objects. Could be even simpler if `get` was defined for tuples - best I could do was to use `collect` to convert the argument tuple into an array.

Fixes #474

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -330,8 +330,8 @@ let
   # for each sweep and maximum truncation cutoff
   # used when adapting internal dimensions:
   sweeps = Sweeps(5)
-  maxdim!(sweeps, 10,20,100,100,200)
-  cutoff!(sweeps, 1E-10)
+  setmaxdim!(sweeps, 10,20,100,100,200)
+  setcutoff!(sweeps, 1E-10)
   @show sweeps
 
   # Run the DMRG algorithm, returning energy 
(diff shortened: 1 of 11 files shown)


mtfishman:
While you're at it could you also fix https://github.com/ITensor/ITensors.jl/issues/513? Should just be a matter of removing the type signatures. I also think you may not need to `collect` here since if you just use the input it will be a tuple, which should be fine to work with.

mtfishman:
Ah I see, `get` isn't defined for `Tuple`, never mind that comment.

emstoudenmire:
Good suggestion to include #513 too. I implemented the alternate `setcutoff!` and `setnoise!` names etc. Those will be easier for new users to understand and is more "self documenting".

mtfishman:
Thanks! Yeah, I want to keep a general convention that if you are setting a field of a struct then the function is of the form `set*!(...)`, like `setstorage!` and `setinds!`.

emstoudenmire:
I just updated the DMRG example codes and the README to use `setmaxdim!`, `setcutoff!` etc as well.

emstoudenmire:
Fixes #513

PR #626, opened 2021-04-09
Account for the new Vector storage of `IndexSet`

mtfishman:
This PR accounts for the new vector storage of `IndexSet`, improving the performance of things like small matrix multiplications.

Tests mostly pass but certain operations like `setindex!(::ITensor, ...)` are still slow.


mtfishman:
Here are some benchmarks:
```julia
using ITensors
using ITensors.NDTensors

i = Index(2)
A = randomITensor(i', i)

function tensor_iterate!(A)
  for I in CartesianIndices(A)
    A[I] = 1 * A[I]
  end
end

println("prime(A)")
@btime prime($A)
println("uniqueinds")
@btime uniqueinds($(A'), $A)
println("commoninds")
@btime commoninds($(A'), $A)
println("unioninds")
@btime unioninds($(A'), $A)
println("noncommoninds")
@btime noncommoninds($(A'), $A)
println("A' * A")
@btime $(A') * $A
println("A[1, 1]")
@btime $A[1, 1]
println("A[i' => 1, i => 1]")
@btime $A[$(i') => 1, $i => 1]
println("A[1, 1] = 1")
@btime $A[1, 1] = 1.0
println("A[i' => 1, i => 1] = 1")
@btime $A[$(i') => 1, $i => 1] = 1.0

j = Index(100)
B = randomITensor(j', j)

println("Iterate ITensor")
@btime tensor_iterate!($B)
println("Iterate Tensor")
@btime tensor_iterate!($(tensor(B)))
println("Iterate Array")
@btime tensor_iterate!($(Matrix(B, j', j)))
```
On this branch:
```julia
prime(A)
  130.749 ns (3 allocations: 464 bytes)
uniqueinds
  161.254 ns (4 allocations: 928 bytes)
commoninds
  171.174 ns (5 allocations: 1.14 KiB)
unioninds
  264.109 ns (6 allocations: 1.78 KiB)
noncommoninds
  208.292 ns (5 allocations: 1.23 KiB)
A' * A
  1.144 μs (23 allocations: 3.11 KiB)
A[1, 1]
  15.791 ns (1 allocation: 16 bytes)
A[i' => 1, i => 1]
  41.888 ns (3 allocations: 272 bytes)
A[1, 1] = 1
  44.167 ns (1 allocation: 224 bytes)
A[i' => 1, i => 1] = 1
  62.184 ns (3 allocations: 480 bytes)
Iterate ITensor
  1.582 ms (70000 allocations: 3.66 MiB)
Iterate Tensor
  5.683 μs (0 allocations: 0 bytes)
Iterate Array
  3.309 μs (0 allocations: 0 bytes)
```
On the latest registered version ITensors v0.1.41:
```julia
prime(A)
  61.680 ns (3 allocations: 480 bytes)
uniqueinds
  852.565 ns (11 allocations: 1.81 KiB)
commoninds
  879.392 ns (11 allocations: 1.81 KiB)
unioninds
  1.357 μs (17 allocations: 4.50 KiB)
noncommoninds
  1.705 μs (20 allocations: 5.48 KiB)
A' * A
  1.160 μs (29 allocations: 3.69 KiB)
A[1, 1]
  237.139 ns (4 allocations: 496 bytes)
A[i' => 1, i => 1]
  392.730 ns (8 allocations: 816 bytes)
A[1, 1] = 1
  413.990 ns (9 allocations: 1008 bytes)
A[i' => 1, i => 1] = 1
  579.016 ns (13 allocations: 1.30 KiB)
Iterate ITensor
  7.931 ms (170000 allocations: 15.56 MiB)
Iterate Tensor
  7.541 μs (0 allocations: 0 bytes)
Iterate Array
  3.296 μs (0 allocations: 0 bytes)
```
and on master:
```julia
prime(A)
  67.205 ns (4 allocations: 352 bytes)
uniqueinds
  696.938 ns (12 allocations: 1.09 KiB)
commoninds
  496.969 ns (9 allocations: 704 bytes)
unioninds
  1.165 μs (18 allocations: 3.77 KiB)
noncommoninds
  1.532 μs (21 allocations: 3.53 KiB)
A' * A
  4.638 μs (54 allocations: 5.45 KiB)
A[1, 1]
  1.159 μs (15 allocations: 2.17 KiB)
A[i' => 1, i => 1]
  1.602 μs (20 allocations: 2.70 KiB)
A[1, 1] = 1
  1.429 μs (21 allocations: 2.72 KiB)
A[i' => 1, i => 1] = 1
  1.781 μs (27 allocations: 3.27 KiB)
Iterate ITensor
  24.962 ms (400002 allocations: 48.98 MiB)
Iterate Tensor
  5.128 μs (0 allocations: 0 bytes)
Iterate Array
  3.320 μs (0 allocations: 0 bytes)
```

The key takeaway is that all of the operations are now the same speed or much faster than the ones on  v0.1.41 or master (except for priming, which is a bit slower as discussed below).

The set functions are mostly faster because I've added optimized implementations that are fast for smallish sets (basically any that we are reasonably expected to see). Either they improve on previous custom implementations, or avoid calling the generic Julia set functions which are designed for the large set limit, since they first convert to `Base.Set` (a `Dict`-like object) which has `O(1)` lookup but the conversion has a noticeable overhead for small sets.

Some speculation on why priming an ITensor is now a bit slower - I believe this is because of the new ITensor type design, which now simply wraps a Tensor. In the new design, a bit more work has to be done for operations that act on indices. Previously the indices could just be extracted directly from the ITensor and then a new ITensor with the new indices could be created, but now first the Tensor has to get extracted, then the indices of that Tensor have to be extracted and modified, and then a new Tensor has to be created and then put back into a new ITensor. I don't think the slowdown is something to be concerned about (since the design allows things like `getindex` and `setindex!` to be much faster), but something to keep in mind. Modifying in-place is faster:
```julia
julia> @btime prime!($A)
  34.101 ns (1 allocation: 224 bytes)
```
which indicates that it has something to do with creating the new ITensor.

emstoudenmire:
Appreciate the detailed benchmarks. I did notice the slower priming performance, though it only looks like a 2x slowdown on an already very fast thing so not the biggest cause for concern. And the tradeoff is much better performance across many other operations, some of which are more likely to appear in loops or "hot" sections of code. And importantly this speeds up over the master branch which includes some redesigns we want less for performance reasons and just as much for simplicity and some bonuses like better JIT times.

mtfishman:
As a summary, this PR has a few (potentially) breaking changes:

1. Delete the `IndexSet` type in favor of `Tuple`/`Vector` of `Index` (depending on the context).
2. Change the `ITensor` type to wrap the `Tensor` type (not technically breaking since the interface is that same as long as you didn't try doing `A.inds` or `A.store`, but see below for performance implications).
3. Making `Tensor`/`ITensor` act like `tensor`/`itensor` by making a view of the input data when possible.

Here are more details about these changes:

1. Delete the `IndexSet` type. Instead, ITensor indices are all `Tuple`s of `Index`. In addition, set functions now return `Vector{<: Index}`. Operations that worked on `IndexSet` like `dag` and `prime` should now work generically on either `Tuple` or `Vector` of `Index`. To help with backwards compatibility, there is a type alias `const IndexSet{IndexT <: Index} = Vector{IndexT}` so if users were writing code like `IndexSet(i, j, k)` it should still "just work" (there may be corner cases but hopefully not many).

2. The `ITensor` type now simply just wraps a `Tensor`. This makes certain operations like `getindex` and `setindex!` of `ITensor`s faster, but slows down things like `prime(::ITensor)` slightly (see above). This has no outward effect on the interface. However, I found that some operations that used to act directly on indices are now faster if they act on the wrapped `Tensor` instead, for example defining:
```julia
dims(A::ITensor) = dims(Tensor(A))
```
is faster than defining it as:
```julia
dims(A::ITensor) = dims(inds(A))
```
My understanding is that the former version is better for the Julia compiler, since it can simply extract the `Tensor` and then call a very fast `dims(::Tensor)` making use of a function barrier, whereas actually extracting `inds(A)` is a bit slower now. This has some implications for how to write very optimized library code, but shouldn't generally be noticeable to users. It generally falls under the category of using function barriers to write optimized ITensor code.

3. Another pretty big breaking design change which is orthogonal to the others is deprecating `tensor`/`itensor` in favor of `Tensor`/`ITensor`. This changes the behavior of `Tensor`/`ITensor` to create views whenever possible. Previously, `Tensor`/`ITensor` provided a guarantee that the type "owned" whatever storage was input. However, I don't think this is generally the behavior that is wanted for `ITensor(::Tensor)` or `Tensor(::ITensor)`, and I think people reach for those functions first. The main subtle choice is what should be meant by `ITensor(A::Array, i::Index...)`. Depending on if these are QNs or not, this could either make a view of the `Array` in a `Dense` storage or it could result in a block sparse tensor that of course in general is not a view of the original data. This asymmetry is shown here:
```julia
julia> using ITensors

julia> using LinearAlgebra

julia> # Non-QN version:
       A = Matrix{Float64}(I, 2, 2)
2×2 Matrix{Float64}:
 1.0  0.0
 0.0  1.0

julia> i = Index(2)
(dim=2|id=178)

julia> T = ITensor(A, i', dag(i))
ITensor ord=2 (dim=2|id=178)' (dim=2|id=178)
NDTensors.Dense{Float64, Vector{Float64}}

julia> T[1, 1] = 2
2

julia> @show T;
T = ITensor ord=2
Dim 1: (dim=2|id=178)'
Dim 2: (dim=2|id=178)
NDTensors.Dense{Float64, Vector{Float64}}
 2×2
 2.0  0.0
 0.0  1.0

julia> A
2×2 Matrix{Float64}:
 2.0  0.0
 0.0  1.0
```
vs.:
```julia
julia> # QN version:
       A = Matrix{Float64}(I, 2, 2)
2×2 Matrix{Float64}:
 1.0  0.0
 0.0  1.0

julia> i = Index([QN(0) => 1, QN(1) => 1])
(dim=2|id=566) <Out>
 1: QN(0) => 1
 2: QN(1) => 1

julia> T = ITensor(A, i', dag(i))
ITensor ord=2
(dim=2|id=566)' <Out>
 1: QN(0) => 1
 2: QN(1) => 1
(dim=2|id=566) <In>
 1: QN(0) => 1
 2: QN(1) => 1

julia> T[1, 1] = 2
2

julia> @show T;
T = ITensor ord=2
Dim 1: (dim=2|id=566)' <Out>
 1: QN(0) => 1
 2: QN(1) => 1
Dim 2: (dim=2|id=566) <In>
 1: QN(0) => 1
 2: QN(1) => 1
 2×2
Block(1, 1)
 [1:1, 1:1]
 2.0

Block(2, 2)
 [2:2, 2:2]
 1.0

julia> A
2×2 Matrix{Float64}:
 1.0  0.0
 0.0  1.0
```
I can't really think of a reason why people would be modifying the `Array` data and `ITensor` separately. I imagine the most common use case for these constructors is to first set up an `Array` and then use it to initialize the `ITensor`, and I don't think it is common to then want to use that `Array` again. I keep seeing people using the syntax `ITensor([0 1; 1 0], i, j)` and I presume they expect it isn't copying. I'll note that Julia itself sometimes treats constructor calls as views and sometimes as copies, but it is generally consistent within the type (for example it depends on whether the type is more of a wrapper or more fundamental, like an `Array` constructor never makes a view but a wrapper type like `PermutedDimsArray` always does). This has been a design that has been bothering me, so discussion is welcome about it.

mtfishman:
Possibly a more realistic example of confusion the view behavior might case:
```julia
julia> i = Index(2)
(dim=2|id=504)

julia> A = [1. 0; 0 1]
2×2 Matrix{Float64}:
 1.0  0.0
 0.0  1.0

julia> T1 = ITensor(A, i', dag(i))
ITensor ord=2 (dim=2|id=504)' (dim=2|id=504)
NDTensors.Dense{Float64, Vector{Float64}}

julia> T2 = ITensor(A, i', dag(i))
ITensor ord=2 (dim=2|id=504)' (dim=2|id=504)
NDTensors.Dense{Float64, Vector{Float64}}

julia> T1 .*= 2
ITensor ord=2 (dim=2|id=504)' (dim=2|id=504)
NDTensors.Dense{Float64, Vector{Float64}}

julia> @show T1;
T1 = ITensor ord=2
Dim 1: (dim=2|id=504)'
Dim 2: (dim=2|id=504)
NDTensors.Dense{Float64, Vector{Float64}}
 2×2
 2.0  0.0
 0.0  2.0

julia> @show T2;
T2 = ITensor ord=2
Dim 1: (dim=2|id=504)'
Dim 2: (dim=2|id=504)
NDTensors.Dense{Float64, Vector{Float64}}
 2×2
 2.0  0.0
 0.0  2.0
```
This behavior isn't necessarily bad, however note that if someone defined `A = [1 0; 01]` or `i = Index([QN() => 2])`, `T2` would not be modified. This could be quite subtle, which was the reason for the special constructor `itensor`, which acts like Julia's [convert](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#conversion-and-promotion) function, i.e. in standard Julia conventions `itensor([...])` would be called `convert(ITensor, [...])`.

An approach would be to tell people to always do `ITensor(copy(A), i', dag(i))` if they were using `A` to construct multiple ITensors and modifying, but the behavior is definitely pretty subtle and in the case when `A` needed to get converted or `i` had QNs, the copying would be unnecessary. That is essentially what the `ITensor` vs. `itensor` constructor conventions was handling before (i.e. `ITensor` never made a view, but handled the logic of not doing too many copies, which was becoming kind of complicated to keep track of).

PR #630, opened 2021-04-14
Allow setting complex element of real ITensor

emstoudenmire:


mtfishman:
So I know we've talked about this design quite a bit, but I have two questions about the particular implementation:

1. Should this handle more generic element type expansions, like `Float32` -> `Float64` or even `Float64` -> `BigFloat`? If so, the design has to be thought about carefully there, since it could be more subtle in those cases to determine when the element type should get expanded, for example:
```julia
julia> A = randn(Float32, 4, 4)
4×4 Matrix{Float32}:
 -0.913951   0.609699  -0.687752  -0.141101
 -0.754697   0.830327  -0.565251   1.16966
  0.160613  -0.529546   1.07754    0.277681
  0.196089  -0.262913  -1.48421    0.201634

julia> A[1, 1] = Float64(3)
3.0

julia> A
4×4 Matrix{Float32}:
  3.0        0.609699  -0.687752  -0.141101
 -0.754697   0.830327  -0.565251   1.16966
  0.160613  -0.529546   1.07754    0.277681
  0.196089  -0.262913  -1.48421    0.201634

julia> A[1, 1] = BigFloat(4)
4.0

julia> A
4×4 Matrix{Float32}:
  4.0        0.609699  -0.687752  -0.141101
 -0.754697   0.830327  -0.565251   1.16966
  0.160613  -0.529546   1.07754    0.277681
  0.196089  -0.262913  -1.48421    0.201634
```
For cases of overflow, Julia handles it by making the element type `Inf`:
```julia
julia> A[1, 1] = 1e100
1.0e100

julia> A
4×4 Matrix{Float32}:
 Inf         0.609699  -0.687752  -0.141101
 -0.754697   0.830327  -0.565251   1.16966
  0.160613  -0.529546   1.07754    0.277681
  0.196089  -0.262913  -1.48421    0.201634
```
I'm bringing this up to think about how general the design is and how we might try to handle other cases. Maybe a more subtle case to think about, for example, is if there is an ITensor with `eltype(T) === Float32` and someone tries to set it with an element like `ComplexF64(2, 3)`, do we want it to expand to element type `ComplexF64` or `ComplexF32`? It's a bit of a corner case, but I still think it's important to outline what the rules of this system are very clearly.

2. In the case of the support question linked to in the original question (http://itensor.org/support/2925/playing-with-mps-itensors-indexing?show=2958#c2958) it seems like that case could be handled with a more conservative approach of making sure the `Empty` storage does the correct storage expansion, without having to allow it for every storage type. That seems like it would handle the main use case for this design, which was the `op` system. I guess the bigger question is, what are the applications for this beyond `op`, and could they be handled sufficiently by `emptyITensor(i, j)` (and then possibly by the `ITensor(i, j)` constructor, if we define that as `emptyITensor(i, j)`)?

mtfishman:
At first this was closed accidentally when changing the default branch from `master` to `main` but in fact this has been superseded by #641 (currently only default constructed ITensors with `EmptyStorage` storage can get set to any element type, but we may expand that functionality to other storage types).

PR #634, opened 2021-04-28
Add code examples from itensor.org to Documenter

emstoudenmire:
Adds nearly all code examples from here: http://itensor.org/docs.cgi?page=formulas&vers=julia
to the Documenter docs. Includes some minor edits to index.md to not reuse the term Code Examples twice in the table of contents.


docs/src/examples/DMRG.md
```diff
@@ -0,0 +1,347 @@
...
+In the last line above we convert the AutoMPO helper object to an actual MPO.
+
+Before beginning the calculation, we need to specify how many DMRG sweeps to do and
+what schedule we would like for the parameters controlling the accuracy.
+These parameters are stored within a sweeps object:
+
+```julia
+sweeps = Sweeps(5) # number of sweeps is 5
+maxdim!(sweeps,10,20,100,100,200) # gradually increase states kept
+cutoff!(sweeps,1E-10) # desired truncation error
```
mtfishman, line 42: Should we start using `setmaxdim!`/`setcutoff!`?

docs/src/examples/DMRG.md
```diff
@@ -0,0 +1,347 @@
...
+
+
+  #
+  # Make sure to do lots of sweeps
+  # when finding excited states
+  #
+  sweeps = Sweeps(30)
+  maxdim!(sweeps,10,10,10,20,20,40,80,100,200,200)
+  cutoff!(sweeps,1E-8)
+  noise!(sweeps,1E-6)
```
mtfishman, line 301: Again, `setmaxdim!`/`setcutoff!`/`setnoise!`?

docs/src/examples/MPSandMPO.md
```diff
@@ -0,0 +1,203 @@
...
+![](updated_mps.png)
+
+As a technical note, if you are working in a context where gauge or orthogonality
+properties of the MPS are important, such as in time evolution using two-site gates, 
+then you may want to call `orthogonalize!(psi,3)`
+before modifying the tensor at site 3, which will ensure that the MPS remains in a 
+well-defined orthogonal gauge centered on site 3. Modifying a tensor which is left- or right-orthogonal
+(i.e. not the "center" tensor of the gauge) will destroy the gauge condition and 
+require extra operations to restore it. (Calling `orthogonalize!` method will automatically
+fix this but will have to do extra work to do so.)
```
mtfishman, line 43: We can also mention `@preserve_ortho psi psi[3] = newA` here as a way to not have to call `orthogonalize!`.

docs/src/examples/ITensor.md
```diff
@@ -0,0 +1,70 @@
...
+and offers a standard interface across various libraries and languages.
+
+**Writing an ITensor to an HDF5 File**
+
+Let's say you have an ITensor `T` which you have made or obtained
+from a calculation. To write it to an HDF5 file named "myfile.h5"
+you can use the following pattern:
+
+```julia
+using HDF5
```
mtfishman, line 42: This would require that users `add HDF5` themselves. We could instead use `using ITensors.HDF5` here which doesn't require that extra step.

emstoudenmire: That's a nice observation - I didn't know that, and was just expecting people to add HDF5, yes.

Do you think if we changed it to `using ITensors.HDF5` that people might think that's a different package than HDF5.jl or be confused by it for some other reason?

mtfishman: Hmm, I don't know. We could just make a comment about what that notation means, and say it is equivalent to `using HDF5` if you `add HDF5`.

mtfishman:
Should we move the sections: https://itensor.github.io/ITensors.jl/dev/#Full-Example-Codes-1 and https://itensor.github.io/ITensors.jl/dev/#ITensor-Code-Samples-1 to the new `Code Examples` section, and just link to the `Code Examples` section from the `Introduction` section?

emstoudenmire:
Yes that's a good idea as long as the link also works from the README page too.

PR #635, opened 2021-04-28
Add RegisterAction

mtfishman:
Add RegisterAction Github Action to make it easier to register a new version through Github.


emstoudenmire:
This is great. I assume the steps to use are the ones here: https://github.com/julia-actions/RegisterAction

At some point I'll update the Checklists.txt file I use to remember how to do stuff to reflect using this new approach.

PR #636, opened 2021-04-28
Docs improvements and more examples

emstoudenmire:
Fix some typos in the docs and expand the set of examples to include more basic ITensor operations.

- Move MPS one site op figures into folder
- Fix some math notation
- Add example of tracing an ITensor
- More basic ITensor operation examples


docs/src/examples/ITensor.md
```diff
@@ -17,6 +17,165 @@ or this
...
+Here are some examples of basic element-wise operations we can do using Julia's dotted operator broadcasting syntax.
+
+```julia
+# Multiply every element of `T` by 2.0:
+T .*= 2.0
+```
+
+```julia
+# Add 1.5 to every element of T
+T .+= 1.5
```
mtfishman, line 115: I'm not sure this will do the correct thing for QN ITensors unfortunately:
```julia
i = Index([QN(0) => 1, QN(1) => 1])
A = emptyITensor(i', dag(i))
A[1, 1] = 1
A .+= 1.5
@show A
```
Gives:
```julia
A = ITensor ord=2
Dim 1: (dim=2|id=571)' <Out>
 1: QN(0) => 1
 2: QN(1) => 1
Dim 2: (dim=2|id=571) <In>
 1: QN(0) => 1
 2: QN(1) => 1
 2×2
Block(1, 1)
 [1:1, 1:1]
 2.5

Block(2, 1)
 [2:2, 1:1]
 1.5

Block(1, 2)
 [1:1, 2:2]
 1.5

Block(2, 2)
 [2:2, 2:2]
 1.5
```
I'm also not even sure how well defined this operation is for QN ITensors, should it only add to the existing QN blocks, any possible nonzero flux block, etc? I guess for now we should just disallow it for QN ITensors.

docs/src/examples/ITensor.md
```diff
@@ -17,6 +17,165 @@ or this
...
+The dotted notation works for functions too:
+
+```julia
+# Replace every element in T by its absolute value:
+T .= abs.(T)
+```
+
+```julia
+# Replace every element in T by the number 1.0
+T .= one.(T)
```
mtfishman, line 127: This operation also is a bit strange for QN ITensors:
```julia
i = Index([QN(0) => 1, QN(1) => 1])
A = emptyITensor(i', dag(i))
A[1, 1] = 2
A .= one.(A)
@show A
```
gives:
```julia
A = ITensor ord=2
Dim 1: (dim=2|id=210)' <Out>
 1: QN(0) => 1
 2: QN(1) => 1
Dim 2: (dim=2|id=210) <In>
 1: QN(0) => 1
 2: QN(1) => 1
 2×2
Block(1, 1)
 [1:1, 1:1]
 1.0
```
Again, the question would be if this should fill in the other possible QN value. Personally I think it should just be ill-defined for QN ITensors. I think the "rule" would be that for `A .= f.(B)`, for QN ITensors, if `f(0) != 0` then the operation is ill-defined, and we should provide a different function like `nonzeros(A) .= f.(nonzeros(B))` if you want the operation to only act on the nonzero values. This logic is a bit tricky to implement the way the ITensor broadcasting is written right now, it needs a rewriting.

Anyway I know this section is about dense tensors where these operations should be fine, I can raise a separate issue but maybe it is just worth making a warning about the QN version of this here.

docs/src/examples/ITensor.md
emstoudenmire: Hm, yes I see. I should slow down on the larger docs PR's and make sure to leave time for review. Thanks for taking a look!

Yes I think for that operation it's best for now to disallow it for QN ITensors. Maybe we could separately make a rule that if you do any function of the form `T .= f.(T)` then it will be carried out on all elements which are compatible with the flux (so it would introduce any missing, but allowed, blocks) but still not act on any disallowed blocks. Just an idea.

docs/src/examples/ITensor.md
emstoudenmire: Yes as a concrete step I'll update that docs section to note that it applies to dense tensors only. Good catch!

See my comment above for another idea of how we could define broadcasting of functions on QN ITensors. There are multiple ways we could define it, so we should discuss more.

docs/src/examples/ITensor.md
mtfishman: Yeah, my issue with that is it doesn't really feel mathematically correct based on the definition of broadcasting being an elementwise operation on all elements of a structure, and the way it generally is defined for sparse structures:
```julia
julia> using LinearAlgebra

julia> D = Diagonal(randn(4))
4×4 Diagonal{Float64, Vector{Float64}}:
 -1.28286   ⋅         ⋅         ⋅ 
   ⋅       0.15365    ⋅         ⋅ 
   ⋅        ⋅       -1.38377    ⋅ 
   ⋅        ⋅         ⋅       -1.87302

julia> D .+= 2.4
ERROR: ArgumentError: cannot set off-diagonal entry (2, 1) to a nonzero value (2.4)
```
I don't think that is a common enough operation to do to warrant making such a convenient and potentially confusing operation. I think for that usage we should provide separate functionality for that, like:
```julia
insert_zero_blocks!(A) # Insert blocks of zero consistent with the flux
nonzeros(A) .+= 1.5  # Now broadcast only on the nonzero values of A
```
Or alternative, you construct another QN ITensor `B` which is filled with the value `1.5` and add it in-place to `A`, which may be just as efficient as the above. Another reason for not making the above operations default behavior for broadcasting is that in general the functionality `insert_zero_blocks!(A)` can be very slow, so now we only do it when necessary, like in constructors. (Note that `insert_zero_blocks!(A)` doesn't actually exist in that form but that kind of operation is used in QN constructors in a different form.)

emstoudenmire: Ok agreed that's the best and clearest approach. It lets you do the thing I was saying, but then see from the code that it exactly what is happening and not have to guess. And it lets you fine-tune the behavior, or we could introduce other similar behaviors and wrappers in the future.

PR #638, opened 2021-04-28
More code examples and improvements

emstoudenmire:
- Example of ITensor construction from arrays
- Improve elementwise examples
- Small improvement
- Add factorizing examples
- Use ITensors.HDF5 in examples with a note about why


docs/src/examples/ITensor.md
```diff
@@ -62,6 +62,36 @@ el = T[j=>1,i=>2,k=>3]
...
+
+```julia
+T = randn(4,7,2)
+
+k = Index(4,"index_k")
+l = Index(7,"index_l")
+m = Index(2,"index_m")
+
+B = ITensor(T,k,l,m)
+```
```
mtfishman, line 93: We could also mention ways to make this more compact, like:
```julia
k, l, m = Index.(size(T), ("index_k", "index_l", "index_m"))
```
That can give people ideas for how to use Julia's built in tools to make things easier. We could put "advanced" tips like that in collapsible blocks that you have to click on to see: https://discourse.julialang.org/t/documenter-jl-and-collapsible-markdown/50898

emstoudenmire: I didn't know that clickable dropdowns were part of html now. That's great! On itensor.org I had been using a more complicated JavaScript trick for that. I wasn't quite able to get it to work like in that example, though. Even when I wrapped the html in a ```@raw html ... ``` block that Documenter uses for raw code, it sort of worked but when I put another code block inside, it got confused and pulled the entire rest of the page's contents into the dropdown.

I was thinking we should have some examples and tips about indices. We could put them in the ITensor examples, for the case of dense indices, and have a separate page all about QN Indices and QN ITensor code examples. (Indices by themselves may not warrant an entire section is what I'm thinking, though maybe they do in order to discuss things like id's and tags and so on.)

mtfishman: Another neat way to make a group of indices:
```julia
is = Index.(size(T), ("i$n" for n in 1:ndims(T)))
# is = [Index(size(T, n), "i$n") for n in 1:ndims(T)]
ITensor(T, is...)
```
since I think people find making a big group of indices tedious but might not realize how to use Julia tools to help with it. I find constructors like the above make writing random tests of contraction much easier, for example:
```julia
using Random
i = [Index(n, "i$n") for n in 1:12]
A = randomITensor(shuffle(i[1:7])...)
B = randomITensor(shuffle(i[4:end])...)
A * B
```

Also, should we mention the `itensor` constructor here as well? That could also be in a collapsible "advanced" section.

emstoudenmire: I do agree these are neat and we should put them somewhere, but I do think it should be in a more advanced sort of section. A dropdown at the end of a closely related section is one good idea. We could also have a separate page or pages called "Advanced ITensor Examples" etc. Some of these techniques are most useful for things like writing tests, whereas a lot of code "in the wild" may only involve making a small handful of indices or using something like `siteinds` though it definitely varies from application to application.

emstoudenmire: I keep being amazed what you can do with broadcasting in Julia. It's not something I've been used to since I've mostly used C++ before!

mtfishman: That's strange, I just tested pasting this into a section of the docs and it works correctly for me:
``````julia
```@raw html
<details><summary>Advanced</summary>
```

```julia
print("hello world!")
```

```@raw html
</details>
```
``````

emstoudenmire: Great, I see. You did it differently from me, by having two @raw blocks. I was trying to use just one that wrapped around all of the contents inside, including the code block.

Issue #593, opened 2021-03-18
Make `ITensor()` default to `emptyITensor()`


mtfishman:
Currently, `ITensor()` is defined as `ITensor(0)`. We may want to change the default to `emptyITensor()` (but make sure that `emptyITensor()` works for the cases that `ITensor(0)` works for now, such as `ITensor() * ITensor(i, j)`).

Issue #629, opened 2021-04-14
Allow setting complex element of real ITensor


emstoudenmire:
We could allow setting an element of a real-valued ITensor to a complex value, which would of course case the storage to be converted to a complex valued storage. 

This change would ease usability in many cases, such as this recent question/comment on the support forum: http://itensor.org/support/2925/playing-with-mps-itensors-indexing?show=2958#c2958

mtfishman:
Closed by #641.

Issue #505, opened 2020-10-07
More missing ITensor broadcasting operations


mtfishman:
The following broadcasting operation errors: `A .= real.(A) .^ 2 .+ im .* imag.(A) .^2`

mtfishman:
Look into [Broadcast.flatten](https://github.com/JuliaLang/julia/blob/v1.6.0/base/broadcast.jl#L282-L318) as a more generic way to forward the function call to NDTensors. Hopefully that wouldn't lead to type instabilities and performance issues.

mtfishman:
Some notes on `A .+= c` and `A .= f.(B)` for QN ITensors, continued from . Right now there is this behavior:
```julia
i = Index([QN(0) => 1, QN(1) => 1])
A = emptyITensor(i', dag(i))
A[1, 1] = 2
A .= one.(A)
@show A
```
gives:
```julia
A = ITensor ord=2
Dim 1: (dim=2|id=210)' <Out>
 1: QN(0) => 1
 2: QN(1) => 1
Dim 2: (dim=2|id=210) <In>
 1: QN(0) => 1
 2: QN(1) => 1
NDTensors.BlockSparse{Float64, Vector{Float64}, 2}
 2×2
Block(1, 1)
 [1:1, 1:1]
 1.0
```
A question would be if this should fill in the other possible QN values, or just not be allowed. Personally I think it should just be ill-defined for QN ITensors. I think the "rule" would be that for A .= f.(B), for QN ITensors, if f(0) != 0 then the operation is ill-defined, and we should provide a different function like nonzeros(A) .= f.(nonzeros(B)) if you want the operation to only act on the nonzero values. This logic is a bit tricky to implement the way the ITensor broadcasting is written right now, it needs a rewriting.

For operations like `A .+= c`, currently this does the incorrect things for QN ITensors:
```julia
i = Index([QN(0) => 1, QN(1) => 1])
A = emptyITensor(i', dag(i))
A[1, 1] = 1
A .+= 1.5
@show A
```
Gives:
```julia
A = ITensor ord=2
Dim 1: (dim=2|id=571)' <Out>
 1: QN(0) => 1
 2: QN(1) => 1
Dim 2: (dim=2|id=571) <In>
 1: QN(0) => 1
 2: QN(1) => 1
NDTensors.BlockSparse{Float64, Vector{Float64}, 2}
 2×2
Block(1, 1)
 [1:1, 1:1]
 2.5

Block(2, 1)
[...]

PR #641, opened 2021-04-29
WIP: Make `ITensor(i, j, k)` constructors create `EmptyStorage` by default

mtfishman:
Fixes https://github.com/ITensor/ITensors.jl/issues/593 and https://github.com/ITensor/ITensors.jl/issues/629.

Most tests are passing now, and it relies on adding more operations for `EmptyStorage` tensors in this PR: https://github.com/ITensor/NDTensors.jl/pull/81.

I still need to define some more corner cases such as contraction of `EmptyStorage` ITensors, as well as change over the constructors with QNs. Generally non-QN ITensors will have very similar behavior after this change, however QN ITensors will be a bit different since right now `ITensor(i, j)` for QN ITensors makes an `ITensor` with flux of `QN()`. This means that operations like:
```julia
A = ITensor(i, j)
randn!(A)
```
or
```julia
A = ITensor(i, j)
fill!(A, 3)
```
will work quite differently now for QN ITensors, though you may argue from these arguments https://github.com/ITensor/ITensors.jl/issues/505#issuecomment-828583195 that those operations are not well defined for QN ITensors anyway, and someone should do that kind of thing in a different way, like use `randomITensor(i, j)`, `ITensor(3, i, j)` (which both default to filling values with `QN()`), or `A = ITensor(QN(), i, j); fill!(nonzeros(A), 3)`.


mtfishman:
A quick update: with `ITensor(i, j, k)` for defaulting to `EmptyStorage` for non-QN indices, all tests are passing now (some tests needed very slight adjustments as you can see). Of course there are some decisions about how the `EmptyStore` `ITensor`s should behave now since they are used in more places, but I've made up some rules for addition, contraction, etc. with other `EmptyStorage` and `Dense` types which seem reasonable (mostly by treating `EmptyStorage` as a true universal zero).

I still need to switch over the QN `ITensor(i, j, k)` constructors to default to `EmptyStorage`, which will surely require additional code additions to `NDTensors` but should hopefully be pretty straightforward, and follow what was already done in the dense case.

mtfishman:
Also fixes #643.

PR #644, opened 2021-05-04
Error messages for get and set index

emstoudenmire:
This PR adds error checks for calls to ITensor getindex and setindex! with the wrong number of index values.
- Fix erroring tests
- Dimension checks for ITensor get and set index
- Extra check and formatting changes
- Add tests and checklist entry on formatting

### DIFF
--- a/Checklists.txt
+++ b/Checklists.txt
@@ -1,3 +1,10 @@
+Checklist for Formatting ITensor Code
+--------------------------------------
+- cd to the top level of the source code
+- open a Julia REPL session
+- run: using JuliaFormatter; format("."; verbose = true)
+- result will be true if code already formatted
+
 Checklist for Tagging a New Release
 --------------------------------------
 - Update the version = "x.y.z" line near the top of Project.toml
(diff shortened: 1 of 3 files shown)


mtfishman:
Maybe better to use dispatch here to ensure there isn't any runtime overhead, for example:
```julia
@propagate_inbounds @inline function _getindex(T::Tensor{<:Any,N}, ivs::Vararg{<:Any,N}) where {N}
  # Implementation
end
```
and then define a fallback version:
```julia
function _getindex(T::Tensor, ivs::Vararg{<:Any,N}) where {N}
  throw(
    DimensionMismatch("$N index values passed to ITensor getindex, expected $(ndims(T))")
  )
end
```

emstoudenmire:
Ok good idea - I can definitely do it that way instead.

So the failing tests include some involving something like `T = emptyITensor()` followed by `T[i=>1,i'=>1] = 1.0` or similar. Are we planning to continue to support this behavior post 0.2? I guess we are at least going to allow empty ITensors to be a "universal zero for addition" but that does leave the separate question of setting elements.

mtfishman:
Actually it looks like the test was supposed to give an error, but now the error is different so just needs to be updated to `DimensionMismatch` from `MethodError`: https://github.com/ITensor/ITensors.jl/blob/f55d05fa4075521e88eacb648d6012a49ef3aa80/test/empty.jl#L49.

mtfishman:
Not sure what the status of `T = emptyITensor(); T[i=>1,i'=>1] = 1.0` will be, I would be in support of adding that.

Issue #618, opened 2021-04-05
`oneITensor` to rule them all


mtfishman:
A common code pattern is:
```julia
A = ITensor(1)
A *= B
A *= C
```
which is useful when initializing a loop, for example. However, it has bothered me that it makes an unnecessary copy of `B`. We decided, I think correctly, that `ITensor(1) * B` should make a copy of `B`, since it could lead to very subtle bugs if `ITensor(1) * B` didn't copy but `ITensor(1.1) * B` (say you were doing floating point operations to compute a constant `a`, and then did `ITensor(a) * B`, and then one time `a` happened to be equal to `1` but you were relying on the copying behavior).

My proposal for this would be to introduce yet another storage type as well as a special number type called `One`. The proposal would be the following:
```julia
using ITensors
using ITensors.NDTensors

# Multiplicative identity
struct One <: Number
end

struct Scalar{ElT} <: TensorStorage{ElT}
  data::Base.RefValue{ElT}
end
Scalar(n::Number) = Scalar(Ref(n))

oneITensor() = itensor(Scalar(One()), IndexSet())
```
The `Base.RefValue` storage allows the value to be modified in-place, i.e. would allow `T[] = 2.0`. That's not totally necessary but I think is helpful for generic code, since a lot of code in `NDTensors` assumes the tensors are mutable.

Then, we can define contraction such that `(oneITensor() * B) = B` (without copying). `Scalar(One())` storage could be interpreted as a universal multiplicative identity. Note that if someone did `1.0 * oneITensor()`, it would create a tensor with storage `Scalar(1.0)`, which would then act as `ITensor(1.0)` with copying behavior with multiplication. This allows the nice property that `(1.0 * oneITensor()) * B` and `oneITensor() * (1.0 * B)` both return copies of `B`.

Another note that the `Scalar` storage isn't strictly necessary here, and we could use a length 1 `Dense` storage with element type `One`, but I've been meaning to add a `Scalar` type anyway since I think it will help clean up some code logic with scalar ITensor contractions.

[...]

PR #648, opened 2021-05-06
Add write-to-disk during DMRG with `SerializedElementArrays.jl`

mtfishman:
This makes use of [SerializedElementArrays.jl](https://github.com/ITensor/SerializedElementArrays.jl) (in the process of being registered) to add an option to write the effective Hamiltonian environments to disk in DMRG. It is enabled with the interface:
```julia
dmrg(H, psi0, sweeps; write_when_maxdim_exceeds=1000)
```
Internally it introduces a new type `DiskProjMPO` which generally acts as a `ProjMPO` but stores the ITensor environments in a `DiskVector`, which is a vector where the elements are written to disk in a serialized format.


mtfishman:
In that latest I added better memory caching of the most recently created environment tensors, so they don't have to be read from disk during Lanczos. This is good to go once [SerializedElementArrays.jl gets registered)[https://github.com/JuliaRegistries/General/pull/36283].

examples/dmrg/write_to_disk/1d_heisenberg.jl
```diff
@@ -0,0 +1,44 @@
+using ITensors
+using ITensors.Strided
+using LinearAlgebra
+using Printf
+using Random
+
+Random.seed!(1234)
+BLAS.set_num_threads(1)
+Strided.set_num_threads(1)
```
emstoudenmire, line 9: Are these settings necessary for doing write-to-disk DMRG? Or just recommended say?

mtfishman: Just copied from another example that was using block sparse multithreading. Though generally it is recommended to turn of Strided threading when using BLAS threading (which is only needed if Julia is started with multiple threads).

src/mps/abstractprojmpo.jl
```diff
@@ -0,0 +1,242 @@
...
+  R = nothing
+  N = length(P.H)
+  while P.rpos > k
+    rl = P.rpos
+    if rl >= N + 1
+      R = psi[N] * P.H[N] * dag(prime(psi[N]))
+      P.LR[N] = R
+      P.rpos = N
+    else
+      R = P.LR[rl] * psi[rl - 1] * P.H[rl - 1] * dag(prime(psi[rl - 1]))
```
emstoudenmire, line 167: Possible to save a call to P.LR here, by either using R if that logic works out or else calling `rproj` ?

src/mps/diskprojmpo.jl
```diff
@@ -0,0 +1,101 @@
...
+
+# Special overload of lproj which uses the cached
+# version of the left projected MPO, and if the
+# cache doesn't exist it loads it from disk.
+function lproj(P::DiskProjMPO)::Union{ITensor,Nothing}
+  (P.lpos <= 0) && return nothing
+  if (P.lpos ≠ P.lposcache) || (P.lpos == 1)
+    # Need to update the cache
+    P.Lcache = P.LR[P.lpos]
+    P.lposcache = P.lpos
```
emstoudenmire, line 65: An alternative to this approach (which definitely works fine but requires managing both pos variables) would be to have a "setter" function `setlpos!` which sets `Lcache=nothing` if the new position variable passed to it is different from the current one. Then similarly defining a function that is for retrieving L that sets Lcache to P.LR[P.lpos] if isnothing(Lcache).

examples/dmrg/write_to_disk/1d_heisenberg.jl
mtfishman: So no, this isn't anything particular to write-to-disk.

emstoudenmire: Thanks - I saw that you removed it in a recent commit. Agreed, since I wouldn't want a beginning user to think these lines are necessary for this example.

src/mps/diskprojmpo.jl
mtfishman: Yeah good point. I don't particularly like how much of this code uses setting the fields directly as opposed to using functions like `setlpos!`, I was just writing it in a way that required the minimal number of changes to get the new functionality.

mtfishman:
Thanks for the comments, definitely the design could be a bit cleaner and use caching a bit better (particularly when making the environments, as you pointed out).

Note that it looks like the benchmark and documenter CI are working again since I've accounted for the change from `master` to `main`. Also updating my local version of `JuliaFormatter` to `v0.14.1` and then formatting locally helped get that CI to pass.

emstoudenmire:
Agreed. I didn't want to nitpick too much about that, but I do think for functions on abstract inputs, using getter and setter patterns is the best way to go, for one thing because then concrete types can execute code like updating caches within these without it having to be in the abstract logic (another more minor reason is so that the field names can be different among various concrete types, but that's pretty minor). 

Good note about JuliaFormatter - probably that's why my recent PR didn't pass. Updating now...

src/mps/abstractprojmpo.jl
mtfishman: Probably. I think using `rproj`, at least how it is design currently, would require `R` to be stored in the `Rcache` field. Probably there would be a way to make that work, but the goal was for this function to be generic between `DiskProjMPO` and `ProjMPO` (to minimize code duplication and make maintenance easier) and `ProjMPO` doesn't have an `Rcache` field.

mtfishman: Though I guess `rproj` may automatically save the correct one in the cache? I have to draw out the diagrams.

src/mps/diskprojmpo.jl
mtfishman: I'm just going to leave this for now, to not mess too much with the current `ProjMPO` code. But we can add that as an issue for further investigation.

src/mps/abstractprojmpo.jl
mtfishman: Alright, I improved caching in `makeL!` and `makeR!` by reusing `R`/`L` and also using `rproj`/`lproj`.

emstoudenmire:
Sounds good - I'd approve merging this. The AbstractProjMPO code is a relatively minor thing as long as you think the code minimizes disk reads as much as possible (referencing your recent commits which reduce that).

mtfishman:
In the latest, I improved memory caching within `makeL!` and `makeR!`. 

Also more generally, I made `lproj` and `rproj` return an instance of a new type `OneITensor` which acts like:
```julia
(::OneITensor * A::ITensor) = A
(A::ITensor * ::OneITensor) = A
```
This is a poor man's version of:  #618.

This allowed a big simplification of functions like `product` which had complicated nested `if` statements accounting for when we were at the edge of the system. Now, since `lproj` and `rproj` return `OneITensor()` that has no cost to contraction, the code can be written generically for those edge cases.

PR #649, opened 2021-05-07
MPS from Tensors

emstoudenmire:
- Add version of MPS constructor taking an array
- Code example for making MPS from a tensor

### DIFF
--- a/src/mps/abstractmps.jl
+++ b/src/mps/abstractmps.jl
@@ -1452,7 +1452,7 @@ function (::Type{MPST})(
     if !isnothing(l)
       Lis = unioninds(Lis, l)
     end
-    L, R = factorize(Ã, Lis; kwargs..., ortho="left")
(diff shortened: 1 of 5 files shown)


src/mps/abstractmps.jl
```diff
@@ -1465,6 +1465,12 @@ function (::Type{MPST})(
   return M
 end
 
+function (::Type{MPST})(
+  A::Array{ElT}, sites; kwargs...
+) where {MPST<:AbstractMPS,ElT<:Number,N}
```
mtfishman, line 1470: No need for `ElT` or `N` here. Also we could make this accept any `AbstractArray` (the `ITensor` constructor may not work with general `AbstractArray` but that is a work in progress).

mtfishman:
Great! This is useful to have.

emstoudenmire:
Ok will fix that & add a test too

PR #655, opened 2021-05-10
WIP: Get some precompilation working

mtfishman:
The goal of this is to write precompilation statements that improve ttfc (time to first contraction).

So far, I've improved type inference of `NDTensors` contraction a little bit, which has made precompilation more effective. Next step is making the precompilation statements more comprehensive, for example cover a range of element types and tensor orders. So far it just helps for contracting two order-2 tensors with `Float64` element type.


emstoudenmire:
Agreed one major takeaway here is a possible refactor of `contract` to ease the burden on the Julia compiler.

One question is about the long precompilation times this PR incurs: for users this will be only once per update I presume, which sounds totally fine and worth it. 
For developers, while it seems like a pain, wouldn't using Revise.jl mostly mitigate this issue? (So only about one long precompilation per coding session, followed by a series of short revises?)

mtfishman:
> Agreed one major takeaway here is a possible refactor of `contract` to ease the burden on the Julia compiler.
> 
> One question is about the long precompilation times this PR incurs: for users this will be only once per update I presume, which sounds totally fine and worth it.
> For developers, while it seems like a pain, wouldn't using Revise.jl mostly mitigate this issue? (So only about one long precompilation per coding session, followed by a series of short revises?)

Yes, I suppose that's right. I was thinking that certain things you do during the development process would trigger precompilation, but I guess that is only if you open a new Julia session after modifying source code. I still will open multiple sessions to test multiple things at once, for example run the test suite while interactively testing out a new feature, so it can still be annoying to have the long precompilation times. I think it was particularly annoying for me during the development of this PR since testing out the precompilation and load times requires opening new Julia sessions each time, so you could imagine that time adds up.

Still, I don't think we want to have precompilation times that are really long, like much longer than other packages, since that makes updates annoying for users. My understanding is that even if any dependency of `ITensors.jl` gets updated, it will trigger precompilation for `ITensors` since in principle we could be precompiling a function from another module that needs to get updated, so even if we aren't updating often, other packages might be. We also wouldn't want to make the precompilation so long that it makes users avoid updating, and over a minute of precompilation might be getting to that point. And unfortunately this doesn't improve things to the point where you would avoid using `ITensors.compile()`, so you add that time on as well, but hopefully that function gets faster from precompilation since it is running things like DMRG.

emstoudenmire:
As a small change which might help compilation times is that some of the functions in NDTensors which are parameterized don't have to be. A tiny example is this function:

```julia
const Labels{N} = NTuple{N,Int}

function contract_labels(T1labels::Labels{N1}, T2labels::Labels{N2}) where {N1,N2}
  ncont = 0 
  for i in T1labels
    i < 0 && (ncont += 1)
  end 
  NR = N1 + N2 - 2 * ncont
  ValNR = Val{NR}
  return contract_labels(ValNR, T1labels, T2labels)
end
```

where I think the function would be almost nearly as fast if the arguments were just of type `::Labels` and if one did `length(T1labels)` instead of using `N1` in the function body. (I bet that length gets inlined for tuples anyway.)

Then the compiler would only have to make one version of this function instead of perhaps 64 different versions (assuming up to 8th order tensors).

Probably this specific function is inconsequential but there may be some bigger ones where the same pattern is occurring, and where if it's really necessary to get the size parameters statically it could be done with some smaller helper functions.

mtfishman:
Whether or not you actually put the type parameter in the type definition doesn't make a difference for how much compilation needs to be done. The types in the signature are mostly used for dispatch, and allow you to use type parameters explicitly in the function body. Julia would still need to run type inference and compile a new specialized function for each new type that gets input into the function, so even if you left off the type parameter in the function definition, it would still need to compile different versions of the function for:
```julia
contract_labels((1, -1), (-1, 2))
contract_labels((1, -1), (-1, 2, 3))
[...]
```
You can analyze this with the `MethodAnalysis` package (see [here](https://stackoverflow.com/questions/64235693/get-a-list-of-all-the-function-signatures-that-have-been-compiled-for-a-generic)):
```julia
julia> tuple_cat(t1, t2) = (t1..., t2...)
tuple_cat (generic function with 1 method)

julia> using MethodAnalysis

julia> methodinstances(tuple_cat)
Core.MethodInstance[]

julia> tuple_cat((1, 2), (3, 4))
(1, 2, 3, 4)

julia> methodinstances(tuple_cat)
1-element Vector{Core.MethodInstance}:
 MethodInstance for tuple_cat(::Tuple{Int64, Int64}, ::Tuple{Int64, Int64})

julia> tuple_cat((1, 2), (3, 4, 5))
(1, 2, 3, 4, 5)

julia> methodinstances(tuple_cat)
2-element Vector{Core.MethodInstance}:
 MethodInstance for tuple_cat(::Tuple{Int64, Int64}, ::Tuple{Int64, Int64})
 MethodInstance for tuple_cat(::Tuple{Int64, Int64}, ::Tuple{Int64, Int64, Int64})
```
A more extreme example of this is this function: https://github.com/ITensor/ITensors.jl/blob/9fe8d5e74cac705a2c69e82f4ab7418df0a44752/src/itensor.jl#L1645-L1650
There are no type parameters explicitly there, but that function has an explosion of different types that could get input (many combinations of storage types, tensor ordered, index types, element types, etc.), each of which needs to have the type inferred and needs to be compiled. That function and the functions it calls are in fact where a lot of the inference time and compilation time happens.

It is similar to having a templated function in C++: you still need to compile it once you have the actual types that you will input. In Julia you are allowed syntactically to leave off type parameters, but really this is only a syntax convenience and Julia still fills it in for you with a free parameter, for example `Labels` gets rewritten to `Labels{N} where {N}` by Julia. Additionally, you can leave off `Labels` entirely, but that doesn't mean Julia won't still compile the function for each new type you input, and you can think of it as Julia rewriting the function from `contract_labels(t1, t2)` to `contract_labels(t1::T1, t2::T2) where {T1, T2}`.

You can avoid specialization on each new type using the [@nospecialize](https://docs.julialang.org/en/v1/base/base/#Base.@nospecialize) macro, so change the function signature to:
```julia
function contract_labels(@nospecialize(T1labels::Labels{N1}), @nospecialize(T2labels::Labels{N2})) where {N1,N2}
  [...]
end
```
though this can make the runtime of the function slower. My understanding is also that this only avoids specialization of actual code generation, not type inference. So Julia will first analyze the function to try to infer the types of all of the objects within the function as well as the return value, and then try to write specialized machine code given those types. I believe `@nospecialize` only controls the second step in that process (in general it still needs to determine the types of functions, for example if the function is called within other functions which need to use the results of type inference).

emstoudenmire:
Makes sense - thanks! Ok so really more of a situation calling for a code refactor perhaps.

mtfishman:
Other things that can help are making sure functions are type stable, splitting up big complicated functions into small functions with simple logic, and giving Julia some hints about types that are within the function (though this can backfire and sometimes make the code run slower if done haphazardly).

mtfishman:
The improvements to type stability will be helpful going forward, so I'm going to merge this once tests pass but with the precompilation turned off.

PR #657, opened 2021-05-11
Docs for expect and correlation_matrix

emstoudenmire:


mtfishman:
> Not sure why the formatting check is failing - it says something in the CI about a few of the examples not being formatted, but when I run the JuliaFormat format function from the top of the source code it visits the examples folder and doesn't report any changes there.

I'm seeing the same thing. Maybe the CI is using a different version from the latest version that you and I are probably running locally.

PR #660, opened 2021-05-13
Update Documenter settings

emstoudenmire:


emstoudenmire:
Despite some warnings and things in the Documentation build output, I think these changes worked (will know for sure if merging it makes the dev docs show the updated docs on main).

But the previews feature definitely worked as can be seen here:
https://itensor.github.io/ITensors.jl/previews/PR660

PR #662, opened 2021-05-13
Fix bug when computing flux where index and Block were getting mixed up

mtfishman:
Fixes a bug when computing flux where the index and Block were getting mixed up (an input `flux(i, 3)` was being treated as `flux(i, Block(3))`.

Also adds `MPS` as alternative syntax for `productMPS`.

Also adds more allowed inputs to `ops` and `op`, like `op(("H", 1), s)`.

### DIFF
--- a/src/NDTensors/blocksparse/blockdims.jl
+++ b/src/NDTensors/blocksparse/blockdims.jl
@@ -153,7 +153,12 @@ end
 
 # Given a CartesianIndex in the range dims(T), get the block it is in
 # and the index within that block
-function blockindex(T, i::Vararg{Int,N}) where {ElT,N}
+function blockindex(T, i::Vararg{Integer,N}) where {ElT,N}
+  # Bounds check.
+  # Do something more robust like:
+  # @boundscheck Base.checkbounds_indices(Bool, map(Base.oneto, dims(T)), i) || throw_boundserror(T, i)
+  @boundscheck any(iszero, i) && Base.throw_boundserror(T, i)
+
   # Start in the (1,1,...,1) block
   current_block_loc = @MVector ones(Int, N)
   current_block_dims = blockdims(T, Tuple(current_block_loc))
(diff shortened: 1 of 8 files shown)


mtfishman:
The one is issue I can think of with using `MPS` as an alias for `productMPS` is the following. With this PR, you have:
```julia
MPS(s, 2) # Create an MPS product state with each value set to 2, same as `MPS(s, fill(2, length(s)))`
randomMPS(s, 2) # Create a random MPS with bond dimension 2
```
I've added a constructor `MPS(s; linkdim=2)` that makes a bond dimension 2 MPS with `EmptyStorage` ITensors. So I would advocate that `randomMPS(s, 2)` gets deprecated in favor of `randomMPS(s; linkdim=2)`, and then `randomMPS(s, 2)` could have a consistent definition with `MPS(s, 2)`. In general, for making code generic between QN and non-QN MPS, it is best to specify the "starting state" of `randomMPS` anyway, so I think it is good to make `randomMPS(s, 2)` have the starting state interpretation.

@emstoudenmire, let me know what you think. The v0.2 release would be a good time to deprecate `randomMPS(s, 2)` in favor of `randomMPS(s; linkdim=2)` so then it can have the "starting state" interpretation in v0.3.

mtfishman:
Here is a summary of the changes made in this PR:

1. Functions like `flux` and `insertblock!` are more strict about expecting a `Block` type instead of a `Tuple`, to avoid ambiguity of something like `flux(T, (1, 2))` vs. `flux(T, Block(1, 2))` (i.e. does the first one mean a block or an index)? This also fixes a bug where the two different meanings were getting confused in dispatch, leading to an out-of-bounds error when computing the flux of a block and causing an error in `setindex!` of a QN ITensor.
2. `eachblock` and `eachdiagblock` now return iterators over `Block`. Before they were inconsistent (one returned `CartesianIndex` and another `Vector{Int}`).
3. `productMPS` is changed to an alias for `MPS`, and `MPS` will be favored as the syntax going forward.
4. `randomMPS(s, 2)` is deprecated in favor of `randomMPS(s; linkdims=2)` and `randomMPS(s, 1, 2)` is deprecated in favor of `randomMPS(s, 1; linkdims=2)` with a deprecation warning. In ITensors v0.3, `randomMPS(s, 2)` will be the same as `MPS(s, 2)`, i.e. `MPS(s .=> 2)` (a Vector of IndexVals specifying the nonzero value of each tensor). This will make the interfaces for the `randomMPS` and `MPS` constructors consistent.

Issue #605, opened 2021-03-25
Deprecate `Base.iterate(i::Index, state = 1)`


mtfishman:
Deprecate iteration of `Index` (which currently iterates through all possible `IndexVal`s of an `Index`) in favor of a more explicit iterator, like `eachindval`.

mtfishman:
Closed by #665.

Issue #281, opened 2020-04-15
[ITensors] `ITensor` iteration features


mtfishman:
It would be good to define better features for iterating over ITensors and other ITensor objects. For example:

- [x] Define `iterate(::ITensor, ...)` to iterate over all of the elements (both structurally zero and nonzero, for the case of sparse ITensors) (implemented in #457).
- [x] Define `Base.eachindex(A::ITensor)` for ITensors, similar to the Julia version which either outputs a linear index `1:dim(A)` or `CartesianIndices` over `1:dim(A,1),1:dim(A,2),...` depending on the indexing of the ITensor storage (implemented in #457).
- [x] Define `eachindvalset(::IndexSet/ITensor)` that returns an iterator over all `IndexValSet`s (implemented as `eachindval` and `eachval` in #665)
- [ ] Implement functions like [keys](https://docs.julialang.org/en/v1/base/collections/#Base.keys) and [pairs](https://docs.julialang.org/en/v1/base/collections/#Base.pairs) for ITensors for iterating over all `IndexValSet`s and values of the ITensor (a question would be if it should iterate over `IndexValSet` or `CartesianIndex`, both of which would be valid keys of an ITensor).
- [ ] Something like the Julia SparseArrays [findnz](https://docs.julialang.org/en/v1/stdlib/SparseArrays/#SparseArrays.findnz) method for iterating over the nonzero IndeValSets and values of sparse ITensors. This could be split up into `eachnz` to get the elements and `eachnzindval` to get just the non-zero IndexValSets.
- [ ] ~~Define `const IndexValSet{IndexT} = IndexSet{IndexVal{IndexT}}`~~ `IndexSet` was removed in #626 and `IndexVal` was removed in #665, so this would just be a `Tuple{Vararg{Pair{Index}}}`.

PR #665, opened 2021-05-17
Replace IndexVal with Pair{<:Index}. Add support for string values.

### LINKED ISSUE #503: More general state function
For many applications, it would be helpful to have a more general definition of `state`, where it can be any vector/order-1 tensor (not just a certain nonzero element, like it represents now). For example, for a "Qubit" site type, it would be helpful to be able to initialize a product state in an arbitrary basis, i.e.:
```julia
s = siteinds("Qubit", 4)
psi = productMPS(s, ["X+", "X-", "X+", "X-"])
```
Unfortunately, this doesn't quite fit into the paradigm of the current use of `state`, where it represents a single nonzero element/an Index set to a certain value.

One proposal for generalizing this would be to make `state` define an order-1 ITensor, i.e.:
```julia
import ITensors: space, state

space(::SiteType"Qubit") = 2

state(::StateName"X+", ::SiteType"Qubit", s::Index) = itensor([1/sqrt(2); 1/sqrt(2)], s)
state(::StateName"X-", ::SiteType"Qubit", s::Index) = itensor([1/sqrt(2); -1/sqrt(2)], s)

# productMPS would call the version of state above:
s = siteinds(4, "Qubit")
psi = productMPS(s, ["X+", "X-", "X+", "X-"])
```
Now, this would of course be breaking, since then it wouldn't be clear when the output of `state` should be an ITensor or just a value. We could introduce another function for the current usage of `state`, such as `val`:
```julia
import ITensors: val

val(::ValName"0", ::SiteType"Qubit") = 1
val(::ValName"1", ::SiteType"Qubit") = 2

s = siteind("Qubit")
T = ITensor(s)
T[s => "0"] = 1.2
T[s => "1"] = 2.4
```
I think something like `val` is a better name than `state` anyway for this usage, since `val` indicates it is used as the value of an IndexVal.

@emstoudenmire, let me what you think about this design change. We can also consider keeping the current usage of `state` the same, and think of another name for making an order-1 ITensor (like just using `op` in some way), but I can't really think of better names than the ones I'm proposing above.

emstoudenmire:
This PR replaces the IndexVal type with `Pair{<:Index}`. Currently `Pair{Index,Int}` and `Pair{Index,String}` work. This change requires less reasoning about extra types, since Pair is a built-in type, and is more forward-looking since we can extend this design to handle things like `Pair{Index{T},UnitRange{Int}}` for slicing with notation like `T[i=>3:5]`.

Overall this change actually reduces the amount of code, and simplifies the remaining code.

Closes #503


mtfishman:
Looks like `val(::Index, ::LastVal)` needs to be defined. As a reminder, I introduced `LastVal` as a way to handle indexing into an `ITensor` with `end`, so the definition should just be `val(i::Index, ::LastVal) = dim(i)`.

src/index.jl
```diff
@@ -586,7 +531,7 @@ function Base.show(io::IO, i::Index)
   end
 end
 
-Base.show(io::IO, iv::IndexVal) = print(io, ind(iv), "=>$(val(iv))")
+Base.show(io::IO, iv::Pair{<:Index}) = print(io, ind(iv), "=>$(val(iv))")
```
mtfishman, line 534: My guess is that we don't need this anymore, since Julia should already print it like this.

src/index.jl
```diff
@@ -439,7 +436,7 @@ Iterating over Index `I` gives the IndexVals `I(1)` through `I(dim(I))`.
 """
 function Base.iterate(i::Index, state::Int=1)
```
mtfishman, line 437: I was actually meaning to get rid of this definition (#605), since I think by reading code like:
```julia
i = Index(5)
for x in i
  @show x
end
```
it would not be clear to me what `x` would be (i.e. you might guess it could be each pair `i => n` or it could be iterating over the values `n` themselves). I think it would be better to deprecate iteration of `Index` in favor of devoted iterator objects `eachval(::Index)` and `eachindval(::Index)`, for example:
```julia
eachval(i::Index) = 1:dim(i)
eachindval(i::Index) = (i => n for n in eachval(i))
```
Would you mind adding those? I think this isn't a straightforward case of deprecating with the `@deprecate` macro, so we could just keep the `Base.iterate` definition for now and use `Base.depwarn` inside it directly, for example just adding the line:
```julia
Base.depwarn("iteration of `Index` is deprecated, use `eachindval` or `eachval` instead.", :iterate)
```
We could just add a `println` but this ensures it gets turned on and off the same way as other deprecation warnings from `@deprecate`.

A bonus would be adding the same iterator definitions but for collections of indices:
```julia
eachval(is::Tuple{Vararg{Index}}) = CartesianIndices(dims(is))
eachindval(is::Tuple{Vararg{Index}}) = (is .=> Tuple(ns) for ns in eachval(is))
```
to allow for something like:
```julia
A = randomITensor(i, j, k)
B = randomITensor(i, x, k, j)
for ivs in eachindval((i, j, k))
  B[x => 1, ivs...] = A[ivs...]
end
```
similar to what was discussed here: #281. I believe there is a similar iterator defined in the C++ version.

mtfishman:
Over all looks great, much simpler now. See comments above for some more work I'm sneaking into this PR :).

src/index.jl
emstoudenmire: Yes good opportunity to add these, and they will be good to have. Just added them.

src/indexset.jl
```diff
@@ -384,6 +384,37 @@ Create a CartesianIndices iterator for an Indices.
...
 
+"""
+    eachval(is::Index...)
+    eachval(is::Tuple{Vararg{Index}})
+
+Create an iterator whose values correspond to a 
+Cartesian indexing over the dimensions
+of the provided `Index` objects.
+"""
+eachval(is::Index...) = eachval(tuple(is...))
```
mtfishman, line 395: FYI if you just write `eachval(is)` Julia already makes `is` into a `Tuple`.

mtfishman:
Interesting, looks like some `getindex` and `setindex!` operations got a bit faster from this PR?

emstoudenmire:
That is interesting. Perhaps something about the Pair type being simpler or better optimized in some way? Or just the chain of functions involved being shorter or simpler? Anyway nice to see that. It does make our library code simpler too and more flexible as I mentioned above.

PR #667, opened 2021-05-20
Automatic fermion system

emstoudenmire:
This PR merges in most of the code for the automatic fermion system, which is off by default (toggled by a global flag). Though this system can correctly handle a wide range of cases, including an entire DMRG calculation and calculation of the single-particle density matrix of an MPS, there are still some known issues such as with the isometries produced by the SVD. Tests specifically for this new system are in the file test/fermions.jl.

- Bring over physics/fermions.jl from auto_fermion
- Introduce use_auto_fermion() global
- Fermion unit tests
- Tests related to get and set index passing
- Add fermion signs to dag
- Remove duplicate isfermionic definitions
- Fix isapprox for fermions
- Prefer to use storage instead of store
- More tests passing
- Make productMPS correct for fermions
- Reverse order of indices in eigen. More tests passing
- Merge auto fermion logic into AutoMPO system All fermion tests passing
- Update correlation_matrix for fermion system and test
- Some cleanup
- Fix bug with sign of AutoMPO link QNs
- Define isfermionic for Index
- Assert eigen dirs only for fermion system for now
- Partial test of MPS product system (other cases still buggy)
- Add fermions.jl to runtests
- Update fermions.jl to reflect NDTensors fixes
- Restore eigen D having same indices up to primelevel
- Formatting


src/physics/fermions.jl
```diff
@@ -18,3 +18,338 @@ function parity_sign(P)::Int
...
+    if fparity(iv_or_qn[p[j]]) == 1
+      n += 1
+      oddp[n] = p[j]
+    end
+  end
+  return parity_sign(oddp[1:n])
+end
+
+# Default implementation for non-QN IndexVals
+NDTensors.permfactor(p, ivs...) = 1.0
```
mtfishman, line 94: Probably should return `1` instead of `1.0` to match the QN version. 

Maybe in general it should accept a type parameter:
```julia
NDTensors.permfactor(T, p, ivs...) = one(T)
```
since `1` doesn't always preserve the type:
```
julia> typeof(one(Int32) * 1)
Int64
```
in the case that an ITensor has type `Int32`.

mtfishman:
Great! Looks like a pretty minimal amount of code.

My main concern is making sure the added logic doesn't affect the performance and behavior of the code when the automatic fermion sign system is turned off.

src/physics/fermions.jl
emstoudenmire: Good catch

src/itensor.jl
```diff
@@ -792,7 +792,9 @@ end
 @propagate_inbounds @inline function _getindex(T::Tensor, ivs::Vararg{<:Any,N}) where {N}
   # Tried ind.(ivs), val.(ivs) but it is slower
   p = NDTensors.getperm(inds(T), ntuple(n -> ind(@inbounds ivs[n]), Val(N)))
-  return _getindex(T, NDTensors.permute(ntuple(n -> val(@inbounds ivs[n]), Val(N)), p)...)
+  fac = NDTensors.permfactor(p, ivs...) #<fermions> possible sign
```
mtfishman, line 795: I thought I had made a comment here but maybe I didn't post it.

Does this affect the performance for the non-QN version?

emstoudenmire: For the non-QN version, this should just call a version of permfactor which is defined as a single-line function that just always returns the number +1 (without even checking a boolean flag or anything). But there's separately the practical question of timing the code to be sure of no hidden slowdowns.

mtfishman:
I see a 20% slowdown in the DMRG benchmark, maybe worth double checking that locally.

emstoudenmire:
Thanks - I realized that I read your question too narrowly. There's some additional code in the ITensor `dag` function that runs regardless of whether there are QN's or not. So I think that code should be moved into a separate function that does nothing in the case of dense ITensors, and only runs when there are QNs.

For the specific line of code you asked about in `getindex`, it shouldn't affect the dense case at all in theory, though this has to be checked. For the QN, but non-fermionic case there will a little more overhead in checking just that none of the QNs are fermionic so we'll have to check this.

But for now my goal is just to check that none of this code has a significant cost when the auto fermion system is turned *off*. So that's the minimal bar for this PR at the moment, if you agree.

mtfishman:
Yes, I am only concerned about whether the code you introduced affects the behavior or performance when the auto fermion system is turned off.

mtfishman:
So agreed that using dispatch on whether or not there are QNs is a good strategy.

emstoudenmire:
This PR should no longer change any performance characteristics or features of the main branch code, as long as `ITensors.using_auto_fermion() == false` which is the default setting. (See benchmark results above showing no DMRG regression.)

mtfishman:
Would it make sense to merge it at this point so it can be tested as an experimental feature?

emstoudenmire:
Yes – just wanted to pause and see if you had any more feedback, but I think it's good to merge now as a feature to be developed more over time.

PR #669, opened 2021-05-21
Minimize the number of swaps when applying nonlocal gates to MPS/MPO

mtfishman:
This adds functionality for looking ahead to the next gate to decide the strategy for swapping. The logic is fairly simple, basically swap towards the position of the next gate if they don't overlap, and as close to the next gate as possible if they do overlap.

Unfortunately (and surprisingly), when compared head to head to the current strategy of always moving the gate to the left (or right), this more advanced strategy actually performs worse in general. Go figures, the best laid plans etc etc. I'll attach some benchmarks after I do some more testing.


emstoudenmire:
Good to see this, though I'm also surprised it doesn't give a speedup.

Does it take into account how gates (usually) act on multiple sites, so additionally looking ahead in terms of how one might want to put the ortho center on either the leftmost or rightmost site of a gate after it acts? I'm thinking of a case like where gates repeatedly act on sites 3,4 then on 4,5, then again on 3,4, then 4,5, over and over. The ideal strategy would be to always put the ortho center on site 4 after applying either of these gates.

mtfishman:
Yeah, I also added logic for looking ahead to the next gate to determine where to put the orthogonality center, I forgot to mention that. The logic is [here](https://github.com/ITensor/ITensors.jl/pull/669/files#diff-30574d2b07b5598a5b432b188569f6d2daacd0e4ddae8a659a0fc257807c31a5R1884-R1894), though it looks like it isn't specialized for the case where the gates overlap right now so I'll have to add that.

PR #685, opened 2021-06-14
MPS product state constructor

### LINKED ISSUE #678: Use `state` instead of `val` in `MPS` product state constructor
It looks like `val` is still being used:
```julia
julia> MPS(s, "X+")
ERROR: ArgumentError: Overload of "val" function not found for Index tags "S=1/2,Site,n=1"
Stacktrace:
  [1] val(s::Index{Int64}, name::String)
    @ ITensors ~/.julia/dev/ITensors/src/physics/sitetype.jl:500
  [2] val
    @ ~/.julia/dev/ITensors/src/index.jl:511 [inlined]
[...]
```

emstoudenmire:
Closes #678

### DIFF
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -242,11 +242,9 @@ function MPS(::Type{T}, ivals::Vector{<:Pair{<:Index}}) where {T<:Number}
     return M
   end
 
-  if hasqns(ind(ivals[1]))
-    links = [Index(QN() => 1; tags="Link,l=$n") for n in 1:N]
-  else
-    links = [Index(1, "Link,l=$n") for n in 1:N]
-  end
+  space = hasqns(ind(ivals[1])) ? QN() => 1 : 1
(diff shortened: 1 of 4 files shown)


emstoudenmire:
The main change here is to just implement the MPS product state constructor taking "state" strings directly now, instead of converting to IndexVals as in 0.1.x. The IndexVal (now Pair{<:Index}) MPS product constructor is still there too as a separate option.

mtfishman:
Looks good to me, thanks!

The only concern I would have here is making sure that all previous usages of `val` now work with `state`, i.e. that the new `state` function has all of the equivalent definitions as the old `state` function/new `val` function. But I presume there is pretty good test coverage and if some definitions are missing then people will complain and we can add them.

emstoudenmire:
Good point – I had the same concern. I'm 99% sure it's covered by the tests, but also I just explicitly checked that the IndexVal (now Pair{<:Index}) MPS constructor does work properly. So this looks good to go!

PR #686, opened 2021-06-15
New interface `contract([A, [B, C]])` for contracting a sequence of tensors

mtfishman:
This adds `contract([A, [B, C]])` as an alternative syntax for `contract([A, B, C]; sequence=[1, [2, 3]])`. It first checks if the input just contains `ITensor`s, and if it doesn't it calls an internal function `deepcontract` that recursively calls contract.

This also adds tests for the different contraction interfaces by checking that the allocations of different contraction sequences are as expected.

Additionally, this adds an unexported function `deepmap` which can be used to convert between the different interfaces, i.e.
```julia
using ITensors.ContractionSequenceOptimization: deepmap
A = [A1, A2, A3] # ITensors to contract
deepmap(n -> A[n], [1, [2, 3]]) == [A[1], [A[2], A[3]]]
```

@LinjianMa let me know if there is other functionality that would be helpful.


mtfishman:
Also note that I've now allowed tags that are just plain integers in this PR, so you can do `Index(2, "1")`, `Index(2, "2")`, etc. We disallowed this very early on just to be on the conservative side, and avoid confusion between plain integer tags and prime levels, as well as avoid confusing things like `Index(2, "1,2")`, `Index(2, "2,3")`, etc. But I think it is nice to have this for simple test cases such as making a set of indices like:
```julia
N = 4
ts = ("$n" for n in 1:N)
d = 3
is = Index.(d, ts)
```
I think there was also an idea floating around that `Index(2, "1")` may actually be interpreted as a prime level, but at this point we are not going to do that so this notation is pretty unambiguous.

It reminds me that it might be nice to have a better syntax for specifying a prime level in the constructor, currently you have to do:
```julia
Index(2, plev=1)
Index(2, "x"; plev=1)
```
We could add:
```julia
```

PR #690, opened 2021-06-16
Make ITensors evaluate as unequal if they have different indices

mtfishman:
For example now:
```julia
i = Index(2)
A = randomITensor(i)
@show A == A'
```
will return `false` instead of throwing an error, which is the current behavior.

This is in line with the general Julia convention that things don't error when evaluating equality, and instead fall back to returning `false` which is helpful for generic code. You can always check `norm(A - B)` if you want to check equality that would error if the indices are different.

### DIFF
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1355,6 +1355,7 @@ adjoint(A::ITensor) = prime(A)
 dirs(A::ITensor, is) = dirs(inds(A), is)
 
 function (A::ITensor == B::ITensor)
+  !hassameinds(A, B) && return false
   return norm(A - B) == zero(promote_type(eltype(A), eltype(B)))
 end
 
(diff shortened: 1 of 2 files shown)


mtfishman:
Some more details on this design decision:

In Julia, `==` is designed such that anything can be compared without erroring:
```julia
julia> 1 == [1, 2]
false

julia> "xyz" == 12
false

julia> i = Index(2)
(dim=2|id=889)

julia> randomITensor(i', i) == 14
false

julia> [1, 2] == [1.0, 2.0]
true

julia> [1, 2]' == @view [1.0 2.0; 3.0 4.0][1:1, 1:2]
true
```
so we should be consistent with that convention for the sake of generic code, since often equality is used in a general way to determine if two objects can be considered the same in a universal sense (such as for hashing or determining if there are repeated objects in a collection, such as in a function like `unique`).

For a "stronger" version of equality, one can use an explicit elementwise comparison that errors if the indices are different like `norm(A - B) == 0`, or `all(A .== B)` once `.==` is implemented for ITensors.

PR #705, opened 2021-07-16
Code examples about making an array from an ITensor, QR with postive R, and sampling an MPS

emstoudenmire:
Improvements to code examples:
- new example about making an Array from an ITensor
- mention `positive=true` option for QR
- new example about sampling from an MPS


mtfishman:
Looks good! The section about conversion to `Array` reminds me that I would like to define `Array(T) = Array(T, inds(T))` for convenience (if you already know the ordering), but that is a separate issue and best to introduce the version where you specify the indices.

Also that section could mention that `array` can also return a copy if a permutation is required.

emstoudenmire:
Great – I wrote a little more about Array copying if a permutation is necessary.

PR #709, opened 2021-07-17
Started work on FAQs

emstoudenmire:


emstoudenmire:
Turns out the failing docs build in some of the commits here and from a previous PR are due to a bug in Documenter 0.27.4. So I pinned the Documenter version to 0.27.3 for now.

PR #711, opened 2021-07-21
Fix `state` function when overloading version that accepts an Index

mtfishman:


mtfishman:
Also started adding a "Qudit"/"Boson" type where this will be used, but I'll leave that for a future PR since I want to get the bug fixed so it can be used in PastaQ.

emstoudenmire:
Oof yes this is a pretty bad bug! Glad it is getting fixed.

PR #715, opened 2021-07-29
Hdf5 read compat

### LINKED ISSUE #714: upgrading v0.1 to v0.2, reading "old" MPS/ITensor format
Hello guys!

I have the following issue: I would like to convert an old MPS / ITensor saved on my disk (inside a HDF5 file) to the new type. Is there a way around downgrading to v0.1 when reading the old HDF5 files?
Could we build a function converting the old ITensor type to the new one?
Just wondering how I can save my computations while still having the bug fixes and features of v0.2 :thinking:

jtschneider:
This fixes #714 .


jtschneider:
Hello @mtfishman,

Let me know if you prefer a different kind of solution! Happy to change this. It was more or less a cheap and fast fix.

src/itensor.jl
```diff
@@ -2565,8 +2565,20 @@ function HDF5.read(
...
   inds = read(g, "inds", IndexSet)
 
-  stypestr = read(attributes(open_group(g, "storage"))["type"])
-  stype = eval(Meta.parse(stypestr))
-  storage = read(g, "storage", stype)
-  return itensor(storage, inds)
+  # check input file for key name of ITensor data
+  # ITensors.jl <= v0.1.x uses `store` as key 
+  # whereas ITensors.jl >= v0.2.x uses `storage` as key
+  if haskey(g, "storage")
```
mtfishman, line 2571: Maybe compress to something like:
```julia
for k in ["store", "storage"]
  if haskey(g, k)
    stypestr = read(attributes(open_group(g, k))["type"])
    stype = eval(Meta.parse(stypestr))
    storage = read(g, k, stype)
    return itensor(storage, inds)
  end
end
```

jtschneider: Okay, that makes sense!
Would you like to quietly leave the if condition without an error in a case (albeit a pathological one) where a stored ITensor does neither have key `storage` nor `store`?

jtschneider: Your above code does however render the function to have a predictable type as output...

mtfishman:
Thanks Jan!

I think there are a few options here:

1. The approach you take of just searching for the different field name options.
2. Reverting the field name to `"store"` in v0.2.
3. Make use of the `"version"` attribute that is set [here](https://github.com/ITensor/ITensors.jl/blob/v0.2.4/src/itensor.jl#L2540). The purpose of that attribute is to indicate breaking changes to the storage format, so then when you are reading you can know which format you are reading.

Unfortunately this breaking change to the storage format was accidental, so it is a bit too late to use options `2.` and `3.`, since there could be files written by people using ITensors v0.2.0 through v0.2.4 where we won't be able to detect that the breaking change happened in the file format. Possibly we could use a combination of `1.` and either `2.` and `3.`. I'm going to refer this to Miles since he designed the HDF5 reading and writing system so might have a preference about how to handle this situation.

In addition, it would be nice to test this functionality. Something you could do is write a file that saves an ITensor using `v0.1.41`, and then upload it as part of a test that then reads the file using `v0.2`.

src/itensor.jl
mtfishman: You could keep the same error message you had and put it after the for loop.

mtfishman:
Thanks @jtschneider!

@emstoudenmire do you think the approach here for handling the (accidental) break to the file format is good or should we make use of the `"version"` field as well?

test/readwrite.jl
```diff
@@ -152,6 +152,20 @@ include("util.jl")
...
+  @testset "DownwardCompat" begin
+    fi = h5open("testfilev0.1.41.h5", "r")
+
+    ITensorName = "ITensorv0.1.41"
+
+    # ITensor version <= v0.1.41 uses the `store` key for ITensor data storage
+    # whereas v >= 0.2 uses `storage` as key
+    @test haskey(read(fi, ITensorName), "store")
+
+    @test typeof(read(fi, ITensorName, ITensor)) == ITensor
```
mtfishman, line 165: Can shorten this to `read(fi, ITensorName, ITensor) isa ITensor`.

emstoudenmire:
The approach here is the right one. As you mentioned, Matt, this could have been a candidate for using the "version" attribute but since now some data could have been written with either "store" or "storage" yet with the same version, here we should keep the version number the same and just expand the definition of what version 1 allows, as is done by this PR.

PR #728, opened 2021-08-27
Generalize ITensor constructors to allow mixtures of collections of indices

mtfishman:
This PR allows ITensor constructors to accept inputs like `ITensor(i, [j, k], (l, m))`, which gets flattened to `ITensor(i, j, k, l, m)`.

This is to help with generic code, for example cases like the following:
```julia
julia> s = siteinds("S=1/2", 2);

julia> H = MPO(s, "Id");

julia> siteinds(H)
2-element Vector{Vector{Index{Int64}}}:
 [(dim=2|id=819|"S=1/2,Site,n=1")', (dim=2|id=819|"S=1/2,Site,n=1")]
 [(dim=2|id=604|"S=1/2,Site,n=2")', (dim=2|id=604|"S=1/2,Site,n=2")]

julia> randomITensor(siteinds(H))
ITensor ord=4 (dim=2|id=819|"S=1/2,Site,n=1")' (dim=2|id=819|"S=1/2,Site,n=1") (dim=2|id=604|"S=1/2,Site,n=2")' (dim=2|id=604|"S=1/2,Site,n=2")
ITensors.NDTensors.Dense{Float64, Vector{Float64}}
```
The above would be generic even if `H` was an `MPS` instead, in which case `siteinds(::MPS)` outputs a `Vector{<:Index}` instead of a `Vector{<:Vector{<:Index}}`. An eventual goal will actually be to make `siteinds(::MPS)` output a `Vector{<:Vector{<:Index}}` since in general `MPS` can have more than one site index, and functionality like this will make that switch easier.

Follow up work will involve making other ITensor functions more generic to accept mixtures of indices and collections of indices (such as MPS and MPO constructors, indexing operations, etc.). An eventual goal will be to make it so that code doesn't have to distinguish between index set operations like `commoninds(A, B)` and `commonind(A, B)`, which can get a bit annoying.


emstoudenmire:
I'm supportive of this extension, as it looks like you've found a clean way to implement it with the `indices` helper functions. 

Regarding the future change of the return value of `siteinds`, I would prefer if possible to keep the simpler and common case of an MPS with one index per site still returning an array of indices so one can do siteinds(psi)[1]. So then siteinds could return different types depending on what kind of MPS is passed to it, standard of non-standard. The only issue would be if this results in a type instability that Julia cannot work around or some other performance issue.

mtfishman:
That does indeed introduce some level of type instability, however I would consider that to be a "higher level" type instability so I don't think it is the primary concern here. My main concern with that design is the code logic may become complicated for a user trying to write code that is generic between "simple" MPS (with just one site Index per site) and more general MPS. If I try to write a generic function for MPS which uses the site indices like this:
```julia
function inner(A::MPS, B::MPS)
  sA = siteinds(A)
  sB = siteinds(B)
  # Do some analysis with sA and sB like check if they are the same:
  @assert hassameinds(sA[1], sB[1])
  # [...]
end
```
then the code that makes use of `siteinds(::MPS)` might need branches where it handles the case of `sA[1]` either being an `Index` or `Vector{<:Index}` separately. I've found that it is better to assume a more general MPS site index structure by default (I've often had to rewrite functions that I first wrote assuming just a simple MPS because I didn't anticipate that function being used with more general MPS that show up in PEPS or LPDO/Choi matrix calculations), which is why I would lean towards making the more general version the default behavior of `siteinds`, but this wouldn't happen until the `v0.3` release so we can take some time to think about it.

For now, the goal will be to replace uses of:
```julia
siteinds(A)
```
with:
```julia
siteinds(all, A)
```
and then see what kinds of changes (like this PR) that we need to make so that the resulting code "just works". We can reevaluate the default behavior of `siteinds(A)` in the future. Note that there are versions like `siteinds(only, A)` and `siteinds(first, A)` which will always just output `Vector{<:Index}`.

PR #731, opened 2021-08-31
Use `Compat` to make `blas_get_num_threads` simpler

mtfishman:
Use the great `Compat.jl` package to make `blas_get_num_threads` simpler (it is a version of `BLAS.get_num_threads()` that works for earlier versions of Julia, since it was only introduced in Julia 1.6).


emstoudenmire:
Less code that works more consistently is always welcome!

PR #735, opened 2021-09-02
Add a general function for writing and reading Arrays of ITensor objects as well as more compact save/load interface

mtfishman:
Add a general function for writing and reading Arrays of ITensor objects.

Also adds an experimental feature for automatically parsing a type when reading from a file with `ITensors.read`, as well as a more compact syntax `ITensors.save`/`ITensors.load` which avoids having to open a file.

For example:
```julia
i = Index(2)
A = randomITensor(i)
B = randomITensor(i)
C = randomITensor(i)
D = randomITensor(i)

X = [A B; C D]

# Original syntax with new `Array{ITensor}` reading/writing functionality
h5open("data.h5", "w") do file
  @write file X
end
X̃ = h5open("data.h5", "r") do file
  read(file, "X", Array{ITensor})
end
@show all(X .== X̃)

# Automatically parse the type, requires using the
# unexported `ITensors.read` function
X̃ = h5open("data.h5", "r") do file
  ITensors.read(file, "X")
end
@show all(X .== X̃)

# New more compact `ITensors.save`/`ITensors.load` syntax
ITensors.save("data.h5", "X", X)
X̃ = ITensors.load("data.h5", "X")
@show all(X .== X̃)
```


mtfishman:
Some follow up thoughts:

1. I created a function `hdf5_type` with definitions like `hdf5_type(::Type{<:Index}) = Index`. The idea is that it takes a general parametric ITensor type and returns the type which will be used in the `"type"` attribute in the HDF5 file. I think this could be used in other definitions of `HDF5.write`.
2. I think we could make a wrapper around `HDF5.read(file, name, ITensors.AutoType)`, for example we could define our own `ITensors.read` which is called as `ITensors.read(file, name)` which calls `HDF5.read(file, name, ITensors.AutoType)`.
3. I noticed that there are limitations with specifying the type being read too strictly, for example `read(file, "X", Index{Int})` doesn't work right now, and neither does `read(file, "X", Array{Index{Int}})`. I think it would be good to support that, but points to the challenge involved with having to pass the type, so the `ITensors.AutoType` version may be better in general.

mtfishman:
In the latest commit I added `ITensors.read` which automatically determines the type by `eval`ing the `"type"` attribute, used as follows:
```julia
i = Index(2)
A = randomITensor(i)
B = randomITensor(i)
C = randomITensor(i)
D = randomITensor(i)

X = [A B; C D]
h5open("data.h5", "w") do file
  @write file X
end

X̃ = h5open("data.h5", "r") do file
  ITensors.read(file, "X")
end
@show all(X .== X̃)
```
This works with any ITensor type.

I like the macro syntax `@write`/`@read` in HDF5.jl, perhaps we can also introduce `ITensors.@read`.

In fact, what I like even more is the `save`/`load` interface in `JLD.jl` and `JLD2.jl` where you don't have to open a file. Perhaps we could define our own `ITensors.save` and `ITensors.load` that wraps `h5open` and `read`/`write`.

mtfishman:
Ok, probably the last change. In the latest I added the more compact `ITensors.save`/`ITensors.load` syntax, inspired by `save`/`load` defined in `JLD`/`JLD2`:
```julia
i = Index(2)
A = randomITensor(i)
B = randomITensor(i)
C = randomITensor(i)
D = randomITensor(i)

X = [A B; C D]
ITensors.save("data.h5", "X", X)
X̃ = ITensors.load("data.h5", "X")

@show all(X .== X̃)
```
I guess the PR has diverged a bit from the original goal. @emstoudenmire, let me know what you think of both the `Array{ITensor}` read/write functionality as well as the more compact `ITensors.save`/`ITensors.load` interface.

mtfishman:
In the latest, I added a fallback to the version of `ITensors.read` where no type is specified, so that if there is not `"type"` attribute in the file it is assumed that the HDF5 file was not made by ITensor and tries to use the general `HDF5.read` function. This allows you to do:
```julia
julia> x = [1, 2, 3]; ITensors.save("data.h5", "x", x); ITensors.load("data.h5", "x")
3-element Vector{Int64}:
 1
 2
 3

julia> x = Index(2); ITensors.save("data.h5", "x", x); ITensors.load("data.h5", "x")
(dim=2|id=800)
```
which I think makes it more useful. A design question that comes up is, what defines the minimal structure of the file such that it is an "ITensor" HDF5 file? I would think ideally there would actually be some sort of "ITensor" attribute that we could check for, but for now the check is basically:
```julia
  type_string = try
    @suppress_err begin
      g = open_group(parent, name)
      type_attribute = attributes(g)["type"]
      HDF5.read(type_attribute)
    end
  catch
    # If the file doesn't have the format
    # expected of ITensor HDF5 files, return `nothing`
    return nothing
  end
```
i.e. it sees if there is a group with the variable name and if it has the `"type"` attribute. It turns out that Julia types saved with HDF5 don't have a group with the variable name, so that already fails.

Also in the latest I added support for saving `Array`s of `MPS`/`MPO`, which is useful for things like boundary MPS.

src/index.jl
```diff
@@ -596,17 +596,17 @@ function readcpp(io::IO, ::Type{Index}; kwargs...)
...
   dir = dir_int < 0 ? In : Out
-  read(io, 8) # Read default IQIndexDat size, 8 bytes
+  HDF5.read(io, 8) # Read default IQIndexDat size, 8 bytes
   return Index(id, dim, dir, tags)
 end
 
 function HDF5.write(parent::Union{HDF5.File,HDF5.Group}, name::AbstractString, I::Index)
   g = create_group(parent, name)
-  attributes(g)["type"] = "Index"
+  attributes(g)["type"] = "$(hdf5_type(I))"
```
emstoudenmire, line 609: Yes I think `hdf5_type` is a good idea

src/readwrite.jl
```diff
@@ -13,3 +13,42 @@ function readcpp(io::IO, ::Type{Vector{T}}; kwargs...) where {T}
...
+  type_string = try
+    @suppress_err begin
+      g = open_group(parent, name)
+      type_attribute = attributes(g)["type"]
+      HDF5.read(type_attribute)
+    end
+  catch
+    # If the file doesn't have the format
+    # expected of ITensor HDF5 files, return `nothing`
+    return nothing
```
emstoudenmire, line 29: One question I would have is whether the functionality in line 40 below (`isnothing(r) && return HDF5.read(parent, name)`) should just be put here in this overload of HDF5.read?

emstoudenmire: Then both this `HDF5.read` overload and `ITensors.read` would have this fallback behavior.

src/readwrite_array.jl
```diff
@@ -0,0 +1,39 @@
...
+  g = create_group(parent, name)
+  attributes(g)["type"] = "$(hdf5_type(d))"
+  attributes(g)["version"] = 1
+  write(g, "size", collect(size(d)))
+  for c in CartesianIndices(d)
+    write(g, "$(Tuple(c))", d[c])
+  end
+end
+
+function HDF5.read(
```
emstoudenmire, line 24: Maybe we should switch to using this for MPS and MPO in version 2 of the ITensor HDF5 file format for those types.

emstoudenmire (approved):
Great design ideas & missing features. This will make it all a lot nicer, and the AutoType feature is something I did have in mind to add eventually.

Here are two questions about the PR:

- Does the `ITensors.save` function append to an existing file? Or does it overwrite the file, so that afterward the data is only the one from the last call to `save`? (I couldn't find the source code for `ITensors.load` and `ITensors.save` for some reason.) 
- The issue you raised about `read` not working with types like `Index{Int}` is a good catch. Could we fix that by just changing the function signature to be more generic? We could still have an internal check inside the function that the actual type returned is the exact type (if concrete) passed by the user. It might be best as a bugfix or future PR.

emstoudenmire:
Ok I figured out my question about save/load. It is renamed to h5write/h5read. So I was able to answer my own question about the file behavior of h5write, which is "cw" and won't overwrite existing contents.

mtfishman:
> Great design ideas & missing features. This will make it all a lot nicer, and the AutoType feature is something I did have in mind to add eventually.
> 
> Here are two questions about the PR:
> 
>     * Does the `ITensors.save` function append to an existing file? Or does it overwrite the file, so that afterward the data is only the one from the last call to `save`? (I couldn't find the source code for `ITensors.load` and `ITensors.save` for some reason.)

Yes, as you figured out I changed the name to `ITensors.h5write`/`ITensors.h5read` to match `HDF5` and also changed the writing behavior to append to the file with `"cw"`. `save` and `load` are actually functions defined in [FileIO.jl](https://github.com/JuliaIO/FileIO.jl) which is supposed to help with automatically detecting file formats from file extensions, so it seems like `HDF5.h5write`/`HDF5.h5read` were a closer match for the functionality we are interested in.

> 
>     * The issue you raised about `read` not working with types like `Index{Int}` is a good catch. Could we fix that by just changing the function signature to be more generic? We could still have an internal check inside the function that the actual type returned is the exact type (if concrete) passed by the user. It might be best as a bugfix or future PR.

Yes, the goal of the `hdf5_type` is to take a general input type and turn it into a standardized type that is stored in the HDF5 file. So I think that should mostly solve this issue, it is just a matter of using it and testing it out. I think a question I was stuck on is that the writing functionality is simple (just map the type with `hdf5_type`), but then when you are reading a file and you explicitly use the version that takes the type, there seemed to be some ambiguity about how specific you should be when specifying the type.

For example, if you save an `Index{Int64}`, internally the type gets saved as `Index`. Then if you load it with `read(file, "index", Index{Vector{Pair{QN, Int64}}})`, probably that shouldn't work, because the type is wrong. So when reading the file, it doesn't make sense to map the type with `hdf5_type`. My conclusion was that maybe we should just make the external interface the automated functions where you don't specify the type, and when reading the file it just takes the type very literally and it should match the type that is saved (otherwise I think it would add a decent amount of code complexity to determine that `Index{Vector{Pair{QN, Int64}}}` is incorrect in that situation, since you would have to try to read the `Index` first and then detect if the result matches the specified type in some way).

src/readwrite_array.jl
mtfishman: Yes, part of the goal of this design would be to use it in more complicated types like MPS and MPO. That would be a good use of the version number.

src/readwrite.jl
mtfishman: Good suggestion, I can change that.

mtfishman:
Another design question I have been thinking about is that in principle it is not necessary for us and other users with custom ITensor types to overload `HDF5.read` and `HDF5.write`. Instead we could overload our own `ITensors.read` and `ITensors.write` functions. I actually prefer this, since `HDF5.read` doesn't really have an analogous version that accepts a type as the last input, from what I could see. For the time being, we could change over to overloading `ITensors.read`/`ITensors.write` internally, and make an internal `HDF5.read`/`HDF5.write` overload that calls the ITensors versions, and recommend overload `ITensors.read`/`ITensors.write` for custom types (I believe that will work, though I'll have to try it out and see if it all works smoothly and doesn't involve a bad case of piracy). @emstoudenmire do you have an opinion on this?

PR #738, opened 2021-09-06
Add syntax highlighting to bibtex entry

paniash:


emstoudenmire:
Thank you -  we’ll merge once we confirm that the updated documentation looks correct.

PR #744, opened 2021-09-20
Limit maxdim in default contract

### LINKED ISSUE #422: Set a good default truncation in `contract(::MPO, ::MPS)`
There is a memory problem in the contraction between an MPS and an MPO using the default "densitymatrix" method:

```julia
function test(N)
  sites = siteinds("S=1/2",N);
  mps = randomMPS(sites);
  mpo = randomMPO(sites);
  @time mpo * mps;
end
```
The result is:
```
N = 8	0.003883 seconds (14.31 k allocations: 3.249 MiB)
N = 9	0.006476 seconds (16.38 k allocations: 8.588 MiB)
N = 10	0.017550 seconds (18.46 k allocations: 29.048 MiB)
N = 11	0.079618 seconds (20.55 k allocations: 109.751 MiB)
N = 12	0.691359 seconds (22.64 k allocations: 431.062 MiB)
N = 13	6.165078 seconds (24.74 k allocations: 1.673 GiB)
N = 14	58.80654 seconds (26.84 k allocations: 6.678 GiB)
```

The test works fine with `*(mpo,psi,method="naive")`.

emstoudenmire:
Fixes #422. Use bound of MPS times MPO bond dimensions to limit bond dimension in default, non-truncating contract(::MPO,::MPS).

- Implement rigorous default maxdims for _contract_densitymatrix
- Test for contract(MPO,MPS) without truncation

### DIFF
--- a/src/mps/mpo.jl
+++ b/src/mps/mpo.jl
@@ -449,7 +449,8 @@ function _contract_densitymatrix(A::MPO, ψ::MPS; kwargs...)::MPS
 
   ψ_out = similar(ψ)
   cutoff::Float64 = get(kwargs, :cutoff, 1e-13)
-  maxdim::Int = get(kwargs, :maxdim, maxlinkdim(A) * maxlinkdim(ψ))
+  maxdim_defined = haskey(kwargs,:maxdim)
+  max_maxdim::Int = get(kwargs, :maxdim, maxlinkdim(A) * maxlinkdim(ψ))
   mindim::Int = max(get(kwargs, :mindim, 1), 1)
   normalize::Bool = get(kwargs, :normalize, false)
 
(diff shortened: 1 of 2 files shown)


mtfishman:
@emstoudenmire could you merge from the `main` branch to update the CI setup on this branch? I was fiddling with setup to start testing Julia 1.7.0-rc1 and there were some issues with the setup (https://discourse.julialang.org/t/julia-v1-7-0-rc1-is-now-available/68278/5) which should now be resolved.

mtfishman:
Some test failures on 1.7 but independent of this PR.

src/mps/mpo.jl
```diff
@@ -491,14 +492,24 @@ function _contract_densitymatrix(A::MPO, ψ::MPS; kwargs...)::MPS
...
   for j in reverse(2:(n - 1))
+    if maxdim_defined
+      maxdim = max_maxdim
+    else
+      maxdim = 1
+      cip = commonind(ψ[j], E[j - 1])
+      ciA = commonind(A[j], E[j - 1])
+      (cip != nothing) && (maxdim *= dim(cip))
+      (ciA != nothing) && (maxdim *= dim(ciA))
+      (cip == nothing && ciA == nothing) && (maxdim = max_maxdim)
```
mtfishman, line 503: Couldn't we always compute this "locally maximum dimension", and than define maxdim as `min(maxdim, max_maxdim)`?

Also if you use:
```julia
      cip = commoninds(ψ[j], E[j - 1])
      ciA = commoninds(A[j], E[j - 1])
```
then you should just be able to generically use `dim(cip)` and `dim(ciA)`, which are defined to be `1` if there are no indices. This should make the code logic simpler.

Finally, in the case when `cip == nothing && ciA == nothing`, can we safely make `maxdim = 1`? That's what would happen with my suggestion above.

emstoudenmire: I like all those suggestions - thanks. Let me see what I can do to simplify that code, as I agree it's kind of unwieldy.

src/mps/mpo.jl
```diff
@@ -450,7 +450,7 @@ function _contract_densitymatrix(A::MPO, ψ::MPS; kwargs...)::MPS
   ψ_out = similar(ψ)
   cutoff::Float64 = get(kwargs, :cutoff, 1e-13)
   maxdim_defined = haskey(kwargs, :maxdim)
```
mtfishman, line 452: I guess we don't need this anymore.

emstoudenmire: Good catch! Removed

mtfishman:
Looks good, thanks! Feel free to merge when tests pass.

Issue #741, opened 2021-09-13
Allow multi-site operators in `OpSum`


mtfishman:
Allow multi-site operators in `OpSum` as well operator-dependent keyword arguments. For example:
```julia
os = OpSum()
os += 0.5, "CX", 1, 3
```
This would make it a bit easier to define Hamiltonians for forming MPOs as well as a Trotter-Suzuki decomposition interface that @GTorlai is working on for PastaQ.

For the sake of making MPOs, it could be implemented by first forming the operator and then doing an expansion in a local operator basis in order to transform it into a sum of products of local operators, but we could still add support for this in the `OpSum` data structure before the MPO constructor support is implemented so it can be used for other applications.

In addition, it would be helpful to allow adding keyword arguments for operators with parameters, for example:
```julia
os += 1.2, "CRz", 1, 4, (ϕ = π/3,)
```

emstoudenmire:
This feature could potentially take advantage of issue #722 when that is implemented. Then one could, say, preprocess "CX" into a sum of products of operators A1*A3 + B1*B3 +... but where now the A, B, etc. operators can be any matrix. So then this feature would not have to rely on a complete basis of operators being defined for the site type and/or could use an SVD to split the operator.

mtfishman:
Closed by #749.

PR #749, opened 2021-09-23
Add support for multisite operators in OpSum

mtfishman:
Add support for multisite operators in OpSum (no MPO construction support yet).

For example:
```julia
julia> os = OpSum()
OpSum:


julia> os += "CX", 1, 2
OpSum:
  1.0 "CX"(1, 2) 


julia> os += "Z", 3
OpSum:
  1.0 "CX"(1, 2) 
  1.0 "Z"(3) 
```
The current application is for creating a trotter decomposition from a Hamiltonian as discussed in #741.

@GTorlai is this sufficient for what you had in mind? Do we also need support for parameters?

### DIFF
--- a/src/physics/autompo.jl
+++ b/src/physics/autompo.jl
@@ -12,7 +12,8 @@ struct SiteOp{N}
   site::NTuple{N,Int}
   params::NamedTuple
 end
-SiteOp(name::String, site::Tuple) = SiteOp(name, site, (;))
+# Change NamedTuple() to (;) when we drop older Julia versions
+SiteOp(name::String, site::Tuple) = SiteOp(name, site, NamedTuple())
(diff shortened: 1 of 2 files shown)


GTorlai:
Thanks for taking care of this. And yes, we do need the parameters as well.

mtfishman:
In the latest, I added support for passing keyword arguments as a `NamedTuple`:
```julia
os = OpSum()
os += "CRz", 1, 2, (ϕ=π/3,)
```
or:
```julia
os += "CRz", (ϕ=π/3,), 1, 2
```

An interesting extension would be to allow:
```julia
os += exp, "CRz", (ϕ=π/3,), 1, 2
```
i.e. a function could be passed as the first argument to specify an operation to perform on the operator.

PR #755, opened 2021-09-30
Fix test truncation test for Julia 1.7

mtfishman:
One of the truncation tests was failing on Julia 1.7 because the RNG changed (https://github.com/JuliaLang/julia/blob/v1.7.0-beta1/NEWS.md#random) and the new one lead to an accidental degeneracy in the singular value spectrum, leading to an extra singular value being truncated. This fixes the tests accounting for that change.


emstoudenmire:
Great. I’ll go ahead and tag a new version after this, since this (from a developer side) and the other PR were pretty serious bugs.

PR #761, opened 2021-10-21
Add `ChainRules` rules for basic reverse mode AD rules for ITensors

mtfishman:
This adds rules for basic operations like adding, contracting, priming, tagging, etc. For example:
```julia
julia> using ITensors

julia> i = Index(2, "i")
(dim=2|id=3|"i")

julia> A = randomITensor(i', dag(i))
ITensor ord=2 (dim=2|id=3|"i")' (dim=2|id=3|"i")
ITensors.NDTensors.Dense{Float64, Vector{Float64}}

julia> f(A) = (A' * A)[1, 1]

julia> f(A)
2.01897986704533

julia> @show f'(A);
(f')(A) = ITensor ord=2
Dim 1: (dim=2|id=3|"i")'
Dim 2: (dim=2|id=3|"i")
ITensors.NDTensors.Dense{Float64, Vector{Float64}}
 2×2
 1.914329191679339   1.109598478276638
 0.9938872710327071  0.0
```
Big missing features are rules for MPS operations and decompositions.

These rules are being moved from: https://github.com/ITensor/ITensorNetworkAD.jl so they can be developed more closely with the development of ITensor (for example we can see if any rules break if we make changes to basic ITensor operations).

### DIFF
--- a/test/ITensorChainRules/test_optimization.jl
+++ b/test/ITensorChainRules/test_optimization.jl
@@ -91,7 +91,7 @@ include("utils/circuit.jl")
     # The variational circuit we want to optimize
     function variational_circuit(θ⃗)
       N = length(θ⃗)
-      return vcat(Rylayer(N, θ⃗), CXlayer(N), Rylayer(N, θ⃗), CXlayer(N))
+      return vcat(Rylayer(N, θ⃗), CXlayer(N))
     end
 
     N = 4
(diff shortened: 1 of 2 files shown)


mtfishman:
Also note that this does not add any dependencies on actual AD or optimization codes, it only relies on the lightweight package `ChainRulesCore` which provides the interface for defining derivative rules (and `ZygoteRules` for technical reasons about overloading derivatives for the `A'` interface for priming ITensors).

PR #770, opened 2021-11-09
Add show for Indices

emstoudenmire:
Defines `show` for collections of Indices (the `Indices` Union type). Very helpful in commands such as `@show inds(T)` where T is an ITensor and when the indices have QNs.

### DIFF
--- a/src/indexset.jl
+++ b/src/indexset.jl
@@ -856,7 +856,16 @@ ITensors.block(is, 1, 2) == (1,1)
 """
 block(inds::Indices, vals::Integer...) = blockindex(inds, vals...)[2]
 
-#show(io::IO, is::IndexSet) = show(io, MIME"text/plain"(), is)
+function show(io::IO, mime::MIME"text/plain", is::Indices)
+  for i in is
+    println(io, i)
+  end
+end
+
+function show(io::IO, is::Indices)
+  !isempty(is) && println(io)
+  return show(io, MIME"text/plain"(), is)
+end
 
 #
 # Read and write


mtfishman:
Thanks, this seems reasonable. Two comments:

1. Should we define a special compact printing for when the Indices are in a collection, like is being proposed for MPS printing in #772? For example these two cases could be handled differently:
```julia
julia> is = (i'', i', i)
((dim=2|id=778)'', (dim=2|id=778)', (dim=2|id=778))

julia> [is, is]
2-element Vector{Tuple{Index{Int64}, Index{Int64}, Index{Int64}}}:
 ((dim=2|id=778)'', (dim=2|id=778)', (dim=2|id=778))
 ((dim=2|id=778)'', (dim=2|id=778)', (dim=2|id=778))
```
2. Should we indicate something about the actual type of the collection, like `Tuple` vs. `Vector`?

mtfishman:
One issue with the plan above is that the way to define the compact version of showing an object is by overloading `show(::IO, args...)` which is what `@show` uses, so `@show` would become less useful. There is an open issue proposing adding a new macro `@display` (https://github.com/JuliaLang/julia/issues/15820) which would be like `@show` but call the non-compact version of `show`. I think we should consider adding that functionality as an internal macro (maybe called `@show_verbose` or something like that).

mtfishman:
Maybe the `compact` attribute of `IO` could be used for this? See:
https://discourse.julialang.org/t/show-and-showcompact-on-custom-types/8493
https://github.com/JuliaLang/julia/blob/9028130434c5fb55e00c65e73aaf13cf0afde380/base/complex.jl#L181-L196

PR #773, opened 2021-11-11
Fix itensor rrule

GTorlai:
This PR fixes a bug in the `rrule` for the `itensor` function, which now takes into account different index ordering. Before this fix, the gradients of these two functions were identical:

```julia
i = Index(2, "i")
j = Index(2, "j")
M = rand(2,2,2,2)

function f1(x)
  T = itensor(M, i',j',i,j)
  ...
end
  
function f2(x)
  T = itensor(M, i,j,i',j')
  ...
end
```


mtfishman:
Looks good, thanks. I'll merge once tests finish running.

emstoudenmire:
Thanks!

PR #777, opened 2021-11-23
Move `NDTensors` to a subdirectory to prepare for registering

mtfishman:
For now this is simply a reorganization, but it is in preparation for going back to having `NDTensors` as a registered Julia package.

The idea is that the code itself for `NDTensors` will live inside this repository, but it will be registered as it's own Julia package, which is possible as a result of some work to handle this use case from the side of Julia's package manager:

- https://github.com/JuliaLang/Pkg.jl/pull/1422
- https://github.com/JuliaLang/Pkg.jl/pull/1766
- https://github.com/JuliaRegistries/Registrator.jl/pull/287

Note this development workflow is used by some high profile packages like [Makie](https://github.com/JuliaPlots/Makie.jl) and [SnoopCompile](https://github.com/timholy/SnoopCompile.jl), so things like CI, package registration, and development should be smooth and we can use those packages as inspiration for how to set things up.

Part of the goal is to test this out for `NDTensors` first and then move packages that are tightly coupled to the development of ITensors.jl into this repository, like:

- https://github.com/ITensor/ITensorsVisualization.jl
- https://github.com/ITensor/ITensorsGPU.jl
- https://github.com/ITensor/ITensorGaussianMPS.jl
- https://github.com/mtfishman/ITensorInfiniteMPS.jl
- https://github.com/ITensor/ITensorsBenchmarks.jl
- etc.

and also maybe consider splitting up the ITensors module into more pieces, for example spin off `ContractionSequenceOptimization` and `LazyApply` into their own registered packages. This will make development easier, so we can immediately see if changes to `ITensors` affects these packages with CI and then update them all at once if needed.


mtfishman:
Some reference PRs from when Makie made this change:

- https://github.com/JuliaPlots/Makie.jl/pull/1020
- https://github.com/JuliaRegistries/General/pull/38233

PR #791, opened 2021-12-02
Move `ITensorGPU` into `ITensors` repository

mtfishman:


mtfishman:
@kshyatt I see a Gitlab CI file, what is that for? I'm going to look into setting up CI so this gets tested against the latest ITensors/NDTensors for each PR/commit.

kshyatt:
That's from back in the day when we were running JuliaGPU CI on gitlab

PR #794, opened 2021-12-13
[ITensorGPU] Add Jenkins testing CI

mtfishman:
@dylex @kshyatt


mtfishman:
@kshyatt Dylan found that the test of `CUTENSOR_AUTOTUNE` was tough to get working with the Jenkins build, is that part needed now? I couldn't find any information about that environment variable in CUDA.jl or the internet at large.

PR #801, opened 2022-01-18
Add docs for enabling debug checks

mtfishman:
Add docs for enabling and disabling debug checks to the Getting Started sections of the docs.


mtfishman:
@emstoudenmire do you think this is a good place in the docs for this?

PR #806, opened 2022-01-26
Change `randomMPS` bond dim error to warning

mtfishman:
Currently, `randomMPS` errors when the bond dimension doesn't reach the desired one. This appears to happen fairly often with quantum numbers, probably because of technicalities of quantum number conservation (for example, to reach the desired bond dimension it might be trying to add a block inconsistent with the flux). This changes it into a more informative warning.


mtfishman:
Related to #763, but this doesn't really close that since we should still investigate the underlying problem of why it doesn't reach the desired bond dimension in some cases.

Issue #803, opened 2022-01-19
[ENHANCEMENT] Improve format Github action


mtfishman:
We could try this Github action for formatting:
```yml
name: Format suggestions

on:
  pull_request:

concurrency:
  # Skip intermediate builds: always.
  # Cancel intermediate builds: only if it is a pull request build.
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}

jobs:
  format:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: julia-actions/setup-julia@latest
        with:
          version: 1
      - run: |
          julia  -e 'using Pkg; Pkg.add("JuliaFormatter")'
          julia  -e 'using JuliaFormatter; format("."; verbose=true)'
      - uses: reviewdog/action-suggester@v1
        with:
          tool_name: JuliaFormatter
          fail_on_error: true
          filter_mode: added
```
used by [ChainRulesCore.jl](https://github.com/JuliaDiff/ChainRulesCore.jl/blob/95612d0860f9356a426e3372aafce7feb4e933d0/.github/workflows/format.yml), which I think actually makes suggestions for formatting issues to change on the PR (whereas ours currently just says if there is a formatting issue, but doesn't suggest a change to make).

emstoudenmire:
I was actually looking the other week to see if there was a way to do this. Would be good to have.

Could the action just run the formatter no matter what? Because in any case when the formatting is wrong, the next step is always to just run the formatter. Or I guess is the issue purely technical about needing to make a new commit after the formatter runs?

mtfishman:
I was hoping that is what the above Github Action will do (make a commit to the PR with proposed changes to fix the formatting), but I haven't tested it.

Currently, the Github Action (https://github.com/ITensor/ITensors.jl/blob/8f8229a3ed56012a0252a30f49a421f157bf338a/.github/workflows/format_check.yml) does run no matter what (at every commit to a PR or the main branch), it just doesn't propose a change or make a new commit with formatting changes.

We also have this Github Action (https://github.com/ITensor/ITensors.jl/blob/8f8229a3ed56012a0252a30f49a421f157bf338a/.github/workflows/format_pr.yml) which makes a new PR periodically with formatting fixes. But it would nice if these behaviors were combined into a single action that directly aids when making a PR.

emstoudenmire:
[...]

PR #811, opened 2022-01-28
[ITensors] [ENHANCEMENT] Start implemented `rrule` for `apply`

mtfishman:
@GTorlai this is very much work in progress, but this should help give you an idea for what the `rrule` would look like.

You can see that it looks very similar to the `rayleigh_quotient` `rrule` you wrote in PastaQ:

https://github.com/GTorlai/PastaQ.jl/blob/560406f060cf5be9eb0c552ff28887d9942f9341/src/autodiff.jl#L72-L97

The next step is to complete the `rrule` for `inner(::MPS, ::MPO, ::MPS)` (you can see I've started a pseudocode version). I'll finish that up and test it with the example in `examples/autodiff/apply.jl`, which has a fairly minimal cost function making use of `apply` and `inner`. It reminds me that I need to dig back into the `Op` system to get AD working with that so we can make these examples simpler...


mtfishman:
@GTorlai I got the `inner` `rrule` working so it looks like the new example is working now (at least it is running).

mtfishman:
@GTorlai in the latest I got the gradient of `op` working, so we can do the same kinds of gradients through operator definitions like we do with `gate` in PastaQ.

I've edited the state preparation and VQE examples and they are a lot simpler now, since they just use basic ITensor operations like `op`, `apply`, and `inner`. That should be sufficient for many of our circuit optimization examples in PastaQ as well.

I'm going to merge this once we have tests passing, so we can build on top of it with things like noisy circuit optimization (for that, I think basically all we need to do is extend the `apply` `rrule` to `MPO` evolution), and test it out with fermions.

mtfishman:
@emstoudenmire I also tweaked the formatting CI setup based on the Github Action I suggested adding in #803, and additionally tweaked the current formatting CI we have to print out the formatting diffs.

So now we have 3 formatting CI runs:

1. `format_check.yml` - This runs for every PR and commit to `main` and checks if the formatting matches the style guide. If it doesn't, as of this PR it will print the issues in the formatting to the Github Actions log file.
2. `format_pr.yml` - This periodically checks if the formatting of `main` is incorrect, and if it is not it makes a PR to fix the formatting.
3. `format_suggestions.yml` - This is a new one based on the [ChainRulesCore.jl formatting Github Action](https://github.com/JuliaDiff/ChainRulesCore.jl/blob/1d076fa7704457dadee9b0b293e63cbe6f0093ad/.github/workflows/format.yml), which uses [reviewdog](https://github.com/reviewdog/reviewdog) to make suggestions on a PR for how to fix the formatting, if the formatting is off. This is a new one I've added in this PR. I believe this closes #803 in a satisfactory way (it gives a simple interface for committing fixes to the PR).

I think in the end we will be able to drop 1. and 2. if 3. works well enough. I'll keep them around for now until we get a chance to test out 3. more. For example, there may be issues getting 3. to work with forks: https://github.com/JuliaDiff/ChainRulesCore.jl/issues/489.

Issue #810, opened 2022-01-28
[ITensors] [BUG] `correlation_matrix` nonsymmetric constraint


mtfishman:
@emstoudenmire, the symmetric constraint in `correlation_matrix` seems odd for cases like this:
```julia
using ITensors
using LinearAlgebra
using Random

Random.seed!(1234)

N = 2
m = 2

s = siteinds("S=1/2", N)
psi = randomMPS(s, n -> isodd(n) ? "Up" : "Dn"; linkdims=m)
C = correlation_matrix(psi, "Sz", "Sx")

C_mpo = zeros(N, N)
for i in 1:N, j in 1:N
  os = OpSum()
  os += "Sz", i, "Sx", j
  C_mpo[i, j] = inner(psi, MPO(os, s), psi)
end

println("correlation_matrix")
display(C)

println("\nMPO")
display(C_mpo)
```
which outputs:
```julia
correlation_matrix
2×2 Matrix{Float64}:
 -3.32551e-18   0.234341
  0.234341     -1.45217e-18

MPO
2×2 Matrix{Float64}:
 -9.34059e-19  0.234341
 -0.219799     1.45217e-18
```

Should we have a keyword argument `ishermitian` so that a user can specify that they expect the correlation matrix to be Hermitian? I would say that it should be off by default to be on the safe side, since the symmetric constraint is just an optimization.

Originally pointed out by @JanReimers.

mtfishman:
In https://github.com/ITensor/ITensor/pull/405, in the end this was handled automatically by forming the operators at the beginning and checking if they are Hermitian conjugates of each other (up to some tolerance), in which case the correlation matrix is deduced to be Hermitian. Additionally, users can set a flag to explicitly tell us if they want the correlation matrix to be constrained to be Hermitian or not. That would be a good way to go here as well.

mtfishman:
Closed by #817.

PR #817, opened 2022-02-05
[ITensors] [BUG] correlation_matrix nonsymmetric constraint #810

JanReimers:
https://github.com/ITensor/ITensors.jl/issues/810
-Right now for <C_i*C_j> is simply calculated the below diag elements explicitly.
-Added unit tests for many possible non hermition correlations.

### DIFF
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -604,8 +604,8 @@ function correlation_matrix(psi::MPS, _Op1::AbstractString, _Op2::AbstractString
 
   # Decide if we need to calculate a non-hermitian corr. matrix which is roughly double the work.
   is_cm_hermitian = false #Assume corr-matrix is non-hermitian
-  if haskey(kwargs, :is_hermitian) #Did the user explicitly request something?
-    is_cm_hermitian = kwargs.is_hermitian #Honour users request
+  if haskey(kwargs, :ishermitian) #Did the user explicitly request something?
+    is_cm_hermitian::Bool = get(kwargs, :ishermitian, false) #Honour users request
   else
     O1 = op(Op1, s, 1)
     O2 = op(Op2, s, 1)
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -590,7 +590,7 @@ function correlation_matrix(psi::MPS, _Op1::AbstractString, _Op2::AbstractString
   ElT = promote_itensor_eltype(psi)
   s = siteinds(psi)
 
-  Op1=_Op1 #make copies into which we can insert "F" string operators.
+  Op1 = _Op1 #make copies into which we can insert "F" string operators.
   Op2=_Op2
   onsiteOp = "$Op1*$Op2"
   fermionic1 = has_fermion_string(Op1, s[1])
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -567,6 +567,7 @@ using efficient MPS techniques. Returns the matrix C.
 
 # Optional Keyword Arguments
 - `site_range = 1:length(psi)`: compute correlations only for sites in the given range
+- `is_hermitian` = false : force indpendent calculation of above and below diag matrix elements.
 
 For a correlation matrix of size NxN and an MPS of typical
 bond dimension m, the scaling of this algorithm is N^2*m^3.
@@ -583,15 +584,15 @@ Czz = correlation_matrix(psi,"Sz","Sz")
 s = siteinds("Electron",N; conserve_qns=true)
 psi = randomMPS(s, n->isodd(n) ? "Up" : "Dn"; linkdims=m)
 Cuu = correlation_matrix(psi,"Cdagup","Cup";site_range=2:8)
-```
+``` 
 """
 function correlation_matrix(psi::MPS, _Op1::AbstractString, _Op2::AbstractString; kwargs...)
   N = length(psi)
   ElT = promote_itensor_eltype(psi)
   s = siteinds(psi)
 
-  Op1 = _Op1 #make copies into which we can insert "F" string operators.
-  Op2 = _Op2
+  Op1=_Op1 #make copies into which we can insert "F" string operators, and then restore.
+  Op2=_Op2
   onsiteOp = "$Op1*$Op2"
   fermionic1 = has_fermion_string(Op1, s[1])
   fermionic2 = has_fermion_string(Op2, s[1])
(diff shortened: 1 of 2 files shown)


src/mps/mps.jl
```diff
@@ -585,9 +585,44 @@ psi = randomMPS(s, n->isodd(n) ? "Up" : "Dn"; linkdims=m)
 Cuu = correlation_matrix(psi,"Cdagup","Cup";site_range=2:8)
 ```
 """
-function correlation_matrix(psi::MPS, Op1::AbstractString, Op2::AbstractString; kwargs...)
+function correlation_matrix(psi::MPS, _Op1::AbstractString, _Op2::AbstractString; kwargs...)
   N = length(psi)
   ElT = promote_itensor_eltype(psi)
+  s = siteinds(psi)
+
+  Op1=_Op1 #make copies into which we can insert "F" string operators.
```
emstoudenmire, line 593: Hi Jan, not sure these variables (`Op1` and `Op2` along with `_Op1` and `_Op2`) are needed. As you may know, this line just makes a reference to the same string, and then later a line like `Op1 = "$Op1*F"` just makes the variable `Op1` "point" to a new string elsewhere in memory while `_Op1` keeps pointing at the original string.

src/mps/mps.jl
```diff
@@ -585,9 +585,44 @@ psi = randomMPS(s, n->isodd(n) ? "Up" : "Dn"; linkdims=m)
...
+    is_proprtional=norm(O1 - O2)<eps;
+    is_hermitian  =norm(O1 - dag(swapprime(O2, 0, 1)))<eps;
+    if is_proprtional || is_hermitian 
+      isHermitian=true
+    end
+    # finally if they are both fermionic and proportional then the corr matrix will
+    # be anti symmetric insterad of Hermitian. Handle things like <C_i*C_j>
+    # at this point we know fermionic2=fermionic1, but we put them both in the if
+    # to clarify the meaning of what we are doing.
+    if is_proprtional && fermionic1 && fermionic2
```
emstoudenmire, line 622: Thanks for thinking through this carefully, to catch the antisymmetric fermionic case.

src/mps/mps.jl
emstoudenmire: Ok nevermind, I see now why these are needed, for the case of computing the lower triangle and fermionic operators. I was just thrown off by the comment which made it sound like it was a Julia string issue.

test/mps.jl
```diff
@@ -659,19 +673,47 @@ end
...
+  
+    # Fermion case
+    s = siteinds("Fermion", N)
+    psi = randomMPS(s; linkdims=m)
+    TestCorreltationMatrix(psi,[("N","N"),("Cdag","C"),("C","Cdag")])
+   
+    s = siteinds("Fermion", N; conserve_qns=false)
+    psi = randomMPS(s; linkdims=m)
+    TestCorreltationMatrix(psi,[("N","N"),("Cdag","C"),("C","Cdag")])
+    # can't test ,("C","C") yet, because AutoMPO thinks this is antisymmetric
```
emstoudenmire, line 711: Remove this comment following our recent discussion?

test/mps.jl
```diff
@@ -481,6 +481,22 @@ function basicRandomMPS(N::Int; dim=4)
   return M
 end
 
+function TestCorreltationMatrix(psi::MPS,ops::Vector{Tuple{String, String}})
```
emstoudenmire, line 484: Minor thing, but should be "Correlation" matrix.

emstoudenmire (dismissed):
Looks good. Since all my comments are very minor, I'll leave it up to you if you want to address them.

src/mps/mps.jl
```diff
@@ -585,9 +585,44 @@ psi = randomMPS(s, n->isodd(n) ? "Up" : "Dn"; linkdims=m)
...
+  onsiteOp = "$Op1*$Op2"
+  fermionic1 = has_fermion_string(Op1, s[1])
+  fermionic2 = has_fermion_string(Op2, s[1])
+  if fermionic1!=fermionic2
+    error("correlation_matrix: Mixed fermionic and bosonic operators are not supported yet.")
+  end
+  
+  # Decide if we need to calculate a non-hermitian corr. matrix which is roughly double the work.
+  isHermitian=false #Assume corr-matrix is non-hermitian
+  if haskey(kwargs, :isHermitian) #Did the user explicitly request something?
```
mtfishman, line 604: Small style change: for user-facing code (like function names, keyword arguments, etc.) we generally try to follow the Julia naming convention of using all lower case. In this case, I would recommend `ishermitian`. For things that are hard to read in that style, you can use underscores (like `has_fermion_string`).

It clashes with the C++ style we use so can be tough switching back and forth, but in general it makes it much easier to follow the Julia style since we often overload Julia functions in generic code.

src/mps/mps.jl
```diff
@@ -585,9 +585,44 @@ psi = randomMPS(s, n->isodd(n) ? "Up" : "Dn"; linkdims=m)
...
+    isHermitian = kwargs.isHermitian #Honour users request
+  else
+    O1=op(Op1, s, 1)
+    O2=op(Op2, s, 1)
+    O1/=norm(O1)
+    O2/=norm(O2)
+    #We need to decide if O1 ∝ O2 or O1 ∝ O2^dagger allowing for some round off errors.
+    eps=1e-10
+    is_proprtional=norm(O1 - O2)<eps;
```
mtfishman, line 613: `is_proprtional` -> `is_proportional`?

test/mps.jl
mtfishman: Also, would be nice to follow the Julia style and call it `test_correlation_matrix`.

In cases like this, I've also gone with the style of defining a function like `check_correlation_matrix` and return either `true` or `false`, then use `@test check_correlation_matrix(...)`. I think it makes it clearer it is using `@test`. An advantage of this is that it is more versatile, for example it could be used outside the context of testing (such as for a debug check in the code).

mtfishman:
Not sure why I'm not seeing the CI test results (such as the formatting, since I can see there are some formatting issues that should have been caught).

mtfishman:
Ok, I pushed a small change to get the CI to trigger. Should also run formatting checks which we can fix on Github.

@JanReimers, just a small thing that we have automated formatting in the Julia version of ITensor. You can run it locally by going to the directory `~/.julia/dev/ITensors` and running:
```julia
julia> using ITensors

julia> cd(pkgdir(ITensors))

julia> pwd()
"/home/mfishman/.julia/dev/ITensors"

julia> using JuliaFormatter

julia> format(".")
false
```
This will automatically format the code, and returns `false` if the code wasn't already formatted properly.

mtfishman:
I pushed a fix for the formatting since I had it ready to go anyway.

src/mps/mps.jl
```diff
@@ -583,11 +584,48 @@ Czz = correlation_matrix(psi,"Sz","Sz")
...
+  fermionic2 = has_fermion_string(Op2, s[1])
+  if fermionic1 != fermionic2
+    error(
+      "correlation_matrix: Mixed fermionic and bosonic operators are not supported yet."
+    )
+  end
+
+  # Decide if we need to calculate a non-hermitian corr. matrix which is roughly double the work.
+  is_cm_hermitian = false #Assume corr-matrix is non-hermitian
+  if haskey(kwargs, :is_hermitian) #Did the user explicitly request something?
```
mtfishman, line 607: Sorry to nitpick, but I would prefer the name `ishermitian` since that is the name of the Julia function:
```julia
julia> using LinearAlgebra

help?> ishermitian
search: ishermitian

  ishermitian(A) -> Bool


  Test whether a matrix is Hermitian.

[...]
```
and we use `ishermitian` as a keyword argument in other functions, like `eigen`.

mtfishman:
> OK I think I fixed all the typos and non conforming variable names. I ran the formatter ... very nice. The one thing I didn't do was Matt's suggestion of re-arranging the testing to something like. @test check_correlation_matrix(...) Right now test_correlation_matrix has two different types of check, and when something is wrong it spits out lots of numbers to look at. If we only test if check_correlation_matrix returns true then I think we lose all that info. If we really want the suggested format I can upgrade check_correlation_matrix to dump lots of info anytime a failure occurs. I prefer to let catch2 do that work. Anyway jenkins is chewing the new code:)

No worries, not a big deal either way. Just a style preference, but it sounds like you have a good reason for why it is designed the way it is.

I just had one more comment but besides that it looks good to me, thanks again!

JanReimers:
OK I changed it to ishermitian.  It is very beneficial to make the user interface as consistent as possible.

mtfishman:
Looks good, thanks Jan!

PR #820, opened 2022-02-10
[ITensors] [ENHANCEMENT] Add `normalize[!](::MPS/MPO)`

mtfishman:


emstoudenmire:
Yes, ok great that sounds right about the scaling, and then I was also going to bring up the potential "spreading the norm out over many sites" idea in case it's large, but sounds like you've got that covered too. And it's something that would be easy to revisit or improve later if it happens to cause a problem.

emstoudenmire:
That's a good plan

mtfishman:
@emstoudenmire I updated it to make it numerically controlled in the case when the norm gets very large:
```julia
julia> using ITensors

julia> s = siteinds("S=1/2", 1000);

julia> psi = randomMPS(s; linkdims=10);

julia> psi = 2 .* psi;

julia> norm(psi)
┌ Warning: The inner product (or norm²) you are computing is very large: NaN, which is greater than 4.503599627370496e15 and may lead to floating point errors when used. You should consider using `lognorm` or `loginner` instead, which will help avoid floating point errors. For example if you are trying to normalize your MPS/MPO `A`, the normalized MPS/MPO `B` would be given by `B = A ./ z` where `z = exp(lognorm(A) / length(A))`.
└ @ ITensors ~/.julia/dev/ITensors/src/mps/abstractmps.jl:1015
NaN

julia> lognorm(psi), length(psi) * log(2)
(693.1471805599457, 693.1471805599452)

julia> psi = normalize(psi);

julia> norm(psi)
```
This is done by computing `lognorm(psi)` and then dividing each tensor within the orthogonality limits by the norm per site.

emstoudenmire:
I think this is a really solid improvement & will pay off down the line in a bunch of places. Thanks for the update!

PR #833, opened 2022-02-27
[ITensors] [ENHANCEMENT]  More flexible `state` syntax (overloading and calling)

mtfishman:
- `state` defined with `Index` now should return an `ITensor`, for consistency with how `op` definitions work (this PR supports backwards compatibility for the now-deprecated syntax which returns a `Vector` that gets automatically converted to an `ITensor`).
- Allow syntax like `state("Up", Index(2, "S=1/2"))` (previously only `state(Index(2, "S=1/2"), "Up")` worked).

### DIFF
--- a/src/physics/site_types/qudit.jl
+++ b/src/physics/site_types/qudit.jl
@@ -31,7 +31,7 @@ function ITensors.state(::StateName{N}, ::SiteType"Qudit", s::Index) where {N}
   n = parse(Int, String(N))
   st = zeros(dim(s))
   st[n + 1] = 1.0
-  return st
+  return itensor(st, s)
 end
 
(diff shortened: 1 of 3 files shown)


emstoudenmire:
Just to check, the preferred (non-deprecated) case of `state` returning an ITensor is only when an Index is passed, but for a `StateName` we still return a Vector, correct? Similar to how for `OpName` one can return a matrix?

mtfishman:
The new convention is:
```julia
state(::StateName"Up", SiteType"S=1/2") = [1, 0]
state(::StateName"Up", SiteType"S=1/2", i::Index) = itensor([1, 0], i)
```
whereas previously it was designed so that the output of the second one was supposed to be `[1, 0]`.

emstoudenmire:
Ok sounds good!

PR #842, opened 2022-03-01
Add normalize! to TEBD Tutorial

emstoudenmire:
Norm should be kept at 1.0 for real-time TEBD Tutorial in the docs.

### DIFF
--- a/docs/src/tutorials/MPSTimeEvolution.md
+++ b/docs/src/tutorials/MPSTimeEvolution.md
@@ -106,6 +106,7 @@ let
   # for Nsteps steps and printing <Sz> on site c
   for step=1:Nsteps
     psi = apply(gates, psi; cutoff=cutoff)
+    normalize!(psi)
     t += tau
     Sz = expect(psi,"Sz";site_range=c:c)
     println("$t $Sz")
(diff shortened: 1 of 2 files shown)


mtfishman:
It would be good to have a flag `normalize` in `apply` that turned this on and off internally.

emstoudenmire:
Yes agreed. And I would argue that the default should be off, since that way the literal meaning is to apply the gates to the state. For TVDP I was kind of debating about a similar thing: should we primarily think of it as "applying" exp(H t) to the state and returning that, or as a "time evolution of quantum states" algorithm where one prefers the state to be normalized. I guess logically the first one, but then it's a little frustrating that for the application it will be used the most for, a best practice is to pass a keyword arg `normalize=true` every time.

mtfishman:
Yeah, I prefer taking the more literal version by default (so `normalize=false` by default). In general people will use unitary evolution anyway, in which case using `normalize=true` wouldn't be so important anyway (and then the norm of the resulting state can give some idea for the truncation/fidelity loss).

emstoudenmire:
Ok so then we're agreed about that. I just switched the default to be not to normalize in TDVP, and we can also nicely add that keyword arg to `apply` without it changing any default behavior that way too.

PR #843, opened 2022-03-01
Unification of `gate` and `op` & some `OpSum` functions

GTorlai:
This PR implements the following changes:

**1)** Change the `SiteOp` type to accept a dense array as input, with appropriate modifications to `MPOTerm`

**2)** Introduce some additional algebra for `OpSum`

**3)** Add support for parametric ops in `MPO`

**4)** Consolidate operators for two-level systems into `qubits.jl`, with `spinhalf.jl` containing aliases.

**5)** Add the `adjoint` keyword argument in the `op` system to generated adjoint operators (used in bases rotations for example).


src/physics/site_types/qubit.jl
```diff
@@ -39,182 +39,293 @@ function ITensors.space(
...
+ITensors.val(::ValName"↑", ::SiteType"Qubit") = 1
+ITensors.val(::ValName"↓", ::SiteType"Qubit") = 2
 
 ITensors.state(::StateName"0", ::SiteType"Qubit") = [1.0, 0.0]
 ITensors.state(::StateName"1", ::SiteType"Qubit") = [0.0, 1.0]
 ITensors.state(::StateName"+", ::SiteType"Qubit") = [1.0, 1.0] / √2
 ITensors.state(::StateName"-", ::SiteType"Qubit") = [1.0, -1.0] / √2
+ITensors.state(::StateName"i", ::SiteType"Qubit") = [1.0, im] / √2
+ITensors.state(::StateName"-i", ::SiteType"Qubit") = [1.0, -im] / √2
+ITensors.state(::StateName"Up", ::SiteType"Qubit") = [1.0, 0.0]
```
mtfishman, line 55: Could we change these to things like:
```julia
ITensors.state(::StateName"Up", t::SiteType"Qubit") = state(StateName("0"), t)
```

src/physics/autompo.jl
```diff
@@ -146,9 +152,13 @@ function Base.show(io::IO, op::MPOTerm)
   end
   for o in ops(op)
     print(io, "\"$(name(o))\"($(string_site_or_sites(o))) ")
+    !isempty(params(o)) && print(io, params(o))
   end
 end
 
+*(α::Number, op::MPOTerm) = MPOTerm(α * coef(op), ops(op))
+*(op::MPOTerm, α::Number) = α * op
```
mtfishman, line 160: Could you add:
```julia
/(op::MPOTerm, α::Number) = MPOTerm(coef(op) / α, ops(op))
```
as well? I came across that for Trotterization.

Also a neat thing is that you can use the notation:
```julia
(α::Number * op::MPOTerm) = MPOTerm(α * coef(op), ops(op))
```

src/physics/autompo.jl
```diff
@@ -203,6 +213,12 @@ end
 
 Base.size(ampo::OpSum) = size(data(ampo))
 
+function Base.iterate(os::OpSum, state=1)
```
mtfishman, line 216: A simpler definition would be:
```julia
iterate(os::OpSum, args...) = iterate(data(os), args...)
```
i.e. just leverage the definition for `Vector`.

mtfishman:
Over all looks good, just made a few comments on code style.

What about using `adjoint` instead of `dag` so that it doesn't get confused with the ITensor `dag` function, which as you know isn't quite the same thing? Or does it need to be compatible with PastaQ notation?

GTorlai:
> What about using `adjoint` instead of `dag` so that it doesn't get confused with the ITensor `dag` function, which as you know isn't quite the same thing? Or does it need to be compatible with PastaQ notation?

Done

src/physics/autompo.jl
```diff
@@ -299,6 +312,29 @@ function (ampo::OpSum - term::Tuple)
...
+/(os::OpSum, α::Number) = OpSum([o / α for o in os])
+
+-(o1::OpSum, o2::OpSum) = o1 + (-1) * o2
+
+"""
+    prune!(os::OpSum; cutoff = 1e-15)
+
+Remove any MPOTerm with norm(coef) < cutoff
+"""
+function prune!(os::OpSum; cutoff=1e-15)
```
mtfishman, line 329: `atol` instead of `cutoff`.

PR #849, opened 2022-03-04
Small improvements and remove Julia 1.3

emstoudenmire:
- Remove unneeded type before array
- Remove Julia 1.3 from testing
- Allow maxdim to be optional in DMRG sweeps keywords

### DIFF
--- a/.github/workflows/test_itensorgaussianmps.yml
+++ b/.github/workflows/test_itensorgaussianmps.yml
@@ -14,7 +14,6 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.3'
           - '1'
         os:
           - ubuntu-latest
--- a/src/mps/dmrg.jl
+++ b/src/mps/dmrg.jl
@@ -316,7 +316,7 @@ function dmrg(PH, psi0::MPS, sweeps::Sweeps; kwargs...)::Tuple{Number,MPS}
   return (energy, psi)
 end
 
-function _dmrg_sweeps(; nsweeps, maxdim, mindim=1, cutoff=1e-8, noise=0.0, kwargs...)
+function _dmrg_sweeps(; nsweeps, maxdim=10_000, mindim=1, cutoff=1E-8, noise=0.0, kwargs...)
   sweeps = Sweeps(nsweeps)
   setmaxdim!(sweeps, maxdim...)
   setmindim!(sweeps, mindim...)
--- a/.github/workflows/test_itensors.yml
+++ b/.github/workflows/test_itensors.yml
@@ -14,7 +14,6 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.3'
           - '1'
         os:
           - ubuntu-latest
@@ -27,9 +26,6 @@ jobs:
         exclude:
           # MacOS not available on x86
           - {os: 'macOS-latest', arch: 'x86'}
-          # Only test all os on the latest release
-          - {version: '1.3', os: 'windows-latest'}
-          - {version: '1.3', os: 'macOS-latest'}
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@latest
--- a/.github/workflows/test_itensors_threaded.yml
+++ b/.github/workflows/test_itensors_threaded.yml
@@ -14,7 +14,6 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.3'
           - '1'
         os:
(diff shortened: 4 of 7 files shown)


src/mps/dmrg.jl
```diff
@@ -316,7 +316,7 @@ function dmrg(PH, psi0::MPS, sweeps::Sweeps; kwargs...)::Tuple{Number,MPS}
   return (energy, psi)
 end
 
-function _dmrg_sweeps(; nsweeps, maxdim, mindim=1, cutoff=1e-8, noise=0.0, kwargs...)
+function _dmrg_sweeps(; nsweeps, maxdim=10_000, mindim=1, cutoff=1E-8, noise=0.0, kwargs...)
```
mtfishman, line 319: What if we just use `typemax(Int)` (which equals `9223372036854775807`)? I'd prefer that over choosing an arbitrary cutoff, it seems either way the calculation would get very slow or run out of memory and the user would find out quickly they did something wrong.

emstoudenmire: Sounds good – just made that change

src/itensor.jl
```diff
@@ -1953,7 +1953,7 @@ end
 
 *(As::ITensor...; kwargs...)::ITensor = contract(As...; kwargs...)
 
-function contract!(C::ITensor, A::ITensor, B::ITensor, α::Number, β::Number=0)::ITensor
```
mtfishman, line 1956: This actually changes the meaning of the code (it changes the positional argument to a keyword argument) and is caused by a bug in an older version of JuliaFormatter: https://github.com/domluna/JuliaFormatter.jl/issues/533.

Could you revert this change and upgrade your version of JuliaFormatter and run it again? Note that just upgrading and running again will not fix it automatically, since it will see the new code with the semicolon as properly formatted code, even if from our perspective it is incorrect.

mtfishman:
Should we also change the `Project.toml`s to have 1.6 as the minimum Julia version in this PR?

https://github.com/ITensor/ITensors.jl/blob/d6ce046e47d5ffe89d2d31b37c352c51e1f5dc49/Project.toml#L47
https://github.com/ITensor/ITensors.jl/blob/d6ce046e47d5ffe89d2d31b37c352c51e1f5dc49/NDTensors/Project.toml#L27

etc.? If we are not testing any more we should also explicitly say they are no longer supported. Once we do that we can bump the versions of all the packages.

src/itensor.jl
emstoudenmire: That's a weird bug. Thanks for catching it! I updated JuliaFormatter, manually reverted this and another one below in a unit test, and reformatted.

emstoudenmire:
I just did the Julia version bumps in the Project.toml's. Good idea. My one concern is whether we should put 1.5 instead of 1.6? It's mainly because when I did that Twitter poll someone indicated they were using a version of Julia at 1.5 or earlier. On the other hand, I think 99% of users if not 100% have control over which version they are using (meaning not through a centralized cluster install – which anyway could get updated too) so they can always fix by just updating their Julia version.

mtfishman:
I would vote for 1.6 since it is currently the long term support (LTS) version. This is the stance that SciML took: https://sciml.ai/news/2021/12/02/lts/index.html.

emstoudenmire:
That's a good argument for 1.6 I agree. So let's go with that.

.github/workflows/test_itensorgaussianmps.yml
```diff
@@ -14,7 +14,6 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.3'
```
mtfishman, line 17: Instead of removing this we should change it to `1.6` so then we make sure to test Julia 1.6. Without it we will only test `1`, which defaults to the latest released version (so for example right now it default to Julia 1.7).

emstoudenmire: Ok will make that change

mtfishman:
Great, thanks! Anything else left to do here?

emstoudenmire:
Hm, not from my end. If you have any small changes you'd like to stick in that had to do with 1.3 features, like keyword argument things, then please go ahead. Otherwise we can merge and then tag a new version soon. We may want to hold off on the new version for a little while, though, in case there are any mildly breaking changes or deprecations that you've been wanting to do.

mtfishman:
Sounds good, let's merge this but wait a bit before registering v0.3.

PR #852, opened 2022-03-08
rrules for `apply`

GTorlai:
This PR adds the following differentiation rules

1. `apply(::Vector{ITensor}, ::MPO)` for MPO conjugate evolution (including noisy gates).
2. `apply` for MPO evolution with `apply_dag = false`.
3. `contract(::MPO, ::MPO)`
4. `tr(::MPO)`


src/ITensorChainRules/ITensorChainRules.jl
```diff
@@ -321,13 +321,17 @@ function ChainRulesCore.rrule(::typeof(apply), x1::Vector{ITensor}, x2::MPS; kwa
   return y, apply_pullback
 end
 
-function ChainRulesCore.rrule(::typeof(apply), x1::Vector{ITensor}, x2::MPO; apply_dag::Bool = true, kwargs...)
+function ChainRulesCore.rrule(
+  ::typeof(apply), x1::Vector{ITensor}, x2::MPO; apply_dag::Bool=true, kwargs...
```
mtfishman, line 325: Putting `apply_dag::Bool=true` wouldn't guarantee that it is true, it is just setting a default. In fact it may be worse and actually change the default behavior of the forward pass, I'm not exactly sure what the interplay is between the keyword arguments defined in the original function vs. the `rrule` of the function.

We should add a check:
```julia
!apply_dag && errror("rrule for apply MPO not yet defined for `apply_dag=false`")
```
or something like that.

GTorlai: This function handles the case with `apply_dag=false` as well.

GTorlai: Sure, just need to grab its value inside the function. Will make this change

src/ITensorChainRules/ITensorChainRules.jl
```diff
@@ -364,4 +410,32 @@ function ChainRulesCore.rrule(::typeof(inner), x1::MPS, x2::MPS; kwargs...)
...
+    return (NoTangent(), x̄1, x̄2)
+  end
+  return y, contract_pullback
+end
+
+function ChainRulesCore.rrule(::typeof(tr), x::MPO; kwargs...)
+  y = tr(x; kwargs...)
+  function contract_pullback(ȳ)
+    s = firstsiteinds(x)
+    #return ȳ * MPO(s, "Id")
```
mtfishman, line 427: What's the problem with `MPO(s, "Id")`? This works for me:
```julia
julia> using ITensors

julia> I = MPO(s, "Id");

julia> I' * I
MPO
[1] ((dim=2|id=573|"S=1/2,Site,n=1")'', (dim=2|id=573|"S=1/2,Site,n=1"), (dim=1|id=198|"Link,l=1"))
[2] ((dim=2|id=104|"S=1/2,Site,n=2")'', (dim=2|id=104|"S=1/2,Site,n=2"), (dim=1|id=709|"Link,l=2"), (dim=1|id=198|"Link,l=1"))
[3] ((dim=2|id=82|"S=1/2,Site,n=3")'', (dim=2|id=82|"S=1/2,Site,n=3"), (dim=1|id=574|"Link,l=3"), (dim=1|id=709|"Link,l=2"))
[4] ((dim=2|id=97|"S=1/2,Site,n=4")'', (dim=2|id=97|"S=1/2,Site,n=4"), (dim=1|id=574|"Link,l=3"))
```

GTorlai:
OK I added the `plev` feature in the `tr(MPO)` rrule.

src/ITensorChainRules/ITensorChainRules.jl
GTorlai: Yes, there was an error that I mistakenly attributed to this. Sometimes it is tricky with Zygote when you have multiple bugs.

src/ITensorChainRules/ITensorChainRules.jl
```diff
@@ -326,6 +319,52 @@ function ChainRulesCore.rrule(::typeof(apply), x1::Vector{ITensor}, x2::MPS; kwa
   return y, apply_pullback
 end
 
+function ChainRulesCore.rrule(::typeof(apply), x1::Vector{ITensor}, x2::MPO; kwargs...)
+  apply_dag = get(kwargs, :apply_dag, true)
```
mtfishman, line 323: Shouldn't this default to `false` since the original `apply` function defaults to `false`?

You may be able to not set a default at all here, could you try just using `kwargs.apply_dag`? Basically my concern is that this keyword argument should match whatever was set by the user for the forward pass of `apply`, so we want to make sure it matches that and not accidentally set it to something different.

src/ITensorChainRules/ITensorChainRules.jl
```diff
@@ -326,6 +319,52 @@ function ChainRulesCore.rrule(::typeof(apply), x1::Vector{ITensor}, x2::MPS; kwa
...
+    x1dag_ȳ = Vector{MPO}(undef, N)
+    x1dag_ȳ[end] = ȳ
+    x1dag = [swapprime(dag(x), 0 => 1) for x in x1]
+    for n in (N - 1):-1:1
+      x1dag_ȳ[n] = apply(x1dag[n], x1dag_ȳ[n + 1]; move_sites_back=true, kwargs...)
+    end
+
+    x̄1 = similar(x1)
+    for n in 1:length(x1)
+      if iseven(length(inds(x1[n])))
```
mtfishman, line 343: Could you explain what this is doing?

GTorlai: This is just checking whether you have a noisy gate, which are currently not differentiable. I left a comment on top of this `if` statement but got deleted somehow between commits

mtfishman: I see, thanks.

mtfishman: Wouldn't that be an issue with applying a noisy gate to an MPS in general, and nothing particular to defining the `rrule`?

mtfishman: To be clear, applying a noisy gate to an MPS right now is not well defined behavior.

GTorlai: Yes correct

PR #856, opened 2022-03-10
Improve Sweeps constructors

emstoudenmire:
Two changes:
 - more flexible Sweeps constructor allowing keyword args which can be numbers or iterables
 - allow default initialization of Sweeps objects


emstoudenmire:
Going to go ahead and merge this because the test failure seems to be a (rare) Lapack SVD failure only. Many other instances of TestITensors did pass on both 1.6 and 1.7

Issue #854, opened 2022-03-09
[ITensors] [ENHANCEMENT] Add `+` operation in the `op` system


GTorlai:
Introduce the addition (over the multiplication currently implemented) for `op`:
```julia
q = siteind("Qudit"; dim = 4)
X = op("a + a†", q)
```
To avoid clashing with operators like "S+", we can check whether there is a space between the `+` and nearby `chars`.

mtfishman:
Closed by #857.

PR #857, opened 2022-03-11
`+` and `-` in the `op` system

GTorlai:
#854

Add `+` and `-` algebra to the `op` system. There needs to be a space between the op symbol and the opnames, as to avoid clashing with `S+` etc.  Right now the product `*` is allows both with and without spaces for backward compatibility, but will throw a warning when used without. I have updated the internal functions so that we don't spam warnings.

Only non-parametric gates can be manipulated this way for now. This syntax is often used for Trotter simulations, where on-site terms can be grouped together and contracted once (same for exchange interactions such as `a†a + aa† - aa - a†a†`).

```julia
q = siteind("Qubit")

X = op("X", SiteType("Qubit"))
Y = op("Y", SiteType("Qubit"))
Z = op("Z", SiteType("Qubit"))

# [X,Y] + [Y,Z] = 2im * (Z+X)
@show op("X * Y - Y * X + Y * Z - Z * Y", q)
# NDTensors.Dense{ComplexF64, Vector{ComplexF64}}
#  2×2
#  0.0 + 2.0im  0.0 + 2.0im
#  0.0 + 2.0im  0.0 - 2.0im
```


src/physics/sitetype.jl
```diff
@@ -224,22 +224,52 @@ function op(name::AbstractString, s::Index...; adjoint::Bool=false, kwargs...)
...
+
+  # the the multiplication come after
+  oploc = findfirst("*", name)
+  if !isnothing(oploc)
+    op1, op2 = nothing, nothing
+    @ignore_derivatives begin
+      op1 = name[1:prevind(name, oploc.start)]
+      op2 = name[nextind(name, oploc.start):end]
+      if !(op1[end] == ' ' && op2[1] == ' ')
+        @warn "composite op definition `A*B` deprecated: please use `A * B` instead (with spaces)"
```
mtfishman, line 270: Let's use `Base.depwarn("composite op definition `A*B` deprecated: please use `A * B` instead (with spaces)", :op)`.

It is the function version of [`@deprecated`](https://docs.julialang.org/en/v1/base/base/#Base.@deprecate) which is the standard way in Julia to deprecate syntax, and can be turned on and off with a flag passed to `julia`.

test/sitetype.jl
```diff
@@ -15,13 +15,62 @@ using ITensors, Test
...
+    Sz = array(op("Sz", q))
+    x = Sp + Sm
+    @test x ≈ array(op("S+ + S-", q))
+    x = Sp - Sm
+    @test x ≈ array(op("S+ - S-", q))
+    x = Sp - Sm - Sp
+    @test x ≈ array(op("S+ - S- - S+", q))
+    x = Sp * Sm + Sm * Sp
+    @test x ≈ array(op("S+ * S- + S- * S+", q))
+    @test x ≈ array(op("S+*S- + S-*S+", q))
```
mtfishman, line 56: We should be able to test that it outputs a deprecation warning, for example: https://discourse.julialang.org/t/help-with-deprecation-warnings-in-tests/50011

emstoudenmire:
Good to see this

Issue #850, opened 2022-03-05
[ITensors] [ENHANCEMENT] More general `site_range` collections in `correlation_matrix`


mtfishman:
It would be nice to support more general `site_range`s in `correlation_matrix`. For example now you can do:
```julia
julia> using ITensors

julia> n = 10;

julia> s = siteinds("S=1/2", n);

julia> psi = randomMPS(s);

julia> correlation_matrix(psi, "Sz", "Sz"; site_range=3:5)
3×3 Matrix{Float64}:
  0.25       -0.0336051   -0.0652171
 -0.0336051   0.25         0.00880679
 -0.0652171   0.00880679   0.25
```
However this is not supported:
```julia
julia> correlation_matrix(psi, "Sz", "Sz"; site_range=[3, 5])
ERROR: MethodError: Cannot `convert` an object of type Vector{Int64} to an object of type UnitRange{Int64}
Closest candidates are:
  convert(::Type{T}, ::LinearAlgebra.Factorization) where T<:AbstractArray at ~/software/julia-1.7.2/share/julia/stdlib/v1.7/LinearAlgebra/src/factorization.jl:58
  convert(::Type{T}, ::AbstractRange) where T<:AbstractRange at ~/software/julia-1.7.2/share/julia/base/range.jl:215
  convert(::Type{T}, ::T) where T<:AbstractArray at ~/software/julia-1.7.2/share/julia/base/abstractarray.jl:16
  ...
Stacktrace:
 [1] correlation_matrix(psi::MPS, _Op1::String, _Op2::String; kwargs::Base.Pairs{Symbol, Vector{Int64}, Tuple{Symbol}, NamedTuple{(:site_range,), Tuple{Vector{Int64}}}})
   @ ITensors ~/.julia/packages/ITensors/5sSxp/src/mps/mps.jl:630
 [2] top-level scope
   @ REPL[7]:1
```
I would picture this would output:
```julia
julia> correlation_matrix(psi, "Sz", "Sz"; site_range=[3, 5])
3×3 Matrix{Float64}:
  0.25        -0.0652171
 -0.0652171    0.25
```
@emstoudenmire @JanReimers

mtfishman:
Also see the discussion in #858, maybe we want to rename the keyword argument to `sites` since it wouldn't always be a range object with this new design.

PR #858, opened 2022-03-12
Implement changes in #800 for expect interface

### LINKED ISSUE #800: [ENHANCEMENT] `expect` interface
I think a better design for the `expect` function interface would be the following:

1. A version `expect(psi::MPS, ops::Vector{<:AbstractString})` that return `Vector{Vector{Float64}}`.
2. A version `expect(psi::MPS, op::AbstractString)` that returns `Vector{Float64}`, which is a wrapper around version 1.

I also think we should always return a `Vector` of results even if there is only a single site in the `site_range`.

The reasoning is twofold:

1. Type stability. In this case, it is not an issue of the performance cost of type instability, but I think it is a design goal we should strive for in general unless there is a very compelling reason (like the site type system, or the `ITensor` type design). Type stability is a good design goal since it helps with predictability of the code: the Julia compiler can predict what types are output based on what types are input, but so can users, which I think in general helps make the code more self-documenting and less surprising.
2. Related to the "surprise" factor in comment 1., it would help with writing generic code. For example, say that someone is writing a generic code that wraps `expect` and does some analysis (say averages all of the expectation values). In the current design, the implementation would have to look something like this:
```julia
function expect_mean(psi::MPS, ops::AbstractString...; kwargs...)
  os = expect(psi, ops...; kwargs...)
  if length(ops) == 1
    return os
  end
  return mean(os)
end
```
while in the current interface proposal, it could look like this:
```julia
function expect_mean(psi::MPS, ops::Vector{<:AbstractString}; kwargs...)
  return mean(expect(psi, ops; kwargs...))
end
```
Similar branching logic might have to be done to handle the special case of `length(site_range) == 1` as well (not in this case, since `mean` is generic for `Vector` or `Number`, but for other situations where that isn't the case).

Also see the discussion in https://github.com/ITensor/ITensor/pull/402.

emstoudenmire:
Implements changes discussed in issue JuliaLang/julia#800 for the `expect` function.
Also implements a new keyword arg `site=s` which is equivalent to passing
`site_range=s:s`. It is an error if both keywords are passed.

Internally, `expect` now uses the `map` function to convert the container
of op names (`ops`) into an equivalent container of vectors of zeros,
then fills these up with the expectation values computed.

There are now two helper functions which convert variable numbers of 
string arguments into tuples to pass into the version of `expect` 
taking a container.

Added a unit test.

### DIFF
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1997,15 +1997,26 @@ function indpairs(T::ITensor; plev::Pair{Int,Int}=0 => 1, tags::Pair=ts"" => ts"
 end
 
 """
-    ishermitian(T::ITensor; eps = 1E-10)
+    transpose(T::ITensor)
+
+Treating an ITensor as a map from a set of indices
+of prime level 0 to a matching set of indices but
+of prime level 1 
+[for example: (i,j,k,...) -> (j',i',k',...)]
+return the ITensor which is the transpose of this map.
+"""
+transpose(T::ITensor) = swapprime(T, 0 => 1)
+
+"""
+    ishermitian(T::ITensor; kwargs...)
 
 Test whether an ITensor is a Hermitian operator,
-up to a numerical tolerance. To be considered an
-operator, an ITensor must have matching pairs
-of indices with prime level 0 and 1.
+that is whether taking `dag` of the ITensor and
+transposing its indices returns numerically 
+the same ITensor.
 """
 function ishermitian(T::ITensor; kwargs...)
-  return isapprox(T, dag(swapprime(T, 0, 1)); kwargs...)
+  return isapprox(T, dag(transpose(T)); kwargs...)
 end
 
 # Trace an ITensor over pairs of indices determined by
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -2003,12 +2003,9 @@ Test whether an ITensor is a Hermitian operator,
 up to a numerical tolerance. To be considered an
 operator, an ITensor must have matching pairs
 of indices with prime level 0 and 1.
-
-# Optional Keyword Arguments
-- `eps = 1E-10`: the numerical tolerance to use 
 """
-function ishermitian(T::ITensor; eps=1E-10)
-  return norm(T - dag(swapprime(T, 0, 1))) < eps
+function ishermitian(T::ITensor; kwargs...)
+  return isapprox(T, dag(swapprime(T, 0, 1)); kwargs...)
 end
 
 # Trace an ITensor over pairs of indices determined by
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1996,6 +1996,21 @@ function indpairs(T::ITensor; plev::Pair{Int,Int}=0 => 1, tags::Pair=ts"" => ts"
   return is_first .=> is_last
 end
 
+"""
+    ishermitian(T::ITensor; eps = 1E-10)
+
+Test whether an ITensor is a Hermitian operator,
+up to a numerical tolerance. To be considered an
+operator, an ITensor must have matching pairs
+of indices with prime level 0 and 1.
+
+# Optional Keyword Arguments
+- `eps = 1E-10`: the numerical tolerance to use 
+"""
+function ishermitian(T::ITensor; eps=1E-10)
+  return norm(T - dag(swapprime(T, 0, 1))) < eps
+end
+
 # Trace an ITensor over pairs of indices determined by
 # the prime levels and tags. Indices that are not in pairs
 # are not traced over, corresponding to a "batched" trace.
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -722,15 +722,23 @@ function correlation_matrix(psi::MPS, _Op1::AbstractString, _Op2::AbstractString
 end
 
 """
-    expect(psi::MPS,ops::AbstractString...;kwargs...)
+    expect(psi::MPS,ops::AbstractString...; kwargs...)
+    expect(psi::MPS,ops; kwargs...)
 
-Given an MPS `psi` and an operator name, returns
+Given an MPS `psi` and a single operator name, returns
 a vector of the expected value of the operator on 
-each site of the MPS. If multiple operator names are
-provided, returns a tuple of expectation value vectors.
+each site of the MPS. 
+
+If multiple operator names are provided, returns a tuple 
+of expectation value vectors.
+
+If a container of operator names is provided, returns the
+same type of container with names replaced by vectors
+of expectation values.
 
 # Optional Keyword Arguments
 - `site_range = 1:length(psi)`: compute expected values only for sites in the given range
+- `site`: compute the expected value on a given site
 
 # Examples
 
@@ -747,14 +755,21 @@ dens = expect(psi,"Ntot")
 updens,dndens = expect(psi,"Nup","Ndn")
 ```
 """
-function expect(psi::MPS, ops::AbstractString...; kwargs...)
+function expect(psi::MPS, ops; kwargs...)
   psi = copy(psi)
   N = length(psi)
   ElT = real(promote_itensor_eltype(psi))
   Nops = length(ops)
   s = siteinds(psi)
 
-  site_range::UnitRange{Int} = get(kwargs, :site_range, 1:N)
+  if haskey(kwargs, :site)
+    haskey(kwargs, :site_range) &&
+      error("Cannot pass both site and site_range keyword args to expect")
+    site = kwargs[:site]
+    site_range = site:site
+  else
+    site_range::UnitRange{Int} = get(kwargs, :site_range, 1:N)
+  end
   Ns = length(site_range)
   start_site = first(site_range)
   offset = start_site - 1
@@ -762,7 +777,7 @@ function expect(psi::MPS, ops::AbstractString...; kwargs...)
   orthogonalize!(psi, start_site)
   norm2_psi = norm(psi)^2
 
-  ex = ntuple(n -> zeros(ElT, Ns), Nops)
+  ex = map(n -> zeros(ElT, Ns), ops)
   for j in site_range
     orthogonalize!(psi, j)
     for n in 1:Nops
@@ -771,11 +786,15 @@ function expect(psi::MPS, ops::AbstractString...; kwargs...)
     end
   end
 
-  if Nops == 1
-    return Ns == 1 ? ex[1][1] : ex[1]
-  else
-    return Ns == 1 ? [x[1] for x in ex] : ex
-  end
+  return ex
+end
+
+function expect(psi::MPS, op::AbstractString; kwargs...)
+  return expect(psi, (op,); kwargs...)[1]
+end
+
+function expect(psi::MPS, op1::AbstractString, ops::AbstractString...; kwargs...)
+  return expect(psi, (op1, ops...); kwargs...)
 end
 
 function HDF5.write(parent::Union{HDF5.File,HDF5.Group}, name::AbstractString, M::MPS)
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -772,7 +772,7 @@ function expect(psi::MPS, ops; kwargs...)
   Ns = length(site_range)
   start_site = first(site_range)
 
-  el_types = map(o -> ishermitian(op(o, s[start_site])) ? Float64 : ElT, ops)
+  el_types = map(o -> ishermitian(op(o, s[start_site])) ? real(ElT) : ElT, ops)
 
   orthogonalize!(psi, start_site)
   norm2_psi = norm(psi)^2
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -722,7 +722,7 @@ function correlation_matrix(psi::MPS, _Op1::AbstractString, _Op2::AbstractString
 end
 
 """
-    expect(psi::MPS,ops::AbstractString...; kwargs...)
+    expect(psi::MPS,op::AbstractString...; kwargs...)
     expect(psi::MPS,ops; kwargs...)
 
 Given an MPS `psi` and a single operator name, returns
@@ -737,8 +737,7 @@ same type of container with names replaced by vectors
 of expectation values.
 
 # Optional Keyword Arguments
-- `site_range = 1:length(psi)`: compute expected values only for sites in the given range
-- `site`: compute the expected value on a given site
+- `sites = 1:length(psi)`: compute expected values only for sites in the given range
 
 # Examples
 
@@ -747,12 +746,13 @@ N = 10
 
 s = siteinds("S=1/2",N)
 psi = randomMPS(s; linkdims=8)
-Z = expect(psi,"Sz";site_range=2:6)
+Z = expect(psi,"Sz";sites=2:4) # compute for sites 2,3,4
+Z3 = expect(psi,"Sz";sites=3)  # compute for site 3 only (output will be a scalar)
 
 s = siteinds("Electron",N)
 psi = randomMPS(s; linkdims=8)
 dens = expect(psi,"Ntot")
-updens,dndens = expect(psi,"Nup","Ndn")
+updens,dndens = expect(psi,"Nup","Ndn") # pass more than one operator
 ```
 """
 function expect(psi::MPS, ops; kwargs...)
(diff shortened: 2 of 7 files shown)


mtfishman:
Overall looks good, thanks!

Maybe we can think about an interface where we don't have both `site_range` and `site`.

My proposal would be to combine them into a single keyword argument `sites`. This could then include passing a non-contiguous range, like `sites=[2, 4]` if you just want the expectation values on sites `2` and `4`. This is something I think we should also have for `correlation_matrix` (#850).

So for example:
```julia

julia> psi = randomMPS(s);

julia> expect(psi, "X"; sites=2:4)
3-element Vector{Float64}:
 0.9627825241221968
 0.9788691729733142
 0.8921243804181297

julia> expect(psi, "X"; sites=2:2)
1-element Vector{Float64}:
 0.9627825241221968

julia> expect(psi, "X"; sites=2)
0.9627825241221968

julia> expect(psi, "X"; sites=[2, 4])
2-element Vector{Float64}:
 0.9627825241221968
 0.8921243804181297
```

test/mps.jl
```diff
@@ -719,6 +719,49 @@ end
...
+
+    res = expect(psi, "Sz"; site_range=2:4)
+    @test res ≈ eSz[2:4]
+
+    res = expect(psi, "Sz", "Sx")
+    @test res[1] ≈ eSz
+    @test res[2] ≈ eSx
+
+    res = expect(psi, "Sz"; site=3)
+    @test typeof(res) == Vector{Float64}
```
mtfishman, line 746: FYI you can write this more succinctly as:
```julia
    @test res isa Vector{Float64}
```

emstoudenmire: Great

emstoudenmire:
Ok I like that idea of having one less keyword arg name to parse and remember.

I know you were just using it as an example, but when you wrote:
```julia
julia> expect(psi, "X"; sites=2) # Not sure if this should return `Float64` or `Vector{Float64}`
0.9627825241221968
```
that would be back to the previous behavior where the single-site case (site_range=2:2) returned a different type, which wasn't type-stable. Technically now I think that passing `sites=2` and getting a number while passing `sites=2:2` and getting a vector *would* be type stable, if we went with that design. It would also be pretty convenient, though it could cause trouble in generic code, perhaps.

What do you think about that issue? Always having the return value be a vector for each operator passed? Or having it switch to being a scalar in the case that there is a single site. (We could still have `sites=[2]` return a length-1 vector.)

mtfishman:
I would see it as an analogy between these two cases:
```julia
expect(psi, ["X"]) isa Vector{Vector{Float64}}
expect(psi, "X") isa Vector{Float64}
```
and:
```julia
expect(psi, "X"; sites=[2]) isa Vector{Float64}
expect(psi, "X"; sites=2:2) isa Vector{Float64}
expect(psi, "X"; sites=2) isa Float64
```

I think that is indeed type stable, since it is basically just this:
```julia
julia> f(; x) = (x isa Number ? 2 * x : 2 * collect(x))

julia> f(; x=2)
4

julia> f(; x=2:2)
1-element Vector{Int64}:
 4

julia> f(; x=2:3)
2-element Vector{Int64}:
 4
 6

julia> @code_warntype f(; x=2)
MethodInstance for (::var"#f##kw")(::NamedTuple{(:x,), Tuple{Int64}}, ::typeof(f))
  from (::var"#f##kw")(::Any, ::typeof(f)) in Main at REPL[5]:1
Arguments
  _::Core.Const(var"#f##kw"())
  @_2::NamedTuple{(:x,), Tuple{Int64}}
  @_3::Core.Const(f)
Locals
  x::Int64
  @_5::Int64
Body::Int64
1 ─ %1  = Base.haskey(@_2, :x)::Core.Const(true)
│         Core.typeassert(%1, Core.Bool)
│         (@_5 = Base.getindex(@_2, :x))
└──       goto #3
2 ─       Core.Const(:(Core.UndefKeywordError(:x)))
└──       Core.Const(:(@_5 = Core.throw(%5)))
3 ┄ %7  = @_5::Int64
│         (x = %7)
│   %9  = (:x,)::Core.Const((:x,))
│   %10 = Core.apply_type(Core.NamedTuple, %9)::Core.Const(NamedTuple{(:x,)})
│   %11 = Base.structdiff(@_2, %10)::Core.Const(NamedTuple())
│   %12 = Base.pairs(%11)::Core.Const(Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}())
│   %13 = Base.isempty(%12)::Core.Const(true)
│         Core.typeassert(%13, Core.Bool)
└──       goto #5
4 ─       Core.Const(:(Base.kwerr(@_2, @_3)))
5 ┄ %17 = Main.:(var"#f#14")(x, @_3)::Int64
└──       return %17

julia> @code_warntype f(; x=2:2)
MethodInstance for (::var"#f##kw")(::NamedTuple{(:x,), Tuple{UnitRange{Int64}}}, ::typeof(f))
  from (::var"#f##kw")(::Any, ::typeof(f)) in Main at REPL[5]:1
Arguments
  _::Core.Const(var"#f##kw"())
  @_2::NamedTuple{(:x,), Tuple{UnitRange{Int64}}}
  @_3::Core.Const(f)
Locals
  x::UnitRange{Int64}
  @_5::UnitRange{Int64}
Body::Vector{Int64}
1 ─ %1  = Base.haskey(@_2, :x)::Core.Const(true)
│         Core.typeassert(%1, Core.Bool)
│         (@_5 = Base.getindex(@_2, :x))
└──       goto #3
2 ─       Core.Const(:(Core.UndefKeywordError(:x)))
└──       Core.Const(:(@_5 = Core.throw(%5)))
3 ┄ %7  = @_5::UnitRange{Int64}
│         (x = %7)
│   %9  = (:x,)::Core.Const((:x,))
│   %10 = Core.apply_type(Core.NamedTuple, %9)::Core.Const(NamedTuple{(:x,)})
│   %11 = Base.structdiff(@_2, %10)::Core.Const(NamedTuple())
│   %12 = Base.pairs(%11)::Core.Const(Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}())
│   %13 = Base.isempty(%12)::Core.Const(true)
│         Core.typeassert(%13, Core.Bool)
└──       goto #5
4 ─       Core.Const(:(Base.kwerr(@_2, @_3)))
5 ┄ %17 = Main.:(var"#f#14")(x, @_3)::Vector{Int64}
└──       return %17
```
and follows the rough rule of having the output collection match as closely as possible to the input collection.

Maybe the most confusing part is calling it `sites` when their might only be one site, but I think that is ok if it means not having to remember multiple keyword argument names.

emstoudenmire:
I made those changes. Overall I really like where this is headed, for example being able to map things like matrices of operator names into matrices of numbers or vectors of expected values, and similar map type behavior with the new `sites` keyword.

I did run into two tricky issues though, and wondered what you thought about them:

1. There seems to be a fundamental type instability (at least the kind @code_warntype doesn't catch) due to the `ElT` variable inside, which grabs the runtime type of the MPS data and allocates vectors with that kind of element type for the output. Temporarily pinning `ElT` to `Float64` allowed me to successfully use `@code_warntype` to confirm type stability of various inputs of the `ops` and `sites` arguments, but then I restored `ElT` to how it was defined before.
2. Because of the nature of this `expect` function, which makes its outer loop over calls to `orthogonalize!(psi,j)` and inner loop over the operators for efficiency reasons, and because of the immutability of tuples, I couldn't easily make cases like `expect(psi,"Sz"; sites=(2,4,7))` return a tuple instead of a vector. I guess it could be done via one more pass over the output to convert vector to tuple, but that seems wasteful and also would lead to pretty cluttered code. Maybe we just shouldn't officially support that case and recommend people instead do `expect(psi,"Sz"; sites=[2,4,7])`? Or do you see a clean way to do it? (Meaning mapping the type of `sites` to the type of the output?)

mtfishman:
> I made those changes. Overall I really like where this is headed, for example being able to map things like matrices of operator names into matrices of numbers or vectors of expected values, and similar map type behavior with the new `sites` keyword.
> 
> I did run into two tricky issues though, and wondered what you thought about them:
> 
>     1. There seems to be a fundamental type instability (at least the kind @code_warntype doesn't catch) due to the `ElT` variable inside, which grabs the runtime type of the MPS data and allocates vectors with that kind of element type for the output. Temporarily pinning `ElT` to `Float64` allowed me to successfully use `@code_warntype` to confirm type stability of various inputs of the `ops` and `sites` arguments, but then I restored `ElT` to how it was defined before.

Fundamentally, the nature of the way we defined the `ITensor` and `MPS` type where they are not parametrized over the element type means this will be the case (since Julia can't possibly infer the element types of an `ITensor` or `MPS` until runtime, since they can switch arbitrarily).

An efficiency concern in this case for not having the element type inferred is if someone was doing heavy duty data analysis of the results of `expect` like looping over many elements, but in that case they should be passing their data to a function (like to do a fourier transform) in which case the type instability shouldn't really be a concern. They can also use a type constraint like `Sz::Vector{Float64} = expect(psi, "Sz")` to fix the element type themselves. So I think this just isn't something we can solve internally, unless we had a design where the user could optionally pass the element type they expect, like `expect(Float64, psi, "Sz")`, but I don't think we should go there.

>     2. Because of the nature of this `expect` function, which makes its outer loop over calls to `orthogonalize!(psi,j)` and inner loop over the operators for efficiency reasons, and because of the immutability of tuples, I couldn't easily make cases like `expect(psi,"Sz"; sites=(2,4,7))` return a tuple instead of a vector. I guess it could be done via one more pass over the output to convert vector to tuple, but that seems wasteful and also would lead to pretty cluttered code. Maybe we just shouldn't officially support that case and recommend people instead do `expect(psi,"Sz"; sites=[2,4,7])`? Or do you see a clean way to do it? (Meaning mapping the type of `sites` to the type of the output?)

I would be ok with either outputting a `Vector` for that case if handling that is too awkward (and specify in the docs that the output type is ill-defined or could change in the future), or just not allowing that as an input. Handling generic code between `Vector` and `Tuple` can be awkward because of mutability vs. immutability, if it can't be written easily to a call to `map` or broadcast.

emstoudenmire:
Thanks, that confirms exactly what I also thought about those two issues.

If you'd like to take a pass or a look at the code to see if there's an easy way to make `;sites=(2,5,8)` map to a Tuple using something like `map` internally, I would be happy if you could figure it out. But the only way I can see would be some (messy) logic at the end of the function where it checks the type of `sites` and the remakes the output to be the right type, after the calculation is done. (That or the more inefficient version of non-trivial calls to `orthogonalize!` more times than needed, which isn't an option.)

mtfishman:
Perhaps we could use a nested call to `map`:
```julia
function expect_site!(psi, s, opname, j)
  orthogonalize!(psi, j)
  return real(scalar(psi[j] * op(opname, s[j]) * dag(prime(psi[j], s[j])))) / norm2_psi
end

ex = map(opname -> map(j -> expect_site!(psi, s, opname, j), site_range), ops)
```
I didn't test that out so I'm not sure if that works (either the code or the idea).

mtfishman:
Also, what if we are taking the expectation value of a non-Hermitian operator, in which case the result isn't guaranteed to be real? Are we just assuming that won't be the case?

emstoudenmire:
Good point about the Hermitian/non-Hermitian thing. Probably we want to not assume that and then check the operators one-by-one like in `correlation_matrix`. 

The double-map code you mentioned above I think runs into the problem that it does multiple passes over the `sites`, meaning `orthogonalize!` will get called multiple times (and not in the way that it can "lazily" just be a do-nothing call to `orthogonalize!`). We could just accept a situation where calling `expect(psi,"Op1","Op2")` has the same cost as two calls to `expect` but I was trying to take advantage of the extra information of the two operators and do something more efficient.

mtfishman:
Yeah, the call to `orthogonalize!` was the part I wasn't sure about. Maybe there is a way to reorganize the code to pull out the call to `orthogonalize!` to make sure it only happens once per site. Agreed we shouldn't sacrifice efficiency to get this corner case working.

src/itensor.jl
```diff
@@ -1996,6 +1996,21 @@ function indpairs(T::ITensor; plev::Pair{Int,Int}=0 => 1, tags::Pair=ts"" => ts"
...
+
+Test whether an ITensor is a Hermitian operator,
+up to a numerical tolerance. To be considered an
+operator, an ITensor must have matching pairs
+of indices with prime level 0 and 1.
+
+# Optional Keyword Arguments
+- `eps = 1E-10`: the numerical tolerance to use 
+"""
+function ishermitian(T::ITensor; eps=1E-10)
```
mtfishman, line 2010: Let's use `atol` here to match Julia's `isapprox`:
```julia
help?> isapprox
search: isapprox

  isapprox(x, y; atol::Real=0, rtol::Real=atol>0 ? 0 : √eps, nans::Bool=false[, norm::Function])


  Inexact equality comparison. Two numbers compare equal if their relative distance or their absolute distance is within
  tolerance bounds: isapprox returns true if norm(x-y) <= max(atol, rtol*max(norm(x), norm(y))). The default atol is zero and the
  default rtol depends on the types of x and y. The keyword argument nans determines whether or not NaN values are considered
  equal (defaults to false).
```
That could anticipate using `rtol` for a relative tolerance, which could make more sense here anyway since it should be in some way relative to the norm of the `ITensor`.

emstoudenmire: Yes, I was wondering if there was a lower-level thing we should be using here, sort of like how the \approx operator we use in tests calls a Base function.

mtfishman: Actually we could just use:
```julia
ishermitian(T::ITensor; kwargs...) = isapprox(T, swapprime(dag(T), 0 => 1); kwargs...)
```

Additionally, this reminds me that I want to define:
```julia
transpose(T::ITensor) = swapprime(T, 0 => 1)
adjoint(T::ITensor) = transpose(dag(T))
```
as a shorthand so we don't have to write `swapprime(T, 0 => 1)` all the time.

Unfortunately the `adjoint` definition clashes with the definition `adjoint(T::ITensor) = prime(T)` to get the notation `T' == prime(T)` (see https://github.com/JuliaLang/julia/issues/44521), but we could add `transpose(T::ITensor)` right now.

emstoudenmire: I just went ahead and implemented it using `isapprox` internally. By the way, we can use all this work on `expect` as a warmup of some of the improvements you mentioned for `correlation_matrix`, including using `isapprox` there for some of the Hermiticity tests etc.

emstoudenmire: Ok we thought of the same thing - I'll go ahead and define transpose too, yes. Later we could make `transpose` more agnostic about the other prime level being 1, but since we use that convention so widely I think it's ok for now.

mtfishman: Yes exactly. I've wanted to come up with a good standard for interfaces with operator-like ITensors which default to prime levels of 0 and 1, but then you can more generally choose the left and right indices with prime levels and tags.

One place I've done that is the `tr(::ITensor)` function: https://github.com/ITensor/ITensors.jl/blob/10d870155ef78ee6c1257ea07f9c1571a651c3bc/src/itensor.jl#L2002 but I think there could be some things to improve about the design there. Ideally we would have a common code pattern that could be shared across all functions that treat ITensors as operators (i.e. matricizing an ITensor), such as `tr`, `tranpose`, `adjoint`, `eigen`, `exp`, etc.

src/mps/mps.jl
```diff
@@ -772,15 +772,17 @@ function expect(psi::MPS, ops; kwargs...)
   Ns = length(site_range)
   start_site = first(site_range)
 
+  el_types = map(o -> ishermitian(op(o, s[start_site])) ? Float64 : ElT, ops)
```
mtfishman, line 775: My guess is that this should be `real(ElT)` instead of `Float64`, for example in case someone is using single precision.

src/itensor.jl
emstoudenmire: Maybe we could chat about this more next week since it's a good design question. By the way, seeing this `tr` function makes me think I should update the Code Example about how to take the trace of an ITensor. (I could leave the existing material there in some form, but really the `tr` function is the proper way not mainly because it's a function but because `tr` should only be formally taken on indices which are identical in meaning).

src/mps/mps.jl
emstoudenmire: Thanks for catching that

mtfishman:
I don't see anything else, is this ready to merge once tests pass?

emstoudenmire:
I think it's good to go. Thanks for all the careful feedback!

emstoudenmire:
As a follow-up, we should look into making similar changes that you suggested for `correlation_matrix`, especially the `sites` keyword arg.

PR #859, opened 2022-03-12
[WIP] Get some AD working for `Ops`

mtfishman:
This gets some limited AD working for the `Ops` system, for example:
```julia
function f(x)
  y = exp(-x * Op("X", 1) * Op("X", 2))
  y *= exp(-x * Op("X", 1) * Op("X", 2))
  U = Prod{ITensor}(y, s)
  return norm(U(V))
end

@show x = 2.0

@show f(x)
@show f'(x)
```

Unfortunately more complicated examples fail, there are some subtle Zygote bugs to work out but I think this is a good start.


mtfishman:
@GTorlai this may clash with some other AD changes you are making, I know we talked about a code reorganization which I've started a bit here but hopefully it doesn't clash too much with what you've already done.

PR #870, opened 2022-03-17
[ITensorMPS] Bond dimensions produced by randomCircuitMPS

markusschmitt:
# Description

Fixes ITensor/ITensorMPS.jl#69

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes.

- [x] Test bond dimensions for L=10. This test is included in `test/mps.jl` as testset "randomCircuitMPS bond dimensions".
```
using ITensors
using Test

L=10
chi=128

sites = siteinds("S=1/2", L)
mps=myRandomCircuitMPS(ComplexF64,sites,chi)

expected_dims = [2, 4, 8, 16, 32, 16, 8, 4, 2]
for i in 1:9
  l = getfirst(x->hastags(x,"Link,l=$(i)"),inds(mps[i]))
  @test dim(l) == expected_dims[i]
end
```

# Checklist:


src/mps/mps.jl
```diff
@@ -144,15 +144,28 @@ function randomCircuitMPS(
 
   l = Vector{Index}(undef, N)
 
+  # Compute the bond dimension for each link.
```
mtfishman, line 147: Would be nice to have this as a separate function, so we could use it other places in the code and simplify the code here.

Maybe `linkdims_upper_bound`? It doesn't have to be exported, just internal use for now.

mtfishman:
Thanks @markusschmitt.

My major concern here would be ensuring that by truncating in this way the orthogonality center is still at site `1`. Would you mind checking that?

markusschmitt:
Yes, let me check that. I am actually not sure.

mtfishman:
If it does turn out to be an issue, a simple fix would be to determine at what site the orthogonality gets broken (say site `n`), set the orthogonality center range as `1:n`, and then call `orthogonalize!(psi, 1)`, so that it does the minimal orthogonalization needed. But hopefully there won't be anything to fix.

dylex:
I thought I fixed this but maybe it's still not right. I'll take a closer look. In the mean time I manually re-ran the test and now it's failing for non-timeout reasons at least.

src/mps/mps.jl
```diff
@@ -127,7 +127,7 @@ function randomizeMPS!(M::MPS, sites::Vector{<:Index}, linkdim=1)
   end
 end
 
-function randomCircuitMPS(
+function myRandomCircuitMPS(
```
mtfishman, line 130: Looks like you need to change this back to `randomCircuitMPS`.

mtfishman:
> I thought I fixed this but maybe it's still not right. I'll take a closer look. In the mean time I manually re-ran the test and now it's failing for non-timeout reasons at least.

Thanks. Indeed, this PR broke a test. Is there a way that I can rerun tests (without making a new commit)? I didn't see a way from the Jenkins online interface (https://jenkins.flatironinstitute.org/blue/organizations/jenkins/ITensors.jl/detail/PR-870/2/pipeline).

dylex:
If you register a user on jenkins I can add you to be able to trigger builds.

mtfishman:
> If you register a user on jenkins I can add you to be able to trigger builds.

That would be great, thanks. I just registered with the User ID "Matthew" and name "Matthew Fishman".

emstoudenmire:
Nice to see this "bounding" approach to solving the issue, Markus. I think if we can verify even theoretically that the extra truncations still preserve or guaranteed the right-orthogonality property of the MPS tensors (so that the orthogonality center remains well-defined and at site 1) then we're in business.

emstoudenmire:
I think the right-orthogonality may be guaranteed to remain true for any value of chi, including the new values being determined by the code in this PR. The reason is that chi is fed into the function _rmatrix, which makes a unitary (or orthogonal if real) matrix of the requested dimensions. So as long as the column dimensions of these matrices are greater than or equal to the row dimensions (in this case, as long as `chi <= dim(sites[j]) * dim(l[j])`) then the resulting tensors will be right-orthogonal. We could add an assertion that checks the above inequality always holds to ensure orthogonality.

markusschmitt:
@emstoudenmire: I think the condition `chi <= dim(sites[j]) * dim(l[j])` is already ensured in the current version when calculating the maximal bond dimensions.

I also checked the orthogonality as follows:
```
using Test
using ITensors

L=6
chi=128

sites = siteinds("S=1/2", L)

mps=randomCircuitMPS(ComplexF64,sites,chi)

right_contracted = mps[L] * conj(prime(mps[L], tags="Link"))
@test isapprox(norm(right_contracted-delta(inds(right_contracted))), 0.0, atol=1e-14)
for j in (L - 1):-1:2
    right_contracted = (mps[j] * right_contracted) * conj(prime(mps[j], tags="Link"))
    @test isapprox(norm(right_contracted-delta(inds(right_contracted))), 0.0, atol=1e-14)
end
```

This could be added as a test, possibly with different smaller `chi` to check for the cases mentioned by Miles.

emstoudenmire:
Great – I just checked and we do already have a similar test in our test/mps.jl file, so no need to add another one, but it's good you did that check yourself. The new test you added is good to have to ensure the bond dimensions now have the minimal values.

The current test isn't passing, though, because `randomCircuitMPS` is not exported by the ITensors module, as it's an internal function only. Please update the test to use the `randomMPS` interface instead so that the tests can pass.

Also it looks like you'll need to make the update pointed out by Matt, about changing the name `myRandomCircuitMPS` back to `randomCircuitMPS` in src/mps/mps.jl.

mtfishman:
This issue should get transferred to [ITensorMPS.jl](https://github.com/ITensor/ITensorMPS.jl), I'll close it since the corresponding issue is being tracked there: https://github.com/ITensor/ITensorMPS.jl/issues/69.

PR #873, opened 2022-03-18
[DO NOT MERGE] Benchmark speed of tests

emstoudenmire:
(DO NOT MERGE THIS PR!) This PR is just for testing the running time of the unit tests when some of the tests are disabled, to verify if the timings from recent PRs are correct and what impact speeding up these tests could have on the overall run time.


emstoudenmire:
Commenting out five tests (ITensorChainRules, itensor, dmrg, mps, mpo) cut the time of running TestITensors in half, from about 40mins to 20mins. From previous timings, these tests were not dominated by compilation time. So it suggests that we could optimize these tests, especially reducing the total number of tests performed in some cases, to significantly decrease the testing time.

PR #874, opened 2022-03-18
Expect for `MPO` and `ITensor`

GTorlai:
This PR implements the `expect` function for both `MPO` and `ITensor`. For the latter, the `ITensor` is interpreted as a quantum state, but the site indices are required in this case.

```julia
s = siteinds("Qubit",4)
ψ = randomMPS(s, 2)
@show expect(ψ, "X")

ψvec = prod(ψ)
@show expect(ψvec, s, "X")

ρ = outer(ψ, ψ)
@show expect(ρ, "X")

ρmat = prod(ρ)
@show expect(ρmat, s, "X")
```
```julia
expect(ψ, "X") = [0.3582026118288978, -0.3268822430450779, -0.28978898369839623, -0.3574296775773233]
expect(ψvec, s, "X") = [0.35820261182889795, -0.3268822430450778, -0.2897889836983962, -0.3574296775773231]
expect(ρ, "X") = [0.35820261182889884, -0.32688224304508123, -0.28978898369839695, -0.3574296775773229]
expect(ρmat, s, "X") = [0.35820261182889895, -0.3268822430450813, -0.289788983698397, -0.3574296775773232]
```

There is some code duplication for `MPS` and `MPO`. Should we just merge them into a single function?


emstoudenmire:
Thanks! I think these will be good to have. There's an efficiency (combined with simplicity of implementation) question we should discuss about the MPO case, though.

Right now if I'm not mistaken the implementation for MPO scales as N^2 where `N==length(rho)`. This is because for each measurement the code traces all of the non-measured sites when `tr` is called. A more efficient implementation would need to store left and right boundary tensors which are the partial trace of the MPO to the left and right of the site being measured. As I'm sure you saw, in the MPS version we handled this issue by calling `orthogonalize!`.

I think in the MPS case it would be reasonable to change the design to work more like the MPO case (only orthogonalize the MPS to the first site of the `sites` range, say, then make and use boundary tensors to do the remaining measurements) then the implementations could also be unified as a bonus of that design change. Merging them could definitely be a separate PR though.

Lastly, I do wonder about some cases where MPOs are used to represent the square root of a state, to where you inner product two MPOs around an operator to get an expected value. I think that usage of MPOs is more rare though, and a solution there could just be to use an MPS with twice as many sites instead. (For example, in the ancilla method for finite T I usually just use an MPS with twice as many sites, even though the MPS actually represents the square root of a mixed state.)

mtfishman:
Are we going to fix the scaling of `correlation_matrix(::MPO, ...)` in this PR or leave that for a future PR?

Issue #815, opened 2022-02-04
[ITensors] [BUG] Some autodiff errors / missing rules


DomCRose:
**Description of bug**
Hi, great work on this library! I found a few issues with the autodiff support while experimenting. I realise this is still a bit in development, but hopefully this helps!

1. Some rules (e.g. replaceind) appear to assume indicies arguments are provided as pairs, rather than as seperate arguments. Seems like it thinks it has a chain rule for the two-argument function, which is then calling the inv_op defined only for the pair argument form. Looks like this might affect a few functions, e.g. replaceind.
2. ~~Differentiation of ITensor construction appears not to work when indices are provided as a vector. Not clear to me from the code why this one doesn't work.~~ **Solved by generalizing ITensor to Array conversion, see:** #818
3. ~~Defining two order 2 ITensors then taking their trace product appears to produce the transpose of the expected gradient for the input matrices. Perhaps a misunderstanding on my end?~~ **Solved by fixing rrule to use correct index order, see:** #818
4. ~~A slightly more complex problem (trace over a transformation from a small parameterized unitary circuit, see a sketched simplified case below) seems to cause a stack overflow error. I noticed the multi-argument contract chain rule explicitly splits the environment construction into left and right parts. While this would be fine for an MPS (assuming a correctly ordered input), perhaps for something a bit more loopy like this its causing issues? Could this be improved for the generic TN contraction case by simply having a single environment of all tensors but 1 and asking the optimal contraction order calculation to figure out how best to contract it?~~ **Solved by adding changes to enable autodiff of arbitrary TN contractions, see:** #878

**How to reproduce**
Examples in the code below.

**Expected behavior**
Expected correct gradients.

**Actual behavior**
Error 1:
```julia
ERROR: LoadError: Trying to differentiate `replaceind` but the inverse of the operation (`inv_op`) `replaceind` with arguments (ITensor ord=2
Dim 1: (dim=2|id=876)
Dim 2: (dim=2|id=832)
NDTensors.Diag{Float64, Float64}
 2×2
 1.0  0.0
 0.0  1.0, (dim=2|id=772), (dim=2|id=876)) and keyword arguments Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}() is not defined.
Stacktrace:
  [1] error(s::String)
    @ Base .\error.jl:33
[...]

PR #878, opened 2022-03-23
[ITensors] Contraction sequence diff

DomCRose:
# Description

Adds a few small changes to allow reverse mode AD through contractions utilizing contraction sequences.

Fixes part of #815.

# How Has This Been Tested?

Checked it maintains the correct gradients for all current tests and added a test for a contraction involving many tensors using a contraction sequence.

I've also done some rough benchmarking. While for many small contractions the gradient costs 10-15 times the forward pass alone (while theory says 3 would be optimal), in some high bond dimension cases I've seen as little as 2 times the forward pass, suggesting the overhead may become small compared to overhead of the forward pass in that case.

# Checklist:

- [x] (NA, there is little to comment on) I have commented my code, particularly in hard-to-understand areas.
- [x] (NA) I have made corresponding changes to the documentation.
- [ ] (TBD) My changes generate no new warnings.
- [x] (NA?) Any dependent changes have been merged and published in downstream modules.


mtfishman:
Very simple, glad to hear that the overhead gets smaller at larger bond dimensions!

test/ITensorChainRules/test_chainrules.jl
```diff
@@ -65,6 +65,21 @@ Random.seed!(1234)
...
+    randomITensor(a, b, k),
+    randomITensor(a, l, m),
+    randomITensor(b, u, n),
+    randomITensor(u, v),
+    randomITensor(k, v),
+    randomITensor(l, m, n),
+  )
+  f(A, B, C, D, E, F) = (A * B * C * D * E * F)[] # Left associative
+  test_rrule(ZygoteRuleConfig(), f, args...; rrule_f=rrule_via_ad, check_inferred=false)
+  seq = ITensors.optimal_contraction_sequence([A, B, C, D, E, F])
```
mtfishman, line 79: Probably you want to use `[args...]` instead of `[A, B, C, D, E, F]` here.

test/ITensorChainRules/test_chainrules.jl
```diff
@@ -65,6 +65,21 @@ Random.seed!(1234)
...
+  f(A, B, C, D, E, F) = contract([A, B, C, D, E, F]; sequence=seq)[] # sequence
```
mtfishman, line 80: You could just use `f(args...) = contract([args...]; sequence=seq)[]` here.

test/ITensorChainRules/test_chainrules.jl
```diff
@@ -65,6 +65,21 @@ Random.seed!(1234)
...
+  a, b, k, l, m, n, u, v = Index.([10, 50, 5, 2, 10, 12, 7, 50])
+  args = (
+    randomITensor(a, b, k),
+    randomITensor(a, l, m),
+    randomITensor(b, u, n),
+    randomITensor(u, v),
+    randomITensor(k, v),
+    randomITensor(l, m, n),
+  )
+  f(A, B, C, D, E, F) = (A * B * C * D * E * F)[] # Left associative
```
mtfishman, line 77: You could just use `f(args...) = contract([args...])[]` here.

mtfishman:
One issue in the tests is that there are some tests of `rrule(::typeof(*), x1::ITensor, x2::ITensor)` which should now be changed to `rrule(::typeof(contract), x1::ITensor, x2::ITensor)`.

I also commented on the test you added, it looks like it had a few issues.

Beyond that, it looks like there are some mysterious errors from deep within `ChainRulesTestUtils`. I have occasionally hit errors like that and it required overloading some functionality inside of `ChainRulesTestUtils` to make it work properly with ITensors.jl objects like `ITensor` and `Index`. Let's see what the tests say after you make the fixes I suggest and hope they get fixed, otherwise it will require some deeper investigation.

DomCRose:
Aha, thanks for spotting these! I think I fixed these on my other PC but must not have pushed to github. Just pushed fixes (I only found one test for the binary * rrule so hopefully thats it).

DomCRose:
Realized some other issues I also had to fix related to the tests I added (from past PRs also) defining a function f, rather than defining an anonymous function and assigning it to the reference f.

DomCRose:
So it appears to be producing the same two errors I was seeing on my local PC. They are particularly odd as they both appear to occur in the second test of a function, while the first is fine. In one case there is a difference of an argument being complex rather than float, but in the other case the two tests appear completely identical. Not familiar enough with the internals of the chain rules tests to say whats happening here.

DomCRose:
To add to the confusion, these errors seem to only occur when tests are run. When I manually create and run these individual tests in a file, they don't seem to error.

mtfishman:
Thanks for investigating. That is very strange...

I know this is bad software engineering practice, but why don't we just comment out the broken tests for now and investigate it later? It seems like the testing issues shouldn't block merging this PR since it will be very nice to have this working, and we can investigate these testing issues and in the meantime add some tests that implicitly test out this functionality.

DomCRose:
While I agree it would be good to understand why this is happening, since I can't reproduce it outside of running the whole test set, it seems like it may be a testing related issue rather than an actual bug. So it seems reasonable to just comment out the tests for the moment if you think thats alright.

Do you know what functionality these two tests are trying to target which other tests don't cover?

mtfishman:
I'm not sure, I'm sure many of the tests have overlaps. There were many subtle bugs related to index ordering we were hitting when first developing the AD code so some of the tests may be extraneous now.

mtfishman:
Let's just go ahead and comment out the broken tests and look into them later.

mtfishman:
Not sure what that final test failure is but it looks unrelated.

Thanks for the contribution @DomCRose! This will be very nice to have.

PR #881, opened 2022-03-27
Define rmul! for AbstractMPS

emstoudenmire:
# Description

Defined `LinearAlgebra.rmul!(::MPS,::Number)` in terms of `_apply_to_orthocenter!`. The need for this function is that OptimKit requires it for the application of using AD to optimize the overlap of two MPS.

# How Has This Been Tested?

Just indirectly through some driver code that is doing AD on the overlap of two MPS. Maybe we should add an explicit test? But it is just a simple pass-through to `_apply_to_orthocenter!`.


mtfishman:
Thanks! I'm sure I had that defined locally somewhere to get examples of AD with MPS working...

PR #886, opened 2022-03-31
[ITensors] [ENHANCMENT] Introduce `Algorithm` type for selecting algorithm backends

mtfishman:
# Description

Introduce a new `Algorithm` type for selecting algorithm backends and allowing functions to have customizable/extendable algorithm backends. It is now used in `contract(::MPO, ::MPS)` and `truncate!(::MPO/MPS)`. For example, you can now extend `truncate!` as follows:
```julia
s = siteinds("S=1/2", 4)
A = MPO(s, "Id")

# Standard algorithm using "Frobenius norm", i.e. treat it as an MPS
truncate!(A)
truncate!(A; alg="frobenius")

# Low-level internal interface, which can be used for overloading `truncate!`
# with a new algorithm
truncate!(ITensors.Algorithm("frobenius"), A)
```
Then to define a new `truncate!` algorithm for MPOs, you can do:
```julia
import ITensors: truncate!

function truncate!(::ITensors.Algorithm"my_new_algorithm", A::MPO; cutoff=1e-15)
  # Implement a new truncation algorithm for MPOs
end
```
which then can be called as:
```julia
truncate!(A; alg="my_new_algorithm", cutoff=1e-8)
```

Also introduce `alg` as an alternative keyword argument for `method` in `contract(::MPO, ::MPS; kwargs...)`.

@JanReimers @emstoudenmire

### DIFF
--- a/ITensorGPU/test/test_cumpo.jl
+++ b/ITensorGPU/test/test_cumpo.jl
@@ -101,7 +101,7 @@ using ITensors, ITensorGPU, Test
     psi = randomCuMPS(sites)
     psi_out = contract(K, psi; maxdim=1)
     @test inner(phi', psi_out) ≈ inner(phi', K, psi)
-    @test_throws ArgumentError contract(K', psi, method="fakemethod")
+    @test_throws MethodError contract(K', psi, method="fakemethod")
 
     badsites = [Index(2, "Site") for n in 1:(N + 1)]
     badpsi = randomCuMPS(badsites)
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,8 +6,17 @@ Note that as of Julia v1.5, in order to see deprecation warnings you will need t
 
 After we release v1 of the package, we will start following [semantic versioning](https://semver.org).
 
+ITensors v0.3.1 Release Notes
+=============================
+
+Bugs:
+
+Enhancements:
+
+- Introduce `Algorithm` type for selecting algorithm backends (#886)
+
 ITensors v0.3.0 Release Notes
-==============================
+=============================
 
 Bugs:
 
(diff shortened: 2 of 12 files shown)


emstoudenmire:
Looks good. I like this pattern, and it reminds me a tiny bit of the idea of "policies" in C++, though this is simpler and better.

One consideration is naming the type `Alg` instead of `Algorithm` to make the connection to the keyword name as clear as possible. But I'm good with `Algorithm` since it's very self-documenting.

mtfishman:
In fact I've been going back and forth about using `alg` vs. `algorithm`. I used `alg` since that is used already in other places (like `svd_alg` in DMRG), which was originally chosen in that way to match Julia's `LinearAlgebra.svd` function: https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.svd

emstoudenmire:
Yes, and I was just meaning possibly using both `alg="name"` and `ITensors.Alg"name"`. That way it's crystal clear that the keyword argument goes with the type. But mixing alg and algorithm is not too confusing.

mtfishman:
Agreed, that could be nice to make them match, I was just saying if we did that I would rather go the other way and use `algorithm` for the keyword argument and `Algorithm"algorithm"` for the type name.

emstoudenmire:
Yes that would be totally fine with me.

mtfishman:
I'll just stick with the current convention calling `alg` the keyword argument and `Algorithm` the type, it looks pretty standard in Julia:

https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.svd
https://docs.julialang.org/en/v1/base/sort/#Sorting-Algorithms

PR #887, opened 2022-04-02
Cleanup outer and add GEMM routing for CUDA

kshyatt:
# Description
This actually turned into a code addition. Added some functions as fall backs for array types that `@strided` doesn't work on, and fixed up some `ITensorGPU` code to be more consistent with `ITensors`.

Fixes # (issue)

# How Has This Been Tested?

`julia test/runtests.jl` passed for `ITensorGPU`, `NDTensors`, and `ITensor`.

# Checklist:

### DIFF
--- a/ITensorGPU/src/cuitensor.jl
+++ b/ITensorGPU/src/cuitensor.jl
@@ -25,8 +25,6 @@ function ITensor(
   inds::Indices{Index{Int}};
   kwargs...,
 )
-  @show size(A)
-  @show dim(inds)
   length(A) ≠ dim(inds) && throw(
     DimensionMismatch(
       "In ITensor(::CuArray, inds), length of AbstractArray ($(length(A))) must match total dimension of IndexSet ($(dim(inds)))",
--- a/ITensorGPU/src/cuitensor.jl
+++ b/ITensorGPU/src/cuitensor.jl
@@ -1,3 +1,7 @@
+import ITensors.NDTensors: NeverAlias, AliasStyle, AllowAlias 
+import ITensors: ITensor
+import CUDA: CuArray
+
 function cuITensor(::Type{T}, inds::IndexSet) where {T<:Number}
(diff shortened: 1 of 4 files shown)


ITensorGPU/src/cuitensor.jl
```diff
@@ -14,6 +18,41 @@ function cuITensor(x::S, inds::IndexSet{N}) where {S<:Number,N}
...
+
+function ITensor(
+  as::AliasStyle,
+  eltype::Type{<:Number},
+  A::CuArray{<:Number},
+  inds::Indices{Index{Int}};
+  kwargs...,
+)
+  @show size(A)
+  @show dim(inds)
```
emstoudenmire, line 30: Looks like there are some @show calls here that need to be commented out?

mtfishman:
Thanks @kshyatt, I'll take a look.

Could you run the formatter? You can do:
```julia
julia> cd(joinpath(ENV["HOME"], ".julia", "dev", "ITensors"))

julia> using JuliaFormatter

julia> format(".")
```

emstoudenmire:
Or if it's handy to either of you, I made a little script that I run from the top level of the folders containing files I want to format. Here's the script:

```
#!/bin/bash
julia --banner=no --color=yes --startup-file=no -e "using JuliaFormatter; format(\".\";verbose=true)"
```
then I put that into a file in my path called `julia_format`, chmod it to 755, and run it as a command.

mtfishman:
Looks good Katie, thanks! Looks like that ITensor constructor you added could get turned into a generic constructor for `AbstractArray` (i.e. both `CuArray` or `Array`). I'm hoping we can start moving in the direction of making the code in ITensors.jl more generic so it work better with `CuArray` data types.

I think the only remaining thing is removing the `@show` lines that Miles pointed out, then I'll merge if you think this is good to go.

mtfishman (approved):

PR #891, opened 2022-04-03
Finite temperature example codes

emstoudenmire:
This PR adds two new example codes for the METTS and purification methods for computing finite temperature properties. Both codes were tested against each other and using ED on a small system size. One improvement could be to add another observable besides the energy. Please let me know of any feedback about the way the codes are written or designed.

I did notice one possible issue with the `apply` function: while it behaves as expected for METTS (nearest-neighbor gates only), for purification which uses second-neighbor gates, `apply` seems extremely slow even for N=12. It might be scaling quadratically with N for this case or encountering some other performance issue.

- Add METTS and purification code examples
- Formatting

### DIFF
--- a/examples/finite_temperature/metts.jl
+++ b/examples/finite_temperature/metts.jl
@@ -1,0 +1,103 @@
+using ITensors
+import ITensors: op
+using Printf
+
+function op(::OpName"expτSS", ::SiteType"S=1/2", s1::Index, s2::Index; τ)
+  h =
+    1 / 2 * op("S+", s1) * op("S-", s2) +
+    1 / 2 * op("S-", s1) * op("S+", s2) +
+    op("Sz", s1) * op("Sz", s2)
+  return exp(τ * h)
+end
+
+"""
+Given a Vector of numbers, returns
+the average and the standard error
+(= the width of distribution of the numbers)
+"""
+function avg_err(v::Vector)
+  N = length(v)
+  avg = v[1]/N
+  avg2 = v[1]^2/N
+  for j=2:N
+    avg  += v[j]/N
+    avg2 += v[j]^2/N
+  end
+  return avg,√((avg2-avg)/N)
+end
+
+function main(; N=10, cutoff=1E-8, δτ=0.1, beta=2.0, NMETTS=3000, Nwarm=10)
+
+  # Make an array of 'site' indices
+  s = siteinds("S=1/2", N)
+
+  # Make gates (1,2),(2,3),(3,4),...
+  gates = ops([("expτSS", (n, n + 1), (τ=-δτ/2,)) for n in 1:(N - 1)], s)
+  # Include gates in reverse order to complete Trotter formula
+  append!(gates, reverse(gates))
+
+  # Make y-rotation gates to use in METTS collapses
+  Ry_gates = ops([("Ry",n,(θ=π/2,)) for n=1:N],s)
+
+  # Arbitrary initial state
+  psi = randomMPS(s)
+
+  # Make H for measuring the energy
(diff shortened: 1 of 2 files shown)


emstoudenmire:
Partly addresses issue #864

mtfishman:
Thanks! I'll take a look.

What about doing purification directly on an MPO (i.e. apply the gates to one side of an identity MPO)? Then the gates would be local. Of course it would be good to investigate why the MPS version is slow.

mtfishman:
You should be able to do:
```julia
rho = MPO(s, "Id") ./ √2
apply(gates, rho; cutoff)
```
`apply` here will apply the gates to one side of the MPO.

emstoudenmire:
Good suggestion. I'll make that change. 

How does one take the expectation value of H with an MPO "state" (density matrix / mixed state). Do we have a function that does that such as `inner`?

Finally, I think I'll still file an issue about the `apply` behavior in the MPS case here, because it should be pretty fast for second-neighbor gates and should scale linearly with N. (I checked that the bond dimensions were small, like 8, yet apply within purification.jl was taking many 10's of seconds even for sizes like N=20.)

mtfishman:
Agreed the MPS version should be fast, but in this case the MPO version may be preferred anyway since the code is simpler.

`inner` should work with `MPO`s, it just contracts over the shared indices. So try `inner(rho, H)`.

Alternatively you could do `tr(apply(rho, H))` but that is not optimized since it does the MPO-MPO contraction before tracing. We should define an optimized version `tr(Apply(rho, H))` but that's not defined now.

mtfishman:
One issue I can imagine in the non-local gates case is that there are too many swaps happening.

Could you try changing the line:
```julia
    psi = apply(gates, psi; cutoff)
```
to:
```julia
    psi = apply(gates, psi; cutoff, move_sites_back=false)
```
This disables the function that returns the sites back to their original positions after the final gate in `gates` is applied. That would be a quick test to see if that final step is the culprit for the poor performance (that would be a good scenario since that step is one that I'm sure could be optimized, it calls a general function I wrote for permuting the sites of an MPS using swap gates but I'm sure we could come up with a better approach than the one I used). If the sites don't get moved back, they will be mixed up so it won't be as easy to compute expectation values.

emstoudenmire:
Ok updated!

Unfortunately the `move_sites_back=false` mode for the MPS case seems to have a bug in it and gave an error about mismatched indices when I tried using it. (The same code without that argument did run successfully.)

mtfishman:
Was the mismatched indices error from the `inner` function? That is expected because it outputs an MPS where the site indices are permuted from the original ones. Try removing the call to `inner`.

emstoudenmire:
Yes it was a call to inner in my code that was causing the error, thanks.

So I did the timings now and I would say that (when using MPS again), both versions of apply and (with move_sites_back=false passed versus not passed) are too slow, however passing move_sites_back=false does make a noticeable difference. For small sizes like N=4, 6, 8 the move_sites_back=false is about 6 times faster (taking 0.01 seconds per apply versus 0.06 times). But then for slightly larger sizes like N=10 the move_sites_back=false grows to about 0.3s per apply while without that it's about 1.4s per apply. 

As I increase the size even just a little past that, even with move_sites_back=false, the time per apply grows to about 5s for N=12, then for N=14 (length of MPS = 28) just one apply took 215 seconds (!), even with move_sites_back=false unfortunately.

emstoudenmire:
Great. Yes I see in the code where I can do that & happy to study it a bit. I might have time this week but not sure, so I made an issue just now and assigned myself to it so hopefully I won't forget.

PR #901, opened 2022-04-19
[ITensors] [ENHANCEMENT] Handle corner case of normalizing a zero MPS

mtfishman:
# Description

This PR checks if the norm of the MPS is nonzero before normalizing. The new behavior is that now if the norm is zero, it performs a no-op. Before, it returned an MPS filled with `NaN`.

<details><summary>Before this PR</summary><p>

```julia
julia> using ITensors

julia> ψ = randomMPS(s);

julia> ψ = 0 * randomMPS(s);

julia> norm(ψ)
0.0

julia> lognorm(ψ)
-Inf

julia> normalize!(ψ);

julia> norm(ψ)
NaN

julia> lognorm(ψ)
NaN
```
</p></details>

<details><summary>After this PR</summary><p>

```julia
julia> using ITensors

julia> ψ = randomMPS(s);

julia> ψ = 0 * randomMPS(s);

julia> norm(ψ)
0.0

julia> lognorm(ψ)
-Inf

julia> normalize!(ψ);

julia> norm(ψ)
0.0

julia> lognorm(ψ)
-Inf
```
</p></details>

Julia goes the route of returning an Array filled with `NaN`:
```julia
julia> using LinearAlgebra

julia> normalize!(zeros(2, 2))
2×2 Matrix{Float64}:
 NaN  NaN
 NaN  NaN
```

The motivation for the new behavior is that we had a user who is doing an application with monitored circuits, i.e. evolving circuits with projected measurements scattered throughout the gate evolution. Some of these projective measurements were leading to states with zero norm. The user had steps calling `normalize!` to normalize the states after the projective measurements, which was introducing `NaN`s into the MPS which was leading to some mysterious errors in different places (for example in a subsequent eigendecomposition inside of `apply`).

This change helped avoid errors like that and the code ran smoothly. A downside is that it may be less clear when a zero state is obtained, so is just putting off the issue for later. But that could get caught in other places in the code. For example, we could check for zero states in `apply`, `correlation_matrix`, etc. or the user will notice when they try to compute properties of the state.

@emstoudenmire I'm curious about your thoughts on this.

# How Has This Been Tested?

Tests added for normalizing zero MPS.

# Checklist:

### DIFF
--- a/src/mps/abstractmps.jl
+++ b/src/mps/abstractmps.jl
@@ -1264,12 +1264,24 @@ Change the MPS or MPO `A` in-place such that `norm(A) ≈ 1`. This modifies the 
 
 In practice, this evenly spreads `lognorm(A)` over the tensors within the range of the orthogonality center to avoid numerical overflow in the case of diverging norms.
 
+If the norm of the input MPS or MPO is 0, normalizing is ill-defined. In this case, we just return the original MPS or MPO. You can check for this case as follows:
+```julia
+s = siteinds("S=1/2", 4)
+ψ = 0 * randomMPS(s)
+lognorm_ψ = []
+normalize!(ψ; (lognorm!)=lognorm_ψ)
+lognorm_ψ[1] == -Inf # There was an infinite norm
+```
+
 See also [`normalize`](@ref), [`norm`](@ref), [`lognorm`](@ref).
 """
 function normalize!(M::AbstractMPS; (lognorm!)=[])
   c = ortho_lims(M)
   lognorm_M = lognorm(M)
   push!(lognorm!, lognorm_M)
+  if lognorm_M == -Inf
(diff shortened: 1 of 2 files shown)


emstoudenmire (approved):
Looks good, as long as the norm being zero really does lead to a log norm of -Inf. I think that's the kind of thing I would be nervous about in C++ but feel like they have handled more robustly with the superior math system in Julia. And you added a test explicitly for this.

mtfishman:
Thanks for checking it over.

I've been going back and forth about this design choice, but ultimately my feeling on this is that normalizing a zero vector/MPS is ill-defined anyway, so we may as well output the MPS which will be the most useful and cause the least amount of trouble, which I think is the original MPS.

PR #902, opened 2022-04-19
[ITensors] [ENHANCEMENT] Allow `Matrix` representations for operators in `expect` and `correlation_matrix`

mtfishman:
# Description

Allow `Matrix` representations for operators in `expect` and `correlation_matrix`. This can make it a bit easier to measure an operator that is not defined already without having to define the operator by overloading `op`.

For inspiration see the discussion here: https://itensor.discourse.group/t/measuring-correlations/51/6

```julia
julia> using ITensors

julia> ψ = randomMPS(s);

julia> expect(ψ, [1/2 0; 0 -1/2])
4-element Vector{Float64}:
  0.1968928836524881
 -0.4253469177197064
  0.4028392528567332
 -0.4026126134149185

julia> expect(ψ, "Sz")
4-element Vector{Float64}:
  0.1968928836524881
 -0.4253469177197064
  0.4028392528567332
 -0.4026126134149185

julia> correlation_matrix(ψ, [1/2 0; 0 -1/2], [1/2 0; 0 -1/2])
4×4 Matrix{Float64}:
  0.25       -0.0837478   0.0793162  -0.0792716
 -0.0837478   0.25       -0.171346    0.17125
  0.0793162  -0.171346    0.25       -0.162188
 -0.0792716   0.17125    -0.162188    0.25

julia> correlation_matrix(ψ, "Sz", "Sz")
4×4 Matrix{Float64}:
  0.25       -0.0837478   0.0793162  -0.0792716
 -0.0837478   0.25       -0.171346    0.17125
  0.0793162  -0.171346    0.25       -0.162188
 -0.0792716   0.17125    -0.162188    0.25
```
</p></details>

# How Has This Been Tested?

Added tests of the behavior shown above.

# Checklist:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,6 +4,22 @@ Please include a summary of the change and which issue is fixed (if applicable).
 
 Fixes #(issue)
 
+If practical and applicable, please include a minimal demonstration of the previous behavior and new behavior below.
+
+<details><summary>Minimal demonstration of previous behavior</summary><p>
+
+```julia
+[YOUR MINIMAL DEMONSTRATION OF PREVIOUS BEHAVIOR]
+```
+</p></details>
+
+<details><summary>Minimal demonstration of new behavior</summary><p>
+
+```julia
+[YOUR MINIMAL DEMONSTRATION OF NEW BEHAVIOR]
+```
+</p></details>
+
 # How Has This Been Tested?
 
 Please add tests that verify your changes to a file in the `test` directory.
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -555,10 +555,18 @@ function sample(m::MPS)
   return result
 end
 
+_op_prod(o1::AbstractString, o2::AbstractString) = "$o1 * $o2"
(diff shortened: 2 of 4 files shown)


emstoudenmire (approved):
Looks good. Not a big change to the code required.

emstoudenmire:
I like the idea so let's merge this. The only possible failure modes are these:

1. for fermions, making the operator from a matrix will lose the information about whether it has J-W string in the current system, so would lead to wrong results if someone tried to make "C" or "Cdag" from a matrix
2. if the sites are of mixed type or dimension, then the operator from a matrix may only be meaningful for some of the sites

I'm ok with #1 because it's just another one of the pitfalls of the current fermion system, and would be fixed by the new one. For #2 it's a little more of a possible silent failure mode if all the sites have the same dimension but a different physical meaning. But I think we're ok there because the user is still being explicit about what they are doing i.e. "sandwich this matrix between every site" so they have to know what to expect.

In summary again, I think this is good to merge.

mtfishman:
Agreed that those are the main downsides, but I also agree with your assessment that 1. should be fixed by the new Fermion system and 2. is acceptable for the reasons you give.

PR #904, opened 2022-04-21
[ITensors] [ENHANCEMENT] Add support for defining MPOs from operators represented as matrices

mtfishman:
# Description

Motivated by: https://itensor.discourse.group/t/generate-maximally-mixed-state/67

Over all it would be good to be able to use matrices interchangeably with the string representation, like #899 and #902, so this helps us move towards that goal.

```julia
using ITensors
s = siteinds("S=1/2", 4)
H = MPO(s, [1 0; 0 1])
H ≈ MPO(s, "Id")

# Function version
H = MPO(s, n -> [1 0; 0 1])
H ≈ MPO(s, "Id")
```
</p></details>

# How Has This Been Tested?

Tests were added for the behavior above.

# Checklist:

### DIFF
--- a/src/mps/mpo.jl
+++ b/src/mps/mpo.jl
@@ -102,7 +102,8 @@ end
 MPO(sites::Vector{<:Index}, op::String) = MPO(Float64, sites, op)
 
 function MPO(::Type{ElT}, sites::Vector{<:Index}, op::Matrix{<:Number}) where {ElT<:Number}
-  return MPO(ElT, sites, fill(op, length(sites)))
+  # return MPO(ElT, sites, fill(op, length(sites)))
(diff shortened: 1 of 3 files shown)


mtfishman:
@emstoudenmire my main concern here is that plain matrix version:
```julia
H = MPO(s, [1 0; 0 1])
```
could get confused with this MPO constructor:
```julia
MPO(randn(2^8), [[sₙ', dag(sₙ)] for sₙ in s])
```
which first reinterprets the input Array as an ITensor and then SVDs the ITensor into an MPO.

We could only allow the version:
```julia
H = MPO(s, n -> [1 0; 0 1])
```
to avoid any confusion. The second version makes it clearer that the operator goes on each site. Of course they can already be disambiguated by the ordering of the inputs.

emstoudenmire:
It's a good point about those constructors being easily confused. My vote would be, of the new ones, to just implement the one taking a function (`n -> [1 0; 0 1]` in your example). I like that version even if the `n` argument doesn't get used because it's both explicit about what it's doing and it makes it clear how to generalize it to the matrices varying with the site number.

PR #906, opened 2022-04-26
[ITensors] [ENHANCEMENT] Clean up QN `svd` code in `ITensors` by handling QN blocks better in `NDTensors`

emstoudenmire:
## Description

Removes the QN fixing logic in `ITensors.svd` in favor of building the new indices of U and V inside of `NDTensors.svd`. This change will be **essential to fixing the SVD for the auto fermion system**. 

Handling the QNs in a generic way takes a bit of care inside NDTensors, so I added a few helper methods that use the concept of "blocks" of indices such as QNIndex or BlockDim, where the blocks for a QNIndex are QNBlocks. The function `addblock!` pushes new blocks onto the end of the space of a block-sparse index.

However, the resulting code is I think simpler than before, since it basically just builds up the new indices by looping over the non-zero blocks of U and V and then reading off the sizes of their column indices. The rest of the new code is just making sure that other information in the spaces such as the QN information gets passed through when making the new indices.

## How Has This Been Tested?

Ran the qnitensor.jl unit tests (plus will run the CI below) and some tests of my own.

### DIFF
--- a/NDTensors/src/blocksparse/blockdims.jl
+++ b/NDTensors/src/blocksparse/blockdims.jl
@@ -202,6 +202,12 @@ blockindex(T) = (), Block{0}()
 # This is to help with ITensor compatibility
 #
 
+block(i::BlockDim,n::Integer) = i[n]
+
+resize(n::Int,newdim::Int) = newdim
+
+addblock!(i::BlockDim,b::Int) = push!(i,b)
+
 setblockdim!(dim1::BlockDim, newdim::Int, n::Int) = setindex!(dim1, newdim, n)
 
 sim(dim::BlockDim) = copy(dim)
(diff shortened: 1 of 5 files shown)


mtfishman:
Great! Very glad to see that hacky code get removed from the SVD function, and glad that this code cleanup makes it easier to implement the fermion logic.

What if we call it `pushblock!` instead of `addblock!`? That makes it clearer what it is doing (since `push!` means insert a new element at the end of the collection) and also wouldn't get confused with arithmetic addition, like adding quantum numbers.

mtfishman:
Besides that suggested name change this is good to go from my end.

emstoudenmire:
I like that suggestion. I actually went in a slightly different direction, where I realized the size of the new indices is known in advance, so rather than pushing blocks one by one, they can just be set directly. It is probably never a bottleneck, but can't be slower so is in the spririt of avoiding "premature pessimization".

The new function I made for this, `setblock!` fits nicely alongside `setblockdim!` and `setblockqn!`.

mtfishman:
Looks good! Feel free to merge once tests pass.

PR #914, opened 2022-05-13
[ITensors] [BUG] Fix some issues with `directsum`

emstoudenmire:
Makes some improvements to the `directsum` function, which I used in the implementation of the global subspace expansion method in a current PR for the ITensorTDVP package.

### DIFF
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -2222,7 +2222,7 @@ function directsum(
 end
 
 function directsum(A::ITensor, B::ITensor, I, J; kwargs...)
-  tags = get(kwargs,:tags,["sum1","sum2"])
+  tags = get(kwargs, :tags, ["sum1", "sum2"])
   N = length(I)
   (N != length(J)) &&
     error("In directsum(::ITensor, ::ITensor, ...), must sum equal number of indices")
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -2221,7 +2221,8 @@ function directsum(
   )
 end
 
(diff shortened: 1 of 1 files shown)


src/itensor.jl
```diff
@@ -2221,16 +2221,17 @@ function directsum(
...
   (N != length(J)) &&
     error("In directsum(::ITensor, ::ITensor, ...), must sum equal number of indices")
   IJ = Vector{Base.promote_eltype(I, J)}(undef, N)
   for n in 1:N
     In = I[n]
     Jn = J[n]
-    In = dir(A, In) != dir(In) ? dag(In) : In
-    Jn = dir(B, Jn) != dir(Jn) ? dag(Jn) : Jn
+    In = dir(A, In.first) != dir(In) ? dag(In) : In
+    Jn = dir(B, Jn.first) != dir(Jn) ? dag(Jn) : Jn
```
mtfishman, line 2234: I don't think this is correct, `I` and `J` should be lists of indices.

emstoudenmire:
I'm going to close this PR and reevaluate what changes (if any) are needed to `directsum`. It might have just been I didn't know how to use it properly, so the fix could just be a docstring. The context is that the global subspace expansion code in ITensorTDVP uses it.

src/itensor.jl
emstoudenmire: Thanks. I wasn't totally clear on what types I and J were expected to have. I'll reevaluate how I tried to use `directsum` in ITensorTDVP and get a better understanding of it.

PR #915, opened 2022-05-13
[ITensors] [ENHANCMENT] Introduce `removeqn` function for removing a specified quantum number

mtfishman:
# Description

Introduce `removeqn` function for removing a specified quantum number.

```julia
julia> using ITensors

julia> s = siteind("Electron"; conserve_qns=true);

julia> T = op("c†↑", s);

julia> @show T;
T = ITensor ord=2
Dim 1: (dim=4|id=685|"Electron,Site")' <Out>
 1: QN(("Nf",0,-1),("Sz",0)) => 1
 2: QN(("Nf",1,-1),("Sz",1)) => 1
 3: QN(("Nf",1,-1),("Sz",-1)) => 1
 4: QN(("Nf",2,-1),("Sz",0)) => 1
Dim 2: (dim=4|id=685|"Electron,Site") <In>
 1: QN(("Nf",0,-1),("Sz",0)) => 1
 2: QN(("Nf",1,-1),("Sz",1)) => 1
 3: QN(("Nf",1,-1),("Sz",-1)) => 1
 4: QN(("Nf",2,-1),("Sz",0)) => 1
 4×4
Block(2, 1)
 [2:2, 1:1]
 1.0

Block(4, 3)
 [4:4, 3:3]
 1.0

julia> R = removeqn(T, "Sz");

julia> @show R;
R = ITensor ord=2
Dim 1: (dim=4|id=685|"Electron,Site")' <Out>
 1: QN("Nf",0,-1) => 1
 2: QN("Nf",1,-1) => 2
 3: QN("Nf",2,-1) => 1
Dim 2: (dim=4|id=685|"Electron,Site") <In>
 1: QN("Nf",0,-1) => 1
 2: QN("Nf",1,-1) => 2
 3: QN("Nf",2,-1) => 1
 4×4
Block(2, 1)
 [2:3, 1:1]
 1.0
 0.0

Block(3, 2)
 [4:4, 2:3]
 0.0  1.0
```
</p></details>

# How Has This Been Tested?

Tests were added checking the elements are the same before and after and that the QNs are the ones we expect.

# Checklist:

### DIFF
--- a/src/qn/qnitensor.jl
+++ b/src/qn/qnitensor.jl
@@ -467,6 +467,9 @@ function δ_split(i1::Index, i2::Index)
 end
 
 function splitblocks(A::ITensor, is=inds(A); tol=0)
+  if !hasqns(A)
+    return A
+  end
   isA = filterinds(A; inds=is)
   for i in isA
     i_split = splitblocks(i)
@@ -480,4 +483,21 @@ function splitblocks(A::ITensor, is=inds(A); tol=0)
   end
   A = dropzeros(A; tol=tol)
   return A
+end
+
+function removeqn(T::ITensor, qn_name::String; mergeblocks=true)
+  if !hasqns(T)
+    return T
+  end
+  inds_R = removeqn(inds(T), qn_name)
+  if mergeblocks
+    inds_R = ITensors.mergeblocks(inds_R)
+  end
+  R = ITensor(inds_R)
+  for iv in eachindex(T)
+    if !iszero(T[iv])
+      R[iv] = T[iv]
+    end
+  end
+  return R
 end
--- a/src/qn/qnitensor.jl
+++ b/src/qn/qnitensor.jl
@@ -489,10 +489,7 @@ function removeqn(T::ITensor, qn_name::String; mergeblocks=true)
   if !hasqns(T)
     return T
   end
-  inds_R = removeqn(inds(T), qn_name)
-  if mergeblocks
-    inds_R = ITensors.mergeblocks(inds_R)
-  end
+  inds_R = removeqn(inds(T), qn_name; mergeblocks)
   R = ITensor(inds_R)
   for iv in eachindex(T)
     if !iszero(T[iv])
--- a/src/exports.jl
+++ b/src/exports.jl
@@ -69,6 +69,7 @@ export
   plev,
   prime,
   removetags,
+  removeqn,
   removeqns,
   replacetags,
   replacetags!,
--- a/src/index.jl
+++ b/src/index.jl
@@ -517,6 +517,21 @@ Removes the QNs from the Index, if it has any.
 """
 removeqns(i::Index) = i
 
+"""
+    removeqn(::Index, qn_name::String)
+
+Remove the specified QN from the Index, if it has any.
+"""
+removeqn(i::Index, qn_name::String) = i
+
+"""
+    mergeblocks(::Index)
+
+Merge the contiguous QN blocks if they have the same
+quantum numbers.
+"""
+mergeblocks(i::Index) = i
+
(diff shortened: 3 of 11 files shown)


emstoudenmire:
Nice to have this! I can see how it can be very useful.

Looking at the code, I guess right now this doesn't merge blocks that end up having the same QN after this procedure correct? That's not necessarily a problem, as we saw with the "splitblocks" experiments with DMRG where packing this as much as possible isn't always an advantage, but I just wanted to understand if that's done here.

mtfishman:
It merges contiguous blocks (you can see that in the example in the first post). There is an option to turn that off with the keyword argument `mergeblocks=false`.

It doesn't permute the blocks, so sometimes it won't fully merge the blocks, for example if you remove the `"Nf"` QN in the example in the first post it won't merge the `("Sz", 0)` sectors since they are not next to each other (they are the first and last states for the `"Electron"` type, as you know). The reason for that is I didn't want this function to produce indices with new Index ids (so that you could for example remove a QN from both an MPS and MPO and keep the indices compatible). It seems ok to keep the Index id the same if QN are just merged as long as the states remain in the original ordering (i.e. the tensor elements are the same, just the sparsity pattern changes).

emstoudenmire:
Yes that's the right behavior that we want I agree.

What I was asking about also was at the storage level. Like are the number of separate blocks in the storage changed at all? Or it's just kept the same and everything still works as long as the new index values are in 1-1 correspondence with the old ones, even if a single QN block now spans over two different blocks in the storage?

mtfishman:
With `mergeblocks=false`, the block structure remains the same (same number of blocks of the same size, just with different QNs).

With `mergeblocks=true` (the default), if there are contiguous blocks with the same QNs after QNs are removed, blocks are merged, so it results in a tensor with fewer blocks and for blocks that were merged they are larger and filled out with zeros.

mtfishman:
Let me know if that sounds reasonable to you, if so I'll merge and register a new version.

emstoudenmire:
Great. Yes please go ahead and merge this PR. 

I was wondering, just for my understanding, which line of code modifies the tensor storage, since I could only see code here that was modifying indices.

mtfishman:
The current implementation is pretty simplistic. It make a new ITensor with empty storage using the new indices and then sets the elements one by one. In this way it dynamically allocates blocks that are nonzero as needed. This is more of a "quick and dirty" solution which could be optimized (for example a simple optimization in the `mergeblocks=false` case would be to just copy the storage). However I doubt this would be a bottleneck in a calculation, I imagine it is usually just done once before doing some measurements or a new DMRG run.

src/qn/qnitensor.jl
```diff
@@ -481,3 +484,17 @@ function splitblocks(A::ITensor, is=inds(A); tol=0)
...
+  if !hasqns(T)
+    return T
+  end
+  inds_R = removeqn(inds(T), qn_name; mergeblocks)
+  R = ITensor(inds_R)
+  for iv in eachindex(T)
+    if !iszero(T[iv])
+      R[iv] = T[iv]
+    end
+  end
```
mtfishman, line 498: This is where the new tensor with the new storage is being made.

emstoudenmire: Ok thanks! I kept missing it when looking through the changes. Yes later on I imagine we can have better facilities for slicing and then could do sliced assignments of entire blocks of T into either blocks or sub-blocks of R. (As a side thought, better slicing for both dense and block sparse tensors might simplify the block sparse combiner code.)

mtfishman: Yeah, this would be a good candidate for making use of better slicing operations.

PR #923, opened 2022-05-25
Implement min_blockdim keyword for blocksparse SVD

emstoudenmire:
# Commits

- Implement block_mindim keyword for QN SVD
- Add test and improve some docs and an export
- Change parameter name to min_blockdim

# Description

Implements an optional keyword `min_blockdim` which defaults to 0, meaning it has no effect by default. When set, at least `min_blockdim` singular values are kept for each original block of the tensor. (Of course once this number exceeds the size of that block then all the singular values are kept from that block regardless of `min_blockdim`.)

One example usage is setting `min_blockdim=1` which ensures all blocks are kept, though they might be truncated to as small as one singular value. This could be useful for cases where one wants all possible quantum numbers in the new indices formed to be present.

There is still also the `mindim` parameter which acts globally on all the singular values, collected together, whereas `min_blockdim` is applied to each block individually.

# How Has This Been Tested?

New test added in the decomp.jl test file.

### DIFF
--- a/NDTensors/src/blocksparse/linearalgebra.jl
+++ b/NDTensors/src/blocksparse/linearalgebra.jl
@@ -4,11 +4,11 @@ const DiagBlockSparseMatrix{ElT,StoreT,IndsT} = DiagBlockSparseTensor{ElT,2,Stor
 const DiagMatrix{ElT,StoreT,IndsT} = DiagTensor{ElT,2,StoreT,IndsT}
 
 function _truncated_blockdim(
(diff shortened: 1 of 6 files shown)


mtfishman:
Great, glad to see it is a very minimal change.

emstoudenmire:
Yes most of the changed lines were me just renaming a few variables and throwing in a new docstring etc. The actual code changes needed were very small, essentially 5 lines or so!

PR #956, opened 2022-07-14
[ITensorGPU] [ENHANCEMENT] Improve functionality for transferring data between CPU and GPU

mtfishman:
# Description

This PR is meant to fix doing automatic differentiation of a circuit simulation with MPS/MPO on GPU, including PastaQ's `runcircuit`.

There were a few MPS/MPO functions that weren't differentiable, including ones for transferring MPS/MPO back and forth between CPU/GPU.

In the process, I also simplified some of the GPU code, for example I rewrote the MPS/MPO constructors to just call the CPU versions and then transfer the results to GPU. I think the tradeoff of any performance issues with that (if there are any) are outweighed by the code simplification.

# How Has This Been Tested?

Need to add tests, possibly including doing AD of `PastaQ.runcircuit` on GPU.

# Checklist:


mtfishman:
@kshyatt the main changes here to `ITensorGPU` are:

1. A more generic `cu` function built on top of `Adapt.adapt` and `Functors.fmap` (and more generally, making `ITensor` objects compatible with [Adapt.jl](https://github.com/JuliaGPU/Adapt.jl)). This also includes some better functionality for adapting the element type and storage/data types in general.
2. Simplify the ITensorGPU `MPS`/`MPO` constructors by just calling the CPU versions and then moving the result to GPU. That seemed to basically be what the previous versions were doing anyway.
3. Use the new `adapt` functionality to make `expect` and `correlation_matrix` work on GPUs. The basic issue with the previous version was that `op` creates ITensors on CPU by default, so then when you try to contract them with the MPS tensors which are on GPU it throws an error since we don't support contract CPU and GPU ITensors (I've also improved the error message when trying to contract CPU and GPU tensors together). The fix here is to use `adapt` to adapt the data type of the ITensor built with `op` to the data type of the tensors of the input MPS. We could consider adding support to `op` where you can input a custom data type (so `op(Vector, "Sz", s, 1)` if you want it on CPU and `op(CuVector, "Sz", s, 1)` if you want it on GPU but I'll leave that for a future PR).
4. Better handling of `EmptyStorage` with `CuVector` data types.

There is a lingering question I have with the `cu` function that the `CUDA.cu` function moves the data to GPU and also converts to single precision, whereas the design in `ITensorGPU` is that for `ITensor` objects it preserves the precision of the input. Maybe we can consider using the behavior of `CUDA.cu` in the future to make things consistent (I'm imagining a case where we have some code that mixes ITensor objects with Julia Arrays or Flux objects and it may confusing if they have different behavior).

Anyway, I'll merge this when tests pass since I want to work on some things that make use of this PR, most of this is behind the scenes and can be changed in the future if needed.

PR #960, opened 2022-08-04
Document and tweak nullspace function

emstoudenmire:
# Description

Document the `nullspace` function and set a higher default `atol` keyword argument. The default one was often resulting in cases where the computed nullspace was too small. 

# How Has This Been Tested?

Added a new regression test of a specific case where the previous code was failing.

# Checklist:

### DIFF
--- a/src/nullspace.jl
+++ b/src/nullspace.jl
@@ -147,11 +147,37 @@ function nullspace(::Order{2}, M::ITensor, left_inds, right_inds; tags="n", kwar
   return Ñ
 end
 
-function nullspace(T::ITensor, is...; tags="n", kwargs...)
+"""
+    nullspace(T::ITensor, left_inds...; tags="n", atol=1E-12, kwargs...)
+
+Viewing the ITensor `T` as a matrix with the provided `left_inds` viewed
+as the row space and remaining indices viewed as the right indices or column space,
+the `nullspace` function computes the right null space. That is, it will return
+a tensor `N` acting on the right indices of `T` such that `T*N` is zero.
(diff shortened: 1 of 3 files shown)


emstoudenmire:
@mtfishman note that I did comment out part of the previous nullspace test. It was failing with the new changes, which I don't totally understand because it was passing an explicit `atol` which should anyway override the new default that I put. What I don't understand is how this test was passing before, basically.

src/nullspace.jl
```diff
@@ -147,11 +147,37 @@ function nullspace(::Order{2}, M::ITensor, left_inds, right_inds; tags="n", kwar
...
+"""
+    nullspace(T::ITensor, left_inds...; tags="n", atol=1E-12, kwargs...)
+
+Viewing the ITensor `T` as a matrix with the provided `left_inds` viewed
+as the row space and remaining indices viewed as the right indices or column space,
+the `nullspace` function computes the right null space. That is, it will return
+a tensor `N` acting on the right indices of `T` such that `T*N` is zero.
+The returned tensor `N` will also have a new index with the label "n" which
+indexes through the 'vectors' in the null space.
+
```
mtfishman, line 159: Should we add a warning that the implementation might change in the future and therefore the tolerance behavior might change?

emstoudenmire: That's a good idea – will add it

mtfishman: Also maybe warn that for input tensors with very small singular values the size of the returned nullspace could be very sensitive to the tolerance? Maybe that is already inherent to any floating implementation of a nullspace function.

PR #968, opened 2022-08-30
[ITensors] Set the `default_warn_order` to 32

GiggleLiu:
Set the default `default_warn_order` to 32. Otherwise the warning is too many in a rank tensor network contraction algorithm.

### DIFF
--- a/src/global_variables.jl
+++ b/src/global_variables.jl
@@ -3,7 +3,7 @@
 # Warn about the order of the ITensor after contractions
 #
 
-const default_warn_order = 14
+const default_warn_order = 32
 
 const warn_order = Ref{Union{Int,Nothing}}(default_warn_order)
 


emstoudenmire:
Thanks Jinguo, but the default warn order is already very high. For specialized applications like yours, we have provided options to raise or turn off the warn order, like these:
```
ITensors.set_warn_order(N)
@disable_warn_order
```

emstoudenmire:
The error message does actually include the information about `@disable_warn_order` and other ways to change or turn off the warn order, so then the only question is about the backtrace. We do want to keep the backtrace happening at least once so that users can find the code that is triggering the warn order.

I think a good solution here would be for us to count the number of times the backtrace gets printed then only print it a small number of total times. (We don't want it to never print by default.) 

A good workaround for now, though, is just to call `@disable_warn_order` for your code.

GiggleLiu:
Ah, that is true. Maybe we can left as it is.

An alternative way to warn users about the tensor rank is only warn users in the debug mode. i.e. enclosing the warning information with a `@debug` macro. Users only see the warning when they set `ENV["JULIA_DEBUG"] = "ITensors"`.

Which one do you prefer?

mtfishman:
We could only enable the backtrace in debug mode.

I've found it is fairly common that people assume this warning is an error or a serious issue in their code, and also don't read the message at all, which may be due to the backtrace looking long and scary. So it would be nice to improve this in some way since I don't think the current situation has been working as well as I would have hoped. Maybe the message explaining how to suppress the warning should be made clearer?

I prefer putting backtraces in warnings since it helps you actually fix the warning and not just have to guess where it is coming from, but I think it scares people (I've found people are scared by warnings in general but that is another issue).

I would be ok increasing the warning cutoff a bit from the current one, it isn't uncommon to reach tensors of order 14 when doing ED calculations.

PR #971, opened 2022-09-07
site_range keyword for truncate!

emstoudenmire:
# Description

Adds a `site_range` keyword argument for `truncate!`. This is helpful for algorithms where it is known that only part of the MPS was changed, so as to avoid doing extra work. (Recent application of this is a specialized routine for applying MPOs which only act on part of an MPS.)

# How Has This Been Tested?

Added a unit test to test/mps.jl

# Checklist:

### DIFF
--- a/src/mps/abstractmps.jl
+++ b/src/mps/abstractmps.jl
@@ -1612,7 +1612,7 @@ function truncate!(::Algorithm"frobenius", M::AbstractMPS; kwargs...)
   orthogonalize!(M, last(sites))
 
   # Perform truncations in a right-to-left sweep
-  for j in reverse(first(sites)+1:last(sites))
+  for j in reverse((first(sites) + 1):last(sites))
     rinds = uniqueinds(M[j], M[j - 1])
     ltags = tags(commonind(M[j], M[j - 1]))
     U, S, V = svd(M[j], rinds; lefttags=ltags, kwargs...)
--- a/src/mps/abstractmps.jl
+++ b/src/mps/abstractmps.jl
@@ -1605,13 +1605,14 @@ end
 
 function truncate!(::Algorithm"frobenius", M::AbstractMPS; kwargs...)
   N = length(M)
+  sites = get(kwargs, :site_range, 1:N)
 
   # Left-orthogonalize all tensors to make
   # truncations controlled
-  orthogonalize!(M, N)
+  orthogonalize!(M, last(sites))
 
   # Perform truncations in a right-to-left sweep
-  for j in reverse(2:N)
+  for j in reverse(first(sites)+1:last(sites))
     rinds = uniqueinds(M[j], M[j - 1])
     ltags = tags(commonind(M[j], M[j - 1]))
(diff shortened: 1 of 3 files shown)


src/mps/abstractmps.jl
```diff
@@ -1605,13 +1605,14 @@ end
 
 function truncate!(::Algorithm"frobenius", M::AbstractMPS; kwargs...)
```
mtfishman, line 1606: For keyword arguments that aren't getting forwarded, I prefer the style of putting them directly in the function signature, i.e:
```julia
function truncate!(::Algorithm"frobenius", M::AbstractMPS; site_range=eachindex(M), kwargs...)
  [...]
end
```
instead of using `get(kwargs, [...])`.

mtfishman:
Seems like a good idea. An alternative interface could be to allow passing a Vector of `maxdim` of the number of links, i.e.:
```julia
s = siteinds("S=1/2", 6)
psi = randomMPS(s; linkdims=4)
truncate!(psi; maxdim=[2, 3, 3, 3, 2])
```
Just bringing that up for debate but I think the `site_range` interface you propose is simpler.

For MPOs that only act on part of the MPS, I was picturing we could support something like this:
```julia
N = 10
site_range = 2:4
s = siteinds("S=1/2", N)
H = MPO(s[site_range])
psi = randomMPS(s)
Hpsi_sub = apply(H, psi[site_range])
Hpsi = copy(psi)
Hpsi[site_range] = Hpsi_sub
```
Not all of these operations are supported right now but it shouldn't be hard to support them. Note that the `truncate!` of a sub-MPS could be done in a similar way (i.e. slicing part of the MPS, truncating it, and then sticking it back into the original MPS).

emstoudenmire:
Interesting comments. I was thinking that we could eventually have an interface like how we have `apply` for gates, but for MPOs which are acting like "multi-site gates". These MPOs would just have a subset of the site indices, and would be assumed to act as the identity on all other sites. (At first we could require that the MPO have all of the sites in some contiguous range.)

Something like
```
G = make_gate_MPO(s[3:5])
psi = apply(G,psi; cutoff)
```

This is really helpful in particular for quantum computing simulations, which is the main use case for where I need it. It makes a big difference in the efficiency of the code.

emstoudenmire:
The issue I'd have with the separate maxdims for the number of links is that to prevent any truncation, would one either have to put very large maxdims on the bonds you want the code to skip and or know the current linkdims in advance? (Not that they are hard to find out, but it just requires extra code.) Relatedly, even if the maxdim is high, the cutoff might still incur a truncation which one doesn't want, just because of some context where the truncation is known to be avoidable.

mtfishman:
Yeah that's a clear disadvantage of specifying a vector of `maxdim`, so let's scratch that idea.

Agreed it would be good to support that kind of interface for `apply`, the slicing interface I show above could be an internal implementation detail for that.

src/mps/abstractmps.jl
mtfishman: Looks good, thanks!

mtfishman:
Looks good to me, feel free to merge when ready.

PR #978, opened 2022-09-20
Kmp5/refactor/refactor and format

kmp5VT:
# Description

The effort of this pull is to reformat the code (Many comments were much longer than formatting guide permits) and to refactor  the code into a more comprehensive package. This includes moving experimental ideas into an experimental folder. Should experimental ideas even be in the master branch of ITensor.jl? Constructing new files with meaningful names. Breaking up large files into digestible pieces. Increasing the documentation of codes. Restructuring existing code to be easily read and understood.

# How Has This Been Tested?

Code should pass existing unit tests. No new ideas will be implemented in this format.


src/itensor.jl
```diff
@@ -2129,28 +1702,12 @@ function indpairs(T::ITensor; plev::Pair{Int,Int}=0 => 1, tags::Pair=ts"" => ts"
...
-
-"""
-    ishermitian(T::ITensor; kwargs...)
-
-Test whether an ITensor is a Hermitian operator,
-that is whether taking `dag` of the ITensor and
-transposing its indices returns numerically 
-the same ITensor.
+      tr(T::ITensor; kwargs)
+Trance of an ITensor over a pair of indices
```
mtfishman: I believe that there needs to be a line separating the function signature and the description for it to format correctly by Documentor.

Also "Trance" -> "Trace".

src/ITensors.jl
```diff
@@ -132,11 +132,18 @@ include("indexset.jl")
...
+include("OPS/tensor_algebra.jl")
 include("broadcast.jl")
 include("decomp.jl")
 include("iterativesolvers.jl")
 include("adapt.jl")
 
+#####################################
+# Experimental ITensor Functions
+#
+include("experimental/itensorcombiner.jl")
```
mtfishman, line 144: The `combiner` constructor isn't an experimental feature so this isn't a good location for this file.

src/Ops/permutations.jl
```diff
@@ -0,0 +1,152 @@
...
+    dag(T::ITensor; allow_alias = true)
+
+Complex conjugate the elements of the ITensor `T` and dagger the indices.
+
+By default, an alias of the ITensor is returned (i.e. the output ITensor
+may share data with the input ITensor). If `allow_alias = false`,
+an alias is never returned.
+"""
+
+function dag(as::AliasStyle, T::Tensor{ElT,N}) where {ElT,N}
```
mtfishman, line 92: `dag` isn't related to permutations so this doesn't seem like a good location for this function.

src/ITensors.jl
```diff
@@ -132,11 +132,18 @@ include("indexset.jl")
 # ITensor
 #
 include("itensor.jl")
+include("OPS/tensor_algebra.jl")
```
mtfishman, line 135: I don't think `Ops` is the right location for this, the theme of the `Ops` directory is for files defining the `Op` type (a representation of an operator on sites of a lattice) and lazy operations between `Op`s.

src/ITensors.jl
```diff
@@ -149,6 +156,7 @@ include("nullspace.jl")
 #####################################
 # Ops to ITensor conversions
 #
+include("Ops/permutations.jl")
```
mtfishman, line 159: Same as above, I don't think `Ops` is the right location for this file.

src/developer_tools.jl
```diff
@@ -28,3 +28,139 @@ function pause()
...
+!!! warning
+    Note that in the future we may return specialized
+    AbstractArray types for certain storage types,
+    for example a `LinearAlgebra.Diagonal` type for
+    an ITensor with `Diag` storage. The specific storage
+    type shouldn't be relied upon.
+
+See also [`array`](@ref), [`matrix`](@ref).
+"""
+vector(T::ITensor, inds...) = vector(permute(T, inds...; allow_alias=true))
```
mtfishman, line 166: I think these could go in the new `permutations.jl` file.

src/itensor.jl
kmp5VT: Ahh got it, thanks!

src/Ops/permutations.jl
kmp5VT: Dagger should compute a complex conjugate which transposes a matrix, correct?

src/ITensors.jl
kmp5VT: Okay, I though you had said this is a function which you were using to test a theory? I left the NDTensor based combine functions untouched in 'itensor.jl'

src/ITensors.jl
kmp5VT: I made a new folder called tensor_operations and moved this file there

src/ITensors.jl
kmp5VT: Same as above

src/ITensors.jl
kmp5VT: This file was moved to tensor_operations

src/Ops/permutations.jl
mtfishman: In ITensor, for non-QN tensors `dag` just performs a complex conjugate, and for QN tensors it performs a complex conjugate and reverses the Index arrow directions. Transposes of ITensors are performed through Index manipulations (replacing indices with each other, swapping prime levels, etc.).

mtfishman:
@kmp5VT looks like the code needs to be formatted, you can do that by running:
```julia
using JuliaFormatter; format(".")
```
in the directory:
```julia
~/.julia/dev/ITensors
```

I've been seeing the Windows test failing recently, I'm not sure why, but it looks like the other tests are passing. Also the GPU test failed, my guess is that it's unrelated to this PR.

src/ITensors.jl
mtfishman: The experimental feature I was referring to is the `combine_contract` function.

Issue #1006, opened 2022-11-09
[ITensors] [BUG] Commit 56e8a1ed7cf31c08c0d9e26d40ab5d34197af56d breaks autompo unit tests


JanReimers:
**Description of bug**
ITensors.jl version 0.3.13

This commit from [WIP] [ITensors] [ENHANCEMENT] Remove `OpSum` in favor of `Ops.OpSum` (#920)
results in a failure in the test/autompo.jl unit tests.

**Minimal code demonstrating the bug or unexpected behavior**
julia test/autompo.jl

Stack dump:
Matrix operator representation - hashing bug: Error During Test at /home/janr/.julia/dev/ITensors/test/autompo.jl:1091
  Got exception outside of a @test
  ArgumentError: No file exists at given path: opsum_hash_bug.jld2
  Stacktrace:
    [1] checkpath_load
      @ ~/.julia/packages/FileIO/aP78L/src/loadsave.jl:167 [inlined]
    [2] load(::String, ::String, ::Vararg{String}; options::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
      @ FileIO ~/.julia/packages/FileIO/aP78L/src/loadsave.jl:110
    [3] load(::String, ::String, ::Vararg{String})
      @ FileIO ~/.julia/packages/FileIO/aP78L/src/loadsave.jl:110
    [4] macro expansion
      @ ~/.julia/dev/ITensors/test/autompo.jl:1092 [inlined]

@testset "Matrix operator representation - hashing bug" begin
    comps, n, dims = load("opsum_hash_bug.jld2", "comps", "n", "dims")  #line 1092
    s = [Index(d) for d in dims]
    for _ in 1:100
      os = components_to_opsum(comps, n)
      # Before defining `hash(::Op, h::UInt)`, this
      # would randomly throw an error due to
      # some hashing issue in `MPO(::OpSum, ...)`
      MPO(os, s)
    end
  end

code tries to load a pr-existing file that does not exist.  The file opsum_hash_bug.jld2 either needs to be created inside the test framework (preferable IMHO), or included in the source tree (not preferable IMHO).

mtfishman:
That file is stored in the repository: https://github.com/ITensor/ITensors.jl/blob/main/test/opsum_hash_bug.jld2

It may be that when you run the test in that way it doesn't find the file because the `load` call here:
[...]

PR #1008, opened 2022-11-10
[ITensors] [BUG] #1006  

JanReimers:
Fix autompo unit tests when pwd!=ITensors/test.  Use a calculated absolute path, as suggest by Matt.


mtfishman:
Thanks!

PR #1009, opened 2022-11-16
[NDTensors] Implement block sparse QR decomposition

JanReimers:
Also use combiners to render ITensors down to rank-2 tensors, which greatly simplifies the QR code at the ITensors level.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ITensors"
 uuid = "9136182c-28ba-11e9-034c-db9fb085ebd5"
 authors = ["Matthew Fishman <mfishman@flatironinstitute.org>", "Miles Stoudenmire <mstoudenmire@flatironinstitute.org>"]
-version = "0.3.30"
+version = "0.3.31"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"


JanReimers:
Need to test locally with Julia 1.6

mtfishman:
Hi Jan, sorry for being so slow about reviewing this. It (finally) looks like we will be using https://github.com/JanReimers/ITensorMPOCompression.jl so it would be good to get this and #1012 merged. I'll leave a few comments but over all I think they both look good, thanks for the contributions!

NDTensors/src/blocksparse/linearalgebra.jl
```diff
@@ -296,6 +296,102 @@ function LinearAlgebra.eigen(
   return D, V, Spectrum(d, truncerr)
 end
 
+# QR a block sparse Rank 2 tensor.
+#  This code thanks to Niklas Tausendpfund https://github.com/ntausend/variance_iTensor/blob/main/Hubig_variance_test.ipynb
+#
+function LinearAlgebra.qr(T::BlockSparseTensor{ElT,2}; kwargs...) where {ElT}
```
mtfishman, line 302: Let's import `qr` and change `LinearAlgebra.qr` to just `qr`, since that's the style we've been trying to use in general. I've been adding comments like:
```julia
# Overload of `LinearAlgebra.qr`
```
above the function signatures to clarify where the function originates.

NDTensors/src/blocksparse/linearalgebra.jl
```diff
@@ -296,6 +296,102 @@ function LinearAlgebra.eigen(
...
+  end
+
+  # can qindr have more blocks than T?
+  if nblocks(qindr) > nnzblocksT
+    resize!(qindr, nnzblocksT)
+  end
+
+  for n in 1:nnzblocksT
+    q_dim_red = minimum(dims(Rs[n]))
+    NDTensors.setblockdim!(qindr, q_dim_red, n)
```
mtfishman, line 345: We shouldn't need `NDTensors` here since we are in that module already.

NDTensors/src/blocksparse/linearalgebra.jl
```diff
@@ -296,6 +296,102 @@ function LinearAlgebra.eigen(
...
+    blockQ = nzblocksQ[n]
+    blockR = nzblocksR[n]
+
+    if VERSION < v"1.5"
+      # In v1.3 and v1.4 of Julia, Ub has
+      # a very complicated view wrapper that
+      # can't be handled efficiently
+      Qb = copy(Qb)
+      Rb = copy(Vb)
+    end
```
mtfishman, line 386: We've dropped support for Julia versions below v1.6 so this shouldn't be necessary any more.

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -360,44 +360,92 @@ function noinds_error_message(decomp::String)
...
+    R *= dag(vαr)
+  end
+  return Q, R
+end
+
 qr(A::ITensor; kwargs...) = error(noinds_error_message("qr"))
 
-# TODO: write this in terms of combiners and then
-# call qr on the order-2 tensors directly
 function qr(A::ITensor, Linds...; kwargs...)
```
mtfishman, line 393: Instead of `kwargs...` and then `get`, let's put the keyword arguments directly in the signature with the defaults, i.e.:
```julia
function qr(A::ITensor, Linds...; tags=ts"Link,qr")
  qtag = TagSet(tags)
  # [...]
end
```

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -360,44 +360,92 @@ function noinds_error_message(decomp::String)
...
-  if isempty(Ris_original)
-    α = trivial_index(Lis)
-    vRα = onehot(eltype(A), α => 1)
-    A *= vRα
-    Ris = [α]
+  lre = isempty(Lis) || isempty(Ris)
+  # make a dummy index with dim=1 and incorporate into A so the Lis & Ris can never
+  # be empty.  A essentially becomes 1D after collection.
+  if (lre)
+    A, vαl, vαr, Lis, Ris = add_trivial_index(A, Lis, Ris)
```
mtfishman, line 401: Can we add the check for `isempty(Lis) || isempty(Ris)` inside `add_trivial_index` to simplify this a bit?

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -360,44 +360,92 @@ function noinds_error_message(decomp::String)
...
+  #
+  #  Use combiners to render A down to a rank 2 tensor ready matrix QR routine.
+  #
+  CL, CR = combiner(Lis...), combiner(Ris...)
+  cL, cR = combinedind(CL), combinedind(CR)
+  AC = A * CR * CL
+  #
+  #  Make sure we don't accidentally pass the transpose into the matrix qr routine.
+  #
+  if inds(AC) != IndexSet(cL, cR)
```
mtfishman, line 413: Let's use `(cL, cR)` instead of `IndexSet(cL, cR)`, since `IndexSet` is deprecated.

mtfishman: Also maybe the check can be replaced if we use:
```julia
AC = permute(AC, cL, cR; allow_alias=true)
```

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -360,44 +360,92 @@ function noinds_error_message(decomp::String)
...
+  # if hasqns(AC)
+  # for b in nzblocks(QT)
+  #     i1 = inds(QT)[1]
+  #     i2 = inds(QT)[2]
+  #     r1 = inds(RT)[1]
+  #     newqn = -dir(i2) * flux(i1 => Block(b[1]))
+  #     ITensors.setblockqn!(i2, newqn, b[2])
+  #     ITensors.setblockqn!(r1, newqn, b[2])
+  # end
+  # end
```
mtfishman, line 431: For now we can just delete this and fix any bugs properly in NDTensors if they come up.

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -360,44 +360,92 @@ function noinds_error_message(decomp::String)
...
+
+  #
+  #  Undo the combine oepration, to recover all tensor indices.
+  #
+  Q, R = itensor(QT) * dag(CL), itensor(RT) * dag(CR)
+
+  # Conditionally remove dummy indices.
+  if (lre)
+    Q, R = remove_trivial_index(Q, R, vαl, vαr)
```
mtfishman, line 440: If we move the logic `isempty(Lis) || isempty(Ris)` inside `add_trivial_index`, this could just be an overload of `remove_trivial_index`:
```julia
remove_trivial_index(Q::ITensor, R::ITensor, vαl::Nothing, vαr::Nothing) = (Q, R)
```

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -360,44 +360,92 @@ function noinds_error_message(decomp::String)
...
   end
+  #
+  # fix up the tag name for the index between Q and R.
+  #  
+  q = commonind(Q, R)
+  settags!(Q, qtag, q)
+  settags!(R, qtag, q)
```
mtfishman, line 447: Let's use the out-of-place versions `Q = settags(Q, qtag, q)`, etc.

emstoudenmire:
Looks like some nice code, especially the lower-level function (the one that works for matrix-like block sparse tensors). 

Is the part about adding and removing trivial indices to handle the case where the tensor doesn't have row or column indices? Or what is that part for? Thanks.

mtfishman:
> 

Yeah, it is based on a bit of a hack I had added to `svd` to handle the case where the row or column indices are empty, I found that really useful for implementing certain algorithms generically (such as code where an MPS may be missing link indices). It could of course be implemented more efficiently but personally I think it is ok to leave in since the functionality is reasonable and we could always optimize it later.

emstoudenmire:
Thanks for explaining - sounds reasonable to leave it in.

src/tensor_operations/matrix_decomposition.jl
JanReimers: I like this because it makes it clear what is supported and the defaults right in the function signature. However it is breaking the _factorize()_ function because keys like _:ortho_ and _:which_decomp_ are inadvertently left in the _kwargs_ dict. when _qr()_ is called.  I can modify _factorize()_ to explicitly remove those keys before calling _qr()_ but it looks pretty clunky.  There is also a test in test_svd.jl that inadvertently passes down cutoff=nothing causing another failure.  
Is there way to indicate that un-used named args are allowed?
`

src/tensor_operations/matrix_decomposition.jl
JanReimers: Actually I have un-commented it, put a mirror version in rq() and updated the tests to verify zero flux between Q&R when the input has zero flux.  This ends up being important for orthogonalization of certain MPOs.

NDTensors/src/blocksparse/linearalgebra.jl
JanReimers: Done

src/tensor_operations/matrix_decomposition.jl
mtfishman: Could we please move comments to #1012 if that is where the QR code will be developed? It is confusing tracking two PRs.

PR #1012, opened 2022-11-20
[NDTensors] [ITensors] Implement QR/RQ/QL/LQ decompositions

JanReimers:


NDTensors/src/blocksparse/linearalgebra.jl
```diff
@@ -296,6 +296,196 @@ function LinearAlgebra.eigen(
...
+    blockview(Q, blockQ) .= Qb
+    blockview(R, blockR) .= Rb
+  end
+
+  return R, Q
+end
+# QR a block sparse Rank 2 tensor.
+#  This code thanks to Niklas Tausendpfund https://github.com/ntausend/variance_iTensor/blob/main/Hubig_variance_test.ipynb
+#
+function LinearAlgebra.qr(T::BlockSparseTensor{ElT,2}; kwargs...) where {ElT}
```
mtfishman, line 396: Could you add `qr` to `src/imports.jl` and change this from `LinearAlgebra.qr` to `qr`?

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -360,46 +360,174 @@ function noinds_error_message(decomp::String)
...
+  #
+  #  Use combiners to render A down to a rank 2 tensor ready matrix QR routine.
+  #
+  CL, CR = combiner(Lis...), combiner(Ris...)
+  cL, cR = combinedind(CL), combinedind(CR)
+  AC = A * CR * CL
+  #
+  #  Make sure we don't accidentally pass the transpose into the matrix qr routine.
+  #
+  if inds(AC) != IndexSet(cL, cR)
```
mtfishman, line 479: `IndexSet` is deprecated. I think this should be: `inds(AC) != (cL, cR)`.

JanReimers: Good point.  Do we even need the if statement?  Presumably calling permute on properly order indices would be fast anyway.

NDTensors/src/blocksparse/linearalgebra.jl
```diff
@@ -296,6 +296,196 @@ function LinearAlgebra.eigen(
...
+
+  Q = BlockSparseTensor(ElT, undef, nzblocksQ, indsQ)
+  R = BlockSparseTensor(ElT, undef, nzblocksR, indsR)
+
+  for n in 1:nnzblocksT
+    Qb, Rb = Qs[n], Rs[n]
+    blockQ = nzblocksQ[n]
+    blockR = nzblocksR[n]
+
+    if VERSION < v"1.5"
```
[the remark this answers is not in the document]
JanReimers: OK fixed, but no pushed yet.

NDTensors/src/blocksparse/linearalgebra.jl
mtfishman: It's in `src/imports.jl` which is used by the `ITensors` module, but it's not in `NDTensors/src/imports.jl` which would be relevant here. If there are places that overload using `LinearAlgebra.f`, that is an older style which we are moving away from.

src/tensor_operations/matrix_decomposition.jl
mtfishman: Good point. The purpose of that check was to avoid unnecessary copying in `permute`, but you can accomplish the same thing by setting `permute(A, inds; allow_alias=true)` (https://itensor.github.io/ITensors.jl/stable/ITensorType.html#ITensors.permute-Tuple{ITensor,%20Any}) to avoid copying when possible.

By default, even if the permutation of the indices is trivial, `permute` will copy the data to avoid any surprising behavior if code modifies the tensor data later on.

NDTensors/src/blocksparse/linearalgebra.jl
JanReimers: OK that works.  I cleaned up the dense version as well.

src/tensor_operations/matrix_decomposition.jl
JanReimers: OK fixed using allow_alias=true

JanReimers: Thanks for fixing the CI tests.

NDTensors/src/blocksparse/linearalgebra.jl
```diff
@@ -296,6 +296,180 @@ function LinearAlgebra.eigen(
...
+  end
+
+  # can qindl have more blocks than T?
+  if nblocks(qindl) > nnzblocksT
+    resize!(qindl, nnzblocksT)
+  end
+
+  for n in 1:nnzblocksT
+    q_dim_red = minimum(dims(Rs[n]))
+    NDTensors.setblockdim!(qindl, q_dim_red, n)
```
mtfishman, line 345: We can drop `NDTensors`.

NDTensors/src/blocksparse/linearalgebra.jl
```diff
@@ -296,6 +296,180 @@ function LinearAlgebra.eigen(
...
+    Qb, Rb = Qs[n], Rs[n]
+    blockQ = nzblocksQ[n]
+    blockR = nzblocksR[n]
+
+    blockview(Q, blockQ) .= Qb
+    blockview(R, blockR) .= Rb
+  end
+
+  return Q, R
+end
```
mtfishman, line 471: I guess this just repeats the code in #1009 so this can be dropped once that is merged?

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -360,46 +360,170 @@ function noinds_error_message(decomp::String)
...
+  Q, L, q = rq(A, uniqueinds(A, Linds...); kwargs...)
+  #
+  # fix up the tag name for the index between Q and R.
+  #  
+  qtag::TagSet = get(kwargs, :tags, "Link,ql") #tag for new index between Q and R
+  settags!(Q, qtag, q)
+  settags!(L, qtag, q)
+  q = settags(q, qtag)
+
+  return L, Q, q
```
mtfishman, line 526: These would have the same kinds of comments/style suggestions as the `qr` function in #1009.

NDTensors/src/linearalgebra.jl
```diff
@@ -402,11 +402,98 @@ function LinearAlgebra.qr(T::DenseTensor{ElT,2,IndsT}; kwargs...) where {ElT,Ind
...
+    RM, QM = rq(matrix(T))
+  end
+
+  # Make the new indices to go onto Q and R
+  r, q = inds(T)
+  q = dim(q) < dim(r) ? sim(q) : sim(r)
+  Qinds = IndsT((q, ind(T, 2)))
+  Linds = IndsT((ind(T, 1), q))
+  Q = NDTensors.tensor(NDTensors.Dense(vec(Matrix(QM))), Qinds) #Q was strided
+  R = NDTensors.tensor(NDTensors.Dense(vec(RM)), Linds)
```
mtfishman, line 426: Drop `NDTensors`.

NDTensors/src/linearalgebra.jl
```diff
@@ -402,11 +402,98 @@ function LinearAlgebra.qr(T::DenseTensor{ElT,2,IndsT}; kwargs...) where {ElT,Ind
...
 
+#
+#  Uses kwargs:positive to decide which rq method to call.
+#
+function rq(T::DenseTensor{ElT,2,IndsT}; kwargs...) where {ElT,IndsT}
+  if get(kwargs, :positive, false)
+    RM, QM = rq_positive(matrix(T))
+  else
+    RM, QM = rq(matrix(T))
+  end
```
mtfishman, line 418: Move `positive` into the function signature.

JanReimers:
Code review comments from Matt's review of PR#1009 that are addressed on this branch/PR:
[Remove NDTensor. and LinearAlgebra. qualifiers](https://github.com/ITensor/ITensors.jl/pull/1009#discussion_r1122554848)
[Remove special code the version <1.5](https://github.com/ITensor/ITensors.jl/pull/1009#discussion_r1122555974)
[clean up code for empty Lis/Ris](https://github.com/ITensor/ITensors.jl/pull/1009#discussion_r1122557376)
[Remove usage of IndexSet](https://github.com/ITensor/ITensors.jl/pull/1009#discussion_r1122557815)
[Use allow_alias when calling permute (less if statements)](https://github.com/ITensor/ITensors.jl/pull/1009#discussion_r1122557815)
[Use out of place versions of settags](https://github.com/ITensor/ITensors.jl/pull/1009#discussion_r1122559984)

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -369,46 +369,168 @@ function noinds_error_message(decomp::String)
...
+  # 
+  if hasqns(AC)
+    for b in nzblocks(QT)
+      i1 = inds(QT)[1]
+      i2 = inds(QT)[2]
+      r1 = inds(RT)[1]
+      newqn = -dir(i2) * flux(i1 => Block(b[1]))
+      ITensors.setblockqn!(i2, newqn, b[2])
+      ITensors.setblockqn!(r1, newqn, b[2])
+    end
```
mtfishman, line 435: In principle this code shouldn't be necessary if the indices are built correctly in NDTensors, but that should not require using ITensors functions like `setblockqn!`. For example, Miles fixed a similar issue with the SVD code a while ago, from within NDTensors: https://github.com/ITensor/ITensors.jl/commit/c12e4f21187524eb18949469a6a4a496c350ca73. Can we do something similar here?

JanReimers:
Moving issue from PR#1009:
Instead of kwargs... and then get, let's put the keyword arguments directly in the signature with the defaults, i.e.:

```
function qr(A::ITensor, Linds...; tags=ts"Link,qr", positive=false)
  qtag = TagSet(tags)
  # pass `positive` to Tensor version of `qr`
end
```

src/tensor_operations/matrix_decomposition.jl
JanReimers: All done except the change of function signature. see https://github.com/ITensor/ITensors.jl/pull/1012#issuecomment-1452522151

mtfishman:
What about:
```julia
function qr(A::ITensor, Linds...; tags=ts"Link,qr", kwargs...)
  qtag = TagSet(tags)
  # pass `kwargs...` to Tensor version of `qr`
end
```

NDTensors/src/linearalgebra.jl
mtfishman: What about:
```julia
function rq(T::DenseTensor{ElT,2,IndsT}; positive=false, kwargs...) where {ElT,IndsT}
  # [...]
end
```

JanReimers:
Yes that worked

NDTensors/src/linearalgebra.jl
JanReimers: Done

NDTensors/src/linearalgebra.jl
JanReimers: Done

src/tensor_operations/matrix_decomposition.jl
mtfishman: Looks like you found a nice fix for this, that's great, thanks for investigating!

JanReimers: Most welcome.  Much more elegant to build the new index at the NDTensors level.  There are two remaining problems for the case where R is order 2:
1) The indices of R should be ordered so that dense(R) displays an upper triangular matrix. (easy to fix)
2) RT returned from NDTensors is triangular, but R=RT*dag(CR) is often not triangular any more.  This is a consequence of the QN space combining and sorting done inside the combiner.  I don't think this can be fixed by building the index between Q&R in a different order because the external index of R (right index of T) could have split blocks, but the data passed to NDTensors has merged blocks.  Anyway that is my current understanding, but as you know my understandings in this area are often wrong!

I plan to add some code to make a simple gauge transform (after the R=RT*dag(CR) steps) to apply to Q and R in order to restore triangularity [word?].  I suppose we could make this optional, or provide a flag to turn it off for users who don't care.

Let me know if you think the gauge transform is the wrong approach,

mtfishman: Before you get started on that, let's discuss it. I don't think either of those are necessary, since the index ordering of ITensors are internal details, and so is the ordering of the QN spaces after combining. Also, I'm not quite sure what you mean by your point 1., since if `R` is upper triangular, `dense(R)` should be as well since it should have the same exact elements (just with the block structure removed).

But in general, for ITensor `R` we should only guarantee that it is upper triangular _if_ it is permuted such that the index shared with `Q` is the first one (for a QR decomposition, it should be the last one for an RQ decomposition). But we don't guarantee which order of indices it is output with, so that we could choose the index ordering to avoid unnecessary permutations.

mtfishman: Perhaps I'm just misunderstanding your point though, could you share a minimal example of what you think is incorrect about the current implementation?

mtfishman:
Thanks @JanReimers, this will be nice to have.

Sorry to be a pain, but I noticed that a lot of the code looks very similar between `qr` and `rq` in NDTensors. Do you think we can try to compress it into a single function? It could be like:
```julia
function _generic_decomp(block_decomp::Function, T::BlockSparseTensor{ElT,2}; kwargs...) where {ElT}
  # Generic code between `qr` and `rq`, `block_decomp = qr` or `block_decomp = rq`.
end

function qr(T::BlockSparseTensor{ElT,2}; kwargs...) where {ElT}
  return _generic_decomp(qr, T; kwargs...)
end

function rq(T::BlockSparseTensor{ElT,2}; kwargs...) where {ElT}
  return _generic_decomp(rq, T; kwargs...)
end
```
Of course that may not work exactly, and you may have to permute the outputs of `rq` to make it generic. But the general design should be that the functions are the same up to calling some inner function that gets passed that decomposed the blocks.

Also, I noticed this package implement a `ql` decomposition: https://github.com/JuliaLinearAlgebra/MatrixFactorizations.jl. It seems better to implement `ql(::BlockSparseTensor)` making use of `ql(::DenseTensor)` which calls `ql(::Matrix)` defined in that package. Using `ql`, it should then be easier to implement a generic function `_generic_decomp` that you pass either `ql` or `qr`. Then we don't have to support our own `rq` LAPACK wrapper.

For context, the general goal of these changes is to minimize the amount of code we have to support while maximizing the functionality and flexibility. That can be done with good generic programming, code reuse, and calling out to external libraries where appropriate. That makes it easier to maintain the code, fix bugs, optimize the code, etc. since if you have a single generic function (like `_generic_decomp`), you can fix a bug in one place, add optimizations like multithreading over blocks in one place, etc.

src/tensor_operations/matrix_decomposition.jl
JanReimers: I agree point #1 is minor, it means a new user will be surprised to sometimes see what *appears* be lower triangular when they do @show dense(R).  An experienced user will understand that index ordering is arbitrary and not be concerned.

I will make a small example showing problem #2.

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -369,48 +369,142 @@ function noinds_error_message(decomp::String)
...
+  Q, L, q = rq(A, uniqueinds(A, Linds...); kwargs...)
+  #
+  # fix up the tag name for the index between Q and R.
+  #  
+  qtag = TagSet(tags)
+  Q = settags(Q, qtag, q)
+  L = settags(L, qtag, q)
+  q = settags(q, qtag)
+
+  return L, Q, q
```
mtfishman, line 505: Shouldn't this output `Q, L, q`?

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -369,48 +369,142 @@ function noinds_error_message(decomp::String)
...
+end
+
+function ql(A::ITensor, Linds...; tags=ts"Link,ql", kwargs...)
+  Q, L, q = rq(A, uniqueinds(A, Linds...); kwargs...)
```
mtfishman, line 496: Shouldn't this output `R, Q, q`?

mtfishman: (Or `L, Q, q`, I guess the name doesn't matter and it makes sense to name it `L`, but the question stands that isn't the 2nd output of `rq` the orthogonal matrix?).

src/tensor_operations/matrix_decomposition.jl
mtfishman: Right, but we shouldn't fix 1. in the code, if ordering the indices in the "counterintuitive" way is more performant.

JanReimers: OK here is a small example:

```
using ITensors
#QN ordering we might see in an MPO
qns=[QN("Sz",0)=>1,QN("Sz",-1)=>1,QN("Sz",1)=>1,QN("Sz",0)=>1]
i=Index(qns,"i")
A=randomITensor(i,i')
Q,R,q=qr(A,i)
@show dense(R)
#QN ordering carefully chosen to not get re-ordered by combiners
qns=[QN("Sz",-1)=>1,QN("Sz",0)=>1,QN("Sz",0)=>1,QN("Sz",1)=>1]
i=Index(qns,"i")
A=randomITensor(i,i')
Q,R,q=qr(A,i)
@show dense(R)

```
and the output:

> dense(R) = ITensor ord=2
> Dim 1: (dim=4|id=257|"i")'
> Dim 2: (dim=4|id=826|"Link,qr")
> Dense{Float64, Vector{Float64}}
>  4×4
>  0.000   1.677   0.000  0.000
>  0.373   0.000   0.000  0.000
>  0.000   0.000   0.000  0.055
>  0.000  -1.615  -0.215  0.000
> dense(R) = ITensor ord=2
> Dim 1: (dim=4|id=773|"i")'
> Dim 2: (dim=4|id=680|"Link,qr")
> Dense{Float64, Vector{Float64}}
>  4×4
>  1.607   0.000   0.000   0.000
>  0.000  -1.093   0.000   0.000
>  0.000   0.303  -1.127   0.000
>  0.000   0.000   0.000  -0.215
> ITensor ord=2 (dim=4|id=773|"i")' (dim=4|id=680|"Link,qr")
> Dense{Float64, Vector{Float64}}

src/tensor_operations/matrix_decomposition.jl
JanReimers: Yes I am all for re-factoring and minimizing the amount of code that must be maintained.  I will have a look.  I noticed the Q,L swap this morning I will fix that.

src/tensor_operations/matrix_decomposition.jl
mtfishman: Thanks @JanReimers, I understand what you mean now. I think there is a more fundamental question here about what it means for an ITensor to be upper triangular, since as you've identified that is very basis-dependent.

For example, if you take a QR decomposition of an ITensor with multiple indices that end up an `R`, how do you define if it is upper triangular? You have to define a basis ordering for those multiple indices. I would argue that the way to "define" if it is upper triangular would be if it is upper triangular after you combine the indices that are unique to `R`, for example:
```julia
using ITensors

space = [QN("Sz",0)=>1, QN("Sz",-1)=>1, QN("Sz",1)=>1, QN("Sz",0)=>1]
i, j, k = Index.((space, space, space), ("i", "j", "k"))
A = randomITensor(i, j, k)
Q, R = qr(A, i)
q = commonind(Q, R)
C = combiner(j, k)
c = combinedind(C)
Matrix(R * C, q, c)
```
which outputs:
```julia
4×16 Matrix{Float64}:
 0.0  0.118356  0.345319  -0.394825  -0.507585   0.0      …   0.0       0.0       0.0       0.0       0.0       0.0
 0.0  0.0       0.0        0.0        0.0       -1.25421      1.62696   0.0       0.0       0.0       0.0       0.0
 0.0  0.0       0.0        0.0        0.0        0.0         -0.825069  0.0       0.0       0.0       0.0       0.0
 0.0  0.0       0.0        0.0        0.0        0.0          0.0       0.700738  0.219431  0.454174  0.329561  0.0
```
i.e. it is only well defined once you matricize it in the correct way, using the index ordering and index combining that the (matrix) QR decomposition used.

For the case where the ITensor is order 2 and the QR is treated as a matrix between the two indices, we _could_ special case that to ouput an upper triangular `R`, but I still think that we shouldn't force that and keep the same convention, that it is only upper triangular after using a combiner on `R` (which should give a deterministic way to order the basis in the same way that the matrix QR used internally).

mtfishman: This leads to another point, which is that in general really people should specify both the left and right indices of the QR to make sure they know the correct index/basis ordering and therefore combine the indices correctly to get an upper triangular `R`. However, I find that this doesn't work:
```julia
using ITensors

space = [QN("Sz",0)=>1, QN("Sz",-1)=>1, QN("Sz",1)=>1, QN("Sz",0)=>1]
i, j, k = Index.((space, space, space), ("i", "j", "k"))
A = randomITensor(i, j, k)
Q, R = qr(A, (i,), (j, k))
q = commonind(Q, R)
C = combiner(j, k)
c = combinedind(C)
Matrix(R * C, q, c)
```
since it puts all of the input indices `(i, j, k)` onto `Q`. I think you need to add a version:
```julia
qr(A::ITensor, Linds, Rinds; kwargs...)
```
like we have for `eigen` where you can specify both the left and right indices. That should be the "core" function, and then we can have some convenience functions:
```julia
qr(A::ITensor, Linds::Index...; kwargs...)
qr(A::ITensor, Linds; kwargs...)
```
which call the version `qr(A::ITensor, Linds, Rinds; kwargs...)` by determining `Rinds` from the set difference of the indices of `A` and `Linds`. That could be for the case where a user doesn't care about the basis ordering, for example if they just want the orthogonal tensor `Q`.

mtfishman: > I don't think there will be much performance effects for fixing item #1, it is just one line in the outer routine:
> 
> ```
>  #
>  # Make sure that q is the first index of R so that dense(R) appears as upper triangular.
>  #
>  R = permute(R, q, noncommoninds(R, q)...; allow_alias=true) #make sure the q index is first.
> ```
> 
> But I think most important is to resolve item #2, where experienced users will be scratching their heads trying figure out what to do with a non-triangular R as the example above demonstrates.

As I said, I don't think this is something that needs to be "fixed" from our end, since we are allowed to output ITensors in any order we want, since mathematically the index ordering doesn't matter. So we should do extra work just to make printing nicer. Users should permute the tensors themselves if that is the ordering they want.

JanReimers: Yes my code inside qr() checks 
`if hasqns(q) && order(R)==2
`
wrapping the gauge transform which is two red flags right there!  I will have to document the basis dependence well, so people understand.

I am pretty sure that for MPO compression I need to gauge fix to get R back to triangular, in order to maintain Parker's regular form.  i.e. those identity operators in the corners, really need to stay in the corners.  There may be some generalization where in they can move around, but that might be much more difficult to understand and program.  Is this a one off situation?  I don't know.  Should we provide a separate interface that users can call after qr, in order to restore triangularity if users think they need it?  Or more friendly a kwargs flag that does it inside qr, and errors 
`if flag==true && order(R)!=2?
`

In any case we can definitely support 
`qr(A::ITensor, Linds, Rinds; kwargs...)`
as the core function.  That will be easy.

JanReimers: OK no problem.  So I think the remaining issues are:

1. Use qr(A::ITensor, Linds, Rinds; kwargs...) interface as the core function.
2. See if can we can reduce lines of code in NDTensors by implementing {qr,ql} instead of {qr,rq}.  If it looks promising it might take a while. 

Does that sound correct?

mtfishman: Yeah, that's a good summary, thanks!

test/base/test_decomp.jl
```diff
@@ -49,6 +156,219 @@ using ITensors, LinearAlgebra, Test
...
+      @test W ≈ Q * L atol = 1e-13
+      @test norm(dense(Q * dag(prime(Q, q))) - δ(Float64, q, q')) ≈ 0.0 atol = 1e-13
+
+      L, Q, q = ITensors.lq(W, ilr)
+      @test flux(Q) == QN("Sz", 0)
+      @test flux(L) == QN("Sz", 0)
+      @test W ≈ Q * L atol = 1e-13
+      @test norm(dense(Q * dag(prime(Q, q))) - δ(Float64, q, q')) ≈ 0.0 atol = 1e-13
+    end
+  end
```
mtfishman, line 370: Could you move tests that make use of MPS/MPO types into `test/ITensorLegacyMPS`? Relatedly, do you feel like the other tests properly cover the basic functionality, and these are just "extra"?

There are two reasons:
1. Divide the tests into a more "minimal" test set just depending on ITensors which are run by default for all PRs, while other ones are triggered manually or on the `main` branch, to minimize CI testing time on PRs.
2. In preparation for moving the MPS/MPO code into a separate package and deprecating in favor of the new tree tensor network type in https://github.com/mtfishman/ITensorNetworks.jl.

JanReimers: No problem moving them.  Yes they can probably be considered "extra" .  I wanted to verify that 1) complicated block structures do not trigger a problem (some combiner effect?), 2) Fermions in the Hubbard model do not trigger something.

Can you think of a reason to have we should have a fermion operator in the core QR tests? If not then they are most likely "extra"

JanReimers: In test/ITensorLegacyMPS/base/test_qnmpo.jl or a separate file like test_qnmpo_decomp.jl ?

mtfishman: `test_qnmpo.jl` is fine.

JanReimers: Done

src/tensor_operations/matrix_decomposition.jl
JanReimers: That was a useful refactor.  A lot of code reduction.  The qr_positive and ql_positive routines can probably be refactored into one, with a slight reduction in performance (*-1 for wholes rows instead of partial rows).

JanReimers: Found a bug in ql_positive not captured by the current unit tests.  Fix and new unit tests in process.

src/tensor_operations/matrix_decomposition.jl
JanReimers: Fixed

src/tensor_operations/matrix_decomposition.jl
JanReimers: If you agree refactoring ql_positive and qr_positive into one generic routine is not useful, then this is done.

JanReimers:
Status upate:

1. Full support for dense and Blocksparse QR/RQ/QL/LQ under review in PR#1012 
2. Rank reduction support for QR/RQ/QL/LQ Will forward PR once PR#1012 is finished.
3. Update factorize() to allow Blocksparse through (currently it errors out before calling qr()) and allow RQ/LQ/QL 
4. Update all docs accordingly

NDTensors/src/linearalgebra.jl
```diff
@@ -388,23 +367,120 @@ function LinearAlgebra.eigen(
...
+    end
+  end
+  return (Q, L)
+end
+
+#
+#  Lapack replaces A with Q & R carefully packed together.  So here we just copy a
+#  before letting lapack overwirte it. 
+#
+function ql(A::AbstractMatrix{T}; kwargs...) where {T}
```
mtfishman, line 446: What about using https://github.com/JuliaLinearAlgebra/MatrixFactorizations.jl for this?

src/tensor_operations/matrix_decomposition.jl
mtfishman: Are you asking if you should refactor `qr_positive(M::AbstractMatrix)` and `ql_positive(M::AbstractMatrix)` into one generic function? I think it could make sense to make a single generic `qx_positive` function, if they are actually implemented the same way (i.e. analyzing the sign of the diagonal entries and then multiplying those signs into `Q`).

However, I'm a bit confused why they look different, it seems like `ql_positive` has some extra logic for the case `nc > nr` that `qr_positive` doesn't have? That already seems like a good reason for making them one function, assuming that logic should be shared by both functions.

Additionally, it doesn't seem like the code as it is written would handle the complex case properly. In general for complex number there would be a phase along the diagonals and the diagonal entries of `R`/`L` should get transformed into the absolute value of the diagonal entry. Julia's `sign` function generically handles that:
```julia
julia> sign(1 + 2im)
0.4472135954999579 + 0.8944271909999159im

julia> sign(-1)
-1
```

NDTensors/src/linearalgebra.jl
JanReimers: Yes I had a look at that.  in ql.jl line 182 they actually the (almost) identical code:
function _ql(layout, axes, A, args...; kwds...)
    require_one_based_indexing(A)
    AA = similar(A, _qleltype(eltype(A)), size(A))
    copyto!(AA, A)
    return ql!(AA, args...; kwds...)
end

I suspect the reason it is identical is becuase they looked at qr(A::AbstractMatrix) inside LinearAlgebra.  In LinearAlgebra the subsequent call to qr!(A::AbstractMatrix) is just a call to lapack like we have.  Whereas the  https://github.com/JuliaLinearAlgebra/MatrixFactorizations.jl  people seem to be doing something more complicated.  

If you are trying to avoid the lapack call I don't see how unless there some algebraic relation between qr and ql that I am not aware of.

mtfishman: I just would prefer to use that package so we don't have to support our own wrapper (I assumed they would both be using LAPACK).

src/tensor_operations/matrix_decomposition.jl
JanReimers: issue #1:
The difference has to do with where the diagonal is for rectangular R and L matricies.  First off for QR/QL, if they rectangular then nc>nr.  Under these conditions for R the diagonal is right where you expect it.  But for L the diagonal is shifted over to the right by dc=nc-nr.  So L is different. It should be easy to make this generic, but the code might be a little harder to comprehend.
issue #2
When we change the sign of a row, R we can ignore all zeros to the left of the diag, and for L we can ignore the zeros to the right of the diag.  A simple generic version with flip the sign of the whole row which is slightly less efficient.  I doubt we would have an easy time measuring the difference though, the time the lapack would probably swamp the difference.
issues #3
complex handling:  Let me investigate.

mtfishman: I see, thanks for the summary. It's a bit more complicated than I was imagining and the functions aren't so complicated so let's just leave them as they are (but we should make sure we handle the complex case properly).

NDTensors/src/linearalgebra.jl
JanReimers: It is giving the wrong answer for rectungular input with N_rows>N_columns.  I also notice they have no unit test for ql, rq?!?
This has prompted me to beef up the unit tests at the NDTensors level to catch this sort of thing.

mtfishman: I see, that's too bad. Would you mind raising an issue in that repository about that? It would be nice to just use that package and have one community-wide version of `ql` but I guess we will need to support our own for now. Really it should be in Julia's standard LinearAlgebra library but I guess no one bothered to make a high level wrapper for it.

JanReimers: Sure.  I can also save a git-stash of this change so we can resurect later.

src/tensor_operations/matrix_decomposition.jl
JanReimers: It turns the complex case is already handled properly, it seems lapack is doing that for us.  It is returning all reals on the diagonals of R and L.
I have beefed up the unit tests at the NDTensor level in include all combinations:

1. Nrows>Ncols, Ncols>Nrows
2. qr,ql
3. Float64, ComplexF64, Float32, ComplexF32
4. positive in [false,true]
5. If positive=true, check all(real(diagX).>=0.0) && all(imag(diagX).==0.0)

So maybe we should leave qr_positive and ql_positive as is, but I can add some comments about the complex case.
I will wait for confirmation (or lack thereof :) before pushing the unit test upgrade to avoid triggering a full CI run.

NDTensors/src/linearalgebra.jl
JanReimers: [Done](https://github.com/JuliaLinearAlgebra/MatrixFactorizations.jl/issues/43)

mtfishman: Great, thanks! Yes hopefully that issue is easy to fix so we can just use their wrapper (though of course that doesn't need to block this PR).

src/tensor_operations/matrix_decomposition.jl
mtfishman: I would rather not rely on LAPACK returning reals on the diagonals of `R` and `L`. My guess is that isn't guaranteed by the API (for example, in https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-dpcpp/2023-0/gerqf.html#API I don't see anything mentioned about guarantees of constraints on the diagonal values of `R`), and could depend on the particular LAPACK backend, architecture, version, etc.

It should be a simple change to make it more general, i.e. changing:
```julia
    if real(R[c, c]) < 0.0
      R[c, c:end] *= -1 #only fip non-zero portion of the row.
      Q[:, c] *= -1
    end
```
to:
```julia
    sign_Rc = sign(R[c, c])
    if !isone(sign_Rc)
      R[c, c:end] *= conj(sign_Rc) #only fip non-zero portion of the row.
      Q[:, c] *= sign_Rc
    end
```
This should be exactly the same for the real diagonal case but then also cover any complex diagonal cases that may arise.

JanReimers: Excellent point ... I forgot about all those vendor specific (partial) dropins for lapack floating around out there.

JanReimers: Done

NDTensors/src/blocksparse/linearalgebra.jl
```diff
@@ -296,6 +296,71 @@ function LinearAlgebra.eigen(
   return D, V, Spectrum(d, truncerr)
 end
 
+ql(T::BlockSparseTensor{ElT,2}; kwargs...) where {ElT} = qx(ql, T; kwargs...)
```
mtfishman, line 299: Let's use the style:
```julia
ql(T::BlockSparseTensor{<:Any,2}; kwargs...) = qx(ql, T; kwargs...)
```
Basically, a "rule" would be try to avoid ` where {...}` unless it is unavoidable, i.e. you actually use the type and the type isn't easy to access through a function. For example, even if we were using `ElT`, you can easily access that same information with the function `eltype(T)`.

We have been moving towards that style, which makes the code cleaner and makes it easier to write generic code.

mtfishman: Please make the same changes to the rest of the code involved in this PR, such as the dense and block sparse `qr`, `qx`, etc.

mtfishman:
@JanReimers we did a reorganization of the NDTensors library in #1061 so you'll need to update this PR appropriately.

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -363,52 +363,134 @@ function eigen(A::ITensor; kwargs...)
...
+function rq(A::ITensor, Linds::Indices, Rinds::Indices; tags=ts"Link,rq", kwargs...)
+  return xq(ql, tags, A, Linds, Rinds; kwargs...)
+end
+function lq(A::ITensor, Linds::Indices, Rinds::Indices; tags=ts"Link,lq", kwargs...)
+  return xq(qr, tags, A, Linds, Rinds; kwargs...)
+end
+#
+#  Generic function implementing both qr and ql decomposition. The X tensor = R or L. 
+#
+function qx(qx::Function, qtags, A::ITensor, Linds::Indices, Rinds::Indices; kwargs...)
```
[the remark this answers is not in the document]
JanReimers: Done

NDTensors/test/linearalgebra.jl
```diff
@@ -20,4 +20,44 @@ end
   @test norm(U2 * U2' - Diagonal(fill(1.0, m))) < 1E-14
 end
 
+Base.eps(::Type{Complex{T}}) where {T<:AbstractFloat} = eps(T)
```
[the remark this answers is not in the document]
JanReimers: Fixed

NDTensors/src/blocksparse/linearalgebra.jl
JanReimers: OK I think I got them all for Elt.   We still have where {IndsT} in the dense qx() function.  Is it better to use
indstype(::Type{<:Tensor{<:Any,<:Any,<:Any,IndsT}}) where {IndsT} = IndsT
indstype(T::Tensor) = indstype(typeof(T))

for this as well?

JanReimers: I tried it and it seems to be introducing subtle bugs, I am probably not doing it right.

NDTensors/src/linearalgebra/linearalgebra.jl
```diff
@@ -388,23 +367,122 @@ function LinearAlgebra.eigen(
...
-function LinearAlgebra.qr(T::DenseTensor{ElT,2,IndsT}; kwargs...) where {ElT,IndsT}
-  positive = get(kwargs, :positive, false)
-  # TODO: just call qr on T directly (make sure
-  # that is fast)
-  if positive
-    QM, RM = qr_positive(matrix(T))
-  else
-    QM, RM = qr(matrix(T))
-  end
+function qr(T::DenseTensor{<:Any,2,IndsT}; positive=false, kwargs...) where {IndsT}
```
mtfishman, line 370: Can drop `IndsT` and `where {IndsT}`.

NDTensors/src/linearalgebra/linearalgebra.jl
```diff
@@ -388,23 +367,122 @@ function LinearAlgebra.eigen(
...
+  qxf = positive ? qr_positive : qr
+  return qx(qxf, T; kwargs...)
+end
+function ql(T::DenseTensor{<:Any,2,IndsT}; positive=false, kwargs...) where {IndsT}
```
mtfishman, line 374: Can drop `IndsT` and `where {IndsT}`.

NDTensors/src/linearalgebra/linearalgebra.jl
```diff
@@ -388,23 +367,122 @@ function LinearAlgebra.eigen(
...
+  qxf = positive ? ql_positive : ql
+  return qx(qxf, T; kwargs...)
+end
+
+#
+#  Generic function for qr and ql decomposition of dense matrix.
+#  The X tensor = R or L.
+#
+function qx(qx::Function, T::DenseTensor{<:Any,2,IndsT}; kwargs...) where {IndsT}
```
mtfishman, line 383: Can drop `IndsT` and `where {IndsT}` in the function signature and use `IndsT = indstype(T)` inside the function.

NDTensors/src/linearalgebra/linearalgebra.jl
```diff
@@ -388,23 +367,122 @@ function LinearAlgebra.eigen(
...
+    end
+  end
+  return (Q, L)
+end
+
+#
+#  Lapack replaces A with Q & R carefully packed together.  So here we just copy a
+#  before letting lapack overwirte it. 
+#
+function ql(A::AbstractMatrix{T}; kwargs...) where {T}
```
mtfishman, line 448: Can drop `T` and `where {T}` in the function signature and use `eltype(A)` instead of `T` inside the function.

NDTensors/src/blocksparse/linearalgebra.jl
mtfishman: I left some other comments on other cases of ` where ` that I think you can remove. Please post in those comments what you tried and how it didn't work if you see issues.

NDTensors/src/linearalgebra/linearalgebra.jl
JanReimers: Done

NDTensors/src/linearalgebra/linearalgebra.jl
JanReimers: Done

NDTensors/src/linearalgebra/linearalgebra.jl
JanReimers: Done

NDTensors/src/linearalgebra/linearalgebra.jl
JanReimers: Done

NDTensors/src/blocksparse/linearalgebra.jl
JanReimers: Fixed

mtfishman:
Looks great, thanks for sticking with all of the comments @JanReimers. It's been a long time coming to have block sparse QR implemented in ITensor, it's been on the TODO list for many years. Is this ready to merge from your end?

JanReimers:
Yes, I have tested everything I can think of in the MPO compression
module.  There are probably more unit tests than we need in NDTensors and
ITensors, but they are all small/fast, so better too many the too few for
now.   Thanks for your patience with me.

Cheers
Jan

PR #1014, opened 2022-11-22
Address issue 1010

kmp5VT:
# Description

Removes the `decomp.jl` file. All articles of this file have been moved to `src/tensor_operations/matrix_algebra.jl`


mtfishman:
Thanks!

Issue #526, opened 2020-11-18
Product MPOs have bond dimension greater than 1 with AutoMPO


mtfishman:
@emstoudenmire, I think this issue has come up before (but maybe it was in the C++ version). Here is a minimal example:
```julia
using ITensors
s = siteinds("S=1/2", 3)
M = MPO(AutoMPO() + ("Sz", 1), s)
@show inds(M[2])
```
returns:
```julia
inds(M[2]) = IndexSet{4} (dim=3|id=714|"Link,l=1") (dim=3|id=687|"Link,l=2") (dim=2|id=144|"S=1/2,Site,n=2")' (dim=2|id=144|"S=1/2,Site,n=2")
```
It seems like any single term `AutoMPO` I input returns an MPO with bond dimension 3. I assume that is because it is making an MPO of the form that _could_ accommodate other MPO terms, and doesn't specialize to the case of just a single term.

The reason this came up is that I want to implement a generic MPO constructor `MPO(s, "Id")` which in general returns an MPO that is a product of specified operators. This is currently available for non-QN MPOs, but as I was generalizing to QN MPOs I realized it would be a lot easier to let `AutoMPO` generate the MPO for me (since it already deals with all of the subtleties of dealing with blocks, flux, etc.). However, I noticed it doesn't output the MPO with the optimal bond dimension as shown above.

In theory we could truncate the MPO output by AutoMPO, but naively truncating an MPO with our `truncate!` function doesn't work in general, since the norm can diverge (like for the case of `MPO(s, "Id")`, the norm of that operator grows as `2^N`). As a separate issue, we should remember to implement the MPO truncation algorithm from the paper https://arxiv.org/abs/1909.06341.

mtfishman:
I know Miles is aware, but just for general knowledge and as a reminder, this is what happens if you try to naively truncate a Hamiltonian-like MPO:
```julia
julia> s = siteinds("S=1/2", 200);

julia> M = MPO(AutoMPO() + ("Id", 1), s);

julia> truncate!(M; cutoff = 1e-15);

julia> @show M[1];
M[1] = ITensor ord=3
Dim 1: (dim=2|id=375|"S=1/2,Site,n=1")'
Dim 2: (dim=2|id=375|"S=1/2,Site,n=1")
Dim 3: (dim=1|id=662|"Link,l=1")
NDTensors.Dense{Float64,Array{Float64,1}}
 2×2×1
[:, :, 1] =
[...]

PR #1015, opened 2022-11-23
Reduce bond dimension of `MPO`s generated from `OpSum`s consisting of one-site operators

leburgel:
# Description

In the current implementation of the `OpSum` to `MPO` converter every bond of the `MPO` is initialized with a dimension of at least 3, even if there are no interactions passed along that bond. I added a small change here that checks if there is actually something passed along each bond, and sets the bond dimension to 2 (for one-site operators and the identities) if this is not the case.

This is related to #526 in the sense that the bond dimension for product `MPO`s is now reduced from 3 to 2 as a side effect, but the bond dimension = 1 case would still need to be handled as a special case to get a real product `MPO`.

```julia
using ITensors
s = siteinds("S=1/2", 4)
os = OpSum()
for i in 1:4
    os += "X", i
end
H = MPO(os, s)
@show linkdims(H)
```
returns:
```julia
linkdims(H) = [3, 3, 3]
```

</p></details>

```julia
using ITensors
s = siteinds("S=1/2", 4)
os = OpSum()
for i in 1:4
    os += "X", i
end
H = MPO(os, s)
@show linkdims(H)
```
returns:
```julia
linkdims(H) = [2, 2, 2]
```

</p></details>

# How Has This Been Tested?

I added a small test for the specific case of Hamiltonians consisting of only one-site operators at the end of test/autompo.jl.

# Checklist:

### DIFF
--- a/src/physics/autompo/opsum_to_mpo.jl
+++ b/src/physics/autompo/opsum_to_mpo.jl
@@ -14,8 +14,7 @@ function svdMPO(os::OpSum{C}, sites; kwargs...)::MPO where {C}
     return (only(site(t[1])) <= n <= only(site(t[end])))
   end
 
-  rightmap = Dict{Vector{Op},Int}()
-  next_rightmap = Dict{Vector{Op},Int}()
+  rightmaps = fill(Dict{Vector{Op},Int}(), N)
 
   for n in 1:N
     leftbond_coefs = MatElem{ValType}[]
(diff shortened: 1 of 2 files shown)


emstoudenmire:
This is great, and I'm pleased that it actually makes the code simpler, or at least not more complicated.

Agreed that the bond dimension 1 case might need special handling.

Issue #1023, opened 2022-11-29
[ITensorGPU] [BUG] Typo in subtraction of 2 dense tensors on GPU


kbidzhiev:
**Description of bug**

Subtraction of two Tensors on GPU gives a wrong sign.

**Minimal code demonstrating the bug or unexpected behavior**

<details><summary>Minimal runnable code</summary><p>

```Julia
using ITensors, ITensorGPU

    i = Index(4,"i")
    j = Index(2,"j")
    k = Index(2,"k")
    v = randomITensor(i,j, k) + 1.0im * randomITensor(i,j, k)
    w = randomITensor(i,j, k) + 1.0im * randomITensor(i,j, k)

    cuv = cu(v)
    cuw = cu(w)

    @show cu(v - w) ≈ (cuv - cuw) 
    @show cu(v - w) ≈ -(cuv - cuw) 
```
</p></details>

**Expected output or behavior**

cu(v - w) ≈ cuv - cuw = true
cu(v - w) ≈ -((cuv - cuw)) = false

**Actual output or behavior**

cu(v - w) ≈ cuv - cuw = false
cu(v - w) ≈ -((cuv - cuw)) = true

<details><summary>Output of minimal runnable code</summary><p>

```julia
cu(v - w) ≈ cuv - cuw = false
cu(v - w) ≈ -((cuv - cuw)) = true
```
</p></details>

PR #1024, opened 2022-11-29
typo in subtraction of 2 dense tensors on GPU

kbidzhiev:
# Description
issue[#1023](https://github.com/ITensor/ITensors.jl/issues/1023)

```julia
function Base.:-(B::CuDenseTensor, A::CuDenseTensor)
```
the code should return B-A, instead it returns A-B


mtfishman:
Thanks, could you please add a test?

kbidzhiev:
I've added test.

Moreover I have couple of questions.
Why it's necessary to copy results in the file ITensorGPU/src/tensor/cudense.jl

```julia
function Base.:-(B::CuDenseTensor, A::CuDenseTensor)
...
  copyto!(data(store(B)), vec(C))
  return B
```

and the same for 
```julia
function Base.:-(A::CuDense, Ais::IndexSet, B::CuDense, Bis::IndexSet)
...
  copyto!(data(B), vec(C))
  return C
```

for which reason two functions are not similar ?

mtfishman:
Thanks for the PR and tests!

Could you raise a separate issue about that? The second definition should probably be removed.

mtfishman:
Also, could you run the formatter to format your code?

kbidzhiev:
> Also, could you run the formatter to format your code?

Done :)

Issue #1034, opened 2022-12-08
[NDTensors] [BUG] array(combiner * tensor) produces garbage


christopherdavidwhite:
**Description of bug**

If I multiply a tensor by a combiner and then turn the result into a Julia array, I get garbage---it appears to be accessing uninitialized memory

**Minimal code demonstrating the bug or unexpected behavior**

```julia
using ITensors
using PkgVersion
@show VERSION
@show PkgVersion.Version(ITensors)

i = Index(2)
j = Index(3)
a = ITensor([1,2],i)

b = ITensor([1,1,1], j)

a |> array |> display
b |> array |> display
a*b |> array |> display

C = combiner(i,j)

C*a*b |> array |> display

```

**Expected output or behavior**
<details><summary>Expected output</summary>

```julia
VERSION = v"1.8.2"
PkgVersion.Version(ITensors) = v"0.3.21"
a |> array = [1.0, 2.0]
b |> array = [1.0, 1.0, 1.0]
a * b |> array = [1.0 1.0 1.0; 2.0 2.0 2.0]
C * a * b |> array = [1.0, 2.0, 1.0, 2.0, 1.0, 2.0]
```
</details>

**Actual output or behavior**

<details><summary>Actual output</summary>

```julia
VERSION = v"1.8.2"
PkgVersion.Version(ITensors) = v"0.3.21"
a |> array = [1.0, 2.0]
b |> array = [1.0, 1.0, 1.0]
a * b |> array = [1.0 1.0 1.0; 2.0 2.0 2.0]
C * a * b |> array = [1.0, 2.0, 6.93383344655886e-310, 6.93407979483234e-310, 6.9340145831688e-310, 6.93401455534304e-310]
```
</details>

**Version information**

 - Output from `versioninfo()`:
```julia
julia> versioninfo()
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, tigerlake)
  Threads: 1 on 8 virtual cores
Environment:
  LD_LIBRARY_PATH = LD_LIBRARY_PATH:/home/christopher/software/lib


```
 - Output from `using Pkg; Pkg.status("ITensors")`:
```julia
[...]

PR #1038, opened 2023-01-05
[ITensorGaussianMPS] Non-number-conserving states for Gaussian MPS

b-kloss:
# Description
Generalized ITensorGaussianMPS to handle generic gaussian fermionic states including pairing correlations (non-number conserving states).

# Potential issues
- In the non-number-conserving case, the order of the returned populations is hardcoded and follows `[1,0,1,0,...,1,0,0,1]` while in the number-conserving case the population order will have arbitrary permutations between pairs.
- If a correlation matrix in non-number-conserving format, but with vanishing pairing correlations is input to `correlation_matrix_to_gmps`, the implementation falls back to the number-conserving implementation and returns the GMPS circuit to be used in the number-conserving case. This means that the GMPS circuit is not compatible with the input correlation matrix (but with a submatrix of the input matrix). This inconsistency is handled automatically in calls to `correlation_matrix_to_mps`.
- The implementation seems unstable for certain (measure-0 set?) of input correlation matrices, but usually an infinitesimal perturbation of the input matrix is sufficient to obtain the correct result. This behaviour is demonstrated in `ITensorGaussianMPS/examples/spinless_fermion_pairing_and_instability.jl`.
- Only spinless fermions are currently supported.

# How Has This Been Tested?

Tests have been added to `ITensorGaussianMPS/test/gmps.jl` to ensure that correlation matrices of MPS generated from correlation matrices match entries of the correlation matrix, for the following cases

- [x] Real, non-number-conserving correlation matrix
- [x] Real, number-conserving correlation matrix (with vanishing, but explicitly present, pairing correlation). This triggers fallback to the number-conserving code.
- [x] Complex, non-number-conserving correlation matrix.

# Checklist:


ITensorGaussianMPS/src/gmps.jl
```diff
@@ -80,7 +132,11 @@ end
 
 function shift!(G::Circuit, i::Int)
   for (n, g) in enumerate(G.rotations)
-    G.rotations[n] = Givens(g.i1 + i, g.i2 + i, g.c, g.s)
+    if typeof(g) <: Givens
```
mtfishman, line 135: I think it would be better to make use of a generic function `shift(g, i)` here, defined as:
```julia
shift(g::Givens, i::Int) = Givens(g.i1 + i, g.i2 + i, g.c, g.s)
shift(g::DiagRotation, i::Int) = DiagRotation(g.i1 + i, g.i2 + i, g.d)
```

ITensorGaussianMPS/test/gmps.jl
```diff
@@ -123,3 +126,181 @@ end
...
+  cc = correlation_matrix(psi, "C", "C")
+  cblocked = ITensorGaussianMPS.reverse_interleave(c)
+
+  @test all(abs.(cblocked[(N + 1):end, (N + 1):end] - cdagc[:, :]) .< 1e-6)
+  @test all(abs.(cblocked[1:N, 1:N] - ccdag[:, :]) .< 1e-6)
+  @test all(abs.(cblocked[1:N, (N + 1):end] - cc[:, :]) .< 1e-6)
+  @test all(abs.(cblocked[(N + 1):end, 1:N] - cdagcdag[:, :]) .< 1e-6)
+end
+
+@testset "Fermion (BCS real - no pairing)" begin
```
emstoudenmire, line 188: I don't think I understand how this test differs from the previous one. I see the title says "No pairing" but much of the code looks similar.

b-kloss: The only difference is in the value of `Delta`. There's only minor differences between the three tests, so they could be combined into one if that is prefered. I split them so it's easier to identify what the actual issue is.

mtfishman: Probably best to combine them by looping over different values of `Delta`.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -132,15 +79,29 @@ end
...
-      G.rotations[n] = Givens(g.i1 + i, g.i2 + i, g.c, g.s)
-    elseif typeof(g) <: DiagRotation
-      G.rotations[n] = DiagRotation(g.i1 + i, g.i2 + i, g.d)
-    end
+    G.rotations[n] = Givens(g.i1 + i, g.i2 + i, g.c, g.s)
   end
   return G
 end
 
+function scale!(G::Circuit, i::Int)
```
mtfishman, line 87: Maybe `shift!` and `scale!` could be replaced by some map functions:
```julia
function map!(f, G::Circuit)
  for i in eachindex(G.rotations)
    G.rotations[i] = f(G.rotations[i])
  end
  return G
end
```
and:
```julia
function map_indices!(f, G::Circuit)
  return map!(g -> Givens(f(g.i1), f(g.i2), g.c, g.s)
end
```
Then we can use `map_indices!(i -> i + j, G)` and `map_indices!(i -> i * j, G)` instead of `shift!(G, j)` and `scale!(G, j)` (also `conj!` could be written in terms of `map!`).

ITensorGaussianMPS/examples/spinless_fermion_pairing_and_instability.jl
```diff
@@ -68,15 +68,15 @@ let
   sympert = 0.5 * (pert + transpose(pert))
   psi = ITensorGaussianMPS.correlation_matrix_to_mps(
     sites,
-    ITensorGaussianMPS.interleave(Complex.(Gamma));
-    eigval_cutof=1e-10,
+    Pairing(ITensorGaussianMPS.interleave(Complex.(Gamma)));
```
mtfishman, line 71: Can't `Pairing` be deduced automatically by analyzing `length(sites)` and `size(Gamma)`?

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -25,61 +25,8 @@ end
...
-    a1, a2 = A[i, G.i1], A[i, G.i2]
-    A[i, G.i1] = a1
-    A[i, G.i2] = G.d * a2
-  end
-  return A
-end
-
 struct Circuit{T} <: LinearAlgebra.AbstractRotation{T}
-  rotations::Array{Union{Givens{T},DiagRotation{T}},1}
+  rotations::Array{Givens{T},1}
```
mtfishman, line 29: You can just use `Vector{Givens{T}}`, that's the same as `Array{Givens{T},1}`.

ITensorGaussianMPS/src/gmps.jl
b-kloss: Sounds good, will take care of it.

ITensorGaussianMPS/examples/spinless_fermion_pairing_and_instability.jl
b-kloss: Yes, the motivation to require it here was to have a common interface for `correlation_matrix_to_mps` and `correlation_matrix_to_gmps` (the latter requires a wrapped correlation matrix). What about adding that check to the version `correlation_matrix_to_mps` for input of type AbstractMatrix, instead of automatically assuming it's number-conserving.

mtfishman: Yeah, that makes sense to have a version specialized to `AbstractMatrix` which tries to deduce the wrapper type automatically.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -303,6 +264,14 @@ end
 # Correlation matrix diagonalization
 #
 
+struct Pairing
```
mtfishman, line 267: What about the name `BCS`? I've also seen the name "Generalized Hartree-Fock" (https://arxiv.org/abs/cond-mat/9312044).

Also I would parametrize this type, i.e.:
```julia
struct BCS{T}
  data::T
end
```
so someone could dispatch on the type parameter if needed.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -188,6 +294,193 @@ function givens_rotations(v::AbstractVector{ElT}) where {ElT}
...
+  LinearAlgebra.lmul!(g2, gs)
+  LinearAlgebra.lmul!(g1, gs)
+  return gs, r
+end
+
+function check_pairing_correlations(Λ0::AbstractMatrix{ElT}) where {ElT<:Number}
+  paired = false
+  Λblocked = reverse_interleave(Λ0)
+  N = div(size(Λblocked, 1), 2)
+  if all(abs.(Λblocked[1:N, (N + 1):end]) .<= eps(Float64))
```
mtfishman, line 363: It would be faster to do:
```julia
all(x -> abs(x) <= eps(eltype(Λ0)), @view Λblocked[1:N, (N + 1):end])
```
which won't allocated temporary matrices.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -196,45 +489,81 @@ stored in a structure as a set of Givens rotations.
...
+    maxblocksize=maxblocksize,
+  )
+end
+
+function correlation_matrix_to_gmps(
+  Λ0::T;
+  eigval_cutoff::Float64=1e-8,
+  minblocksize::Int=1,
+  maxblocksize::Int=size(Λ0.data, 1),
+) where {T<:Union{Pairing,ConservingNf}}
```
mtfishman, line 526: Maybe make these `AbstractSymmetry` subtypes (or something like that).

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -196,45 +489,81 @@ stored in a structure as a set of Givens rotations.
...
+  Λ = T(Hermitian(copy(Λ0.data)))
```
mtfishman, line 527: Could define `copy`, `Hermitian`, `eltype`, `size`, etc. for `ConservesNfParity`/`ConservesNf`.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -196,45 +489,81 @@ stored in a structure as a set of Givens rotations.
...
+  ElT = eltype(Λ.data)
   V = Circuit{ElT}([])
-  ns = Vector{real(ElT)}(undef, N)
   err_tot = 0.0
-  for i in 1:N
-    blocksize = 0
-    n = 0.0
+  Λ = check_pairing_correlations(Λ)
```
mtfishman, line 531: I would call this `maybe_drop_pairing_correlations` or something a bit more descriptive like that. When I read `check_*` I think it is just returning `true`/`false` or throws an error if a check fails.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -196,45 +489,81 @@ stored in a structure as a set of Givens rotations.
...
+  N = size(Λ.data, 1)
+  calctype = typeof(Λ)
+  factor = calctype <: Pairing ? 2 : 1
```
mtfishman, line 534: Could turn this into a generic function `site_stride(::ConservesNf) = 1`, `site_stride(::ConservesNfParity) = 2`.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -188,6 +294,193 @@ function givens_rotations(v::AbstractVector{ElT}) where {ElT}
...
+function check_pairing_correlations(Λ0::Pairing)
+  is_paired, Λ = check_pairing_correlations(Λ0.data)
+  if !is_paired
+    return ConservingNf(Λ)
+  else
+    return Pairing(Λ)
+  end
+end
+
+function get_subblock(
```
mtfishman, line 384: What about `subblock_eig`?

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -188,6 +294,193 @@ function givens_rotations(v::AbstractVector{ElT}) where {ElT}
...
+  startind::Int;
+  eigval_cutoff::Float64=1e-8,
+  minblocksize::Int=1,
+  maxblocksize::Int=size(_Λ.data, 1),
+)
+  blocksize = 0
+  n = 0.0
+  err = 0.0
+  p = Int[]
+  nB = 0.0
```
mtfishman, line 429: This confused me since it's actually of type `Vector{eltype(_Λ)}`. To initialize I think you can just type:
```julia
nB::Vector{eltype(_Λ)}
```

ITensorGaussianMPS/src/gmps.jl
mtfishman: Also I would return the eigenvalues as the first output since that's the convention that `eigen` uses.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -188,6 +294,193 @@ function givens_rotations(v::AbstractVector{ElT}) where {ElT}
...
+  v = _v.data
+
+  p = sortperm(nB)
+  n1 = nB[first(p)]
+  n2 = nB[last(p)]
+  to_break = false
+  ns[2 * i] = n1
+  ns[2 * i - 1] = n2
+  if length(v) == 2
+    to_break = true
```
mtfishman, line 475: I think it would be better to make this `to_break` logic a separate function outside of `process_populations!`

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -196,45 +489,81 @@ stored in a structure as a set of Givens rotations.
...
-      err = min(n, 1 - n)
-      err ≤ eigval_cutoff && break
+    v, nB, err = get_subblock(
+      Λ,
+      i;
+      eigval_cutoff=eigval_cutoff,
+      minblocksize=minblocksize,
+      maxblocksize=maxblocksize,
+    )
+    ns, to_break = process_populations!(ns, nB, v, i)
```
mtfishman, line 545: I found the logic of these two functions hard to follow.

It seems cleaner to have an interface like this:
```julia
    v, n, err = isolate_subblock_eigs(Λ, i; eigval_cutoff, minblocksize, maxblocksize) # Outputs only the isolated/disentangled eigenvector(s)/eigenvalue(s)
    set_occupations!(ns, i, n)
    stop_gmp_sweep([Whatever is needed to determine when to stop the sweep])
```

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -251,12 +580,51 @@ function ITensors.ITensor(u::Givens, s1::Index, s2::Index)
...
+  s2 = sites[u.i2]
+  return ITensor(u, s1, s2, is_boguliobov)
+end
+
+function to_manybody_gates(s::Vector{<:Index}, C::Pairing)
+  is_bog = g -> abs(g.i2 - g.i1) == 2 ? false : true
+  s1 = g -> div(g.i1 - 1, 2) + 1
+  s2 = g -> div(g.i2 - 1, 2) + 1
+  U = [
+    ITensor(g, s[s1(g)], s[s2(g)], is_bog(g)) for
```
mtfishman, line 614: Could we make use of the `ConservesNfParity` and `ConservesNf` wrappers here? Then we can dispatch on them with:
```julia
ITensor(::ConservesNfParity{<:Givens}, [...]) # Boguliobov rotation
ITensor(::ConservesNf{<:Givens}, [...]) # Number conserving rotation
```

In ITensors.jl the closest function to `to_manybody_gates` is called `ops`, i.e. https://github.com/ITensor/ITensors.jl/blob/v0.3.23/examples/gate_evolution/quantum_simulator.jl. We could also just call it `itensors`, since it constructs a list of ITensors.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -292,16 +660,47 @@ function correlation_matrix_to_mps(
...
+  Λ0::T;
+  eigval_cutoff::Float64=1e-8,
+  maxblocksize::Int=size(Λ0.data, 1),
+  minblocksize::Int=1,
+  kwargs...,
+) where {T<:Union{Pairing,ConservingNf}}
+  if eltype(Λ0.data) <: AbstractFloat
+    MPS_Elt = Float64
+  else
+    MPS_Elt = ComplexF64
```
mtfishman, line 687: Can't you just use `eltype(Λ0)` for the eltype of the MPS instead of hardcoding it to `Float64` or `ComplexF64`?

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -292,16 +660,47 @@ function correlation_matrix_to_mps(
...
+  @assert size(Λ.data, 1) == size(Λ.data, 2)
   ns, C = correlation_matrix_to_gmps(
-    Λ; eigval_cutoff=eigval_cutoff, maxblocksize=maxblocksize
+    Λ; eigval_cutoff=eigval_cutoff, minblocksize=minblocksize, maxblocksize=maxblocksize
   )
   if all(hastags("Fermion"), s)
-    U = [ITensor(s, g) for g in reverse(C.rotations)]
-    ψ = MPS(s, n -> round(Int, ns[n]) + 1, U; kwargs...)
+    U = to_manybody_gates(s, calctype(C))
+    ψ = MPS(MPS_Elt, s, n -> round(Int, ns[calctype <: Pairing ? 2 * n : n]) + 1)
```
mtfishman, line 697: Again, I would use a generic `site_stride` function here instead of an if-statement.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -333,6 +732,12 @@ function slater_determinant_to_mps(s::Vector{<:Index}, Φ::AbstractMatrix; kwarg
   return correlation_matrix_to_mps(s, conj(Φ) * transpose(Φ); kwargs...)
 end
 
+function slater_determinant_to_mps(
+  s::Vector{<:Index}, Φ::T; kwargs...
+) where {T<:Union{Pairing,ConservingNf}}
+  return correlation_matrix_to_mps(s, T(conj(Φ.data) * transpose(Φ.data)); kwargs...)
```
mtfishman, line 738: I think for things like `conj` and `transpose` we should define overloads for `ConservesNf`/`ConservesNfParity`. That would make this code a bit nicer.

ITensorGaussianMPS/src/gmps.jl
mtfishman: Never mind, you would have to use:
```julia
local nB::Vector{eltype(_Λ)}
```
to define a variable without assigning a value (https://docs.julialang.org/en/v1/manual/variables-and-scoping/#Local-Scope). I think assignment is a bit simpler and more canonical so:
```julia
nB = eltype(_Λ)[]
```
seems best.

ITensorGaussianMPS/src/gmps.jl
b-kloss: Ok. What about the return signature here for AbstractMatrices and AbstractSymmetry types? The former returns ::bool,::AbstractMatrix, the latter returns ::AbstractSymmetry --- probably best to go with ::bool,::Any in both cases?

ITensorGaussianMPS/src/gmps.jl
mtfishman: Actually `replace!` and `replace_indices!` would be better names since that matches the Julia Base definitions (https://docs.julialang.org/en/v1/base/collections/#Base.replace!).

ITensorGaussianMPS/src/gmps.jl
mtfishman: What if they both returned `AbstractSymmetry`? It seems like that covers the purpose of the `Bool` anyway.

ITensorGaussianMPS/src/gmps.jl
b-kloss: I agree it makes sense to have these overloads, but in this specific case it doesn't help much unless we want to implement most of the AbstractMatrix Interface (* requires unwrapping)

mtfishman: Ah right, I guess once you get `*` involved it gets more complicated. You can still define something relatively generic like:
```julia
(x::AbstractSymmetry * y::AbstractSymmetry) = set_data(x, x.data * y.data)
```

b-kloss: Could be done. For this concrete example, I don't particularly like that that one doesn't error if x,y are different concrete types of AbstractSymmetry.

mtfishman: Yeah, I was thinking that too. You could restrict them to have exactly the same type like this:
```julia
(x::T * y::T) where {T<:AbstractSymmetry} = set_data(x, x.data * y.data)
```
but that is already too strict since even in this case one is transposed.

We could do:
```julia
function (x::AbstractSymmetry * y::AbstractSymmetry)
  if !has_same_symmetry(x, y)
    error("Can't multiply two symmetric objects with different symmetries.")
  end
  return set_data(x, x.data * y.data)
end

has_same_symmetry(::AbstractSymmetry, ::AbstractSymmetry) = false
has_same_symmetry(::ConservesNf, ::ConservesNf) = true
has_same_symmetry(::ConservesNfParity, ::ConservesNfParity) = true
```
That's not such a burden to require of new symmetry types.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -287,21 +661,57 @@ a matrix product state (MPS).
 The correlation matrix should correspond to a pure state (have all eigenvalues
 of zero or one).
 """
-function correlation_matrix_to_mps(
+function convert_correlation_matrix(Λ::AbstractMatrix, s::Vector{<:Index})
```
mtfishman, line 664: Maybe `symmetric_correlation_matrix`? `convert_*` is a bit general here (if I saw this function out of context I wouldn't know what it was converting to).

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -196,44 +485,90 @@ stored in a structure as a set of Givens rotations.
...
+    LinearAlgebra.lmul!(g, V)
+    Λ = set_data(Λ, Hermitian(g * Matrix(Λ.data) * g'))
   end
-  return ns, V
+  ###return non-wrapped populations for backwards compatibility
+  return ns.data, V
 end
 
+##ToDo: This restricts the data of AbstractSymmetry to have same eltype, but could also promote eltypes
+(x::AbstractSymmetry * y::AbstractSymmetry) =
```
mtfishman, line 556: What about:
```julia
function (x::AbstractSymmetry * y::AbstractSymmetry)
  if !has_same_symmetry(x, y)
    error("Can't multiply two symmetric objects with different symmetries.")
  end
  return set_data(x, x.data * y.data)
end

has_same_symmetry(::AbstractSymmetry, ::AbstractSymmetry) = false
has_same_symmetry(::ConservesNf, ::ConservesNf) = true
has_same_symmetry(::ConservesNfParity, ::ConservesNfParity) = true
```
instead? That should be more general and extensible to other types wrapped by `AbstractSymmetry`.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -165,6 +265,33 @@ end
 #
 # Correlation matrix diagonalization
 #
+abstract type AbstractSymmetry end
+
+struct ConservingNfParity{T} <: AbstractSymmetry
```
mtfishman, line 270: I would vote for `ConservesNf` and `ConservesNfParity` instead of `Conserving*`, I think that makes it sound more like a property of the object and also is closer to how we might actually say it (i.e. "the state conserves particle number" vs. "the state is particle number conserving").

ITensorGaussianMPS/src/ITensorGaussianMPS.jl
```diff
@@ -8,10 +8,15 @@ using LinearAlgebra
...
+  correlation_matrix_to_mps,
   slater_determinant_to_gmps,
+  correlation_matrix_to_gmps,
   hopping_hamiltonian,
+  pairing_hamiltonian,
   slater_determinant_matrix,
-  slater_determinant_to_gmera
+  slater_determinant_to_gmera,
+  Pairing,
+  ConservingNf
```
mtfishman, line 19: I think we shouldn't export these since they should be mostly hidden from external users.

ITensorGaussianMPS/src/gmps.jl
b-kloss: Yes, that looks better.

ITensorGaussianMPS/src/gmps.jl
b-kloss: Ok.

ITensorGaussianMPS/src/ITensorGaussianMPS.jl
b-kloss: Ok.

ITensorGaussianMPS/src/gmps.jl
b-kloss: Ok.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -267,31 +277,31 @@ end
...
-site_stride(::ConservingNf) = 1
-site_stride(::ConservingNfParity) = 2
-Base.copy(A::T) where {T<:AbstractSymmetry} = T(Base.copy(A.data))
-Base.size(A::T) where {T<:AbstractSymmetry} = Base.size(A.data)
-Base.size(A::T, dim::Int) where {T<:AbstractSymmetry} = Base.size(A.data, dim)
+set_data(::ConservesNf, x) = ConservesNf(x)
+set_data(::ConservesNfParity, x) = ConservesNfParity(x)
+site_stride(::ConservesNf) = 1
+site_stride(::ConservesNfParity) = 2
+copy(A::T) where {T<:AbstractSymmetry} = T(Base.copy(A.data))
```
mtfishman, line 296: By the way you don't need `Base.copy` on the right hand side.

mtfishman: (Just `copy`.)

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -451,7 +463,7 @@ function set_populations!(_ns::ConservingNf, _nB::ConservingNf, _v::ConservingNf
 end
 
 function set_populations!(
```
mtfishman, line 465: I usually refer to them as `occupations`, is one more common than the other?

ITensorGaussianMPS/test/gmps.jl
```diff
@@ -2,8 +2,9 @@ using ITensorGaussianMPS
 using ITensors
 using LinearAlgebra
 using Test
-
-const Pairing = ITensorGaussianMPS.ConservingNfParity
+import ITensorGaussianMPS.ConservesNf
+import ITensorGaussianMPS.ConservesNfParity
+const Pairing = ConservesNfParity
```
mtfishman, line 7: Should be able to replace this with:
```julia
using ITensorGaussianMPS: ConservesNf, ConservesNfParity
```

mtfishman: `import` let's you overload the type/function, while `using` just brings it into the namespace.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -391,6 +817,103 @@ function interleave(xs...)
...
+    )
+    ns = interleave(
+      interleave(ns_up[1:2:end], ns_dn[1:2:end]), interleave(ns_up[2:2:end], ns_dn[2:2:end])
+    )
+  elseif Λ_up <: ConservesNf && Λ_dn <: ConservesNf
+    C = Circuit(interleave(C_up.rotations, C_dn.rotations))
+    ns = interleave(ns_up, ns_dn)
+  else
+    error("Λ_up and Λ_dn have incompatible subtypes of AbstractSymmetry")
+  end
```
mtfishman, line 909: This looks like a good candidate for splitting this off by overloading `interleave(::AbstractSymmetry, ::AbstractSymmetry)`.

b-kloss: Agreed. I'll address it when the spinful case is implemented properly.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -196,44 +539,103 @@ stored in a structure as a set of Givens rotations.
...
-    Λ = Hermitian(g * Λ * g')
+    LinearAlgebra.lmul!(g, V)
+    Λ = set_data(Λ, Hermitian(g * Matrix(Λ.data) * g'))
   end
-  return ns, V
+  ###return non-wrapped occupations for backwards compatibility
+  return ns.data, V
+end
+
+##ToDo: This restricts the data of AbstractSymmetry to have same eltype, but could also promote eltypes
```
mtfishman, line 618: I guess this shouldn't be a restriction anymore.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -391,6 +844,116 @@ function interleave(xs...)
...
+  second_half = Vector((n + 1):(2 * n))
+  interleaved_inds = interleave(first_half, second_half)
+  ordered_inds = sortperm(interleaved_inds)
+  return M[ordered_inds, ordered_inds]
+end
+
+function correlation_matrix_to_mps(
+  s::Vector{<:Index},
+  Λ_up0::AbstractSymmetry,
+  Λ_dn0::AbstractSymmetry;
```
b-kloss, line 886: Dispatching on number of sites being a (1,2 x) multiple of the size of the correlation matrix like in the spinless case is non-unique here. For example s could be 2x the size of the correlation matrix if two species of spinless `Fermion` is used and the state is number conserving, or it could be 2x the size of the correlation matrix if `Electron` sites are used and the state is only parity conserving. Possible fix could be dispatching on sitetypes ( `"Electron"` , `"Fermion"`)?

mtfishman: Maybe you figured it out, but yes I had been dispatching on the site types if I remember correctly.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -391,50 +812,110 @@ function interleave(xs...)
...
+    ###ToDo: Not sure what to do here yet. Just copied code from old interface below. 
     @assert length(s) == N_up
     @assert length(s) == N_dn
     if isspinful(s)
-      space_up = [QN(("Nf", 0, -1), ("Sz", 0)) => 1, QN(("Nf", 1, -1), ("Sz", 1)) => 1]
-      space_dn = [QN(("Nf", 0, -1), ("Sz", 0)) => 1, QN(("Nf", 1, -1), ("Sz", -1)) => 1]
+      if typeof(Λ_up)<:ConservesNf
+        space_up = [QN(("Nf", 0, -1), ("Sz", 0)) => 1, QN(("Nf", 1, -1), ("Sz", 1)) => 1]
+        space_dn = [QN(("Nf", 0, -1), ("Sz", 0)) => 1, QN(("Nf", 1, -1), ("Sz", -1)) => 1]
+      elseif typeof(Λ_up)<:ConservesNfParity
```
b-kloss, line 899: Is this correct?

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -391,50 +812,110 @@ function interleave(xs...)
...
     end
-    U = [ITensor(sf, g) for g in reverse(C.rotations)]
-    ψf = MPS(sf, n -> round(Int, ns[n]) + 1, U; kwargs...)
+    U = itensors(sf, set_data(Λ_up, C))
+    ψf = MPS(MPS_Elt, sf, n -> round(Int, ns[site_stride(Λ_up) * n]) + 1)
+    @show norm(ψf)
+    ψf = apply(U, ψf; kwargs...)
+    @show norm(ψf)
     ψ = MPS(N_up)
     for n in 1:N_up
```
b-kloss, line 920: State is normalized up to application of the combiners. Norm explodes after application of the combiners. Probably an issue with the way I defined the `space` above? Not an issue for number-conserving state.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -181,82 +171,61 @@ function quadratic_hamiltonian(os::OpSum)
     sites[n] = ntuple(n -> ITensors.site(term[n]), Val(2))
     nsites = max(nsites, maximum(sites[n]))
   end
-  quadrants_present = unique(quads)
-  # currently compatible formats:
-  # [(2,2)] -> hopping only
-  # [(1,2),(2,2)] -> hopping and pairing
-  @show quadrants_present
   ElT = all(isreal(coefs)) ? Float64 : ComplexF64
```
mtfishman, line 174: Can we use:
```julia
coef_type = reduce(Base.promote_typeof, coefs)
ElT = all(isreal(coefs)) ? real(coef_type) : coef_type
```
to make it more agnostic about the precision (i.e. more gracefully handle `Float32`, etc.)?

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -181,82 +171,61 @@ function quadratic_hamiltonian(os::OpSum)
...
+  other_quad = i -> i==2 ? 1 : 2
+  for n in 1:nterms
+    quad = quads[n]
+    offsets = nsites .* (quad .- 1)
+    h[(sites[n] .+ offsets)...] += coefs[n]
+    other_offsets = nsites .* (other_quad.(quad) .- 1)
+    h[(sites[n] .+ other_offsets)...] += -conj(coefs[n])
   end
-  return h
+  return Hermitian(interleave(h))
```
mtfishman, line 184: Should we check that it is approximately Hermitian? We could use `ishermitian` and `IsApprox.jl` (https://github.com/jlapeyre/IsApprox.jl).

ITensorGaussianMPS/src/gmps.jl
mtfishman: That looks ok, though the modules should be `-2` to indicate it is fermionic. You can use the `ITensors.space` functions to check the internal definitions we have:
```julia
julia> space(SiteType("Fermion"))
2

julia> space(SiteType("Fermion"); conserve_qns=true)
2-element Vector{Pair{QN, Int64}}:
 QN("Nf",0,-1) => 1
 QN("Nf",1,-1) => 1

julia> space(SiteType("Fermion"); conserve_nfparity=true)
2-element Vector{Pair{QN, Int64}}:
 QN("NfParity",0,-2) => 1
 QN("NfParity",1,-2) => 1
```

ITensorGaussianMPS/src/gmps.jl
mtfishman: Did you check the indices/QN spaces of the combiners and the MPS tensors before/after the application, say for a very small MPS?

Maybe related to https://github.com/ITensor/ITensors.jl/issues/1034?

b-kloss: I located the problem to be in `identity_blocks_itensor` but have no idea why it fails for non-number conserving case while it works in the number-conserving one. I split the combiner application from the identity_blocks_itensor application for testing purposes and the norm only diverges after the latter. Unlikely that it's related to #1034

b-kloss: I think the mapping inside `identity_blocks_itensor` may go wrong when the dimension of the subspaces is > 1. I'll test that independently.

b-kloss: `identity_blocks` is only works for blocksizes of 1 or if input and output indices have the same (compatible layout). Basically mapping between the combined `Fermion` space with only `NfParity` as quantum number and an `Electron` space with only NfParity quantum number is not a well-defined operation for two `QNIndex`.

mtfishman: Yes, I could believe that the current code doesn't handle that properly and that is a bit tricky to write with current BlockSparse/QN functionality

b-kloss:
I inserted error messages when `Electron`-sites are used together with `ConservesNfParity` and documented in the code what to address in the future. I also reduced the hamiltonian interface to `quadratic_hamiltonian` (general including the pairing terms) and `hopping_hamiltonian` (only the number-conserving terms), with a keyword-arg for the latter whether we want to convert to a hopping hamiltonian (with check) or extract the hopping term from a general quadratic `OpSum`.

ITensorGaussianMPS/src/gmps.jl
b-kloss: The first line returns any for me on Julia v1.8. I wrote an alternative but it's much less elegant than this solution.

b-kloss: I pushed my fix to this, it's a bit more verbose.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -111,52 +136,111 @@ is_annihilation_operator(::OpName"c") = true
...
+  h[(N + 2):2:(2 * N), (N + 2):2:(2 * N)] = h_dn[(n + 1):N, (n + 1):N]
+  # super-quadrant (1,2)
+  h[1:2:N, (N + 1):2:(2 * N)] = h_up[1:n, (n + 1):(2 * n)]
+  h[2:2:N, (N + 2):2:(2 * N)] = h_dn[1:n, (n + 1):(2 * n)]
+
+  #convert from blocked to interlaced format. Odd base-rows are spin-up, even are spin-down.
+  return Hermitian(interleave(h))
+end
+
+function hopping_hamiltonian(os::OpSum; extract=false)
```
mtfishman, line 221: What does `extract` mean here? Does `extract=true` mean that it drops any pairing terms? If so, we could call it `ignore_pairing_terms` or `drop_pairing_terms` (in the future it would make sense to provide a tolerance cutoff for when to drop the terms but we could leave that for future work, and then `drop_pairing_terms=true` would be the same as `drop_pairing_terms_tol=Inf`).

ITensorGaussianMPS/src/gmps.jl
mtfishman: @b-kloss I guess these comments were related to trying to get `Electron` site types working with general Gaussian states with pairing correlations which we are leaving for future work so this thread can be resolved?

ITensorGaussianMPS/src/gmps.jl
mtfishman: (EDIT: Let's use `drop_pairing_terms` to be consistent with the terminology used in the function `maybe_drop_pairing_correlations`.)

b-kloss: `extract=true` means it drops pairing terms, while otherwise it would raise an error if pairing terms with weight above eps() are present. I'll address this.

b-kloss: I decided to go just with `drop_pairing_terms_tol[=0.0]` for brevity. It's not as intuitive that you have to set it to Inf if you want to drop them always, but i think it's worth the tradeoff of having only one keyword-argument and not needing to handle potential contradictions between the two.

ITensorGaussianMPS/src/gmps.jl
```diff
@@ -111,52 +136,110 @@ is_annihilation_operator(::OpName"c") = true
...
+  #convert from blocked to interlaced format. Odd base-rows are spin-up, even are spin-down.
+  return Hermitian(interleave(h))
+end
+
+function hopping_hamiltonian(os::OpSum; drop_pairing_terms_tol=0.0)
+  # convert to blocked format
+  h = reverse_interleave(Matrix(quadratic_hamiltonian(os)))
+  # check that offdiagonal blocks are 0
+  N = div(size(h, 1), 2)
+  tol=max(drop_pairing_terms_tol,eps(real(eltype(h))))
```
mtfishman, line 223: Instead of this, shouldn't we make the default value equivalent to `drop_pairing_terms_tol=eps(real(eltype(h)))`? I think we should let people set the tolerance to zero if they want, but pick a reasonable default value.

We will have to write a function that gets the equivalent of `eltype(h)` based on `os`. We could define a function `coefficient_type(os::OpSum)`, for example:
```julia
function coefficient_type(os::OpSum)
  return mapreduce(typeof ∘ coefficient, promote_type, terms(os))
end
```
or something like that. Then we can set the default `drop_pairing_terms_tol=eps(real(coefficient_type(os)))`.

Otherwise, we could set the default to `drop_pairing_terms_tol=nothing` and then have a line:
```julia
if isnothing(drop_pairing_terms_tol)
  drop_pairing_terms_tol = eps(real(eltype(h)))
end
```
Maybe that makes the most sense for efficiency (i.e. not having to loop over the coefficients multiple times).

b-kloss: I opted for the latter, since it's simpler.

mtfishman:
[gaussianmps]

ITensorGaussianMPS/test/gmps.jl
```diff
@@ -2,124 +2,192 @@ using ITensorGaussianMPS
...
+    if 1 < i < N
+      js = [i - 1, i + 1]
+    elseif i == 1
+      js = [i + 1]
+    else
+      js = [i - 1]
+    end
+    for j in js
+      #os .+= pairing, "Cdag", i, "Cdag", j
+      os .+= -conj(pairing), "C", i, "C", j
```
mtfishman, line 44: Why don't these both have to be specified?

mtfishman: I think that for both hopping and pairing terms we should take the inputs very literally and not add in any terms that weren't specified in the `OpSum`. So if someone inputs just `"Cdag", 1, "Cdag" 2` we shouldn't infer that they meant to also add `"C", 1, "C", 2`, and if someone specifies just `"Cdag", 1, "C", 2` we shouldn't infer that they meant to also add `"C", 1, "Cdag", 2`. That is how it works for MPO construction and I think we should be consistent with that (for example, so that a user can input that same `OpSum` into `quadratic_hamiltonian` and `MPO` and get consistent results), and additionally I think we shouldn't be too presumptuous since someone might actually want a non-Hermitian quadratic operator. We can add a check that the hopping Hamiltonian that is extracted is in fact Hermitian where appropriate, or people can check that themselves.

mtfishman: Could you explain what is different about how pairing and hopping terms have to be handled? I'm not following that part of the story.

b-kloss: `"Cdag", 1, "C", 2` and `"C", 1, "Cdag", 2` are the same term (up to a minus sign due to commutation relations). It's sufficient to have one of them in the `OpSum`. If it's hermitian conjugate is also present, the resulting contribution to the Hamiltonian is also Hermitian, no matter in which order the term was specified (again assuming that the sign from the commutation relations has been taken into account in the coefficient).
`"Cdag", 1, "Cdag", 2` and `"C", 1, "C", 2` are terms which both need to be specified explicitly in the OpSum in order to have a hermitian operator. The old handling of `quadratic_hamiltonian` did not require this, in analogy to the hopping terms above. However, this lead to a mismatch in the definition of the coefficient between an `OpSum` passed to `MPO` and `quadratic_hamiltonian`, so good that you caught this! I fixed it and will push after fixing a weird bug I am running into.

mtfishman: Ok thanks for the explanation, that makes sense.

In principle we _could_ have some concept of terms being equivalent up to Hermitian conjugation, which would require defining what the Hermitian conjugate/dagger of a certain operator is, for example with definitions like`dag(::OpName"C") = OpName"Cdag"()`. Then we could take the `dag` of terms of an `OpSum`, and therefore determine which terms are equivalent up to Hermitian conjugation. Of course we don't have to do that for this PR and that would be more involved, just putting it out there in case it inspires some way of designing the code to make it more generic between hopping and pairing terms.

b-kloss: The "bug" turned out to be just a normal variation of the quality of the result with parameters, so I think this is resolved, unless you would like me to add a test that actually test for the equivalence of the operator returned by `quadratic_operator(os::OpSum)` and `MPO(os::OpSum)` (`examples/spinless_fermion_pairing.jl` verifies this, but I didn't make it into a test).

ITensorGaussianMPS/test/gmps.jl
```diff
@@ -10,116 +10,180 @@ using Test
...
+      os_p2 .+= -Delta2 / 2.0, "C", n, "C", n + 1
+      os_p2 .+= Delta2 / 2.0, "C", n + 1, "C", n
     end
-  end
-  H = MPO(os, s)
 
-  @test inner(ψ', H, ψ) ≈ E rtol = 1e-5
-  @test inner(ψ', H, ψ) / norm(ψ) ≈ E rtol = 1e-5
+    h = ITensorGaussianMPS.quadratic_hamiltonian(os_h + os_p)
+    h2 = ITensorGaussianMPS.quadratic_hamiltonian(os_h + os_p2)
```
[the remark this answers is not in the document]
b-kloss: Taken care of with last commit.

ITensorGaussianMPS/test/gmps.jl
mtfishman: Thanks for adding the test, that will be a good reference to make sure we're being consistent about the conventions.

mtfishman:
[gaussianmps]

PR #1040, opened 2023-01-13
Kmp5/enhancements/inits

kmp5VT:
# Description

With the goal to make more versatile code, create a default _storage_type function that take type and index information and returns the appropriate storage type. Any missing information will follow default behavior defined (for now) by me. Utilize this default_storage_type variable in all constructors of NDTensirs and ITensors

# Checklist:

### DIFF
--- a/NDTensors/src/default_storage.jl
+++ b/NDTensors/src/default_storage.jl
@@ -1,6 +1,6 @@
 ## This is a fil which specifies the default storage type provided some set of parameters
 ## The parameters are the element type and storage type
-function default_storage_type(storetype::Type{<:TensorStorage} = Dense, eltype::Type = Float64, datatype::Type{<:AbstractArray} = Vector, indextype::Type = Tuple{Int,})
+function default_storage_type(storetype::Type{<:TensorStorage} = Dense, eltype::Type = Float64, datatype::Type{<:AbstractArray} = Vector, index = Tuple{Int,})
   #TODO check if indices are block sparse or not
   type = nothing
   try
@@ -12,13 +12,10 @@ function default_storage_type(storetype::Type{<:TensorStorage} = Dense, eltype::
   return type
 end
 
-function default_storage_type(eltype::Type{<:Number})
-  default_storage_type(Dense, eltype)
-end
+default_storage_type(eltype::Type{<:Number}) =  default_storage_type(Dense, eltype)
 
-function default_storage_type(datatype::Type{<:AbstractArray})
+default_storage_type(datatype::Type{<:AbstractArray}) =
   default_storage_type(Dense, Float64, datatype)
-end
 
 ## This function needs to be defined for each vector type
 default_storage_type(datatype::Type{<:Vector{ElT}}) where {ElT} = default_storage_type(Dense, ElT, Vector)
@@ -27,4 +24,4 @@ default_storage_type(eltype::Type, datatype::Type{<:AbstractArray}) = default_st
 
 default_storage_type(datatype::Type{<:AbstractArray}, eltype::Type) = default_storage_type(Dense, eltype, datatype)
 
-function def
+default_storage_type(index) = default_storage_type(Dense, Float64, Vector, index)
--- a/NDTensors/src/default_storage.jl
+++ b/NDTensors/src/default_storage.jl
@@ -1,0 +1,30 @@
+## This is a fil which specifies the default storage type provided some set of parameters
+## The parameters are the element type and storage type
+function default_storage_type(storetype::Type{<:TensorStorage} = Dense, eltype::Type = Float64, datatype::Type{<:AbstractArray} = Vector, indextype::Type = Tuple{Int,})
+  #TODO check if indices are block sparse or not
+  type = nothing
+  try
+    type = storetype{eltype, datatype{eltype}}
+  catch e
+    println("Cannot construct storetype", storetype)
+    throw(e)
+  end
+  return type
+end
+
+function default_storage_type(eltype::Type{<:Number})
+  default_storage_type(Dense, eltype)
+end
+
+function default_storage_type(datatype::Type{<:AbstractArray})
+  default_storage_type(Dense, Float64, datatype)
+end
+
+## This function needs to be defined for each vector type
+default_storage_type(datatype::Type{<:Vector{ElT}}) where {ElT} = default_storage_type(Dense, ElT, Vector)
+
+default_storage_type(eltype::Type, datatype::Type{<:AbstractArray}) = default_storage_type(Dense, eltype, datatype)
+
+default_storage_type(datatype::Type{<:AbstractArray}, eltype::Type) = default_storage_type(Dense, eltype, datatype)
+
+function def
--- a/NDTensors/src/dense/dense.jl
+++ b/NDTensors/src/dense/dense.jl
@@ -18,6 +18,10 @@ struct Dense{ElT,VecT<:AbstractArray} <: TensorStorage{ElT}
   function Dense{ElT,VecT}(data::Array) where {ElT,VecT<:AbstractArray{ElT}}
     return new{ElT,VecT}(vec(data))
   end
+
+  function Dense{ElT, VecT}() where {ElT, VecT<:AbstractArray{ElT}}
+    return new{ElT, VecT}(VecT(undef, 0))
+  end
 end
 
 function Dense(data::VecT) where {VecT<:AbstractArray{ElT}} where {ElT}
--- a/NDTensors/src/NDTensors.jl
+++ b/NDTensors/src/NDTensors.jl
@@ -31,6 +31,7 @@ include("imports.jl")
 # DenseTensor and DiagTensor
 #
 include("aliasstyle.jl")
+include("default_storage.jl")
 include("similar.jl")
 include("tupletools.jl")
 include("dims.jl")
--- a/NDTensors/src/NDTensors.jl
+++ b/NDTensors/src/NDTensors.jl
@@ -31,11 +31,11 @@ include("imports.jl")
 # DenseTensor and DiagTensor
 #
 include("aliasstyle.jl")
-include("default_storage.jl")
 include("similar.jl")
 include("tupletools.jl")
 include("dims.jl")
 include("tensorstorage.jl")
+include("default_storage.jl")
 include("tensor.jl")
 include("adapt.jl")
 include("generic_tensor_operations.jl")


NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,39 @@
...
+  #TODO check if indices are block sparse or not
+  type = nothing
+  try
+    type = storetype{eltype,datatype{eltype}}
+  catch e
+    println("Cannot construct storetype", storetype)
+    throw(e)
+  end
+  return type
+end
```
mtfishman, line 18: I would avoid this version of the function for now since the logic could become quite complicated making the element type, vector type, and storage type compatible with each other. For example, this would have to handle inputs like:
```julia
default_storage_type(Dense{Float32,Vector{Float32}}, ComplexF64, CuVector{ComplexF32})
```
which I think would benefit from a function like:
```julia
adapt_type(Float32, Vector{Float64}) -> Vector{Float32}
```
as we discussed the other day.

NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,39 @@
...
+
+default_storage_type(eltype::Type{<:Number}) = default_storage_type(Dense, eltype)
```
mtfishman, line 20: I would write this as:
```julia
default_storage_type(eltype::Type{<:Number}) = default_storage_type(default_data_type(eltype))
```
and then define:
```julia
default_data_type(eltype::Type{<:Number}) = Vector{eltype}
```

NDTensors/src/default_storage.jl
mtfishman: Also maybe we should use `default_storagetype` to be consistent with the existing functions `storagetype` and `set_storagetype`.

NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,39 @@
...
+
+function default_storage_type(datatype::Type{<:AbstractArray})
+  return default_storage_type(Dense, Float64, datatype)
```
mtfishman, line 23: Here I would use the existing `set_datatype` function, for example:
```julia
default_storagetype(datatype::Type{<:AbstractArray}) = set_datatype(Dense, datatype)
```

NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,39 @@
...
+end
+
+## This function needs to be defined for each vector type
+function default_storage_type(datatype::Type{<:Vector{ElT}}) where {ElT}
```
mtfishman, line 27: This doesn't seem necessary if the `AbstractArray` is written in a generic way.

NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,39 @@
...
+  return default_storage_type(Dense, ElT, Vector)
+end
+
+function default_storage_type(eltype::Type, datatype::Type{<:AbstractArray})
```
mtfishman, line 31: Is this needed? I would hope that for simplicity we could avoid versions where you input multiple types, at least for now.

NDTensors/src/dense/dense.jl
```diff
@@ -18,8 +18,24 @@ struct Dense{ElT,VecT<:AbstractArray} <: TensorStorage{ElT}
   function Dense{ElT,VecT}(data::Array) where {ElT,VecT<:AbstractArray{ElT}}
     return new{ElT,VecT}(vec(data))
   end
+
+  function Dense{ElT, VecT}() where {ElT, VecT<:AbstractArray{ElT}}
+    return new{ElT, VecT}(VecT(undef, 0))
```
mtfishman, line 23: Can this be written as `VecT()` instead of `VecT(undef, 0)`?

NDTensors/src/dense/dense.jl
```diff
@@ -18,8 +18,24 @@ struct Dense{ElT,VecT<:AbstractArray} <: TensorStorage{ElT}
...
+  end
 end
 
+## We need a general typeinfo constructor for Dense
+## This constructor will be called by all other versions
+Dense(eltype::Type, datatype::Type) = default_storage_type(Dense, eltype, datatype)
```
mtfishman, line 29: These constructors are a bit funny to me, since they are returning types, not objects.

I think these should be functions:
```julia
dense_type(eltype::Type{<:Number}) = adapt_type(eltype, Dense)
dense_type(datatype::Type{<:AbstractArray}) = adapt_type(datatype, Dense)
```
Also I would try to avoid the version that accepts multiple types to keep things simple (though maybe there is a need for it that I'm not seeing right now).

NDTensors/src/default_storage.jl
mtfishman: Then it is expected that `set_datatype` is defined for each storage type, which is already expected by my implementation of `adapt`.

Possibly we can think about making a default fallback version of `set_datatype` for all storage types using Julia introspection tools, see for example: https://discourse.julialang.org/t/error-calling-similar-array-complexf32-float32/86248 and https://discourse.julialang.org/t/manipulating-parametric-types/6893, however it doesn't seem like there is an official Julia API for manipulating type parameters of a type.

NDTensors/src/dense/dense.jl
```diff
@@ -18,8 +18,24 @@ struct Dense{ElT,VecT<:AbstractArray} <: TensorStorage{ElT}
...
+
+Dense(datatype::Type{<:AbstractArray}) = Dense(Float64, datatype)
+
+Dense(eltype::Type{<:Number}) = Dense(eltype, Vector)
```
mtfishman, line 33: Maybe this could be:
```julia
dense_type(eltype::Type{<:Number}) = dense_type(default_datatype(eltype))
```

NDTensors/src/dense/dense.jl
```diff
@@ -18,8 +18,24 @@ struct Dense{ElT,VecT<:AbstractArray} <: TensorStorage{ElT}
...
+  function Dense{ElT, VecT}() where {ElT, VecT<:AbstractArray{ElT}}
+    return new{ElT, VecT}(VecT(undef, 0))
+  end
 end
 
+## We need a general typeinfo constructor for Dense
+## This constructor will be called by all other versions
+Dense(eltype::Type, datatype::Type) = default_storage_type(Dense, eltype, datatype)
+
+Dense(datatype::Type{<:AbstractArray}) = Dense(Float64, datatype)
```
mtfishman, line 31: This looks like it would be wrong for inputs like `Dense(Vector{Float32})`.

Maybe this should be:
```julia
dense_type(datatype::Type{<:AbstractArray}) = adapt_type(datatype, Dense)
```
or:
```julia
dense_type(datatype::Type{<:AbstractArray}) = set_datatype(Dense, datatype)
```

NDTensors/src/default_storage.jl
kmp5VT: Yeah, I messaged the Julia slack and realized writing the function in this way made things much more difficult for myself. I have made some updates which should simplify the idea. I'm still going through your `Dense` comments

NDTensors/src/default_storage.jl
kmp5VT: This makes a lot of sense, I have updated that code, thanks!

NDTensors/src/default_storage.jl
kmp5VT: Yeah I have a more generic method, this was a result of the original code taking the `abstractarray` type without an element type. Once you add element type data its really difficult to separate array type and array element type information

NDTensors/src/default_storage.jl
kmp5VT: I have updated this and now have 2 functions
`default_datatype(eltype::Type{<:Number})` and `default_storagetype(datatype::Type{<:AbstractArray{ElT}}) where {ElT}`
In this way, if you call `default_storagetype(Vector{Vector{Float64}})` the code correctly constructs a `Dense{Vector{Float64}, Vector{Vector{Float64}}}` but it is not possible to call `default_datatype(Vector{Float64})` so it might make sense to choose one and use it for both definitions above. I would think it would be more flexible to change the first definition to `default_datatype(eltype::Type)`

NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,19 @@
+## This is a fil which specifies the default storage type provided some set of parameters
+## The parameters are the element type and storage type
+default_datatype(eltype::Type{<:Number}) = Vector{eltype}
+default_eltype() = Float64
+
+default_storagetype(datatype::Type{<:AbstractArray{ElT}}) where {ElT} = Dense{ElT, datatype}
```
mtfishman, line 6: I prefer to write this as:
```julia
default_storagetype(datatype::Type{<:AbstractArray}) = Dense{eltype(datatype), datatype}
```
In general I think it is best to use functions like that instead of `where` as much as possible, both as a style choice and also it can make it easier to write generic code.

NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,19 @@
...
+
+default_storagetype(datatype::Type{<:AbstractArray}) = default_storagetype(datatype{default_eltype()})
```
mtfishman, line 8: Wouldn't this overwrite the version above, and also overwrite an existing element type?

NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,19 @@
...
+default_storagetype(eltype::Type{<:Number}) = default_storage_type(default_datatype(eltype))
+default_storagetype() = default_storagetype(default_datatype(default_eltype()))
```
mtfishman, line 10: We could define:
```julia
default_datatype() = default_datatype(default_eltype())
```
to simplify this a bit.

NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,19 @@
...
+default_storage_type(storagetype::Type{<:TensorStorage}) =
```
mtfishman, line 11: This function may be beyond the type of functionality we want for the `default_storage_type`. I think `default_storage_type` should only handle hierarchies, not converting/adapting storage types.

Instead, I think this functionality could be handled by a function `adapt_type` (which would supersede the current `NDTensors.similartype` function):
```julia
adapt_type(default_datatype(), storagetype)
```

mtfishman: Open to discussion about it though. A lot of this design depends on how it will actually get used in the constructors, so I think it will be helpful to see how that works out.

NDTensors/src/dense/dense.jl
```diff
@@ -20,38 +20,43 @@ struct Dense{ElT,VecT<:AbstractArray} <: TensorStorage{ElT}
...
+#Start with high information constructors and move to low information constructors
+function Dense{ElT, VecT}() where {ElT, VecT<:AbstractArray{ElT}}
+  return default_Densetype(VecT)(VecT())
 end
 
-function Dense(data::Array{ElT}) where {ElT}
-  return Dense{ElT,Vector{ElT}}(vec(data))
+# Construct from a set of indices
+function Dense{ElT,VecT}(inds::Tuple) where {ElT,VecT<:AbstractArray{ElT}}
+  return default_Densetype(VecT)(VecT(undef, dim(inds)))
```
mtfishman, line 36: The function `default_Densetype` doesn't seem necessary here since the element type and data type are already specified by `ElT` and `VecT`, it seems like you can just call the inner constructor directly here.

NDTensors/src/dense/dense.jl
```diff
@@ -20,38 +20,43 @@ struct Dense{ElT,VecT<:AbstractArray} <: TensorStorage{ElT}
...
+  #eturn Dense(VecT(undef, dim(inds)))
 end
 
+# This function is ill-defined. It cannot transform a complex type to real...
 function Dense{ElR}(data::AbstractArray{ElT}) where {ElR,ElT}
-  return ElT == ElR ? Dense(data) : Dense(ElR.(data))
+  return default_Densetype(ElR)(data)
```
mtfishman, line 42: Again, I don't think `default_Densetype` is necessary here, this is better handled by a type conversion or `adapt`.

mtfishman: We should think about the `default_storagetype` function as only used when we don't have enough information based on the inputs about the full storage type, which is not the case here (since this construct a storage type `Dense{ElR,AbstractArray{ElR}}` which is fully inferrable from the inputs).

NDTensors/src/dense/dense.jl
```diff
@@ -20,38 +20,43 @@ struct Dense{ElT,VecT<:AbstractArray} <: TensorStorage{ElT}
...
 end
 
-# Construct from a set of indices
-function Dense{ElT,VecT}(inds) where {ElT,VecT<:AbstractArray{ElT}}
-  return Dense(VecT(dim(inds)))
-end
+#Do we want this to be zero? 
+Dense{ElT}(dim::Integer) where {ElT} = default_Densetype(ElT)(default_datatype(ElT)(undef, dim))
```
mtfishman, line 46: I don't think `default_Densetype(ElT)` is needed here, it seems sufficient to just use a `Dense` constructor directly.

By default I think this should make a storage filled with zeros, and only make `undef` if `undef` is specified. But that touches on a question about how to make a generic data that is filled with zeros.

We may need to define additional generic functions for that like:
```julia
zeros(datatype::Type{<:AbstractArray}, dims) = fill!(datatype(undef, dims), zero(eltype(datatype)))
```

NDTensors/src/dense/dense.jl
```diff
@@ -20,38 +20,43 @@ struct Dense{ElT,VecT<:AbstractArray} <: TensorStorage{ElT}
...
 
-Dense{ElT}(dim::Integer) where {ElT} = Dense(zeros(ElT, dim))
+Dense{ElT}() where {ElT} = default_Densetype(ElT)()
```
mtfishman, line 48: I think this should avoid `default_Densetype` and just be:
```julia
Dense{ElT}() where {ElT} = Dense(default_datatype(ElT)())
```

NDTensors/src/dense/dense.jl
```diff
@@ -20,38 +20,43 @@ struct Dense{ElT,VecT<:AbstractArray} <: TensorStorage{ElT}
   end
 end
 
-function Dense(data::VecT) where {VecT<:AbstractArray{ElT}} where {ElT}
-  return Dense{ElT,VecT}(data)
+## We need a general typeinfo constructor for Dense
+default_Densetype(datatype::Type{<:AbstractArray{ElT}}) where {ElT} = default_storagetype(datatype)
```
mtfishman, line 24: Looking through the code below, I don't think this function is really necessary, and the same functionality can be implemented with a combination of the `default_datatype` function and the `Dense` constructor.

NDTensors/src/default_storage.jl
kmp5VT: I'll try to reconfigure and move forward without using where (when possible)

NDTensors/src/default_storage.jl
kmp5VT: No so these are actually 2 different cases. The first one covers if you provide a datatype with a elementtype hard coded or deduced from previous lines of code. The second one covers if you just provide a datatype without any type information. By default `Vector` is a `Array{T, 1} where {T}` so `eltype(Vector) = Any`. To mitigate problems of multiple types in a vector I require a call with an untyped datatype to use the default element type

NDTensors/src/default_storage.jl
kmp5VT: I think that this functionality would be useful, but at the moment complicated to implement in this matter. I saw your comments the day before and didn't modify this `default_storage_type` function from the previous push. For now I will remove it and I think it makes sense to utilize an `adapt_type` function to perform this operation.

NDTensors/src/default_storage.jl
mtfishman: Indeed you're correct, I didn't appreciate the difference in how Julia dispatches one those two signatures:
```julia
julia> f(T::Type{<:AbstractArray{ElT}}) where {ElT} = "f1"

julia> f(T::Type{<:AbstractArray})= "f2"
f (generic function with 2 methods)

julia> f(Vector{Int})
"f1"

julia> f(Vector)
"f2"
```
That's a nice way to differentiate whether or not an element type is already specified!

I think instead of:
```julia
default_storagetype(datatype::Type{<:AbstractArray}) = default_storagetype(datatype{default_eltype()})
```
we should do something like:
```julia
default_storagetype(datatype::Type{<:AbstractArray}) = default_storagetype(adapt_type(default_eltype(), datatype))
```
(here I'm using `adapt_type` which is the next generation of the current function `similartype`: https://github.com/ITensor/ITensors.jl/blob/v0.3.23/NDTensors/src/similar.jl). Otherwise it wouldn't preserve other type parameters like the dimensions of the AbstractArray, or handle array wrapper correctly.

kmp5VT: Is adapt_type currently functional/implemented?

mtfishman: Unfortunately no, I was just talking to Miles and he needs the same kind of functionality so we should meet about implementing it. It could even get turned into a general Julia package `AdaptTypes.jl`.

But for the sake of this PR for now you may be able to use `similartype`.

kmp5VT: Okay, I know `similartype` does not work with CuVectors currently but I will still include it on this PR

NDTensors/src/dense/dense.jl
kmp5VT: Ahh, I see!

NDTensors/src/dense/dense.jl
```diff
@@ -20,53 +20,67 @@ struct Dense{ElT,VecT<:AbstractArray} <: TensorStorage{ElT}
...
 
-Dense(::UndefInitializer, dim::Integer) = Dense(Float64, undef, dim)
+Dense(VecT::Type{<:AbstractArray{ElT}}, dim::Integer) where ElT = Dense{ElT, VecT}((dim,))
 
-Dense{ElT}() where {ElT} = Dense(ElT[])
-Dense(::Type{ElT}) where {ElT} = Dense{ElT}()
+Dense(::Type{ElT}, dim::Integer) where {ElT} = Dense{ElT}(dim)
 
-setdata(D::Dense, ndata) = Dense(ndata)
+Dense(x::ElT, dim::Integer) where {ElT<:Number} = Dense{ElT}((fill(x, dim)))
```
mtfishman, line 66: This implicitly creates CPU data so for consistency we should think about how to write it in terms of `default_datatype`. Maybe:
```julia
function Dense(x::Number, dim::Integer)
  data = default_datatype(typeof(x))(undef, dim)
  return Dense(fill!(data, x))
end
```

NDTensors/src/dense/dense.jl
```diff
@@ -89,8 +89,8 @@ function Base.real(::Type{Dense{ElT,DataT}}) where {DataT<:AbstractArray{ElT}} w
    default_Densetype( adapt(real(ElT), DataT))
 end
 
-function complex(::Type{Dense{ElT,DataT{ElT}}}) where {ElT}
-  return Dense{complex(ElT),DataT{complex(ElT)}}
+function complex(::Type{Dense{ElT,Vector{ElT}}}) where {ElT}
+  return Dense{complex(ElT),Vector{complex(ElT)}}
```
mtfishman, line 93: Seems like this could be:
```julia
function complex(T::Type{<:Dense})
  return set_datatype(T, similartype(datatype(T), complex(eltype(T))))
end
```

ITensorGPU/src/cuitensor.jl
```diff
@@ -2,8 +2,10 @@ import ITensors.NDTensors: NeverAlias, AliasStyle, AllowAlias
 import ITensors: ITensor
 import CUDA: CuArray
 
-function cuITensor(::Type{T}, inds::IndexSet) where {T<:Number}
-  return ITensor(Dense{float(T)}(CUDA.zeros(float(T), dim(inds))), inds)
+function cuITensor(eltype::Type{<:Number}, inds::IndexSet)
+  d = Dense(CUDA.zeros(eltype, dim(inds)))
```
mtfishman, line 6: Can't we use `default_storagetype(CuVector{eltype})` here?

NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,14 @@
+## This is a fil which specifies the default storage type provided some set of parameters
+## The parameters are the element type and storage type
+default_datatype(eltype::Type{<:Number}) = Vector{eltype}
+default_eltype() = Float64
+
+default_storagetype(datatype::Type{<:AbstractArray{ElT}}) where {ElT} = Dense{ElT,datatype}
```
mtfishman, line 6: `default_storagetype(datatype::Type{<:AbstractArray}) = Dense{eltype(datatype),datatype}`

NDTensors/src/dense/dense.jl
```diff
@@ -3,70 +3,101 @@
...
+
+#Start with high information constructors and move to low information constructors
+function Dense{ElT,DataT}() where {ElT,DataT<:AbstractArray{ElT}}
+  return Dense{ElT,DataT}(DataT())
+end
+
+# Construct from a set of indices
+# This will fail if zero(ElT) is not defined for the ElT
+function Dense{ElT,DataT}(inds::Tuple) where {ElT,DataT<:AbstractArray{ElT}}
+  return Dense{ElT,DataT}(zeros(DataT, dim(inds)))
```
mtfishman, line 31: I guess this should use `generic_zeros`? Would the plan be to use `generic_zeros(Dense{ElT,DataT}, dim(inds))` for this kind of thing?

NDTensors/src/dense/dense.jl
```diff
@@ -3,70 +3,101 @@
...
+end
+
+function Dense{ElT,DataT}(
+  ::UndefInitializer, inds::Tuple
+) where {ElT,DataT<:AbstractArray{ElT}}
+  return Dense{ElT,DataT}(DataT(undef, dim(inds)))
```
mtfishman, line 37: I guess there will be a problem in functions like this where if `DataT` is not `AbstractVector` (i.e. a more general `AbstractArray`) then this construct would fail since you only provide a single dimension. For example:
```julia
julia> Array{Int,3}(undef, 4)
ERROR: MethodError: no method matching Array{Int64, 3}(::UndefInitializer, ::Int64)
Closest candidates are:
  Array{T, 3}(::UndefInitializer, ::Int64, ::Int64, ::Int64) where T at boot.jl:463
  Array{T, N}(::UndefInitializer, ::Int64...) where {T, N} at boot.jl:465
  Array{T, N}(::UndefInitializer, ::Integer...) where {T, N} at baseext.jl:25
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[4]:1
```
A fix could be to use `dims(inds)` instead, which would require that `length(inds) == ndims(DataT)`.

Then we could have another version that makes `length(inds) == ndims(DataT)` if they are not already equal. But this is a tricky case now that we're thinking of making `DataT` more general.

NDTensors/src/dense/dense.jl
```diff
@@ -3,70 +3,101 @@
...
+end
+
+function Dense{DataT}(inds::Tuple) where {DataT<:AbstractArray}
+  return default_storagetype(DataT)(inds)
 end
 
-function Dense(data::VecT) where {VecT<:AbstractArray{ElT}} where {ElT}
-  return Dense{ElT,VecT}(data)
+function Dense{DataT}(::UndefInitializer, inds::Tuple) where {DataT<:AbstractArray}
+  return default_storagetype(DataT)(undef, inds)
```
mtfishman, line 49: Shouldn't this be `dims(inds)`?

NDTensors/src/dense/dense.jl
```diff
@@ -3,70 +3,101 @@
...
 function Dense{ElR}(data::AbstractArray{ElT}) where {ElR,ElT}
-  return ElT == ElR ? Dense(data) : Dense(ElR.(data))
+  return Dense{ElR,similartype(typeof(data), ElR)}(data)
 end
 
-# Construct from a set of indices
-function Dense{ElT,VecT}(inds) where {ElT,VecT<:AbstractArray{ElT}}
-  return Dense(VecT(dim(inds)))
+function Dense{ElT}(dim::Integer) where {ElT<:Number}
+  return default_storagetype(ElT)(zeros(default_datatype(ElT), dim))
```
mtfishman, line 62: I guess this will now use `generic_zeros`?

NDTensors/src/fill.jl
```diff
@@ -1,14 +1,54 @@
...
   StoreT::Type{<:Dense{ElT,VecT}}, dim::Integer
 ) where {VecT<:AbstractArray{ElT}} where {ElT<:Number}
   return StoreT(randn(ElT, dim))
 end
 
-function zeros(datatype::Type{<:AbstractArray{ElT}}, dim::Integer=0) where {ElT<:Number}
-  return fill!(datatype(undef, dim), zero(ElT))
+function generic_zeros(
+  StoreT::Type{<:Dense{ElT,DataT}}, dim::Integer=0
+) where {DataT<:AbstractArray{ElT}} where {ElT}
```
emstoudenmire, line 37: Don't think you need ElT and `where {ElT}` here and can just use `<:AbstractArray`

NDTensors/src/dense/dense.jl
kmp5VT: I think either calling `generic_zeros` with datatype or storagetype here would be fine. My thought was that someone somewhere else might just want to do something like
```
Function f(eltype::Type{<:Number>}, datatype::Type{<:AbstractArray{eltype}})
DenseDataStruct = Dense{eltype, datatype}
d = generic_zeros(DenseDataStruct)
### Do work with d
end
```

NDTensors/src/dense/dense.jl
kmp5VT: Yes it should, thanks!

NDTensors/src/dense/dense.jl
mtfishman: (Same for `real` above.)

NDTensors/src/fill.jl
```diff
@@ -0,0 +1,64 @@
+# This is a file to write generic fills for NDTensors.
+#  This includes random fills, zeros, ...
+
+function generic_randn(
+  StoreT::Type{<:Dense{ElT,DataT}}, dim::Integer=0
+) where {DataT<:AbstractArray{ElT}} where {ElT}
+  data = generic_randn(DataT, dim)
```
mtfishman, line 7: I would drop the `where` and use:
```julia
data = generic_randn(datatype(StoreT), dim)
```

NDTensors/src/dense/dense.jl
kmp5VT: Yes, this is updated!

NDTensors/src/fill.jl
```diff
@@ -0,0 +1,64 @@
...
+  return StoreT(data)
+end
+
+function generic_randn(StoreT::Type{<:Dense}, dim::Integer=0)
```
mtfishman, line 11: Isn't this function the same as the one above?

mtfishman: Oops forgot that this is used to distinguish `Dense` when the eltype/datatype are specified or not specified.

NDTensors/src/default_storage.jl
mtfishman: For example when we incorporate indices we might be:
```julia
functino default_storagetype(datatype::Type{<:AbstractArray{ElT}}, inds::Tuple) where {ElT}
  return set_datatype(default_storagetype(inds), datatype)
end
```
where `default_storagetype(::Tuple)` then analyzes the indices for block indices or dense indices and returns `BlockSparse` or `Dense`.

mtfishman: Then this case could be a special case where:
```julia
function default_storagetype(datatype::Type{<:AbstractArray{ElT}}) where {ElT}
  return default_storagetype(datatype, ())
end
```
and empty indices defaults to `Dense`.

NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,14 @@
...
+
+#currently similartype does not work for cuvector but the goal is for it to work with
```
mtfishman, line 8: Is it sufficient to add a definition:
```julia
buffertype(arraytype::Type{<:CuArray{<:Any,<:Any,B}}) where {B} = B
similartype(arraytype::Type{<:CuArray}, eltype::Type) = CuArray{eltype,ndims(arraytype),buffertype(arraytype)}
```
to handle that case? Same question to you @emstoudenmire since I know you hit a similar issue.

Also note that `CuArray` has an extra type parameter beyond `Array` which is a buffer type: https://github.com/JuliaGPU/CUDA.jl/blob/v3.13.1/src/array.jl#L25 which already speaks to why this in general needs to be overloaded for different `AbstractArray` types.

NDTensors/src/dense/dense.jl
```diff
@@ -3,70 +3,101 @@
...
+  return Dense{ElT,DataT}((dim,))
+end
 
 Dense(::Type{ElT}, dim::Integer) where {ElT} = Dense{ElT}(dim)
 
-Dense(x::ElT, dim::Integer) where {ElT<:Number} = Dense(fill(x, dim))
-
-Dense(dim::Integer) = Dense(Float64, dim)
+function Dense(ElT::Type{<:Number}, ::UndefInitializer, dim::Integer)
+  return Dense{ElT,default_datatype(ElT)}(undef, (dim,))
```
mtfishman, line 82: I would use `set_datatype(Dense, default_datatype(ElT))(undef, (dim,))`.

Though maybe better to just use:
```julia
Dense(eltype::Type{<:Number}, ::UndefInitializer, dim::Integer) = Dense(eltype, undef, (dim,))
```
and defer the logic to:
```julia
Dense(eltype::Type{<:Number}, ::UndefInitializer, dim::Tuple)
```

NDTensors/src/dense/dense.jl
```diff
@@ -3,70 +3,101 @@
...
+end
 
-Dense(::Type{ElT}, ::UndefInitializer, dim::Integer) where {ElT} = Dense{ElT}(undef, dim)
+Dense(x::Number, dim::Integer) = Dense(fill!(default_datatype(typeof(x))(undef, dim), x))
 
-Dense(::UndefInitializer, dim::Integer) = Dense(Float64, undef, dim)
+Dense(dim::Integer) = Dense(default_eltype(), dim)
 
-Dense{ElT}() where {ElT} = Dense(ElT[])
 Dense(::Type{ElT}) where {ElT} = Dense{ElT}()
```
mtfishman, line 89: Should this be specialized to `Number`, i.e.:
```julia
Dense(eltype::Type{<:Number}) = Dense{eltype}()
```

NDTensors/src/fill.jl
```diff
@@ -0,0 +1,64 @@
...
+  data = DataT(undef, dim)
+  for i in 1:dim
+    @inbounds data[i] = randn(ElT)
+  end
+  return data
+end
+
+function generic_randn(DataT::Type{<:AbstractArray}, dim::Integer=0)
+  return generic_randn(DataT{default_eltype()}, dim)
+end
```
mtfishman, line 25:
```julia
function generic_randn(datatype::Type{<:AbstractArray}, dim::Integer=0)
  return generic_randn(similartype(datatype, default_eltype()), dim)
end
```

NDTensors/src/fill.jl
mtfishman: Or just drop `<:AbstractArray` since that is a constraint already put on the `Dense` type itself.

NDTensors/src/fill.jl
```diff
@@ -0,0 +1,64 @@
...
+function randn(
+  StoreT::Type{<:Dense{ElT,VecT}}, dim::Integer
+) where {VecT<:AbstractArray{ElT}} where {ElT<:Number}
+  return StoreT(randn(ElT, dim))
+end
+
+function generic_zeros(
+  StoreT::Type{<:Dense{ElT,DataT}}, dim::Integer=0
+) where {DataT<:AbstractArray{ElT}} where {ElT}
+  data = generic_zeros(DataT, dim)
```
mtfishman, line 38: I would use `datatype(storagetype)` instead of `DataT` (also change `StoreT` to `storagetype`, I've preferred that style but I know that's inconsistent throughout NDTensors right now).

NDTensors/src/fill.jl
```diff
@@ -0,0 +1,64 @@
...
+function generic_zeros(StoreT::Type{<:Dense}, dim::Integer=0)
+  return generic_zeros(StoreT{default_eltype(),default_datatype(default_eltype())}, dim)
+end
+
+function generic_zeros(DataT::Type{<:AbstractArray{ElT}}, dim::Integer=0) where {ElT}
+  return fill!(DataT(undef, dim), zero(ElT))
+end
+
+function generic_zeros(DataT::Type{<:AbstractArray}, dim::Integer=0)
+  return generic_zeros(DataT{default_eltype()}, dim)
```
mtfishman, line 51: `set_eltype(datatype, default_eltype())` (I think other places I've suggested `similartype`/`adapt_type` here but I think we could just define a `set_eltype` for this)

NDTensors/src/fill.jl
kmp5VT: So I tried rewriting this as 
```
  function generic_zeros( StoreT::Type{<:Dense{Any, DataT}}, dim::Integer = 0) where {DataT<:AbstractArray{Any}}
  ``` 
But you cannot access this function, if you call say `generic_zeros(Dense{Float32, Vector{Float32}}, 20)` the only function which is called is 
 ```
function generic_zeros(StoreT::Type{<:Dense}, dim::Integer = 0)
```
And fails. It would be possible to make a function. 
```
function generic_zeros(StoreT::Type{<:Dense{DataT}}, dim::Integer = 0) where {DataT<:AbstractArray}
  data = default_zeros(DataT{default_eltype()}, dim)
  return Dense(data)
end
```
But obviously the provided argument is not a "proper" dense type

mtfishman: What about:
```julia
function generic_zeros(
  StoreT::Type{<:Dense{<:Any,DataT}}, dim::Integer=0
) where {DataT}
  [...]
end
```
`Dense{Any,DataT}` is taken literally as a `Dense` with an element type of `Any`.

kmp5VT: Oh, yes `<:Any` worked. I thought I was doing something wrong. Thanks!

kmp5VT: Hmm actually there was another problem with `<:Any` It can easily get passed by for a less specific constructor (i.e. it can fall into a more generic constructor) but it seems to work better if I specify `<:Number` instead of `<:Any`

NDTensors/src/dense/dense.jl
mtfishman: Also I guess we should drop things like `<:Number`.

NDTensors/src/fill.jl
mtfishman: What's the particular issue, i.e. what inputs and then what function accidentally gets called? That seems strange and may indicates another function has a `<:Number` constraint that should be removed.

NDTensors/src/dense/dense.jl
kmp5VT: I've matched `real` and `complex`

ITensorGPU/src/cuitensor.jl
kmp5VT: Yes we can, and I have updated the code

NDTensors/src/dense/dense.jl
```diff
@@ -3,70 +3,101 @@
...
-function Dense(data::VecT) where {VecT<:AbstractArray{ElT}} where {ElT}
-  return Dense{ElT,VecT}(data)
+function Dense{DataT}(::UndefInitializer, inds::Tuple) where {DataT<:AbstractArray}
+  return default_storagetype(DataT)(undef, inds)
 end
 
-function Dense(data::Array{ElT}) where {ElT}
-  return Dense{ElT,Vector{ElT}}(vec(data))
+function Dense{DataT}(x::Number, dim::Integer) where {DataT<:AbstractArray}
+  return Dense(fill!(DataT{typeof(x)}(undef, dim), x))
```
mtfishman, line 53: In general you should avoid using `DataT{...}`, i.e. explicitly listing type parameters of an unknown type with that notation, since that won't be generic to all data types, for example if they have other type parameters like a tensor order or buffer type. Instead use generic functions, i.e.: `similartype(DataT, typeof(x))` (or `adapt_type` or `set_eltype` or whatever we ultimately decide is the best practice).

NDTensors/src/fill.jl
mtfishman: I guess we should leave `set_eltype` and `adapt_type` for a future PR so for now we should use `similartype` for this kind of thing, i.e. replacing or setting the element type of a data type or storage type.

NDTensors/src/fill.jl
```diff
@@ -0,0 +1,64 @@
...
+end
+
+function zeros(datatype::Type{<:AbstractArray}, dim::Integer=0)
+  return zeros(datatype{default_eltype()}, dim)
```
mtfishman, line 55: Again, we should use `similartype` here.

NDTensors/src/default_storage.jl
kmp5VT: Okay I think I may have figured out the answer to this `{ElT}` problem. We can use the `set_eltype_if_unspecified` function to set the eltype of the specific vectors. Then we only need one 
``` 
Function default_storagetype(datatype::Type{<:AbstractArray} inds::Tuple)
  datatype = set_eltype_if_unspecified(datatype, default_eltype())
  if typeof(inds) == Integer || length(inds) == 0
    return Dense{eltype(datatype), datatype)
  else
    return #blocksparse 
 end
 ```

Comment (mtfishman) [reply NDTensors/src/default_storage.jl]: Yeah, that's seem like a good pattern to make a function `set_eltype_if_unspecified` (I guess that isn't defined right now, correct)? That should be useful in other places as well. We could also have a shorthand:
```julia
set_default_eltype_if_unspecified(datatype) = set_eltype_if_unspecified(datatype, default_eltype())
```

I could imagine this could be based on a special adaptor as part of the future `adapt_type` function, i.e.:
```julia
set_eltype_if_unspecified(datatype::Type, eltype::Type) = adapt_type(SetIfUnspecifiedAdaptor(eltype), datatype)
```

However to clarify we should find a better design based on something like `default_storagetype(inds::Tuple)` which dispatches on the `inds` rather than `if` statements to choose the storage type.

NDTensors/src/dense/dense.jl
kmp5VT: I would imagine it would be a good idea to throw an error in general if length(inds) != ndims(DataT) because it is possible for us to cut inds up to make something the right size but we can't in general know what the right inds will be

mtfishman: Yeah I think that's right. The `DataT` should get fixed to the correct `ndims` in higher up functions based on `inds` (or something else) before it gets to this function, say from constructors that don't specify the data type like `Dense{Float64}(undef, (3, 3, 3))`. That could generically get handled by defining `default_datatype(inds::Tuple)` for example.

NDTensors/src/default_storage.jl
kmp5VT: `set_eltype_if_unspecified` is defined in `NDTensors/src/adapt.jl`.  I will write the `set_eltype_if_unspecified(datatype) = set_eltype_if_unspecified(datatype, default_eltype())` function but I don't know yet if it will be useful. We will see!

I agree with the dispatches. I don't really know what Dense vs Blocksparse inds look like so I don't know how to implement those different dispatches.

mtfishman: Ah I forgot that I had already started to define `set_eltype_if_unspecified`. Looks like a good start but we'll need to generalize that functionality a bit.

No worries, we can handle that in a future PR that extends this stuff to `BlockSparse`. The designs we're developing here should help a lot once we get to that!

kmp5VT: I have already pushed some updates to `set_eltype_if_unspecified` to work for a `AbstractArray` and to use `similartype`

NDTensors/src/dense/dense.jl
```diff
@@ -3,73 +3,145 @@
...
+  return Dense{ElT,typedDataT}(similar(typedDataT, dim(inds)))
 end
 
-function Dense(data::Array{ElT}) where {ElT}
-  return Dense{ElT,Vector{ElT}}(vec(data))
+function Dense{ElR,DataT}(data::AbstractArray) where {ElR,DataT<:AbstractArray}
+  if eltype(data) != ElR
+    data = similartype(typeof(data), ElR)(data)
+  end
+  return Dense{ElR,DataT}(data)
```
mtfishman, line 55: It seems like this would be better handled like:
```julia
function Dense{ElR,DataT}(data::AbstractArray) where {ElR,DataT}
  return Dense{ElR,DataT}(convert(DataT, data))
end
```
`if`-statements with type comparisons like that are basically always better handled through dispatch.

Also, for the version you wrote, if `eltype(data) == ElR`, would it lead to an infinite loop?

NDTensors/src/dense/dense.jl
```diff
@@ -3,73 +3,145 @@
...
 end
 
-function Dense{ElR}(data::AbstractArray{ElT}) where {ElR,ElT}
-  return ElT == ElR ? Dense(data) : Dense(ElR.(data))
+function Dense{DataT}() where {DataT<:AbstractArray}
+  typedDataT = set_eltype_if_unspecified(DataT, default_eltype())
+  ElT = eltype(typedDataT)
+  return Dense{ElT,typedDataT}()
```
mtfishman, line 61: This seems strange to me, I don't think we want to ever treat the first type parameter of `Dense` as the data type, only as the element type.

mtfishman: Also we shouldn't be changing the `DataT` if it is specified. What's the purpose of this function?

NDTensors/src/dense/dense.jl
```diff
@@ -3,73 +3,145 @@
...
+  typedDataT = set_eltype_if_unspecified(DataT, typeof(x))
+  ElT = eltype(typedDataT)
+  return Dense{ElT,typedDataT}(fill!(similar(typedDataT, dim(inds)), x))
+end
 
-Dense(::Type{ElT}, dim::Integer) where {ElT} = Dense{ElT}(dim)
+# This function is ill-defined. It cannot transform a complex type to real...
+function Dense{ElR}(data::AbstractArray{ElT}) where {ElR,ElT}
+  return Dense(similartype(typeof(data), ElR)(data))
+end
```
mtfishman, line 85: Note that we should generally use `convert` for this since it doesn't copy data if it doesn't need to, and also I would call `Dense{ElR}` inside with the pattern that the constructors should keep getting more specific, i.e.:
```julia
function Dense{ElR}(data::AbstractArray) where {ElR}
  return Dense{ElR}(convert(similartype(typeof(data), ElR), data))
end
```

NDTensors/src/dense/dense.jl
```diff
@@ -3,73 +3,145 @@
...
-# Random constructors
-#
+function Dense(ElT::Type{<:Number}, ::UndefInitializer, dim::Integer)
+  return Dense{ElT,default_datatype(ElT)}(undef, (dim,))
+end
 
-function randn(::Type{StoreT}, dim::Integer) where {StoreT<:Dense}
-  return Dense(randn(eltype(StoreT), dim))
+function Dense(::UndefInitializer, dim::Integer)
+  return Dense{default_eltype(),default_datatype(default_eltype())}(undef, (dim,))
```
mtfishman, line 122: Maybe a slightly simpler way to write this would be:
```julia
  datatype = default_datatype()
  return Dense{eltype(datatype),datatype}(undef, (dim,))
```

NDTensors/src/dense/dense.jl
mtfishman: If you want to define versions where the data type is specified but the element type isn't, the way to do that is:
```julia
function Dense{<:Any,DataT}()
  return Dense{eltype(DataT),DataT}()
end
```
which would be called like `Dense{<:Any,Vector{Float32})()`. It's a bit verbose but the best we have since type parameters are positional.

NDTensors/src/dense/dense.jl
mtfishman: Of course there should then be a version:
```julia
function Dense{ElR}(data::AbstractArray{ElR}) where {ElR}
  # [...]
end
```
which catches when the element type is already correct, otherwise we end up in an infinite loop.

mtfishman:
@kmp5VT I just noticed the ITensors.jl tests are passing, that's great! Seems like once you get the ITensorGPU.jl tests passing and we do some more rounds of reviewing and cleaning up the code we could merge this.

ITensorGPU/src/cuitensor.jl
```diff
@@ -12,9 +12,10 @@ cuITensor(inds::Index...) = cuITensor(IndexSet(inds...))
 
 cuITensor() = ITensor()
 function cuITensor(x::S, inds::IndexSet{N}) where {S<:Number,N}
-  dat = CuVector{float(S)}(undef, dim(inds))
-  fill!(dat, float(x))
-  return ITensor(Dense{S}(dat), inds)
+  #dat = CuVector{float(S)}(undef, dim(inds))
+  #fill!(dat, float(x))
+  d = Dense{CuVector}(x, dim(inds))
```
mtfishman, line 17: Would the plan be for this to be:
```julia
generic_fill(default_storagetype(CuVector{typeof(x)}), x)
```
?

NDTensors/src/dense/dense.jl
kmp5VT: @mtfishman In this function I don't actually change the datatype. I don't know if this is something that would happen often but my thought was that you might want to run a program completely with a `default_eltype` but change the datatype say from `Vector` to `DistibutedVector`. Then you could construct with `Dense{ArrayType}()` where Arraytype is specified. 
Also this code does not actually change the datatype. What it does is cover the case if you call `Dense{Vector}` vs `Dense{Vector{Float64}}`. I updated the `set_eltype_if_unspecified` to take an `AbstractArray`. It does nothing if `eltype` is specified or sets the eltype to the a provide type if unspecified (in general this is `default_eltype()`). This way we don't need to have `{DataT<:AbstractArray{ElT}} where {ElT}`

NDTensors/src/dense/dense.jl
kmp5VT: Okay, I have updated this. Thanks!

NDTensors/src/dense/dense.jl
mtfishman: There are two issues:

1. If you call `Dense{T}([...])`, `T` should never be interpreted as the `datatype`, only as the `eltype`. We should consider this a strict rule about constructor calls. For specifying the `datatype` and not the `eltype`, we should use the notation `Dense{<:Any,DataT}([...])`. I would argue that this should only ever return an object of type `Dense{eltype(DataT),DataT}` (maybe in some cases it is better to infer the element type from an argument input but in general I think `eltype(DataT)` is best).
2. If the behavior is that `Dense{<:Any,Vector}()` returns an object of type `Dense{<:Any,Vector{Float64}}`, I consider that changing the data type, since `datatype(Dense{<:Any,T}()) != T` (I would take a strict definition that the data type is the result of calling `datatype`, so you can't ignore the element type).

As a principle (I would say a rare case when there is a very strict rule to follow), type parameter inputs of constructors should be taken very literally, unless they are completely unspecified. So for any type, we should have constructors that follow the pattern:
```julia
struct T{A,B}
  # Some fields
end
T{A,B}(args...) = # Must return an object of type `T{A,B}`
T(args...) = T{infer_A(args...),infer_B(args...)}(args...)
T{A}(args...) = T{A,infer_B(A, args...)}(args...)
T{<:Any,B}(args...) = T{infer_A(B, args...),B}(args...)
```
For behavior that is less strict about type inputs, we should rely on functions that perform conversions (such as `set_eltype` or `adapt`) or passing types as arguments, not as type parameters.

NDTensors/src/dense/dense.jl
mtfishman: Also as I lay out in my comment below, I don't think we should ever treat the first type parameter as the data type (it should always be interpreted as the element type).

NDTensors/src/dense/dense.jl
kmp5VT: 1. I can modify the the `Dense{DataType}` constructors to match these principles.
However, I just want to point out that from my constructor tree for dense it is not possible for `Dense{T}([...])` to infer `T` as the `datatype` because there is no `Dense{DataT<:AbstractArray}(data::AbstractArray)` constructor, only a `Dense{ElT}(data::AbstractArray) where {ElT}` constructor. This is currently some functionality of the code as its written
```
julia> NDTensors.Dense{Vector{Float32}}()
0-element NDTensors.Dense{Float32, Vector{Float32}}
julia> NDTensors.Dense{Vector{Float32}}([1,2,3])
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type Vector{Float32}
julia> NDTensors.Dense{Vector{Float32}}([[1,2,3],[2],[3,4,3]])
typeof(d) = Vector{Vector{Float32}}
3-element NDTensors.Dense{Vector{Float32}, Vector{Vector{Float32}}}:
 [1.0, 2.0, 3.0]
 [2.0]
 [3.0, 4.0, 3.0]
 ``` 
 Any time actual data is provided to `Dense` it will consider it the eltype because `ElT` is general and not required to be a number. 
 
 2.
 The behavior of `Dense{<:Any, Vector}()` would return an object of `Dense{eltype(Dense), Vector{eltype(Dense)}`. What I am trying to do in these constructors is be more stringent to ensure that `ElT = eltype(DataT) = eltype(data)` and `typeof(data) = DataT`. There is an example in the ITensors tests where previously the code constructed a dense via `Dense{ComplexF64, Vector{ComplexF64})(Vector{Float64}([...])`. Also I found that we can do something like this still 
 ```
 julia> NDTensors.Dense{Float32, Vector{Float64}}(Vector{Float64}([2,3,4]))
3-element NDTensors.Dense{Float32, Vector{Float64}}:
 2.0
 3.0
 4.0
 ``` 
I know that these cases aren't "normal behavior" for us or anyone using ITensors but I just to ensure that it is not possible to construct such ambiguous objects because they could cause problems down the line.

Comment (mtfishman) [reply NDTensors/src/dense/dense.jl:61]: Right, I can see that in principle there are cases when we can unambiguously decide that for a call `Dense{T}(args...)`, `T` _could_ be interpreted as the `datatype`, but it is a question of what we _should_ do, and I think it is playing way to fast-and-loose with constructor calls to ever interpret `T` (the first type parameter) as the `datatype`.

I'm confused by you statement that `Dense{<:Any, Vector}()` should return an object of type `Dense{eltype(Dense),Vector{eltype(Dense)}`, I'm not really sure what that means. I would vote for a very general definition:
```julia
Dense{<:Any,DataT}(args...) where {DataT} = Dense{eltype(DataT},DataT}(args...)
```
in which case we always have the property:
```julia
datatype(Dense{<:Any,T}(args...)) == T
eltype(Dense{<:Any,T}(args...)) == eltype(T)
```
Then, by that definition, `Dense{<:Any,Vector}(args...)` returns an object of type `Dense{Any,Vector}`. Even if we could infer a more specific `eltype` than `Any` based on the `args...`, I don't think we should, and in general we should opt for taking the specified type parameters very literally and also keeping them consistent with each other.

That's helpful context that you are trying to deal with those types of corner cases where the type parameters are inconsistent with each other, and it is worth thinking about how to deal with that. I would say that for now, we just try to ensure that our own code always does the correct/safest thing (i.e. we should never generate a type `Dense{Float32,Vector{Float64}}` in our code, for example by using rules I lay out above). It seems like we could catch some of those type parameter consistency mistakes in the inner constructor, for example we could add a check:
```julia
function Dense{ElT,DataT}(data)
  @assert ElT == eltype(DataT)
  return new{ElT,DataT}(convert(DataT, data))
end
```

mtfishman: One way to think about this is that we could have thought about designing the `Dense` type with just one type parameter:
```julia
struct Dense{DataT}
  data::DataT
end
```
in which case the definition of `eltype` would be determined directly from the `DataT` parameter:
```julia
datatype(::Dense{DataT}) where {DataT} = DataT
eltype(x::Dense) = eltype(datatype(x))
```
We could think of the `ElT` parameter in `Dense` as just a convenient way of working with the `eltype` of `DataT` for dispatch, constructor calls, etc. but for all intents and purposes it is really just `eltype(DataT)` by definition.

NDTensors/src/adapt.jl
```diff
@@ -36,5 +36,13 @@ to_vector_type(arraytype::Type{<:AbstractVector}) = arraytype
 to_vector_type(arraytype::Type{Array}) = Vector
 to_vector_type(arraytype::Type{Array{T}}) where {T} = Vector{T}
 
-set_eltype_if_unspecified(arraytype::Type{Vector{T}}, eltype::Type) where {T} = arraytype
-set_eltype_if_unspecified(arraytype::Type{Vector}, eltype::Type) = Vector{eltype}
+function set_eltype_if_unspecified(
+  arraytype::Type{<:AbstractArray{T,N}}, eltype::Type=default_eltype()
```
mtfishman, line 40: I guess we don't need the type parameters `{T,N}` here.

NDTensors/src/dense/dense.jl
```diff
@@ -3,73 +3,111 @@
...
-Dense(x::ElT, dim::Integer) where {ElT<:Number} = Dense(fill(x, dim))
+function Dense{ElT}(dim::Integer) where {ElT}
+  return Dense{ElT,default_datatype(ElT)}(dim)
+end
 
-Dense(dim::Integer) = Dense(Float64, dim)
+Dense{ElT}() where {ElT} = Dense{ElT,default_datatype(ElT)}()
 
-Dense(::Type{ElT}, ::UndefInitializer, dim::Integer) where {ElT} = Dense{ElT}(undef, dim)
+function Dense(data::DataT) where {DataT<:AbstractArray{<:Any,1}}
```
mtfishman, line 65: You can use `DataT<:AbstractVector`. Better yet:
```julia
function Dense(data::AbstractVector)
  return Dense{eltype(data)}(data)
end
```

NDTensors/src/dense/dense.jl
```diff
@@ -3,73 +3,111 @@
...
+  return Dense{eltype(DataT),DataT}(data)
+end
 
-Dense(::UndefInitializer, dim::Integer) = Dense(Float64, undef, dim)
+function Dense(data::DataT) where {DataT<:AbstractArray{<:Any,N}} where {N}
+  #println("Warning: Only vector based datatypes are currenlty supported by Dense. The data structure provided will be vectorized.")
+  vecData = vec(data)
+  return Dense{eltype(vecData),typeof(vecData)}(vecData)
+end
```
mtfishman, line 73: This should do the same thing:
```julia
function Dense(data::AbstractArray)
  return Dense(vec(data))
end
```
and just call the function above.

NDTensors/src/dense/dense.jl
```diff
@@ -3,73 +3,111 @@
...
+## End Dense initializers
+
+setdata(D::Dense, ndata) = Dense(ndata)
+
 copy(D::Dense) = Dense(copy(data(D)))
 
-Base.real(::Type{Dense{ElT,Vector{ElT}}}) where {ElT} = Dense{real(ElT),Vector{real(ElT)}}
+#This is getting closer but is still broken...
+function Base.real(T::Type{Dense})
+  return set_datatype(T, similartype(datatype(T), complex(eltype(T))))
```
mtfishman, line 105: Looks like this should be `real(eltype(T))`. Also `Type{<:Dense}` in the signature.

NDTensors/src/default_storage.jl
```diff
@@ -0,0 +1,23 @@
+## This is a fil which specifies the default storage type provided some set of parameters
+## The parameters are the element type and storage type
+default_datatype(eltype::Type) = Vector{eltype}
+default_datatype() = Vector{default_eltype()}
```
mtfishman, line 4: I would write this as:
```julia
default_datatype() = default_datatype(default_eltype())
```
Even though these are short functions, I find it is best to think about "implementation functions" that are at the bottom that do the work, and processing functions that process the inputs with the goal of calling the "implementation functions". Here it is useful since there will only be one point in the code where `Vector` is actually defined as the default data type (and therefore would be easier to customize, if we decide on that direction).

NDTensors/src/adapt.jl
kmp5VT: We need at least the T parameter for the dispatch system to choose between the two `set_eltype_if_unspecified` functions. Else it will always convert `arraytype` to `arraytype{eltype}`

mtfishman: Right, good point.

NDTensors/src/dense/dense.jl
```diff
@@ -3,104 +3,135 @@
...
-function Dense(data::Array{ElT}) where {ElT}
-  return Dense{ElT,Vector{ElT}}(vec(data))
+function Dense{ElR,DataT}(data::AbstractArray) where {ElR,DataT<:AbstractArray}
+  data = convert(DataT, data)
+  return Dense{ElR,DataT}(data)
 end
 
+# This function is ill-defined. It cannot transform a complex type to real...
 function Dense{ElR}(data::AbstractArray{ElT}) where {ElR,ElT}
   return ElT == ElR ? Dense(data) : Dense(ElR.(data))
```
mtfishman, line 45: Should be written in terms of `convert`?

kmp5VT: Oh yes, this accidentally got converted back to the old version which is what was on your branch

ITensorGPU/src/tensor/culinearalgebra.jl
```diff
@@ -83,8 +83,8 @@ function LinearAlgebra.svd(T::CuDenseTensor{ElT,2,IndsT}; kwargs...) where {ElT,
   Sdata = CUDA.zeros(ElT, dS * dS)
   dsi = diagind(reshape(Sdata, dS, dS), 0)
   Sdata[dsi] = MS
-  S = tensor(Dense(Sdata), Sinds)
-  V = tensor(Dense(vec(MV)), Vinds)
+  S = Tensor(Dense(Sdata), Sinds)
+  V = Tensor(Dense(vec(copy(MV))), Vinds)
```
mtfishman, line 87: These should be lower case `tensor` to avoid unnecessary copying.

NDTensors/src/dense/dense.jl
mtfishman: Still seems like we need to constrain this, otherwise it would be ambiguous with calls like `Dense(Vector{Float32})`.

NDTensors/src/fill.jl
```diff
@@ -0,0 +1,54 @@
...
+end
+
+##TODO replace randn in ITensors with generic_randn
+## and replace zeros with generic_zeros
+function generic_zeros(
+  StoreT::Type{<:Dense{ElT,DataT}}, dim::Integer=0
+) where {DataT<:AbstractArray,ElT}
+  @assert ElT == eltype(DataT)
+  data = generic_zeros(DataT, dim)
+  StoreT = NDTensors.set_datatype(StoreT, typeof(data))
```
mtfishman, line 38: I don't think you should need `NDTensors.` here.

NDTensors/src/fill.jl
mtfishman: I see this case explicitly uses `DataT` to indicate to dispatch that the data type is specified so it probably makes more sense to just use ` DataT` and not `datatype(StoreT)`/`datatype(storagetype)` in this case.

NDTensors/src/fill.jl
```diff
@@ -0,0 +1,54 @@
...
+  end
+  return data
+end
+
+##TODO replace randn in ITensors with generic_randn
+## and replace zeros with generic_zeros
+function generic_zeros(
+  StoreT::Type{<:Dense{ElT,DataT}}, dim::Integer=0
+) where {DataT<:AbstractArray,ElT}
+  @assert ElT == eltype(DataT)
```
mtfishman, line 36: Do we need this check here? Can't this be handled by the inner constructor for `Dense`?

NDTensors/src/fill.jl
mtfishman: Also, I'm confused by this line, doesn't `StoreT` already have a data type of type `DataT`, which is the same type as `data`? Or is this catching some cases where `data` isn't quite the same as `DataT`?

kmp5VT: Its catching the case when they aren't quite the same. This problem case came when you provide a `SubArray{Vector...}}` When you call `generic_zeros` it returns a `Vector` because `generic_zeros` uses `similar` which will unwrap types like we talked about yesterday. So then `DataT` doesn't match `typeof(data)` and I assumed that we shouldn't have the wrapped type as the DataT so I "fix" the  problem here

mtfishman: Ok I think that makes sense, it would be weird to want to make a wrapper array filled with zeros. We could consider making that case error since we wouldn't be taking the input storage type very literally (like we try to do with constructor calls) but unwrapping seems like a reasonable behavior in this case.

NDTensors/src/fill.jl
kmp5VT: If you remove the assert, this is what happens
```
julia> NDTensors.generic_zeros(NDTensors.Dense{Float32, Vector{Float64}}, 20)
20-element NDTensors.Dense{Float64, Vector{Float64}}:
```
My argument is that, this isn't correct because theres no way to know if `ElT` or `eltype(DataT)` is the correct type.

mtfishman: Yes I agree we don't want that behavior, I was just hoping that we could catch these kinds of mistakes farther down in an inner constructor so that we wouldn't have to spread checks like that around the codebase. But I see that because of the line:
```julia
StoreT = set_datatype(StoreT, typeof(data))
```
`StoreT` would start with contradictory element types but then `set_datatype` would fix it to the eltype of `data`/`DataT`. Maybe it should be `set_datatype` that checks for inconsistent element types?

It doesn't have to be dealt with in this PR, but we can think about good practices for catching inconsistent type parameters in future PRs. Something we could think about is having a catch-all generic function `check_consistent_type_parameters` that defines checks like:
```julia
function check_consistent_type_parameters(storagetype::Type{<:TensorStorage})
  @assert eltype(storagetype) === eltype(datatype(storagetype))
end
```

NDTensors/src/fill.jl
mtfishman: (Another case of specializing for dispatch.)

ITensorGPU/src/tensor/culinearalgebra.jl
```diff
@@ -84,7 +84,7 @@ function LinearAlgebra.svd(T::CuDenseTensor{ElT,2,IndsT}; kwargs...) where {ElT,
   dsi = diagind(reshape(Sdata, dS, dS), 0)
   Sdata[dsi] = MS
   S = tensor(Dense(Sdata), Sinds)
-  V = tensor(Dense(vec(MV)), Vinds)
```
mtfishman, line 87: Also I added a `copy(MV)` up above so it isn't needed here.

PR #1044, opened 2023-01-24
[ITensorGPU] Try newer version of Ubuntu for GPU CI tests

mtfishman:
# Description

Try fixing ITensorGPU CI tests.

See comments in #1035.

### DIFF
--- a/jenkins/Dockerfile
+++ b/jenkins/Dockerfile
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:11.4.2-cudnn8-devel-ubuntu20.04
+FROM nvidia/cuda:12.0.0-devel-ubuntu20.04
 
 # julia
 RUN apt-get update && \
--- a/jenkins/Dockerfile
+++ b/jenkins/Dockerfile
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:11.4.2-cudnn8-devel-ubuntu20.04
+FROM nvidia/cuda:12.0.0-devel-ubuntu22.04
 
 # julia
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


mtfishman:
`nvidia/cuda:11.8.0-cudnn8-devel-ubuntu22.04` also fails with `E: Unable to locate package libcutensor-dev`.

dylex:
Yeah I was just looking at this -- it looks like nvidia has not released any libcutensor packages for ubuntu22.04... https://developer.nvidia.com/cutensor/downloads

mtfishman:
That would explain it, that's unfortunate.

dylex:
But you can just substitute `ubuntu20.04` on those instead (`nvidia/cuda:12.0.0-devel-ubuntu20.04`)

mtfishman:
Using `nvidia/cuda:12.0.0-devel-ubuntu20.04` fixed the build issue but when the tests run I see the same error `This functionality is unavailabe as CUTENSOR is missing.` For some reason even though `libcutensor-dev` is being installed, `CUDA.jl` can't find it. Maybe it is some combination of how `CUDA.jl` is looking for the library (an automated process defined in https://github.com/JuliaGPU/CUDA.jl/blob/v3.13.1/deps/bindeps.jl) and something that changed in the upgrade to CUDA 12?

@kshyatt any suggestions?

mtfishman:
Oh wait, I was still looking at the CI logs from a previous commit that was using `nvidia/cuda:11.4.2-cudnn8-devel-ubuntu20.04`. Looks like `nvidia/cuda:12.0.0-devel-ubuntu20.04` ran successfully! Thanks for you help @dylex.

PR #1046, opened 2023-01-25
[ITensors] Rewrite `ProjMPOSum` and `DiskProjMPOSum` more generically in terms of `AbstractSum`

emstoudenmire:
# Description

Defines a generic type `ProjSum{T}` and defines `ProjMPOSum` and `DiskProjMPOSum` in terms of this type. 

# How Has This Been Tested?

Uses unit test from previous PR testing that disk functionality works for `ProjMPOSum`.

# Checklist:


emstoudenmire:
I think the tests are actually passing, and there was just some issue (timeout or something else) for one of the test runners, but the rest passed.

emstoudenmire:
I removed the docstrings for now, since in this refactor they refer to generic code, so it's less clear to me whether they should mention the concrete `ProjMPOSum` type. I can reintroduce them easily though.

src/mps/projmposum.jl
```diff
@@ -23,115 +26,63 @@ sum Σⱼ on the left):
      o--o--o-      -o--o--o--o--o--o |psi>
 ```
 """
-mutable struct ProjMPOSum
-  pm::Vector{ProjMPO}
-end
 
-copy(P::ProjMPOSum) = ProjMPOSum(copy.(P.pm))
+copy(P::ProjSum{T}) where {T} = ProjSum{T}(copy.(P.pm))
```
mtfishman, line 30: It should work without the `where {T}`:
```julia
copy(P::ProjSum) = ProjSum(copy.(P.pm))
```
Julia should be able to deduce the type automatically through default constructors that it makes.

src/mps/projmposum.jl
```diff
@@ -23,115 +26,63 @@ sum Σⱼ on the left):
...
   nt = noiseterm(P.pm[1], phi, dir)
   for n in 2:length(P.pm)
     nt += noiseterm(P.pm[n], phi, dir)
   end
   return nt
 end
+
+const ProjMPOSum = ProjSum{ProjMPO}
+
+ProjMPOSum(mpos::Vector{MPO}) = ProjMPOSum([ProjMPO(M) for M in mpos])
```
mtfishman, line 79: Maybe we can make this generic, i.e.:
```julia
ProjSum{T}(mpos::Vector{MPO}) where {T} = ProjSum{T}([T(M) for M in mpos])

ProjSum{T}(Ms::MPO...) where {T} = ProjSum{T}([Ms...])
```

src/mps/projmposum.jl
```diff
@@ -128,10 +103,56 @@ ProjMPOSum `P`, and `ortho` is a String which can take
...
+"""
+    disk(ps::ProjMPOSum; kwargs...)
+
+Convert a ProjMPOSum into a DiskProjMPOSum,
+which will automatically start caching most 
+stored tensors onto the hard drive.
+"""
+function disk(ps::ProjMPOSum; kwargs...)
+  return DiskProjMPOSum([disk(pm; kwargs...) for pm in ps.pm])
+end
```
mtfishman, line 156: We could also make this generic:
```julia
function disk(ps::ProjSum; kwargs...)
  return ProjSum([disk(pm; kwargs...) for pm in ps.pm])
end
```
I would also vote for dropping `disk(P::DiskProjMPOSum; kwargs...) = P` since this generic version covers that (i.e. `disk` is already a no-op for `DiskProjMPO`).

src/mps/projmposum.jl
```diff
@@ -128,10 +103,56 @@ ProjMPOSum `P`, and `ortho` is a String which can take
...
+```
+"""
+const ProjMPOSum = ProjSum{ProjMPO}
+
+"""
+A `DiskProjMPOSum` functions the same as a `ProjMPOSum`
+(see the docstring for `ProjMPOSum`) but automatically caches most of
+the tensors it stores onto the hard drive to save RAM memory.
+"""
+const DiskProjMPOSum = ProjSum{DiskProjMPO}
```
mtfishman, line 145: Given all the functions are/will be generic in terms of `ProjSum`, I don't think we even need to define this type. It shouldn't be something any user needs anyway since they can just use the `disk` function.

src/mps/projmposum.jl
```diff
@@ -128,10 +98,68 @@ ProjMPOSum `P`, and `ortho` is a String which can take
...
   end
   return nt
 end
+
+#
+# Definition of concrete, generic ProjSum type
+#
+
+mutable struct ProjSum{T} <: AbstractSum
+  terms::Vector{T}
```
mtfishman, line 114: I definitely prefer the field name `terms` here but I worry this will break code.

Also I would remove `mutable` since there is not reason for it since `Vector` is already mutable, but I also worry someone may be relying on that.

Maybe we could just go ahead and make these changes and assume it is only advanced users that might rely on these.

Finally, what about `SerialSum` in contrast with `ParallelSum` that I will define in ITensorParallel.jl?

mtfishman: (Maybe `SequentialSum` since `Folds.jl` uses `SequentialEx`.)

mtfishman: While we're at it, I think we should define generic field getter and setter functions:
```julia
terms(sum::AbstractSum) = sum.terms
set_terms(sum::AbstractSum, terms) = typeof(sum)(terms)
```
and replace the use of `P.terms` with `terms(P)` in the code in this file.

That will help with code I'm writing in `ITensorParallel.jl` for a generic `ParallelSum` type.

src/mps/projmposum.jl
```diff
@@ -128,10 +98,68 @@ ProjMPOSum `P`, and `ortho` is a String which can take
...
+"""
+    disk(ps::ProjMPOSum; kwargs...)
+
+Convert a ProjMPOSum into a DiskProjMPOSum,
+which will automatically start caching most 
+stored tensors onto the hard drive.
+"""
+function disk(ps::ProjMPOSum; kwargs...)
+  return DiskProjMPOSum([disk(pm; kwargs...) for pm in ps.terms])
+end
```
mtfishman, line 163: I think my previous comment got lost, but I think we should make this a generic `AbstractSum` function:
```julia
function disk(sum::AbstractSum; disk_kwargs...)
  return set_terms(sum, [disk(term; disk_kwargs...) for term in terms(sum)])
end
```
based on the generic `set_terms` function I propose above. That will be helpful for the new `ITensorParallel.ParallelSum` type.

src/mps/projmposum.jl
mtfishman: We can also be on the safer side and make `set_terms` part of the interface that needs to be overloaded:
```julia
set_terms(sum::AbstractSum, terms) = error("Please implement `set_terms` for the `AbstractSum` type `$(typeof(sum))`.")
```
since the definition above won't work for `AbstractSum` types with other fields (like the Folds executor object that I plan to add to `ITensorParallel.ParallelSum`).

emstoudenmire:
I made all of those changes. I have a couple of questions as well:

- what about making the function 
  ```
  SequentialSum{T}(mpos::Vector{MPO}) where {T} = SequentialSum([T(M) for M in mpos])
  ```
  generic over `MPO` also. Is there a reason not to? Then it could work for any type where it makes sense to construct a `T` from an `M`.
- does the definition of `length` and `size` here (as calling down to the length of one of the terms) really make sense anymore for this more general or abstract design? Maybe it does but I could survey the use of `length` and `size` for the other "Projected" types to see if we can maybe deprecate or "repurpose" one of these (e.g. `length`) to return the number of terms instead
- I still am wondering slightly about the fate of the docstrings here (and I'm realizing it's a question I have about generic code and docstrings more generally in Julia – in C++ it was a non-question since the docs were just hand written). My question is whether it makes more sense for the docstrings to either (1) still refer to a specific type, like ProjMPOSum even though the code below it is much more general or (2) change the docstrings to refer to the general code, but then they may become very hard for users (or me reading them sometime months) to understand since they would refer to more advanced, abstract, developer-oriented concepts

src/mps/projmposum.jl
```diff
@@ -1,48 +1,18 @@
...
-instead when the `product` method of a ProjMPOSum is invoked,
-each projected MPO in the set of MPOs is multiplied by
-the input tensor one-by-one in an efficient way.
+abstract type AbstractSum end
 
-Drawing of the network represented by a ProjMPOSum
-`P([H1,H2,...])`, showing the case of `nsite(P)==2`
-and `position!(P,psi,4)` for an MPS `psi` (note the
-sum Σⱼ on the left):
+copy(P::AbstractSum) = AbstractSum(copy.(P.terms))
```
mtfishman, line 4: This would be wrong in general, it should be: `copy(P::AbstractSum) = typeof(P)(copy.(terms(P)))` to construct a copy with the same type as the input.

emstoudenmire: Thanks – just about to push the new code & I'll fix this

mtfishman:
>     * what about making the function
>       ```
>       SequentialSum{T}(mpos::Vector{MPO}) where {T} = SequentialSum([T(M) for M in mpos])
>       ```
>       generic over `MPO` also. Is there a reason not to? Then it could work for any type where it makes sense to construct a `T` from an `M`.

I guess that's the awkward part about using such a general name, and why we need to rethink all this in a more general way using a general lazy system (i.e. https://github.com/ITensor/ITensors.jl/blob/v0.3.23/src/LazyApply/LazyApply.jl). For now I would vote for not overthinking this aspect (i.e. how generic should this type should _really_ be), and think of these types as a stepping stone to a new better system that will be in `ITensorNetworks.jl`, and this code is all still code that will be deprecated and moved to `ITensorLegacyMPS.jl`. My goal here is mostly to help clean up and simplify code in `ITensorParallel.jl` which had become bug-ridden and difficult to maintain, as well as simplifying this code to keep it easier to maintain while adding new features like the `disk` feature.

>     * does the definition of `length` and `size` here (as calling down to the length of one of the terms) really make sense anymore for this more general or abstract design? Maybe it does but I could survey the use of `length` and `size` for the other "Projected" types to see if we can maybe deprecate or "repurpose" one of these (e.g. `length`) to return the number of terms instead

Again, I would vote for not thinking about this aspect too much, and keeping the interface requirements for projected types as-is to not break user code (for users that defined their own projected types), and additionally not force us to change too much of the other code in the library. I would worry that changes like that would i.e. effect code in `dmrg` or in other projected MPO types.

>     * I still am wondering slightly about the fate of the docstrings here (and I'm realizing it's a question I have about generic code and docstrings more generally in Julia – in C++ it was a non-question since the docs were just hand written). My question is whether it makes more sense for the docstrings to either (1) still refer to a specific type, like ProjMPOSum even though the code below it is much more general or (2) change the docstrings to refer to the general code, but then they may become very hard for users (or me reading them sometime months) to understand since they would refer to more advanced, abstract, developer-oriented concepts

I think it's ok to keep the docstrings as-is, since they are still the best locations for them, and those functions still work on the `ProjMPOSum` type but also accept more general types. But maybe the question is more fundamentally about what the purpose of these docstrings are, i.e. are they for end users of DMRG who want to understand a bit more about how things work or for developers adding new projected MPO types?

src/mps/projmposum.jl
```diff
@@ -73,15 +49,15 @@ Deduce the element type (such as Float64
 or ComplexF64) of the tensors in the ProjMPOSum
 `P`.
 """
-function Base.eltype(P::ProjMPOSum)
-  elT = eltype(P.pm[1])
-  for n in 2:length(P.pm)
-    elT = promote_type(elT, eltype(P.pm[n]))
+function Base.eltype(A::AbstractSum)
```
mtfishman, line 52: Should be able to drop `Base.` here since it is imported.

src/mps/projmposum.jl
```diff
@@ -95,7 +71,7 @@ indices `(a,s1,s2,b)` to the space `(a',s1',s2',b')`
 then the size is `(d,d)` where
 `d = dim(a)*dim(s1)*dim(s1)*dim(b)`
 """
-Base.size(P::ProjMPOSum) = size(P.pm[1])
+Base.size(A::AbstractSum) = size(first(terms(A)))
```
mtfishman, line 74: Should be able to drop `Base.` here since it is imported.

src/mps/projmposum.jl
```diff
@@ -1,48 +1,24 @@
...
+function set_nsite!(A::AbstractSum, nsite)
+  for t in terms(A)
+    set_nsite!(t, nsite)
   end
-  return Ps
+  return A
 end
 
-Base.length(P::ProjMPOSum) = length(P.pm[1])
+Base.length(A::AbstractSum) = length(terms(A)[1])
```
mtfishman, line 21: Should be able to drop `Base.` here since it is imported.

src/mps/projmposum.jl
```diff
@@ -109,9 +85,9 @@ The MPS `psi` must have compatible bond indices with
 the previous projected MPO tensors for this
 operation to succeed.
 """
-function position!(P::ProjMPOSum, psi::MPS, pos::Int)
-  for M in P.pm
-    position!(M, psi, pos)
+function position!(A::AbstractSum, psi::MPS, pos::Int)
```
mtfishman, line 88: We should return `A` from this function, see #1047.

mtfishman:
Perhaps a simpler solution to the awkward generality of:
```julia
SequentialSum{T}(mpos::Vector{MPO}) where {T} = SequentialSum([T(M) for M in mpos])
```
is to constrain that automated conversion to only work with `AbstractProjMPO` subtypes, i.e.:
```julia
SequentialSum{T<:AbstractProjMPO}(mpos::Vector{MPO}) where {T} = SequentialSum([T(M) for M in mpos])
```

src/mps/projmposum.jl
```diff
@@ -128,10 +105,60 @@ ProjMPOSum `P`, and `ortho` is a String which can take
...
 sweeping direction of the DMRG calculation.
 """
-function noiseterm(P::ProjMPOSum, phi::ITensor, dir::String)
-  nt = noiseterm(P.pm[1], phi, dir)
-  for n in 2:length(P.pm)
-    nt += noiseterm(P.pm[n], phi, dir)
+function noiseterm(A::AbstractSum, phi::ITensor, dir::String)
+  nt = noiseterm(terms(A)[1], phi, dir)
+  for n in 2:length(terms(A))
+    nt += noiseterm(terms(A)[n], phi, dir)
```
mtfishman, line 111: Could rewrite as:
```julia
sum(term -> noiseterm(term, phi, dir), terms(A))
```

src/mps/projmposum.jl
```diff
@@ -128,10 +105,60 @@ ProjMPOSum `P`, and `ortho` is a String which can take
...
+# Definition of concrete, generic SequentialSum type
+#
+
+struct SequentialSum{T} <: AbstractSum
+  terms::Vector{T}
+end
+
+SequentialSum{T}(mpos::Vector{MPO}) where {T} = SequentialSum([T(M) for M in mpos])
+
+SequentialSum{T}(Ms::MPO...) where {T} = SequentialSum{T}([Ms...])
```
mtfishman, line 136: I would vote for adding a type constraint:
```julia
SequentialSum{T<:AbstractProjMPO}(mpos::Vector{MPO}) where {T} = [...]
SequentialSum{T<:AbstractProjMPO}(Ms::MPO...) where {T} = [...]
```

src/mps/projmposum.jl
```diff
@@ -58,12 +34,12 @@ returned ITensor will have the same indices
...
 shorthand for `product(P,v)`.
 """
-function product(P::ProjMPOSum, v::ITensor)::ITensor
-  Pv = product(P.pm[1], v)
-  for n in 2:length(P.pm)
-    Pv += product(P.pm[n], v)
+function product(A::AbstractSum, v::ITensor)::ITensor
+  Av = product(first(terms(A)), v)
+  for n in 2:length(terms(A))
+    Av += product(terms(A)[n], v)
```
mtfishman, line 40: Could use:
```julia
sum(term -> term(v), terms(A)) 
```

src/mps/projmposum.jl
```diff
@@ -73,15 +49,15 @@ Deduce the element type (such as Float64
...
 `P`.
 """
-function Base.eltype(P::ProjMPOSum)
-  elT = eltype(P.pm[1])
-  for n in 2:length(P.pm)
-    elT = promote_type(elT, eltype(P.pm[n]))
+function eltype(A::AbstractSum)
+  elT = eltype(first(terms(A)))
+  for n in 2:length(terms(A))
+    elT = promote_type(elT, eltype(terms(A)[n]))
```
mtfishman, line 55: Could use:
```julia
mapreduce(eltype, promote_type, terms(A))
```

src/mps/projmposum.jl
```diff
@@ -109,10 +85,11 @@ The MPS `psi` must have compatible bond indices with
...
 operation to succeed.
 """
-function position!(P::ProjMPOSum, psi::MPS, pos::Int)
-  for M in P.pm
-    position!(M, psi, pos)
+function position!(A::AbstractSum, psi::MPS, pos::Int)
+  for t in terms(A)
+    position!(t, psi, pos)
   end
+  return A
```
mtfishman, line 92: Better to use:
```julia
  new_terms = map(term -> position!(term, psi, pos), terms(A))
  return set_terms(A, new_terms)
```
This is more general since some `position!` functions don't modify in-place, but instead return new environments (such as in the distributed case).

This is how it is written in https://github.com/ITensor/ITensorParallel.jl/pull/15.

src/mps/projmposum.jl
```diff
@@ -128,10 +105,60 @@ ProjMPOSum `P`, and `ortho` is a String which can take
...
 end
+
+"""
+    disk(ps::AbstractSum; kwargs...)
+
+Call `disk` on each term of an AbstractSum, to enable
+saving of cached data to hard disk.
+"""
+function disk(sum::AbstractSum; disk_kwargs...)
+  return set_terms(sum, [disk(term; disk_kwargs...) for term in terms(sum)])
```
mtfishman, line 123: To follow the style I suggest above, I would change this to use `map`:
```julia
set_terms(sum, map(term -> disk(term; disk_kwargs...), terms(sum)))
```
It's slightly more general since `map` could output something besides a `Vector`, and will try to stick to the type returned by `terms`.

src/mps/projmposum.jl
```diff
@@ -1,48 +1,24 @@
...
+nsite(P::AbstractSum) = nsite(first(terms(P)))
 
-nsite(P::ProjMPOSum) = nsite(P.pm[1])
-
-function set_nsite!(Ps::ProjMPOSum, nsite)
-  for P in Ps.pm
-    set_nsite!(P, nsite)
+function set_nsite!(A::AbstractSum, nsite)
+  for t in terms(A)
+    set_nsite!(t, nsite)
```
mtfishman, line 16: I would use:
```julia
return set_terms(A, map(term -> set_nsite!(term, nsite), terms(A))
```

src/mps/projmposum.jl
```diff
@@ -1,48 +1,24 @@
...
+  return error("Please implement `set_terms` for the `AbstractSum` type `$(typeof(sum))`.")
 end
 
-copy(P::ProjMPOSum) = ProjMPOSum(copy.(P.pm))
-
-ProjMPOSum(mpos::Vector{MPO}) = ProjMPOSum([ProjMPO(M) for M in mpos])
+copy(P::AbstractSum) = typeof(P)(copy.(terms(P)))
 
-ProjMPOSum(Ms::MPO...) = ProjMPOSum([Ms...])
+nsite(P::AbstractSum) = nsite(first(terms(P)))
```
mtfishman, line 12: Probably best to add a check `@assert allequal(nsite.(terms(P))`.

src/mps/projmposum.jl
```diff
@@ -95,7 +71,7 @@ indices `(a,s1,s2,b)` to the space `(a',s1',s2',b')`
 then the size is `(d,d)` where
 `d = dim(a)*dim(s1)*dim(s1)*dim(b)`
 """
-Base.size(P::ProjMPOSum) = size(P.pm[1])
+size(A::AbstractSum) = size(first(terms(A)))
```
mtfishman, line 74: Add a check `@assert allequal(size.(terms(P))`.

mtfishman: (Note that `allequal` was introduced in Julia 1.8 but should be available with `Compat.jl`).

src/mps/projmposum.jl
```diff
@@ -1,48 +1,24 @@
...
+function set_nsite!(A::AbstractSum, nsite)
+  for t in terms(A)
+    set_nsite!(t, nsite)
   end
-  return Ps
+  return A
 end
 
-Base.length(P::ProjMPOSum) = length(P.pm[1])
+length(A::AbstractSum) = length(terms(A)[1])
```
mtfishman, line 21: Add a check `@assert allequal(length.(terms(P))`. Also use `length(first(terms(A)))`.

src/mps/projmposum.jl
emstoudenmire: I see, great. And in the new code is the use of an in-place `position!` justified in the sense that if it does do an in-place modification, then `new_terms` will just contain a bunch of new references, but still to the old, now modified, objects?

src/mps/projmposum.jl
emstoudenmire: Agreed, and less verbose too

src/mps/projmposum.jl
mtfishman: Yeah, that's right. Unfortunately it could be slightly breaking since someone could have written a `position!` overload that doesn't return the modified `term` but hopefully they will reach out if they actually have that issue.

Really we should be using `position!!` following the [BangBang.jl](https://juliafolds.github.io/BangBang.jl/stable/) convention, which we use for functions like `contract!!` and `permutedims!!`, which represents that the function either is in-place or allocates the output, but let's just keep it as `position!` for now to not overcomplicate this PR.

mtfishman:
Looks good, thanks! Not sure why the `ITensorGLMakie` tests fail but looks unrelated, some sort of installation problem.

PR #1052, opened 2023-02-04
save to disk for position! before dmrg sweeps

kojionrzl:
Hi Matt,
  This is a issue we once discussed over emails and below is the solution you provided at that time.  I think it will be nice to have it implemented in ITensor.

Best,
Shengtao

# Description
The current DMRG function calls `position!` before the sweeps. This would require storing environment tensors PH in memory, and can potentially run out of memory if the input MPS has a large bond dimension.

The fix writes the environment tensors PH to disk before `position!` , if the maximum bond dimension of the MPS or the dimension set in the sweep exceeds write_when_maxdim_exceeds.


src/mps/dmrg.jl
```diff
@@ -276,6 +276,10 @@ function dmrg(PH, psi0::MPS, sweeps::Sweeps; kwargs...)
   end
   @assert isortho(psi) && orthocenter(psi) == 1
 
+  if !isnothing(write_when_maxdim_exceeds) &&
+    (maxlinkdim(psi) > write_when_maxdim_exceeds) || (maxdim(sweeps, 1) > write_when_maxdim_exceeds)
```
mtfishman, line 280: The tests are failing which makes me think this isn't quite doing the logic you're expecting. I think it makes the code clearer if you split it into nested if-statements:
```julia
if !isnothing(write_when_maxdim_exceeds)
  if (maxlinkdim(psi) > write_when_maxdim_exceeds) || (maxdim(sweeps, 1) > write_when_maxdim_exceeds)
    PH = disk(PH)
  end
end
```

kojionrzl: Yes I think that is a better format.

kojionrzl:
I'm not sure what else can be done regarding this format. It is not obvious from its suggestion.

mtfishman:
You can run the formatter with:
```julia
julia> using ITensors, JuliaFormatter

julia> format(pkgdir(ITensors))
```

kojionrzl:
Everything seems to work now. Thanks for the help throughout the process!

PR #1053, opened 2023-02-07
Bump to CUDA v4.0

mtfishman:
Bump to NDTensors v0.1.46, ITensors v0.3.26, CUDA v4.0, ITensorGPU v0.1.1.

CUDA v4.0 split off `cuTENSOR` into a separate Julia package so now `ITensorGPU` explicitly depends on `cuTENSOR.jl` as well. In the future we should make `cuTENSOR` an optional dependency (through a [package extension](https://docs.julialang.org/en/v1.10-dev/manual/code-loading/#man-extensions)) by providing generic non-cuTENSOR fallbacks for tensor permutation and contraction (which would get overloaded when `cuTENSOR` is loaded).


mtfishman:
Closes #1050.

PR #1055, opened 2023-02-09
[ITensorGPU] Revert to CUDA 3 to try to get CI working again

mtfishman:
@kmp5VT

### DIFF
--- a/ITensorGPU/Project.toml
+++ b/ITensorGPU/Project.toml
@@ -17,11 +17,11 @@ SimpleTraits = "699a6c99-e7fa-54fc-8d76-47d257e15c1d"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Strided = "5e0ebb24-38b0-5f93-81fe-25c709ecae67"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
-cuTENSOR = "011b41b2-24ef-40a8-b3eb-fa098493e9e1"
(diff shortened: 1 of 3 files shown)


mtfishman:
@kmp5VT it looks like tests are passing with Julia 1.6 and 1.7, but not 1.8. It seems related to the issue you brought up in the ITensorGPU SVD code.

I think it is best to merge this to have the GPU CI tests running, and you can fix that issue in #1040.

It looks like the issue is this line:
https://github.com/ITensor/ITensors.jl/blob/e7740ebb8df26cbacc1671113656e702030e43bb/ITensorGPU/src/tensor/culinearalgebra.jl#L75
which is the one we were talking about. I see a few issues happening there.

First of all I think these should be using the lower case `tensor` constructor to avoid extraneous copying, and also the `vec` shouldn't be necessary since the generic constructors you're developing in #1040 should reshape to a vector if needed. So ideally, these lines would be:
```julia
  S = tensor(Dense(Sdata), Sinds)
  V = tensor(Dense(MV), Vinds)
```
I would be curious if that works, since it avoids copying the data and it seems like copying `vec(MV)` is causing the scalar indexing error.
 
We should be able to reproduce that issue with something like:
```julia
using CUDA
copy(vec(CUDA.rand(2, 2)'))
```
That seems like an issue we should raise with CUDA.jl.

If the above doesn't work, we may have to use:
```julia
  S = tensor(Dense(Sdata), Sinds)
  V = tensor(Dense(copy(MV)), Vinds)
```
as we discussed and then we can think about solving this in a more robust way later.

PR #1056, opened 2023-02-09
[ITensorGPU] Try fixing GPU SVD [no ci]

mtfishman:


mtfishman:
@dylex sorry to both you again about the Jenkins setup but I'm seeing some really weird behavior with the Jenkins CI.

You can see that in this PR, the tests pass, but then when I merge into `main` (https://github.com/ITensor/ITensors.jl/commit/42ecd9fbf4f7da419644fca43382d1c7d33a70e9) the same tests fail when testing against certain versions of Julia with an error about resolving incompatible versions of our Julia package dependencies. It seems to be following old package compatibility rules that we set in previous commits to main, and therefore is not running the tests in a totally clean environment. The only thing I can think is that some files (like a `Project.toml` or a `Manifest.toml` which define the Julia package dependencies) are being left around from previous Jenkins CI runs, or getting pointed to erroneously in the `jenkins/Jenkinsfile` script. Do you have any suggestions about this issue?

mtfishman:
The error message:
```julia
+ julia --project=monorepo -e using Pkg; pkg"dev ./NDTensors . ./ITensorGPU";
┌ Warning: The Pkg REPL interface is intended for interactive use, use with caution from scripts.
└ @ Pkg.REPLMode /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Pkg/src/REPLMode/REPLMode.jl:378
[ Info: Resolving package identifier `.` as a directory at `/home/jenkins/workspace/ITensors.jl_main/`.
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package CUDA [052768ef]:
 CUDA [052768ef] log:
 ├─possible versions are: 0.1.0-3.5.0 or uninstalled
 └─restricted to versions 3.13.0-3 by ITensorGPU [d89171c1] — no versions left
   └─ITensorGPU [d89171c1] log:
     ├─possible versions are: 0.1.1 or uninstalled
     └─ITensorGPU [d89171c1] is fixed to version 0.1.1
```
makes me think that it has an old version of the Julia General package registry (https://github.com/JuliaRegistries/General), since in that case I know that there exist CUDA.jl v3.13 but for some reason it thinks that the latest version available is CUDA.jl v3.5. Maybe it's storing an old version of the General registry in `$HOME/.julia` and it isn't being triggered to update for some reason?

If so we could use:
```julia
pkg"registry up General"
```
to try to force it to update, or:
```julia
pkg"registry rm General"
pkg"registry add General"
```
to remove the old registry and add it back.

dylex:
I can also just manually wipe out all the `.julia` directories as a one-time fix if you'd like.

mtfishman:
Maybe that would be best, just to be on the safe side. It's just odd since I've never had to do that before so I'm not sure what it is about the Jenkins setup might make that necessary.

But by "one-time fix" do you mean that you would delete `.julia` just once, or each time Jenkins runs?

dylex:
I mean once, which is now done, just to see if it changes anything.

mtfishman:
Ok thanks. I reverted this commit: https://github.com/ITensor/ITensors.jl/commit/2d717ac2ec162377e21e449e864ea35a43ef0e46 to see if the tests still breaks now that `.julia` is deleted.

mtfishman:
It looks like deleting `.julia` worked (https://github.com/ITensor/ITensors.jl/commit/3c84ffb4bd814fa6f49893662a75b3d6d84013b5) so my guess is that the General registry saved in `.julia` was outdated and not being updated properly. I'll try adding a registry update command to the Jenkins script.

Issue #1059, opened 2023-02-10
[NDTensors] [ENHANCEMENT] `NDTensors`/`ITensorGPU` refactor


mtfishman:
As a follow up to efforts like #1040 and #1049, refactor `NDTensors` to share more code across storage types and data types (`Dense`, `BlockSparse`, `Array`, `CuArray`, `MtlArray`, etc.).

- [ ] Make `default_storagetype` more generic by dispatching based on the indices, for example to construct `Dense` vs. `BlockSparse` if the indices have block structure (like `Vector{Int}` or `Index{Vector{Pair{QN,Int64}}}`).
- [ ] Interfaces like:
```julia
Dense(Float32, dims)
Dense(CuVector, dims)
Dense(CuVector{Float32}, dims)
Tensor(Float32, dims)
Tensor(CuVector, dims)
Tensor(CuVector{Float32}, dims)
Tensor(Dense{CuVector}, dims)
```
which interpret the input as an element type, data type, or storage type based on dispatch and fill in missing type information with defaults as needed.
- [ ] `set_unspecified_parameters` and `default_parameters` as more generic versions of functions `set_eltype_if_unspecified`, default_eltype`, etc.
- [ ] More formal handling of unspecified type parameters, for example using an `UnspecifiedType` singletone type.
- [ ] Make use of package extensions (being introduced in Julia 1.9) to turn packages like `ITensorGPU` into package extensions, where specialized overloads of generic `NDTensors` functionality for certain data backends like `CUDA.jl` and `Metal.jl` are loaded when those packages are loaded.

PR #1062, opened 2023-02-14
[NDTensors] [Enhancements] `SetParameters` module providing tools for setting type parameters

mtfishman:
The aim of this is to provide some basic tools and common interfaces for getting and setting type parameters, with the hopes of making parts of #1059 easier to implement.

The basic idea is that a new type needs to define overloads for accessing and setting type parameters, for example for `Array`:
```julia
get_parameter(::Type{<:Array{P1}}, ::Position{1}) where {P1} = P1
get_parameter(::Type{<:Array{<:Any,P2}}, ::Position{2}) where {P2} = P2

set_parameter(::Type{<:Array}, ::Position{1}, P1) = Array{P1}
set_parameter(::Type{<:Array{<:Any,P2}}, ::Position{1}, P1) where {P2} = Array{P1,P2}
set_parameter(::Type{<:Array}, ::Position{2}, P2) = Array{<:Any,P2}
set_parameter(::Type{<:Array{P1}}, ::Position{2}, P2) where {P1} = Array{P1,P2}
```
Additionally, if you want to work with default type parameters, you can overload functions like this:
```julia
default_parameter(::Type{<:Array}, ::Position{1}) = Float64
default_parameter(::Type{<:Array}, ::Position{2}) = 1

nparameters(::Type{<:Array}) = Val(2)
```

Then, a number of generic functions become available for manipulating the parameters of that type, for example setting multiple parameters, setting only unspecified parameters, setting default parameters, etc., for example:
```julia
# Set parameters
set_parameters(Array{<:Any,3}, Float32, 1) = Vector{Float32}
set_parameters(Array{<:Any,3}, Float32) = Array{Float32,3}
set_parameters(Array{Float32,3}, Position(2), 2) = Matrix{Float32}

# Set only unspecified parameters
set_unspecified_parameters(Array{Float32}, Float64, 2) = Matrix{Float32}

# Set default parameters
set_parameters(Array{Float32}, Position(2), DefaultParameters()) = Vector{Float32}
set_unspecified_parameters(Matrix, DefaultParameters()) = Matrix{Float64}
```
@kmp5VT @emstoudenmire

These functions only work for a single layer of type parameters. Tools for working with nested type parameters would be left for future work. A possibility would be to have an `Adapt.jl`-like interface for working with nested type structures (`AdaptParameters.jl`) which would help with functionality like recursively modifying the element type of a wrapper type. In general we would have to define an interface for defining how to unwrap and rewrap a type, which is something I started working on in the context of the `IsWrappedArray` trait in https://github.com/ITensor/ITensors.jl/blob/v0.3.26/NDTensors/src/abstractarray/similar.jl. Hopefully even without that, the current functionality can be used to simplify some code in `NDTensors`, such as using it to help implement `eltype` definitions.

### DIFF
--- a/NDTensors/src/NDTensors.jl
+++ b/NDTensors/src/NDTensors.jl
@@ -18,6 +18,9 @@ using Strided
 using TimerOutputs
 using TupleTools
 
+include("SetParameters/src/SetParameters.jl")
+using .SetParameters
+
 using Base: @propagate_inbounds, ReshapedArray, DimOrInd, OneTo
 
 using Base.Cartesian: @nexprs
--- a/NDTensors/src/SetParameters/README.md
+++ b/NDTensors/src/SetParameters/README.md
@@ -1,0 +1,3 @@
+# SetParameters
+
+A module defining functionality for getting and setting the parameters of parametric types.
--- a/NDTensors/src/SetParameters/TODO.md
+++ b/NDTensors/src/SetParameters/TODO.md
@@ -1,0 +1,29 @@
+https://discourse.julialang.org/t/extract-type-name-only-from-parametric-type/14188/23
+```julia
+for type in subtypes(AbstractArray)
+  @eval getname(x::$type) = $type
+end
+```
+
+https://discourse.julialang.org/t/stripping-parameter-from-parametric-types/8293/16
+```julia
+getname(type::Type) = Base.typename(type).wrapper
+```
+
+https://docs.julialang.org/en/v1/manual/methods/#Extracting-the-type-parameter-from-a-super-type
+
+# Overloads needed for `AbstractArray` type
+```julia
+parameter(::Type{<:AbstractArray{P1}}, ::Position{1}) where {P1} = P1
+parameter(::Type{<:AbstractArray{<:Any,P2}}, ::Position{2}) where {P2} = P2
+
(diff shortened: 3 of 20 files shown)


mtfishman:
@kmp5VT I'll merge this so that you can try it out in NDTensors and see if it is useful and determine if there are other features we should add.

emstoudenmire:
This looks really nice, a mini-library for manipulating types in a similar way that one manipulates fields of objects.

The interface makes sense, and some functions like the `set_unspecified_parameters` or the `DefaultParameters` adaptor (correct term?) seem especially powerful.

I'd have one hypothetical question about whether passing a keyword arg would be more simple that passing the `Position(n)` object specifying the position that should be set (so something like `set_parameters(Array{Float32,3}, 2; position=2) = Matrix{Float32}`).

mtfishman:
That wouldn't be type stable, the `Position(n)` object is designed to pass the information at "compile time" (or the JIT equivalent), like Julia's built-in `Val` type, so the Julia compile can deduce it is the `n`th parameters you want to manipulate and compile specialized code for that. In general I'm trying to keep NDTensors type stable, and especially with the "type manipulation functions" I think it is important.

emstoudenmire:
Glad to know the reason. I agree that's a good reason! Could definitely impact performance.

PR #1060, opened 2023-02-13
[NDTensors] Reducing compilation time of dense tensor contraction

shinaoka:
This PR is based on the discussion [here](https://itensor.discourse.group/t/strange-compilation-behavior-for-the-function-mps/391/15).

# Description

As summarized in the above page, ITensors.jl suffers from a long compilation time when decomposing a tensor into MPS. The compilation time increases as the number of indices is increased. Furthermore, even after naive code for a small number of indices is generated, the compilation time for a larger number of indices is not reduced. The same applies to reconstructing a tensor from an MPS.

I found that a substantial portion of the compilation time is spent on *compiling the code for analyzing contraction labels* in `contraction_logic.jl`. To address this issue with a minimal change, I fuse contraction labels before the actual contraction code is called.

Original labels:

```
(1, 2, -1, -2) * (-1, -2, 3, 4) =>  (1, 2, 3, 4)
```

After fusion:
```
(1, 2) * (2, 3) =>  (1, 3)
```

This fusing trick substantially reduces the number of concrete types for `ContractionProperties{NA,NB,NC}`.
The following benchmarks were performed on Macbook Pro 14 (M1) with a single thread.

## Benchmark 1

```Julia
using ITensors
ITensors.disable_warn_order()

function run1(ElemType::Type{T}, nbit) where {T}
    sites = siteinds("Qubit", nbit)
    M = randomMPS(sites)
    fulltensor = reduce(*, M)
    return nothing
end

for nbit in [4, 8, 16, 24]
    @time run1(Float64, nbit)
end
```

v0.3.24:
```
7.673702 seconds (48.42 M allocations: 2.709 GiB, 8.71% gc time, 99.98% compilation time)
10.310319 seconds (51.52 M allocations: 3.140 GiB, 6.30% gc time, 99.99% compilation time)
21.941443 seconds (111.89 M allocations: 6.543 GiB, 6.95% gc time, 99.99% compilation time)
26.910477 seconds (134.17 M allocations: 7.751 GiB, 6.21% gc time, 99.70% compilation time)
```

New code:
```
 4.346442 seconds (28.34 M allocations: 1.551 GiB, 9.49% gc time, 99.96% compilation time)
 0.753839 seconds (4.43 M allocations: 229.441 MiB, 4.92% gc time, 99.85% compilation time)
 2.053619 seconds (11.85 M allocations: 595.432 MiB, 4.50% gc time, 99.90% compilation time)
[...]

PR #1071, opened 2023-02-16
[CI] Split off MPO/MPS tests so they can be run separately

mtfishman:
# Description

Split off MPS tests (by which I mean any tests using the `ITensors.MPS` and `ITensors.MPO` types) so they can be run separately with a comment trigger.

Here is a summary of the new CI behavior after this PR is merged:
- On PRs, every commit triggers non-MPS ITensors tests, NDTensors tests, and ITensorGPU tests on Julia 1.6 and the latest Julia release (only on Ubuntu).
- On PRs, you can trigger tests of ITensors MPS as well as subdirectory packages (ITensorGaussianMPS, ITensorVisualizationBase, ITensorUnicodePlots, ITensorGLMakie) with the following "comment triggers":
  - `[mps]`: Run ITensors MPS tests.
  - `[gaussianmps]`: Run ITensorGaussianMPS tests.
  - `[visualization]`: Run ITensorVisualizationBase tests.
  - `[unicodeplots]`: Run ITensorUnicodePlots tests.
  - `[glmakie]`: Run ITensorGLMakie tests.
 - Every commit to the `main` branch triggers all of the above tests, as well as "ITensors base" tests on macOS (and eventually Windows, but those tests have been finicky in the past so are currently disabled).

Additionally, the behavior when running tests locally is now:
- `Pkg.test("ITensors")`: Run both non-MPS and MPS ITensors tests. Same as `Pkg.test("ITensors"; test_args=["all"])` and `Pkg.test("ITensors"; test_args=["base", "mps"])`.
- `Pkg.test("ITensors"; test_args=["base"])`: Run only non-MPS ITensors tests.
- `Pkg.test("ITensors"; test_args=["mps"])`: Run only MPS ITensors tests.

This makes use of https://github.com/JuliaLang/Pkg.jl/pull/1226 which provides an interface for running subsets of tests specified by command line arguments (the global `ARGS` in Julia), and https://github.com/marketplace/actions/pull-request-comment-trigger which is a Github Action for conditionally running parts of a workflow when a "trigger word" is used in a comment.

@kmp5VT @emstoudenmire this should generally make things nicer when making PRs since commits to PRs will by default only run a much smaller set of tests. That time can be improved even more by assessing which tests are taking a long time (and hopefully compressing/simplifying the test suite), as well as improvements to compilation time like #1060.


mtfishman:
Testing on PRs (just testing NDTensors and non-MPS ITensors) takes about 30 minutes to run (at least on Github Actions, it is more like 15 minutes on my Macbook), compared to around 1 hour on Github Actions before this PR, so it is much better but still not great. The slowest test files are:

- 6 minutes for `test_qnitensor.jl`
- 4 minuts for `test_itensor.jl`
- 6 minutes for `ITensorChainRules` (testing Zygote-based AD)

and then a long tail of test files that take around 1 minute or less each, so there isn't a totally obvious place to cut down a significant amount of testing time (it seems more like "death by a thousand cuts"). `@time` indicates that about 40-50% of the time is just compilation, so that would be a big gain.

Issue #1026, opened 2022-11-29
[ITensorGPU] [BUG] Adding complex and real ITensors on GPU


hjkqubit:
**Description of bug**

Adding complex (`ComplexF64`) and real (`Float64`) ITensors on GPU causes an error.

**Minimal code demonstrating the bug or unexpected behavior**

<details><summary>Minimal runnable code</summary><p>

```julia
using ITensors
using ITensorGPU

gpu = cu
i = Index(2)
A = cu(randomITensor(Float64, i))
B = cu(randomITensor(ComplexF64, i))
A + B
```
</p></details>

**Expected output or behavior**

We'd expect the code to run without raising an error.

**Actual output or behavior**

<details><summary>Output of minimal runnable code</summary><p>

```julia
ERROR: LoadError: GPU compilation of kernel broadcast_kernel(CUDA.CuKernelContext, CUDA.CuDeviceVector{Float64, 1}, Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{1}, Tuple{Base.OneTo{Int64}}, ITensors.var"#286#287"{Base.Broadcast.Broadcasted{ITensors.ITensorStyle, Tuple{Base.OneTo{Int64}}, typeof(+), Tuple{ITensor, ITensor}}}, Tuple{Base.Broadcast.Extruded{CUDA.CuDeviceVector{Float64, 1}, Tuple{Bool}, Tuple{Int64}}, Base.Broadcast.Extruded{CUDA.CuDeviceVector{ComplexF64, 1}, Tuple{Bool}, Tuple{Int64}}}}, Int64) failed
KernelError: passing and using non-bitstype argument

Argument 4 to your kernel function is of type Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{1}, Tuple{Base.OneTo{Int64}}, ITensors.var"#286#287"{Base.Broadcast.Broadcasted{ITensors.ITensorStyle, Tuple{Base.OneTo{Int64}}, typeof(+), Tuple{ITensor, ITensor}}}, Tuple{Base.Broadcast.Extruded{CUDA.CuDeviceVector{Float64, 1}, Tuple{Bool}, Tuple{Int64}}, Base.Broadcast.Extruded{CUDA.CuDeviceVector{ComplexF64, 1}, Tuple{Bool}, Tuple{Int64}}}}, which is not isbits:
[...]

PR #1075, opened 2023-02-23
[NDTensors][ITensors][ITensorsGPU] Address issue #1026

kmp5VT:
# Description

Non Base.Vector Array types promote_rules were broken inside of `dense.jl`. In issue, an error occurred when trying to addtwo CUDA vector based ITensors. Recent PR's fixed similar allowing a CuVector based Dense to be properly constructed but promote_rules accidentally constructed a `CuVector{T} where {T}` when attempting to promote a `Float64` to `ComplexF64`. In all other locations the eltype was properly determine. This was a known issue for `promote_types` and a specific `Vector` fix (`promote_rules` function) was found in `dense.jl`

Fixes #(issue)

To fix this problem in general, similartype is used in a type generic dense `promote_rules` to make sure that the DataT uses the promoted eltype. 

Heres how I tested the code 
```
using ITensors, ITensorsGPU, Test
i = Index(2)
A = randomITensor(Float64, i)
B = randomITensor(Float64, i)

cA = cu(A)
cB = cu(B)

C = A + B
cC = cA + cB

@test NDTensors.data(storage(C)) \approx cpu(NDTensors.data(storage(cC)))
```


NDTensors/src/abstractarray/ndims.jl
```diff
@@ -0,0 +1,7 @@
+ndims(x) = Base.ndims(x)
```
mtfishman, line 1: Hmm so maybe this is too general, should we use:
```julia
ndims(array::AbstractArray) = Base.ndims(array)
ndims(arraytype::Type{<:AbstractArray}) = Base.ndims(arraytype)
```

mtfishman: Also let's add a comment:
```julia
# NDTensors.ndims (not imported from Base)
```
to remind us what's going on here.

NDTensors/src/abstractarray/ndims.jl
```diff
@@ -0,0 +1,7 @@
...
+
+if VERSION < v"1.7"
+  ndims(x::AbstractArray) = ndims(typeof(x))
```
mtfishman, line 4: Maybe better to use:
```julia
ndims(array::AbstractArray) = Base.ndims(array)
```
so that it explicitly uses the `Base` Julia definition.

Also, let's add a comment linking to https://github.com/JuliaLang/julia/pull/40682 so we remember why we had to do this.

PR #981, opened 2022-09-23
[ITensors] [BUG] AD numerical error when calling apply on MPS/MPO multiple times

danielalcalde:
# Description

Fixes the gradient of apply. The gradient d MPS2 * apply(gates, MPS) / dMPS was wrong leading to wrong derivatives when applying the function 'apply' more than once.

Fixes #(issue)
#924 

<details><summary>Minimal demonstration</summary><p>

```julia
using PastaQ
using ITensors
using Zygote

N = 1
hilbert = siteinds("Qubit", N)

os = OpSum()
os += -1, "Y", 1
H = MPO(os, hilbert)

ψs = productstate(hilbert, fill(0, N))

function evolve(ρs, θ)
    gates = ITensor[]
    for i in 1:length(θ)
        if mod1(i, 2) == 1
            g = gate("Rx", siteind(ρs, 1), (θ=θ[i]))
        else
            g = gate("Ry", siteind(ρs, 1), (θ=θ[i]))
        end
        gates = vcat(gates, g)
    end
    ρ = apply(gates, ρs)
    return ρ
end

function evolve2(ρs, θ)
    ρ = ρs
    for i in 1:length(θ)
        if mod1(i, 2) == 1
            g = gate("Rx", siteind(ρs, 1), (θ=θ[i]))
        else
            g = gate("Ry", siteind(ρs, 1), (θ=θ[i]))
        end
        ρ = apply([g], ρ)
    end
    
    return ρ
end

function loss(θ)
    Uψ = evolve(ψs, θ)
    return real(inner(Uψ', H, Uψ))
end

function loss2(θ)
    Uψ = evolve2(ψs, θ)
    return real(inner(Uψ', H, Uψ))
end

θ = π .* 0:0.1:2

println(loss(θ))
#0.7218500130938642

println(loss2(θ))
#0.7218500130938642

loss_grad1 = gradient(loss, θ)
@btime gradient(loss, θ)
#24.040 ms (37236 allocations: 4.23 MiB)

loss_grad2 = gradient(loss2, θ)
@btime gradient(loss2, θ)
#25.280 ms (35459 allocations: 4.12 MiB)

loss_grad1 == loss_grad2
# true  was false before
```
</p></details>

# Todo
Add Tests to verify behaviour

# Checklist:

- [x] My code follows the style guidelines of this project. Please run `using JuliaFormatter; format(".")` in the base directory of the repository (`~/.julia/dev/ITensors`) to format your code according to our style guidelines.
[...]

Issue #924, opened 2022-05-26
[ITensors] [BUG] AD numerical error when calling `apply` on MPS/MPO multiple times


mtfishman:
**Description of bug**

Taking derivatives of functions involving applying gates to an MPS/MPO with `apply` where `apply` gets called multiple times gives the wrong result for the derivative.

**Minimal code demonstrating the bug or unexpected behavior**

<details><summary>Minimal runnable code</summary><p>

```julia
using ITensors
using Zygote

function main(; n, θ, p)
  s = siteinds("S=1/2", n)

  ψ₀ₘₚₛ = MPS(s, "↑")
  ψ₀ = contract(ψ₀ₘₚₛ)

  U(θ) = [θ * op("Z", s, 1)]

  function f(θ, ψ)
    ψθ = ψ
    Uθ = U(θ)
    for _ in 1:p
      ψθ = apply(Uθ, ψθ)
    end
    return inner(ψ, ψθ)
  end

  function g(θ, ψ)
    Uθ = U(θ)
    Utot = Uθ
    for _ in 2:p
      Utot = [Utot; Uθ]
    end
    ψθ = apply(Utot, ψ)
    return inner(ψ, ψθ)
  end

  f_itensor(θ) = f(θ, ψ₀)
  f_mps(θ) = f(θ, ψ₀ₘₚₛ)
  g_itensor(θ) = g(θ, ψ₀)
  g_mps(θ) = g(θ, ψ₀ₘₚₛ)

  @show f_itensor(θ), θ^p
  @show f_mps(θ), θ^p
  @show f_itensor'(θ), p * θ^(p - 1)
  @show f_mps'(θ), p * θ^(p - 1)
  @show g_itensor(θ), θ^p
  @show g_mps(θ), θ^p
  @show g_itensor'(θ), p * θ^(p - 1)
  @show g_mps'(θ), p * θ^(p - 1)
  return nothing
end
```
works fine when `apply` is called once:
```julia
julia> main(; n=1, θ=3.0, p=1)
(f_itensor(θ), θ ^ p) = (3.0, 3.0)
(f_mps(θ), θ ^ p) = (3.0, 3.0)
((f_itensor')(θ), p * θ ^ (p - 1)) = (1.0, 1.0)
((f_mps')(θ), p * θ ^ (p - 1)) = (1.0, 1.0)
(g_itensor(θ), θ ^ p) = (3.0, 3.0)
[...]

PR #1077, opened 2023-02-24
[ITensors] Add regression test for AD through repeated calls to apply on MPS

mtfishman:
Regression test for #924, which was fixed in #981.


mtfishman:
[mps]

PR #1080, opened 2023-03-01
[ITensors] Add RXY gate, fix sqrt(iSWAP), add tests

kshyatt:
# Description

Please include a summary of the change and which issue is fixed (if applicable). Please also include relevant motivation and context. List any dependencies that are required for this change.

- Added tests for various `op`s on `Qubit`s
- Added `Rxy` gate for `Qubit`s
- Made `Rz`'s angle input consistent with `Rx` and `Ry`
- Fix bad division factor for sqrtISWAP gate
- Added various tests elsewhere for `sitetype` and `lattices`

# How Has This Been Tested?

New tests pass locally.

# Checklist:

### DIFF
--- a/src/physics/site_types/qubit.jl
+++ b/src/physics/site_types/qubit.jl
@@ -179,10 +179,10 @@ function ITensors.op(::OpName"Rz", ::SiteType"Qubit"; θ=nothing, ϕ=nothing)
   isone(count(isnothing, (θ, ϕ))) || error(
     "Must specify the keyword argument `θ` (or the deprecated `ϕ`) when creating an Rz gate, but not both.",
   )
-  isnothing(ϕ) && (ϕ = θ)
+  isnothing(θ) && (θ = ϕ)
   return [
-    exp(-im * ϕ / 2) 0
-    0 exp(im * ϕ / 2)
+    exp(-im * θ / 2) 0
+    0 exp(im * θ / 2)
   ]
 end
 
--- a/src/physics/site_types/qubit.jl
+++ b/src/physics/site_types/qubit.jl
@@ -175,13 +175,16 @@ function ITensors.op(::OpName"Ry", ::SiteType"Qubit"; θ::Number)
 end
 
 # Rotation around Z-axis
-function ITensors.op(::OpName"Rz", ::SiteType"Qubit"; θ::Number)
+function ITensors.op(::OpName"Rz", ::SiteType"Qubit"; θ=nothing, ϕ=nothing)
+  isone(count(isnothing, (θ, ϕ))) || error(
+    "Must specify the keyword argument `θ` (or the deprecated `ϕ`) when creating an Rz gate, but not both.",
+  )
+  isnothing(ϕ) && (ϕ = θ)
   return [
-    exp(-im * θ / 2) 0
-    0 exp(im * θ / 2)
+    exp(-im * ϕ / 2) 0
+    0 exp(im * ϕ / 2)
   ]
 end
-ITensors.op(::OpName"Rz", t::SiteType"Qubit"; ϕ::Number) = op(OpName("Rz"), t; θ=ϕ)
 
 # Rotation around generic axis n̂
 function ITensors.op(::OpName"Rn", ::SiteType"Qubit"; θ::Real, ϕ::Real, λ::Real)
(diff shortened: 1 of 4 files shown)


src/physics/site_types/qubit.jl
```diff
@@ -331,6 +331,17 @@ function ITensors.op(::OpName"Ryy", ::SiteType"Qubit"; ϕ::Number)
 end
 ITensors.op(::OpName"RYY", t::SiteType"Qubit"; kwargs...) = op("Ryy", t; kwargs...)
 
+# Ising (XY) coupling gate
+function ITensors.op(::OpName"Rxy", t::SiteType"Qubit"; ϕ::Number)
```
mtfishman, line 335: Wikipedia defines this as the complex conjugate of the definition you add here: https://en.wikipedia.org/wiki/List_of_quantum_logic_gates#Two_qubit_interaction_gates. I don't know what's standard but it seems like our other definitions are consistent with Wikipedia (besides the factor of 2 in the angles), and the Wikipedia definitions seem self-consistent.

src/physics/site_types/qubit.jl
```diff
@@ -303,7 +303,7 @@ function ITensors.op(::OpName"√iSWAP", t::SiteType"Qubit")
   return [
     1 0 0 0
     0 1/√2 im/√2 0
-    0 im/2 1/√2 0
+    0 im/√2 1/√2 0
```
mtfishman, line 306: Yikes, thanks for catching that.

src/physics/site_types/qubit.jl
```diff
@@ -175,10 +175,10 @@ function ITensors.op(::OpName"Ry", ::SiteType"Qubit"; θ::Number)
 end
 
 # Rotation around Z-axis
-function ITensors.op(::OpName"Rz", ::SiteType"Qubit"; ϕ::Number)
+function ITensors.op(::OpName"Rz", ::SiteType"Qubit"; θ::Number)
```
mtfishman, line 178: This would be breaking, though I think it would be nice to change it to be consistent with "Rx" and "Ry".

Maybe we can deprecate it quietly for now, say with:
```julia
ITensors.op(::OpName"Rz", ::SiteType"Qubit"; θ=nothing, ϕ=θ)
  isnothing(θ) && error("Must specify gate angle in Rz gate.")
  # [...]
end
```
so then people can use `θ` going forward, but it won't break if they are still using `ϕ` and don't specify `θ`.

kshyatt: Made it a separate pass-through so people don't try to pass `theta` and `phi` simultaneously

mtfishman: The way you wrote it, one method will overwrite the other since the arguments have the same signatures (and kwargs can't be used for dispatch).

We could check for that explicitly:
```julia
ITensors.op(::OpName"Rz", ::SiteType"Qubit"; θ=nothing, ϕ=nothing)
  isone(count(isnothing, (θ, ϕ))) || error("Must specify the keyword argument `θ` (or the deprecated `ϕ`) when creating an Rz gate, but not both.")
  # [...]
end
```

kshyatt: > The way you wrote it, one method will overwrite the other since the arguments have the same signatures (and kwargs can't be used for dispatch).

lol forgot about this one. i love kwargs.

src/physics/site_types/qubit.jl
```diff
@@ -175,13 +175,16 @@ function ITensors.op(::OpName"Ry", ::SiteType"Qubit"; θ::Number)
 end
 
 # Rotation around Z-axis
-function ITensors.op(::OpName"Rz", ::SiteType"Qubit"; θ::Number)
+function ITensors.op(::OpName"Rz", ::SiteType"Qubit"; θ=nothing, ϕ=nothing)
+  isone(count(isnothing, (θ, ϕ))) || error(
+    "Must specify the keyword argument `θ` (or the deprecated `ϕ`) when creating an Rz gate, but not both.",
+  )
+  isnothing(ϕ) && (ϕ = θ)
```
mtfishman, line 182: I think this should have the opposite logic, i.e.
```julia
isnothing(θ) && (θ = ϕ)
```
and then use `θ` since `ϕ` is the deprecated one.

kshyatt: sure, can swap

mtfishman:
Thanks!

PR #1081, opened 2023-03-01
[ITensors] Generalize `correlation_matrix` by ensuring it doesn't create ops on unused sites

mtfishman:
Generalize `correlation_matrix` by ensuring it doesn't create ops on unused sites. Before it was hard-coded to use the first site to check Hermiticity of operators, which could cause trouble if the particular op isn't defined for that site. Now it uses the first site in the `sites` input.

This helps with calling `correlation_matrix` on MPS that have dangling bonds, such as ones sliced from other MPS (like infinite MPS, see https://github.com/ITensor/ITensorInfiniteMPS.jl/pull/72). In that case, you can make "dummy" MPS tensors as the first and last sites that are delta tensors made from the dangling indices, since dangling indices aren't handled correctly right now (though it would be nice to support dangling indices directly). These act like sites, but then can be skipped over with the `sites` input.

@nbaldelli

### DIFF
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -687,7 +687,7 @@ using efficient MPS techniques. Returns the matrix C.
 
 # Optional Keyword Arguments
 
-  - `site_range = 1:length(psi)`: compute correlations only
+  - `sites = 1:length(psi)`: compute correlations only
      for sites in the given range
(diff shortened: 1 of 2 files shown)


mtfishman:
[mps]

PR #1094, opened 2023-03-11
[ITensorGPU] Make it possible to use metal with ITensorGPU

kmp5VT:
# Description

This is a change which shows how simple it now is to use different Array backends.  I have added functionality to ITensorGPU and it is now possible to use Metal based GPU arrays as a backend to `TensorStorage` objects, such as dense.

Fixes #(issue)

# How Has This Been Tested?

This will be tested with new unit tests added for metal. As well I will add a readme guide of the necessary steps to create a new datatype backend.  

# Checklist:

### DIFF
--- a/ITensorGPU/Project.toml
+++ b/ITensorGPU/Project.toml
@@ -18,6 +18,7 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Strided = "5e0ebb24-38b0-5f93-81fe-25c709ecae67"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 cuTENSOR = "011b41b2-24ef-40a8-b3eb-fa098493e9e1"
+Metal = "dde4c033-4e86-420c-a63e-0dd931031962"
 
 [compat]
 Adapt = "3.5"
@@ -33,6 +34,7 @@ Strided = "1.1.2"
 TimerOutputs = "0.5.13"
 cuTENSOR = "1.0.1"
 julia = "1.6"
+Metal = "0.2.0"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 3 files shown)


kmp5VT:
I need to make a unit test for this functionality. Also I need to make a function like `cu` (maybe `mtl`?)

mtfishman:
Very cool @kmp5VT. So this let's us construct Tensors/ITensors with `MtlArray` data but those would still fail if you tried to add them or contraction them with each other?

Also, I guess ultimately we would want to use `SetParameters` (#1062) for this?

Let's discuss where this code should go. Specifically, my vision for this is that we should use Requires.jl [1] for Julia versions below 1.9 and package extensions [2,3] for Julia versions 1.9 and later. This part of the docs [4] explains how we can use either one, depending on which version of Julia the user is using.

To get started with that, we should put this code in a new file `ext/NDTensorMetal/NDTensorMetal.jl`, and follow the instructions in [2,4] to turn this into a package extension.

We could also start doing that with the CUDA code, and start developing a new CUDA backend from scratch in a file `ext/NDTensorCUDA/NDTensorCUDA.jl`. Then, using Requires.jl/package extensions, that code would be conditionally loaded when someone loads one of those GPU backends (i.e. `using CUDA, NDTensors` would load the package extensions `NDTensorCUDA` and `using Metal, NDTensors` would load the package extension `NDTensorMetal`).

Good idea to have a `mtl` function in analogy to the `cu` function. For now we could hold off on that and use the `Adapt.jl` interface for that, i.e. just make sure that `adapt(Metal, x)` works for `x isa Tensor` or `x isa ITensor`.

[1] https://github.com/JuliaPackaging/Requires.jl
[2] https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)
[3] https://discourse.julialang.org/t/package-extensions-for-julia-1-9/93397
[4] https://pkgdocs.julialang.org/dev/creating-packages/#Backwards-compatibility

mtfishman:
@kmp5VT can we close this in favor of #1097?

kmp5VT:
> @kmp5VT can we close this in favor of #1097?

Yes!

PR #1097, opened 2023-03-17
[NDTensors][ITensorGPU] create extensions based optional back-end for NDTensors

kmp5VT:
# Description

The goal here is to create an extension directory which allows us to introduce code that works with non-dependent packages and improves functionality/usability of the ITensors library.  Here I move TBLIS and Octavian to their own NDTensor package and create back-ends for GPU libraries like CUDA and Metal.

All previous test cases should pass and new cases for optional backends will be introduced in the same manner that the base package is tested.

# Checklist:


emstoudenmire:
Karl, regarding code like 
```julia
C = ITensor(NDTensors.generic_randn(MtlVector{Float64}, dim(dim3)), dim3)
```
will there be in the future a version of `randomITensor` that can do this instead? Something like `randomITensor(MtlVector, i, j)` ?

kmp5VT:
@emstoudenmire the short answer is yes something like this will exist in ITensors soon.
The long answer is that I will need to do a large refactor of the ITensor system much like I have done for the NDTensors system.  In this PR, I am trying to limit the scope of my changes to NDTensors with only necessary changes being introduced to the ITensors directory.

ITensorGPU/src/ITensorGPU.jl
```diff
@@ -1,101 +1,104 @@
...
-using LinearAlgebra
 using NDTensors
-using Random
-using SimpleTraits
-using StaticArrays
-using Strided
-using TimerOutputs
-using cuTENSOR
+NDTensors.allow_ndtensorcuda(false);
+if !NDTensors.allow_ndtensorcuda()
```
mtfishman, line 5: I was picturing that we would not do this, and just tell people in the documentation that they shouldn't set `NDTensors.allow_ndtensorcuda(true)` if they want to use `ITensorGPU`.

mtfishman: I.e. let's just keep `ITensorGPU` the way it was before.

NDTensors/ext/NDTensorMetal/array.jl
```diff
@@ -0,0 +1,41 @@
...
+
+function LinearAlgebra.norm(x::Metal.MtlArray)
+  v = real(eltype(x))(0.0)
+  for i in x
+    v += real(conj(i) * i)
+  end
+  return sqrt(v)
+end
+
+function transform(old::Metal.MtlArray, new::Metal.MtlArray, func::Function)
```
mtfishman, line 12: This looks like it is that same as `map!`, which is already defined:
```julia
julia> using Metal

julia> A = mtl(randn(2, 2))
2×2 MtlMatrix{Float32}:
 -0.524251  1.0942
  2.86618   0.0602248

julia> B = mtl(zeros(2, 2))
2×2 MtlMatrix{Float32}:
 0.0  0.0
 0.0  0.0

julia> map!(x -> 2x, B, A)
2×2 MtlMatrix{Float32}:
 -1.0485   2.1884
  5.73237  0.12045
```

NDTensors/ext/NDTensorMetal/array.jl
```diff
@@ -0,0 +1,41 @@
...
+  @assert length(old) == length(new)
+  for i in 1:length(old)
+    new[i] = func(old[i])
+  end
+  return new
+end
+
+function Base.real(x::Metal.MtlArray)
```
mtfishman, line 20: This is already defined:
```julia
julia> A = mtl(randn(ComplexF32, 2, 2))
2×2 MtlMatrix{ComplexF32}:
 -0.420252+0.722033im   1.17723+1.01282im
 -0.227524+0.253891im  -0.49268+2.55224im

julia> real(A)
2×2 MtlMatrix{Float32}:
 -0.420252   1.17723
 -0.227524  -0.49268
```
(I would guess the definitions below this one are also already defined for `MtlArray`).

NDTensors/ext/NDTensorCUDA/adapt.jl
```diff
@@ -0,0 +1,25 @@
+#
+# Used to adapt `EmptyStorage` types
+#
+
+to_vector_type(arraytype::Type{CuArray}) = CuVector
```
mtfishman, line 5: Can we remove `to_vector_type` and use `set_ndims(arraytype, 1)` instead?

NDTensors/ext/examples/NDTensorCUDA.jl
```diff
@@ -0,0 +1,91 @@
+using CUDA
```
mtfishman, line 1: Let's run these package extension examples from the tests. I've found that running examples in the test suite (even if you don't explicitly test anything, which still checks for errors) can help make sure the examples stay up to date.

We can use `include(joinpath(pkgdir(NDTensors), "ext", "examples", "NDTensorCUDA.jl"))`, for example, as a way to get the correct file location in a robust way.

ITensorGPU/src/ITensorGPU.jl
kmp5VT: Ahh okay I see. Do you mean `NDTensors.allow_ndtensorcuda(true)` if they want to use `NDTensorCUDA` and not `ITensorGPU` or if `NDTensors.allow_ndtensorcuda(true)` we should turn off the components of ITensorGPU that clash with NDTensorCUDA?

NDTensors/ext/NDTensorCUDA/adapt.jl
kmp5VT: Yes!

ITensorGPU/src/ITensorGPU.jl
mtfishman: `NDTensors.allow_ndtensorcuda()` just narrowly means that the extension `NDTensorCUDA` can get loaded if it is set to `true` and the extension `NDTensorCUDA` can't get loaded if it is set to `false`.

If `NDTensors.allow_ndtensorcuda()` is `true`, it will of course cause problems if someone tries to load `ITensorGPU`, but we just tell people they shouldn't do that (in the docs/examples) and don't try to fix it automatically.

kmp5VT: Okay I updated this to be literal and just turn off `NDTensorCUDA` if `!NDTensors.allow_ndtensorcuda()` I also added a warning to ITensorGPU and a print to say if NDtensorCUDA is compiled

.github/workflows/test_ndtensors.yml
```diff
@@ -34,7 +34,8 @@ jobs:
         shell: julia --project=monorepo {0}
         run: |
           using Pkg;
-          pkg"dev ./NDTensors"
+          Pkg.develop(path=joinpath(pwd(), "NDTensors"))
+          Pkg.develop(path=pwd())
       - name: Run the tests
         run: |
-          julia --project=monorepo --depwarn=yes -e 'using Pkg; Pkg.test("NDTensors")'
+          julia --project=monorepo --depwarn=yes -e 'using Pkg; Pkg.instantiate(); Pkg.test("NDTensors")'
```
mtfishman, line 41: Did this need to change?

ITensorGPU/src/ITensorGPU.jl
```diff
@@ -1,13 +1,19 @@
...
+using NDTensors
+
 using Adapt
 using CUDA
 using CUDA.CUBLAS
 using CUDA.CUSOLVER
+if CUDA.runtime_version() ≥ v"12.0"
+  println(
+    "Warning: currently cuTENSOR fails with CUDA versions 12.0 and higher.\n To deal with this problem suggest reducing CUDA version to 11.8 using the command `CUDA.set_runtime_version!(\"11.8\")`",
+  )
```
mtfishman, line 12: Maybe instead:
```julia
  println(
    "Warning: currently cuTENSOR does not support CUDA runtime versions 12.0 and higher.\n We suggest reducing the CUDA runtime version to below version 12 using the command `CUDA.set_runtime_version!`, for example `CUDA.set_runtime_version!(\"11.8\")`.",
  )
```
to be more agnostic about which version we recommend, since I don't think there is anything special about CUDA runtime version 11.8.

ITensorGPU/test/runtests.jl
```diff
@@ -1,6 +1,22 @@
-using ITensorGPU, Test, CUDA
+if VERSION < v"1.8"
+  itensorgpu_path = joinpath(pwd(), "..")
```
mtfishman, line 2: This makes a path relative to where the file is called from, so will probably fail if someone includes the `"runtests.jl"` file if they are in a different directory from `test/`, for example if they were in the `ITensorGPU/` directory and called `include("test/runtests.jl")`.

This should be more general:
```julia
itensorgpu_path = joinpath(@__FILE__, "..")
```
`@__FILE__` is a macro that expands to the path of the file where it is located, see [the Julia docs](https://docs.julialang.org/en/v1/base/base/#Base.@__FILE__) for more details.

ITensorGPU/test/runtests.jl
```diff
@@ -1,6 +1,22 @@
...
+  cp(
+    joinpath(pwd(), "..", "LocalPreferences.toml"),
+    joinpath("$(LOAD_PATH[2])", "JuliaLocalPreferences.toml"),
+  )
+  #old_path = pop!(LOAD_PATH)
+  #push!(LOAD_PATH, itensorgpu_path)
+  #push!(LOAD_PATH, old_path)
```
mtfishman, line 9: Probably should remove these lines.

ITensorGPU/test/runtests.jl
```diff
@@ -1,6 +1,22 @@
-using ITensorGPU, Test, CUDA
+if VERSION < v"1.8"
+  itensorgpu_path = joinpath(pwd(), "..")
+  cp(
+    joinpath(pwd(), "..", "LocalPreferences.toml"),
+    joinpath("$(LOAD_PATH[2])", "JuliaLocalPreferences.toml"),
```
mtfishman, line 5: Could you explain what `LOAD_PATH` is being used for here? Maybe there is a better way to do this, this seems fragile.

.github/workflows/test_ndtensors.yml
kmp5VT: So I was running into a compat issue when `NDTensors` tried to compile `ITensors` in `monorepo`. But I just ran the tests again on my machine and I don't see this problem so I will revert the change

ITensorGPU/test/runtests.jl
```diff
@@ -1,6 +1,22 @@
...
+end
+
+using CUDA
+println("Running ITensorGPU tests with a runtime CUDA version: $(CUDA.runtime_version())")
+
+using ITensorGPU, Test
 
 CUDA.allowscalar(false)
+using Pkg
+Pkg.add("Combinatorics")
```
mtfishman, line 19: We shouldn't add packages like this inside the tests. Either they will get installed automatically with `Pkg.test`, or you should activate and instantiate the test environment to install the dependencies.

NDTensors/ext/NDTensorCUDA/NDTensorCUDA.jl
```diff
@@ -0,0 +1,23 @@
+module NDTensorCUDA
+
+println("Compiling NDTensors with CUDA")
```
mtfishman, line 3: This print statement should be removed.

NDTensors/ext/NDTensorCUDA/adapt.jl
```diff
@@ -0,0 +1,30 @@
+#
+# Used to adapt `EmptyStorage` types
+#
+
+struct NDTensorCuArrayAdaptor{B} end
```
mtfishman, line 5: Could you write a comment about why this type was needed? I remember there was some behavior with handling types with some unspecified parameters that CUDA's adaptor wasn't handling properly, but it would be good to make a reminder about it.

NDTensors/ext/NDTensorCUDA/adapt.jl
```diff
@@ -0,0 +1,30 @@
...
+@inline function NDTensors.cu(xs; unified::Bool=false)
+  return fmap(
+    x -> adapt(NDTensorCuArrayAdaptor{unified ? Mem.UnifiedBuffer : Mem.DeviceBuffer}(), x),
+    xs,
+  )
+end
+
+function Adapt.adapt_storage(
+  ::NDTensorCuArrayAdaptor{B}, xs::AbstractArray{T,N}
+) where {T,N,B}
```
mtfishman, line 16: Can we remove the `where` and use `buffertype`, `eltype, `ndims`, etc.?

ITensorGPU/test/runtests.jl
kmp5VT: Ahh good point. I was only testing by launching `Pkg.test` and didn't consider `include`

NDTensors/ext/NDTensorCUDA/adapt.jl
```diff
@@ -0,0 +1,30 @@
...
+function NDTensors.adapt_storagetype(
+  ::NDTensorCuArrayAdaptor{B}, xs::Type{EmptyStorage{ElT,StoreT}}
+) where {ElT,StoreT,B}
+  return NDTensors.emptytype(NDTensors.adapt_storagetype(CuVector{ElT,B}, StoreT))
+end
+
+## In house patch to deal issue of calling ndims with an Array of unspecified eltype
+## https://github.com/JuliaLang/julia/pull/40682
+if VERSION < v"1.7"
+  ndims(::Type{<:CuArray{<:Any,N}}) where {N} = N
```
mtfishman, line 29: Can't this be defined for any `AbstractArray`:
```julia
ndims(::Type{<:AbstractArray{<:Any,N}}) where {N} = N
```
so we don't need a specialized version here for `CuArray`?

ITensorGPU/test/runtests.jl
kmp5VT: Okay, this was just another viable solution to the `LocalPreferences` + `Pkg.test` issue in julia v"1.6" and v"1.7"

NDTensors/ext/NDTensorCUDA/adapt.jl
```diff
@@ -0,0 +1,30 @@
...
+
+function Adapt.adapt_storage(
+  ::NDTensorCuArrayAdaptor{B}, xs::AbstractArray{T,N}
+) where {T,N,B}
+  return isbits(xs) ? xs : CuArray{T,1,B}(xs)
+end
+
+function NDTensors.adapt_storagetype(
+  ::NDTensorCuArrayAdaptor{B}, xs::Type{EmptyStorage{ElT,StoreT}}
+) where {ElT,StoreT,B}
```
mtfishman, line 22: Same comment as above about trying to avoid `where` as much as possible.

NDTensors/ext/NDTensorCUDA/adapt.jl
mtfishman: We may have to define:
```julia
buffertype(::NDTensorCuArrayAdaptor{B}) where {B} = B
```

NDTensors/ext/examples/NDTensorCUDA.jl
```diff
@@ -0,0 +1,128 @@
...
+using NDTensors
+
+using ITensors
+using Test
+
+using Pkg;
+Pkg.add("Zygote");
```
mtfishman, line 8: We shouldn't `add` packages like this in examples (users should install it themselves).

NDTensors/src/NDTensors.jl
```diff
@@ -220,13 +223,31 @@ function disable_tblis()
...
+    end
+
+    @require CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba" begin
+      if CUDA.functional()
+        include("../ext/NDTensorCUDA/NDTensorCUDA.jl")
+      end
+    end
+
+    @require Metal = "dde4c033-4e86-420c-a63e-0dd931031962" begin
+      println("NDTensors is compiled with Metal.jl")
```
mtfishman, line 248: Print statement should be removed.

NDTensors/test/dense.jl
```diff
@@ -1,220 +1,221 @@
 using NDTensors
+
 using Test
 
 @testset "Dense Tensors" begin
-  @testset "DenseTensor basic functionality" begin
-    A = Tensor((3, 4))
+  for op in ops
+    @testset "DenseTensor basic functionality" begin
```
mtfishman, line 7: You can make an `@testset` into a loop:
```julia
@testset "DenseTensor basic functionality on $op" for op in ops
  # Tests
end
```

NDTensors/ext/NDTensorCUDA/adapt.jl
mtfishman: Maybe you can use the `storagetype` function you defined below for `EmptyStorage`.

NDTensors/Project.toml
```diff
@@ -36,4 +37,28 @@ StaticArrays = "0.12, 1.0"
 Strided = "0.3, 1"
 TimerOutputs = "0.5.5"
 TupleTools = "1.2.0"
+Preferences = "1.3.0"
+ZygoteRules = "0.2.2"
```
mtfishman, line 41: Are these needed as direct dependencies?

NDTensors/Project.toml
```diff
@@ -20,6 +20,7 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Strided = "5e0ebb24-38b0-5f93-81fe-25c709ecae67"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 TupleTools = "9d95972d-f1c8-5527-a6e0-b4b365fa01f6"
+Preferences = "21216c6a-2e73-6563-6e65-726566657250"
```
mtfishman, line 23: Is this needed as a direct dependency?

NDTensors/Project.toml
mtfishman: Oh I see, I thought because it was in `[compat]` it meant that it was a direct dependency, but you are using `[compat]` to set the compatible versions of the `[extras]` as well. But I don't see that you are setting compat entries for the other `[extras]`, is this one in particular needed?

ITensorGPU/src/adapt.jl
```diff
@@ -3,10 +3,23 @@
...
+  return NDTensors.emptytype(
+    NDTensors.similartype(store, CuVector{eltype(x),CUDA.Mem.DeviceBuffer})
+  )
+end
+
+function NDTensors.storagetype(
+  x::Type{<:NDTensors.EmptyStorage{ElT,StoreT}}
+) where {ElT,StoreT}
+  return StoreT
+end
```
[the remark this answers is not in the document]
kmp5VT: These were in NDTensors but weren't being properly picked up because of the diverting compat numbers. This will be addressed in the new PR

NDTensors/ext/NDTensorCUDA/adapt.jl
kmp5VT: It already is, but was failing for some reason. I think it was a testing environment error and not related to the code. I removed this and will monitor

NDTensors/src/NDTensors.jl
kmp5VT: I didn't know if we wanted to warn/inform users that they instantiated extensions when they add certain packages.

NDTensors/Project.toml
kmp5VT: I was just matching the compat versions from `ITensors`.

mtfishman: In general, I haven't been putting `[compat]` entries for test dependencies unless it is necessary, like if our tests fail with a newer version of a package and we want to specify the last version where the tests work. In those cases where we do want to specify the `[compat]` of a test dependency, I think the right place to put it would be in `NDTensors/test/Project.toml`, not in `NDTensors/Project.toml`.

Project.toml
```diff
@@ -41,7 +41,7 @@ HDF5 = "0.14, 0.15, 0.16"
 IsApprox = "0.1"
 KrylovKit = "0.4.2, 0.5, 0.6"
 LinearMaps = "3"
-NDTensors = "0.1.50"
+NDTensors = "0.2.0"
```
mtfishman, line 44: Can we use `"0.1.50, 0.2"`? Otherwise, ITensorGPU won't be able to access updated versions of ITensors.jl.

kmp5VT: Yes, this works and is updated.

NDTensors/test/dense.jl
```diff
@@ -106,23 +108,23 @@ using Test
     @test Array(J * K) ≈ Array(J) * Array(K)
   end
 
-  @testset "Random constructor" begin
-    T = randomTensor((2, 2))
+  @testset "Random constructor" for op in ops
```
mtfishman, line 111: Instead of having to define each `@testset` with a loop over `ops`, what about having a larger outer loop over `ops`? That would make it a bit more convenient for writing new tests.

kmp5VT: Does it matter that some tests that are not dependent on ops but will be in the ops loop?

mtfishman: Good question. Maybe those could just be moved to a different part of the tests, outside of the outer loop over `ops`, so that they aren't run twice unnecessarily. That could also have the advantage that we can more easily track which tests are run for all devices and which ones are run only on CPU.

.github/workflows/documentation.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         with:
           version: '1'
       - name: Install dependencies
-        run: julia --project=docs/ -e 'using Pkg; Pkg.develop(path="./NDTensors"); Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
+        run: julia --project=docs/ -e 'using Pkg; Pkg.develop(path="./NDTensors"); Pkg.develop(PackageSpec(path=pwd()));'
```
mtfishman, line 26: Could you explain this change?

NDTensors/Project.toml
```diff
@@ -37,3 +37,26 @@ Strided = "0.3, 1"
 TimerOutputs = "0.5.5"
 TupleTools = "1.2.0"
 julia = "1.6"
+
+[weakdeps]
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
+Metal = "dde4c033-4e86-420c-a63e-0dd931031962"
+TBLIS = "48530278-0828-4a49-9772-0f3830dfa1e9"
+Octavian = "6fd5a793-0b7e-452c-907f-f8bfe9c57db4"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
```
mtfishman, line 46: Is Zygote supposed to be a weakdep?

ITensorGPU/test/runtests.jl
```diff
@@ -1,4 +1,4 @@
-if VERSION ≤ v"1.8" && "@v#.#" ∉ LOAD_PATH
+if isless(VERSION, v"1.8") && "@v#.#" ∉ LOAD_PATH
```
mtfishman, line 1: Maybe `<` instead of `isless`?

.github/workflows/test_itensors_base_ubuntu.yml
```diff
@@ -34,15 +34,17 @@ jobs:
...
+          Pkg.activate(".")
+          pkg"dev ./NDTensors"
       - name: Run the tests
-        shell: julia --project=monorepo {0}
+        shell: julia {0}
         run: |
           using Pkg;
-          Pkg.test("ITensors"; coverage=true, test_args=["base"])
+          Pkg.activate(".")
+          Pkg.test(; coverage=true, test_args=["base"])
```
mtfishman, line 47: Could you explain these changes?

.github/workflows/documentation.yml
kmp5VT: No this must have been an accidental change from when I was looking into a problem ryan had

NDTensors/Project.toml
kmp5VT: No this was for testing in `ext/NDTensorCUDA/examples.jl`. This is outdated from when I was trying to add the dependencies need in `NDTensors/test/` . I removed all Zygote from this file.

ITensorGPU/test/runtests.jl
kmp5VT: I was getting a weird error thrown in `Jenkins` that it didn't have the `<` operator and to use `isless`. I think this bug was related to a different issue so I migrated from `isless` to `<`

NDTensors/ext/NDTensorCUDA/set_types.jl
```diff
@@ -0,0 +1,34 @@
+buffertype(datatype::Type{<:CuArray{<:Any,<:Any,B}}) where {B} = B
+function buffertype(datatype::Type{<:CuArray})
+  println(
+    "CuArray definitions require a CUDA.Mem buffer try $(datatype{default_buffertype()})"
+  )
+  throw(TypeError)
```
mtfishman, line 6: I think `$(datatype{default_buffertype()})` won't always expand to what you hope it will here, since if someone just passes `CuArray` it would make it look like the buffer type should be passed as the first type parameter.

Also, I don't think you're using `TypeError` correctly, it would be used like this:
```julia
julia> throw(TypeError(:typeassert, Float64, 1))
ERROR: TypeError: in typeassert, expected Float64, got a value of type Int64
Stacktrace:
 [1] top-level scope
   @ REPL[46]:1
```
which gets output by type assertions such as:
```julia
julia> 1::Float64
ERROR: TypeError: in typeassert, expected Float64, got a value of type Int64
Stacktrace:
 [1] top-level scope
   @ REPL[43]:1
```

I would go with something simpler, like:
```julia
function buffertype(datatype::Type{<:CuArray})
  return error("No buffer type specified in the type $(datatype)")
end
```

NDTensors/ext/NDTensorCUDA/set_types.jl
```diff
@@ -0,0 +1,34 @@
...
+function set_eltype(arraytype::Type{<:CuArray}, eltype::Type)
+  return CuArray{eltype,NDTensors.ndims(arraytype),buffertype(arraytype)}
+end
+
+function set_ndims(arraytype::Type{<:CuArray{T,<:Any,<:Any}}, ndims) where {T}
+  return CuArray{eltype(arraytype),ndims,buffertype(arraytype)}
+end
+
+function set_ndims(arraytype::Type{<:CuArray{T}}, ndims) where {T}
+  return CuArray{eltype(arraytype),ndims,default_buffertype()}
```
mtfishman, line 20: It seems too opinionated to fill in the `default_buffertype()` here, shouldn't it leave it unspecified if it wasn't specified?

In a future PR we should start using `SetParameters`, it should make these types of functions simpler to define.

NDTensors/ext/NDTensorCUDA/set_types.jl
```diff
@@ -0,0 +1,34 @@
...
+end
+
+function set_ndims(arraytype::Type{<:CuArray}, ndims)
+  return CuArray{NDTensors.default_eltype(),ndims,default_buffertype()}
```
mtfishman, line 24: Again, it seems to opinionated to fill in `default_buffertype()`.

NDTensors/ext/NDTensorMetal/adapt.jl
```diff
@@ -0,0 +1,5 @@
+#
+# Used to adapt `EmptyStorage` types
+#
```
mtfishman, line 3: Is this comment right? This looks more general than `EmptyStorage`.

NDTensors/ext/NDTensorMetal/set_types.jl
```diff
@@ -0,0 +1,24 @@
...
+  return MtlArray{eltype,ndims(arraytype)}
+end
+
+function set_ndims(arraytype::Type{<:MtlArray{T,<:Any}}, ndims) where {T}
+  return MtlArray{eltype(arraytype),ndims}
+end
+
+function set_ndims(arraytype::Type{<:MtlArray{T}}, ndims) where {T}
+  return MtlArray{eltype(arraytype),ndims}
+end
```
mtfishman, line 11: Aren't these the same? For example:
```julia
julia> f(x::Type{<:Array{T}}) where {T} = 1

julia> f(x::Type{<:Array{T,<:Any}}) where {T} = 2

julia> f(Array{Float64})
2

julia> f(Array{Float64,2})
2
```

NDTensors/ext/examples/NDTensorCUDA.jl
mtfishman, line 1: @kmp5VT looks like this file needs to get cleaned up a but (there are extra lines and comments that don't seem to be used).

NDTensors/src/NDTensors.jl
```diff
@@ -220,13 +223,30 @@ function disable_tblis()
...
-  end
-  @require Octavian = "6fd5a793-0b7e-452c-907f-f8bfe9c57db4" begin
-    include("linearalgebra/octavian.jl")
+  @static if !isdefined(Base, :get_extension)
+    @require TBLIS = "48530278-0828-4a49-9772-0f3830dfa1e9" begin
+      enable_tblis()
+      include("../ext/NDTensorTBLIS/NDTensorTBLIS.jl")
+    end
+    @require Octavian = "6fd5a793-0b7e-452c-907f-f8bfe9c57db4" begin
+      #include("linearalgebra/octavian.jl")
```
mtfishman, line 237: Looks like this can be removed.

NDTensors/src/NDTensors.jl
```diff
@@ -220,13 +223,30 @@ function disable_tblis()
...
 function __init__()
-  @require TBLIS = "48530278-0828-4a49-9772-0f3830dfa1e9" begin
-    enable_tblis()
-    include("tensoralgebra/tblis.jl")
-  end
-  @require Octavian = "6fd5a793-0b7e-452c-907f-f8bfe9c57db4" begin
-    include("linearalgebra/octavian.jl")
+  @static if !isdefined(Base, :get_extension)
+    @require TBLIS = "48530278-0828-4a49-9772-0f3830dfa1e9" begin
+      enable_tblis()
```
mtfishman, line 233: Do we need this any more? It seems like `using TBLIS; using NDTensors` achieves the same thing. Maybe we want to deprecate it and remove it in a future release in case people are using it.

jenkins/Jenkinsfile
```diff
@@ -27,7 +27,10 @@ pipeline {
           }
           steps {
             sh '''
-              julia -e 'using Pkg; Pkg.Registry.update(); Pkg.develop(path="./NDTensors"); Pkg.develop(;path="."); Pkg.develop(; path="./ITensorGPU"); Pkg.add("CUDA"); using CUDA; CUDA.set_runtime_version!("v11.8"); Pkg.test("ITensorGPU");'
+              julia -e 'using Pkg; Pkg.add("CUDA"); using CUDA; CUDA.set_runtime_version!("v11.8"); Pkg.activate("./ITensorGPU"); Pkg.develop(path="."); Pkg.add(name="NDTensors", version="0.1.50"); Pkg.instantiate(); Pkg.test();'
```
mtfishman, line 30: Maybe we should use `Pkg.add(name="NDTensors", version="0.1")` instead of hardcoding to `"0.1.50"`, in case we update the version of NDTensors 0.1.

src/itensor.jl
```diff
@@ -673,7 +673,9 @@ copy(T::ITensor)::ITensor = itensor(copy(tensor(T)))
 #
 
 # Helper functions for different view behaviors
-Array{ElT,N}(::NeverAlias, A::AbstractArray) where {ElT,N} = Array{ElT,N}(A)
+function Array{ElT,N}(::NeverAlias, A::AbstractArray) where {ElT,N}
+  return NDTensors.similartype(A, ElT)(A)
+end#Array{ElT,N}(A)
```
mtfishman, line 678: Extraneous comment.

src/mps/adapt.jl
```diff
@@ -1 +1,5 @@
 adapt_structure(to, x::Union{MPS,MPO}) = map(xᵢ -> adapt(to, xᵢ), x)
+
+# function NDTensors.cu(x::Union{MPS,MPO}; unified=false)
+#   return map(xᵢ -> NDTensors.cu(xᵢ; unified=unified), x)
+# end
```
mtfishman, line 5: Looks like this could be removed.

NDTensors/test/device_list.jl
```diff
@@ -0,0 +1,21 @@
...
+  devs = Vector{Function}(undef, 0)
+  if isempty(test_args) || "base" in test_args
+    push!(devs, NDTensors.cpu)
+  end
+
+  if "cuda" in test_args || "all" in test_args
+    CUDA.allowscalar()
+    if CUDA.functional()
+      push!(devs, NDTensors.cu)
+      #include(joinpath(pkgdir(NDTensors), "ext", "examples", "NDTensorCUDA.jl"))
```
mtfishman, line 11: Should this be removed?

NDTensors/test/device_list.jl
```diff
@@ -0,0 +1,21 @@
...
+    end
+  end
+
+  if "metal" in test_args || "all" in test_args
+    push!(devs, NDTensors.mtl)
+    Metal.allowscalar()
+    include(joinpath(pkgdir(NDTensors), "ext", "examples", "NDTensorMetal.jl"))
```
mtfishman, line 18: Should this be removed?

NDTensors/test/device_list.jl
```diff
@@ -0,0 +1,21 @@
+function devices_list(test_args)
+  devs = Vector{Function}(undef, 0)
+  if isempty(test_args) || "base" in test_args
+    push!(devs, NDTensors.cpu)
+  end
+
+  if "cuda" in test_args || "all" in test_args
+    CUDA.allowscalar()
+    if CUDA.functional()
```
mtfishman, line 9: Maybe we should output a warning if it isn't functional?

mtfishman: I think I would be surprised if I tried to run the CUDA backend tests but they just silently didn't run since CUDA wasn't set up properly.

NDTensors/test/device_list.jl
mtfishman: Oops sorry, I misread this, I thought it was loading the extension package but I see it is loading the example.

NDTensors/ext/NDTensorMetal/set_types.jl
```diff
@@ -0,0 +1,20 @@
+function set_eltype(arraytype::Type{<:MtlArray}, eltype::Type)
+  return MtlArray{eltype,ndims(arraytype)}
+end
+
+function set_ndims(arraytype::Type{<:MtlArray{T}}, ndims) where {T}
+  return MtlArray{eltype(arraytype),ndims}
+end
+
+function set_ndims(arraytype::Type{<:MtlArray}, ndims)
+  return MtlArray{NDTensors.default_eltype(),ndims}
```
mtfishman, line 10: Shouldn't this avoid setting a default as well?

src/itensor.jl
```diff
@@ -673,7 +673,9 @@ copy(T::ITensor)::ITensor = itensor(copy(tensor(T)))
 #
 
 # Helper functions for different view behaviors
-Array{ElT,N}(::NeverAlias, A::AbstractArray) where {ElT,N} = Array{ElT,N}(A)
+function Array{ElT,N}(::NeverAlias, A::AbstractArray) where {ElT,N}
+  return NDTensors.similartype(A, ElT)(A)
```
mtfishman, line 677: I don't think this is right. Constructor calls like `Array{Float64,3}(...)` should be very literal and only output an object with the type `Array{Float64,3}`.

src/itensor.jl
```diff
@@ -345,7 +345,7 @@ function ITensor(
       "In ITensor(::AbstractArray, inds), length of AbstractArray ($(length(A))) must match total dimension of IndexSet ($(dim(inds)))",
     ),
   )
-  data = Array{eltype}(as, A)
+  data = NDTensors.similartype(A, eltype)(A)
```
mtfishman, line 348: I think this should be:
```julia
data = NDTensors.similartype(A, eltype)(as, A)
```
so it preserves the desired copying behavior.

NDTensors/ext/NDTensorMetal/adapt.jl
```diff
@@ -0,0 +1 @@
+@inline NDTensors.mtl(xs; unified::Bool=false) = NDTensors.adapt_structure(MtlArray, xs)
```
mtfishman, line 1: Should this just be `adapt(MtlArray, xs)` instead of `adapt_storage(MtlArray, xs)`?

mtfishman: Also seems like `unified` isn't needed/used here.

NDTensors/ext/examples/NDTensorCUDA.jl
```diff
@@ -0,0 +1,124 @@
...
+  @test inner(cm', cH, cm) ≈ inner(m', H, m)
+end
+
+### To run the NDTensorCUDA tests in the NDTensors test suite. use the following commands in the NDTensors directory.
+if false # false so we don't have an infinite loop
+  using Pkg
+  Pkg.add("CUDA")
+  using CUDA
+  Pkg.test("NDTensors"; test_args=["cuda"])
+end
```
mtfishman, line 122: Seems strange to put this in an example code, can this be removed?

PR #1100, opened 2023-03-22
[ITensorGaussianMPS] [Bug] Fix mismatched convention for single particle Hamiltonians.

b-kloss:
# Description
This adds a missing factor of 2 converting the single-particle operator (Hamiltonian) returned from `quadratic_operator` to its number-conserving single-particle operator (Hamiltonian) form within `hopping-operator`. Ensures full consistency between the operator generated in the many-body picture via `MPO(os::OpSum)` and the single-particle picture via `quadratic_operator(os::OpSum)` and `hopping_operator(os::OpSum)`. The tests have been adjusted accordingly.
Fixes failing tests in previous PR #1038 .

PR #1101, opened 2023-03-22
[CI] Add example of an improved comment trigger workflow

mtfishman:
This adds an example of an improved comment trigger workflow, based on https://dev.to/zirkelc/trigger-github-workflow-for-comment-on-pull-request-45l2.

Currently the comment triggers only run on the latest commit of the `main` branch (see #1100), this new design would hopefully fix that. Additionally, the new design should show the tests running in the PR, and make a comment about the results of the workflow. I think it only runs once it is actually on `main` so I'll merge this and test it in future PRs.


mtfishman:
[gaussianmps]

PR #1103, opened 2023-03-24
[CI] Test out new comment trigger example workflow

mtfishman:


emstoudenmire:
Interesting. Does this run arbitrary Github actions based on special comments in PR's? Or just specific kinds of actions?

mtfishman:
I don't see why you couldn't run any action, for example by sticking any commands you want into: https://github.com/ITensor/ITensors.jl/blob/d45126ca8aa31b46a3eb744e7cdfa0e22ee5663a/.github/workflows/comment_trigger_example.yml#L32-L34

mtfishman:
[test ITensorGaussianMPS]

PR #1106, opened 2023-03-30
[ITensorGaussianMPS] [Enhancement] Constrain diagonalization to fermionic transformations. 

b-kloss:
# Description

This PR implements diagonalization of quadratic operators via Schur decomposition in the Majorana representation. This improves numerical stability across all of parameter space, especially in cases where standard eigendecomposition does not result in gaussian fermionic eigenstates due to spectral (quasi)degeneracies. Furthermore, `correlation_matrix_to_gmps` is now returning a circuit which diagonalizes the input correlation matrix even when a number-conserving correlation matrix is input as a parity-conserving correlation matrix.
A few non-exhaustive tests for the newly introduced linear algebra routines are added (the functionality being otherwise tested indirectly in the `*gmps` routine tests).

### DIFF
--- a/ITensorGaussianMPS/examples/spinless_fermion_pairing.jl
+++ b/ITensorGaussianMPS/examples/spinless_fermion_pairing.jl
@@ -52,14 +52,14 @@ let
   @assert ishermitian(h)
   e = eigvals(Hermitian(h))
   @show e
-  E, V = gaussian_eigen(h)
+  E, V = eigen_gaussian(h)
   @show sum(E[1:N])
   Φ = V[:, 1:N]
-  c = conj(Φ) * transpose(Φ)
-
+  c = real.(conj(Φ) * transpose(Φ))
+  
   #Get (G)MPS
   psi = ITensorGaussianMPS.correlation_matrix_to_mps(
-    sites, real.(c); eigval_cutoff=1e-10, maxblocksize=14, cutoff=1e-11
+    sites, c; eigval_cutoff=1e-10, maxblocksize=14, cutoff=1e-11
   )
   @show eltype(psi[1])
   cdagc = correlation_matrix(psi, "C", "Cdag")
--- a/ITensorGaussianMPS/examples/spinless_fermion_pairing.jl
+++ b/ITensorGaussianMPS/examples/spinless_fermion_pairing.jl
@@ -6,17 +6,17 @@ using ITensorGaussianMPS
 
 ITensors.disable_contraction_sequence_optimization()
 let
-  N = 10
-  sites = siteinds("Fermion", N; conserve_qns=false)
+  N = 8
+  sites = siteinds("Fermion", N; conserve_qns=false, conserve_nfparity=true)
   _maxlinkdim = 100
   # DMRG cutoff
   _cutoff = 1e-13
   # Hopping
-  t = -1.2
+  t = -1.0
   # Electron-electron on-site interaction
   U = 0.0
   # Pairing
-  Delta = 0.7
+  Delta = 1.00
   @show t, U, Delta
   # Free fermion Hamiltonian
   os_h = OpSum()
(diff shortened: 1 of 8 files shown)


ITensorGaussianMPS/examples/spinless_fermion_pairing.jl
```diff
@@ -49,16 +49,22 @@ let
...
   #Get Ground state 
-  E, V = eigen(Hermitian(h))
-  @show E
+  @assert ishermitian(h)
+  e = eigvals(Hermitian(h))
+  @show e
+  E, V = gaussian_eigen(h)
+  @show sum(E[1:N])
   Φ = V[:, 1:N]
   c = conj(Φ) * transpose(Φ)
```
mtfishman, line 58: I would put the call to `real` here, i.e. `c = real(conj(Φ) * transpose(Φ))` to emphasize that step more.

ITensorGaussianMPS/src/ITensorGaussianMPS.jl
```diff
@@ -16,9 +15,13 @@ export slater_determinant_to_mps,
   quadratic_hamiltonian,
   quadratic_operator,
   slater_determinant_matrix,
-  slater_determinant_to_gmera
+  slater_determinant_to_gmera,
+  gaussian_eigen,
+  gaussian_diag_corr,
+  get_gaussian_GS_corr
```
mtfishman, line 21: I would vote for not exporting these two functions.

ITensorGaussianMPS/src/linalg.jl
```diff
@@ -0,0 +1,201 @@
...
+    # make another pass with rotation in the complex plane per eigenvector
+    U .*= exp.(-im * angle.(U[1:1, :]))
+    @assert norm(imag.(U)) < sqrt(eps(real(ElT)))
+    U = real(real.(U))
+  end
+  return d, U
+end
+
+"""Takes a single-particle Hamiltonian in interlaced Dirac format and finds the complex fermionic transformation U that diagonalizes it."""
+function gaussian_eigen(H; pert=nothing)
```
mtfishman, line 42: I would vote for naming this `eigen_gaussian` instead.

mtfishman: Also to make the keyword argument more descriptive, I would call it something like `noise_scale` instead of `pert`.

ITensorGaussianMPS/src/linalg.jl
```diff
@@ -0,0 +1,201 @@
...
+  for e in approx_unique_eigvals
+    mask = rounded_halfspectrum .== e
+    if abs(e) < eps(real(eltype(U0)))
+      # handle values close to zero separately
+      # rotate subspace for both positive and negative eigenvalue if they are close enough to zero
+      mask = vcat(mask, mask)
+      subspace = copy(U[:, :][:, mask])
+      subspace = make_subspace_real_if_possible(subspace)
+      v = @views U[:, :][:, mask]
+      v .= subspace
```
mtfishman, line 138: I think this can just be written as `U[:, mask] = subspace`.

ITensorGaussianMPS/src/linalg.jl
```diff
@@ -0,0 +1,201 @@
...
+  rounded_halfspectrum = round.(spectrum[1:n], sigdigits=sigdigits)
+  approx_unique_eigvals = unique(rounded_halfspectrum)
+  # loop over degenerate subspaces
+  for e in approx_unique_eigvals
+    mask = rounded_halfspectrum .== e
+    if abs(e) < eps(real(eltype(U0)))
+      # handle values close to zero separately
+      # rotate subspace for both positive and negative eigenvalue if they are close enough to zero
+      mask = vcat(mask, mask)
+      subspace = copy(U[:, :][:, mask])
```
mtfishman, line 135: You should be able to just use `subspace = U[:, mask]` here, slicing makes a copy by default (i.e. if you don't use `@view`).

b-kloss:
[test ITensorGaussianMPS]

PR #1107, opened 2023-03-31
[ITensorGPU][Bug] Restrict to CUDA.jl v4.0.1 to fix compatibility issues with cuTENSOR

kmp5VT:
# Description

It looks like there is an issue with the updated version of cuTENSOR where it is not properly linking to the lib binaries. To get around that issue, use an older version of CUDA and cuTENSOR


mtfishman:
I was wondering what was going on with the CI, glad you figured out the cause.

Instead of changing the compat entries in `ITensorGPU/test/Project.toml`, I think it makes more sense to change them in `ITensorGPU/Project.toml`. Otherwise we are just fixing the tests, but we're still allowing people to install a version of `cuTENSOR` that we know breaks `ITensorGPU`.

I think this should work:
```julia
cuTENSOR = "1.0.0 - 1.0.1"
```
to not allow upgrading to the problematic `cuTENSOR v1.0.2` (the notation for `compat` entries is specified here: https://pkgdocs.julialang.org/v1/compatibility/).

Also, I don't see any issues about this in `CUDA.jl`, should we raise an issue over there?

ITensorGPU/test/Project.toml
```diff
@@ -1,5 +1,6 @@
 [deps]
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
+cuTENSOR = "011b41b2-24ef-40a8-b3eb-fa098493e9e1"
```
mtfishman, line 3: Do we need to add this as an explicit test dependency?

kmp5VT: I am verifying that now on my machine. I don't think we do

mtfishman:
Looks like this issue reported about CUDA.jl v4 and cuDNN may be related: https://github.com/JuliaGPU/CUDA.jl/issues/1850

PR #1110, opened 2023-04-06
Fix periodic square_lattice point

ryanlevy:
# Description

Fix `square_lattice` from https://itensor.discourse.group/t/multiple-exchange-interactions-in-2d-lattices/692
Previously the point pair was `(x,1)->(x,1+Ny)` and now will be `(x,1)->(x,Ny)`


emstoudenmire:
Perfect, thanks Ryan!

PR #1113, opened 2023-04-17
[ITensors] drop maxQNs, allow for dynamic setting of QN register

brian-dellabetta:
# Description

This is a draft PR to improve functionality for ITensor's QN registers to allow for more than values, currently hard-coded to `maxQNs=4`. This is needed for larger scale machine learning / optimization tasks. Rather than maintain our own fork, I want to see if this is something you'd be interested to have in the original repo. At larger register size, it is easier to handle much of the arithmetic with hash maps (`Dict` or `OrderedCollections.OrderedDict` type) rather than `StaticArray` types. If you are interested in merging this in, I can update the data struct to use Dict or OrderedDict entirely, rather than moving between `SVector`/`MVector` types. I wanted to post for feedback first before completing the full PR.

cc @yzcj105/@jingc769 and @ArtemStrashko

Fixes #(issue)

Allows for QN struct to have more than 4 QNVals

<details><summary>Previously, QN fails out for 5+ qnvals</summary><p>

```julia
julia> QN(("Sz", 0), ("Sx", 0), ("Sy", 2), ("X", 3), ("B", 4))
ERROR: BoundsError: attempt to access 4-element StaticArraysCore.MVector{4, ITensors.QNVal} with indices SOneTo(4) at index [5]
...
```
</p></details>

<details><summary>New implementation will succeed</summary><p>

```julia
julia> QN(("Sz", 0), ("Sx", 0), ("Sy", 2), ("X", 3), ("B", 4))
QN(("B",4),("Sx",0),("Sy",2),("Sz",0),("X",3))
```
</p></details>

# How Has This Been Tested?

Current tests already provide equality/inequality and ordering checks to ensure new code didn't break anything

- [x] Dropped `@test_throws BoundsError` when QN with 5 qnvals is created
- [x] Changed it to an equality check

# Checklist:

### DIFF
--- a/src/exports.jl
+++ b/src/exports.jl
@@ -386,6 +386,7 @@ export
   isfermionic,
   modulus,
   val,
+  QNVal,
 
   # qn/qnindex.jl
   blockdim,
--- a/src/qn/qn.jl
+++ b/src/qn/qn.jl
@@ -59,7 +59,7 @@ end
 const ZeroVal = QNVal("", 0, 0)
 
 """
-A QN object stores a collection of up to four
+A QN object stores a collection of up to N
 named values such as ("Sz",1) or ("N",0).
 These values can include a third integer "m"
 which makes them obey addition modulo m, for
@@ -71,13 +71,15 @@ addition and subtraction element-wise on each of
 the named values. If a name is missing from the
 collection, its value is treated as zero.
 """
-struct QN
-  data::SVector{<:Any,QNVal}
+struct QN{T}
+  data::T
 
-  QN() = new(SVector{0,QNVal}())
-  QN(s::SVector{<:Any,QNVal}) = new(s)
-  QN(m::MVector{<:Any,QNVal}) = new(SVector(m))
-  QN(mqn::NTuple{N,QNVal}) where {N} = new(SVector(mqn))
+  QN() = new{SVector{0,QNVal}}(SVector{0,QNVal}())
+  QN(s::SVector{N,QNVal}) where {N} = new{SVector{N,QNVal}}(s)
+  QN(m::MVector{N,QNVal}) where {N} = new{SVector{N,QNVal}}(SVector(m))
+  QN(mqn::NTuple{N,QNVal}) where {N} = new{SVector{N,QNVal}}(SVector(mqn))
+  QN(mqn::AbstractArray{<:QNVal}) = new{SVector{length(mqn),QNVal}}(SVector(mqn))
+  QN(d::Dict{TKey,QNVal}) where {TKey} = new{Dict{TKey,QNVal}}(d)
 end
 
 function hash(obj::QN, h::UInt)
@@ -85,7 +87,7 @@ function hash(obj::QN, h::UInt)
   # for performance here; put non-zero QNVals
   # to front and slice when passing to hash
   nzqv = QNVal[]
-  for qv in obj.data
+  for qv in data(obj)
     if val(qv) != 0
       push!(nzqv, qv)
     end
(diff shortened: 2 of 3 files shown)


mtfishman:
I could definitely imagine that limiting to 4 isn't great for more general applications. Out of curiosity, how many symmetries have you encountered in practice?

An alternative design suggestion I have would be to parametrize `QN` type by the maximum number of QNs:
```julia
struct QN{N}
  data::SVector{N,QNVal}
end
```
where `N` is the maximum number allowed. We could then still default it to `4` (or whatever we want), and then users could do `QN{5}(("Sz", 0), ("Sx", 0), ("Sy", 2), ("X", 3), ("B", 4))` if they want more (say potentially at the expense of some performance).

Usually we try to go for a one-size fits all in ITensor, especially for basic types like this, but it makes me uncomfortable going to a dynamic object like a `Dict` for this (I would worry about performance).

Also, we're essentially using sorting as a way to avoid using a hash map, if we went with a dynamically-sized storage type, what would be the issue with using a `Vector` that we keep sorted as opposed to a `Dict`?

brian-dellabetta:
@mtfishman I think I know the resolution not requiring a Dict. Lemme try and push a resolution later today

mtfishman:
I'm fine with using a `Dict` to implement functionality if that is the best way and it is fast, I was mostly referring to what the actual data structure of the `QN` object should be.

With 1000s of elements, a `StaticVector` or any fixed size array would be pretty bad, they aren't meant for storing that much data and become slow for both runtime and compile times.

In principle we could just parametrize the `QN` object to be anything, i.e.:
```julia
struct QN{T}
  data::T
end
```
and then you could use a dynamic object like a `Vector` or `Dict` if you want for those use cases. Our standard constructors could choose a `StaticVector`. We would just need to write the internal code generically so that it works efficiently for those different data storage types (maybe with some minimal required interface).

mtfishman:
Sounds good. I may not have time for an in-depth review in the near-term, but I'll try to give some feedback.

Julia doesn't have a trait syntax like that yet, at least not in the main language. There are packages like https://github.com/mauro3/SimpleTraits.jl but I don't think that's necessary here. Instead, we can just use "duck typing", i.e. don't constrain the type of `T` in `QN{T}`, and just expect the data of `QN` to implement some minimal set of functions (and the code will fail if they don't). Probably a requirement that they are `AbstractArray`s will be sufficient. I think the trickiest part will be making the code work generically and be performant for both mutable dynamic data storage like `Vector` and static, immutable data storage like `StaticVector`, but I think it won't be a problem if we design the code well and come up with good generic code patterns.

emstoudenmire:
Thanks, Brian for the suggestion and being flexible about how it might get implemented in ITensor.

A small side note (mostly for @mtfishman) is that a more flexible `QN{T}` type could be helpful for some future developments around non-Abelian and other symmetries too.

brian-dellabetta:
@mtfishman I just pushed some updates, so that both `AbstractArray{<:QNVal}` and `AbstractDict{<:QNVal}` types are allowed to store the QN information. I still can't figure out how to get the array-based addition to work when maxQNs is not set, I left a TODO and commented out the code snippet I thought would work. Can you take a look? 

This PR might take some more thought, it's not as clean as I had hoped, but I think in the meantime I can cherry pick what we need into our repo and drop our private fork, so no rush. Thanks for your time and consideration!

mtfishman:
I'll close this. We are in the process of rewriting the QN type to make it more agnostic about the storage backend with this kind of application in mind, so this should be possible in the near future once that lands.

PR #1119, opened 2023-04-26
[NDTensors][ITensorGPU] Debug CUDA and upkeep

kmp5VT:
Branch to remove `to_vector_type` fix some of the default typing problems and also fix the CUDA runtime problem without restricting CUDA version


ITensorGPU/test/Project.toml
```diff
@@ -5,3 +5,7 @@ ITensors = "9136182c-28ba-11e9-034c-db9fb085ebd5"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```
mtfishman, line 8: Do we need `Pkg` as a test dependency? I don't see it used anywhere.

kmp5VT: I don't think it is necessary

mtfishman:
Looks good, thanks!

jenkins/Jenkinsfile
```diff
@@ -27,7 +27,7 @@ pipeline {
           }
           steps {
             sh '''
-              julia --project=monorepo -e 'using Pkg; Pkg.Registry.update(); Pkg.develop(; path="./NDTensors"); Pkg.develop(; path="."); Pkg.develop(; path="./ITensorGPU"); Pkg.test("ITensorGPU");'
+              julia -e 'using Pkg; Pkg.Registry.update(); Pkg.develop(path="./NDTensors"); Pkg.develop(; path="./ITensorGPU"); Pkg.add("CUDA"); using CUDA; CUDA.set_runtime_version!("v11.8"); Pkg.test("ITensorGPU");'
```
mtfishman, line 30: This new version seems to drop testing against the development version of `ITensors` (the operation `Pkg.develop(; path=".")`), I think we should keep that.

kmp5VT: Okay, I added it back!

Issue #696, opened 2021-06-30
Generalize `inner(::MPO, ::MPS, ::MPO, ::MPS)`


mtfishman:
Currently, `inner(::MPO, ::MPS, ::MPO, ::MPS)` only works for MPO and MPS with standard prime level conventions, it should be generalized to arbitrary indices.

This is helpful for PEPS as well as applications in PastaQ, such as getting the fidelity between a mixed state represented as an LPDO and a pure state (@GTorlai).

PR #1124, opened 2023-05-11
[ITensors] Add prime indices for inner in documentation

ryanlevy:
# Description

`inner(x,A,x)` was [deprecated](https://github.com/ITensor/ITensors.jl/blob/1ef8f05566711127fc5d6aaad5b3d9c7e39ee582/src/mps/mpo.jl#L285) but the docs aren't updated to show this. I added this to the docs and `inner` documentation page. I left out the `Apply(A,x)` option but can add that in if you want to mention it. 

I should also note that `inner(A,x,A,x)` only works without primes at the moment (#696 )

# Checklist:


mtfishman:
Thanks! Too bad about the AD test failures, probably there was a regression in Zygote, which has been the case for previous AD test regressions (i.e. https://github.com/GTorlai/PastaQ.jl/issues/300).

emstoudenmire:
Thanks Ryan!

PR #1130, opened 2023-05-31
fix a bug: sum of MPOs via `directsum` algorithm falsely returns an MPS

phymann:
Previously `+(M1, M2; alg="directsum")` will return an `MPS`-type object despite that both M1 and M2 are `MPO`-type objects. After this one-line modification, the bug is fixed : )

### DIFF
--- a/src/mps/abstractmps.jl
+++ b/src/mps/abstractmps.jl
@@ -1474,7 +1474,7 @@ function +(::Algorithm"directsum", ψ⃗::MPST...) where {MPST<:AbstractMPS}
   )
   ϕj = replaceind(ϕj, ljm => dag(ljm_prev))
   ϕ[j] = ϕj
-  return ϕ
+  return convert(MPST, ϕ)
 end
 
 """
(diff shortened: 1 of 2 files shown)


src/mps/abstractmps.jl
mtfishman, line 1: Instead we could change this line:

https://github.com/ITensor/ITensors.jl/pull/1130/files#diff-30574d2b07b5598a5b432b188569f6d2daacd0e4ddae8a659a0fc257807c31a5L1449

to:
```julia
  ϕ = MPST(n)
```

phymann: Agreed 😉

mtfishman:
Thanks @phymann. Let's separate the changes to `MPO + MPO` and the `expect(::MPO, ...)` into different PRs to keep things simpler, I think the `expect` one will be more involved to review.

phymann:
Sure, I switch to a new PR and close this one, since I am not yet very familiar with this git/GitHub workflow.

PR #1133, opened 2023-06-15
ITensorGaussianMPS - bugfixes

ryanlevy:
# Description

Fixing some tiny things in GaussianMPS package:

- `hopping_hamiltonian`
  - Catch incorrect length terms before it hits quadrant
  - Better error message for terms like `"Ntot",i,"Ntot",i+1`
- API: I believe `blocksize` should be `maxblocksize`


mtfishman:
Thanks!

Issue #1127, opened 2023-05-24
[ITensors] [BUG] CI test fails for Julia 1.9 in test/ITensorChainRules/test_chainrules_ops.jl


JanReimers:
ITensor main branch: CI test fails for Julia 1.9 in `test/ITensorChainRules/test_chainrules_ops.jl`.
Partial stack dump:
```julia
test_rrule: #215 on Float64: Error During Test at /home/janr/.julia/packages/ChainRulesTestUtils/lERVj/src/testers.jl:202
  Got exception outside of a @test
  TypeError: in typeassert, expected Tuple{T, T} where T, got a value of type Tuple{Tangent{Any, NamedTuple{(:which_op, :sites, :params), Tuple{ZeroTangent, ZeroTangent, Tangent{Any, NamedTuple{(:θ,), Tuple{Float64}}}}}}, ZeroTangent}
  Stacktrace:
    [1] _totuple
      @ ./tuple.jl:386 [inlined]
    [2] Tuple
      @ ./tuple.jl:369 [inlined]
    [3] vect_pullback
      @ ~/.julia/packages/ChainRules/aKxNz/src/rulesets/Base/array.jl:41 [inlined]
    [4] (::Zygote.ZBack{ChainRules.var"#vect_pullback#1367"{2}})(dy::Vector{Union{Nothing, NamedTuple{(:which_op, :sites, :params), Tuple{Nothing, Nothing, NamedTuple{(:θ,), Tuple{Float64}}}}}})
      @ Zygote ~/.julia/packages/Zygote/IoW2g/src/compiler/chainrules.jl:205
    [5] Pullback
      @ ~/.julia/dev/ITensors/src/Ops/Ops.jl:188 [inlined]
    [6] (::typeof(∂(+)))(Δ::NamedTuple{(:f, :args, :kwargs), Tuple{Nothing, Tuple{Vector{Union{Nothing, NamedTuple{(:which_op, :sites, :params), Tuple{Nothing, Nothing, NamedTuple{(:θ,), Tuple{Float64}}}}}}}, Nothing}})
      @ Zygote ~/.julia/packages/Zygote/IoW2g/src/compiler/interface2.jl:0
    [7] Pullback
      @ ~/.julia/dev/ITensors/test/ITensorChainRules/test_chainrules_ops.jl:47 [inlined]
    [8] (::typeof(∂(#215)))(Δ::Float64)
[...]

PR #1135, opened 2023-06-20
[NDTensors] Fix output type of Empty BlockSparse contraction

mtfishman:
The output type of Empty BlockSparse contraction was a bit off, and not adjusting for the dimensions of the output tensor. That caused some subtle bugs in other operations that were expecting the types to be the same, see:

https://github.com/ITensor/ITensorInfiniteMPS.jl/pull/77#discussion_r1232428688

@LHerviou @kmp5VT


mtfishman:
Closes #1127 by marking that test as broken for Julia versions 1.9 and above. @kmp5VT I know you also addressed that one in #1097 so your fix can be reverted.

PR #1142, opened 2023-07-07
[NDTensors][Feature] Introduce FillArrays to replace EmptyStorage

kmp5VT:
# Description

The goal here is to work on developing a better system for empty storage type and replace this structure. Fill arrays are a lazy way to create full arrays that are actually a single value. This will also be useful for uniform diagonal tensors.

# Checklist:


kmp5VT:
So what I am doing here is implementing `EmptyStorage` as a true storage type which holds a data type that is always a `FillArray.Zeros`. the `Zeros` is a placeholder for an array full of zeros and has little physical memory. The zeros is nice because memory can be allocated using something like this `Array(Zeros{ElT}(dimension))`.  In general I am imagining that the EmptyType will get "transformed" into another storage type manually (like `Diag(emptystorage)`) or via inspection of the indices of the `EmptyStorageTensor`

mtfishman:
I was actually picturing just removing `EmptyStorage` entirely, and that functionality would get replaced by other storage types like `Dense` or `Diag` which happen to have `FillArray.Zeros` as their data.

mtfishman:
The motivation would be that having extra storage types is a pain, and it has been hard to keep up with all of the operations that need to get defined for the `EmptyStorage` type (basically having fewer types is just easier both for us and for users). I haven't thought through all of the implications of getting rid of `EmptyStorage` but I don't see why it would be an issue, since we can always write specialized functions by dispatching on when the data is `FillArray.Zeros`.

kmp5VT:
As of now, I can't imagine any issues with removing the EmptyStorage/EmptyTensor type. Would you still want the EmptyNumber type? Its possible to construct `Zeros{EmptyNumber}`

mtfishman:
I also am not a huge fan of the `EmptyNumber` type. If I recall correctly, it served two purposes:

1. As a placeholder for when you don't know what the element type is going to be (since otherwise you can just use a concrete element type), for example when making an ITensor with `ITensor(i, j, k)`, where the element type is left unset and it is determined by the first element that gets set. However, I think we could use `Union{}` as a placeholder for not knowing what the element type will be, since that is Julia's universal "bottom" type (the opposite of `Any`).
2. As a "universal zero value", in which case it could be replaced by the `Zero` type from [Zeros.jl](https://github.com/perrutquist/Zeros.jl). I believe I may have tried that at some point but I can't remember why I didn't end up using that.

So let's make it a goal to try to remove `EmptyNumber` as well.

Yes, it will definitely have some deep implications throughout the code, but in the end I think it will make the library much simpler (i.e. remove a lot of code, and also "automatically fix" many lingering bugs related to `EmptyStorage`), and as you say make it easier to implement new storage and data type backends.

kmp5VT:
I started from scratch to make life easier.

PR #1148, opened 2023-07-12
[ITensors] Ignore Tag and TagSet overflow

mtfishman:
# Description

Tags currently have a limit of 16 characters and TagSets have a limit of 4 tags. We can investigate making these limits customizable or increasing their defaults, but before we do that it is a pain to run into errors from using too many tags or tags that are too long, which is becoming pretty common in high dimensional tensor networks in ITensorNetworks. This PR just ignores extra tag characters and tags, for example:
```julia
julia> TagSet("ijklmnopqabcdefgh")
"ijklmnopqabcdefg"

julia> TagSet("a,b,c,d,e")
"a,b,c,d"
```

@emstoudenmire curious what you think about this. I think the only downside is if someone is relying on making unique tags/TagSets that then don't end up unique because some tags or characters get ignored, but that is pretty easily avoidable in practice.

### DIFF
--- a/src/tagset.jl
+++ b/src/tagset.jl
@@ -86,8 +86,7 @@ function GenericTagSet{T,N}(str::AbstractString) where {T,N}
         nchar = 0
       end
     elseif current_char != ' ' # TagSet constructor ignores whitespace
-      nchar == length(current_tag) &&
-        error("Currently, tags can only have up to $(length(current_tag)) characters")
+      nchar == length(current_tag) && continue
       nchar += 1
       @inbounds current_tag[nchar] = current_char
     end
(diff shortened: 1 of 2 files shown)


mtfishman:
Good idea. I added the following so a user can optionally enable strict tag checking:
```julia
julia> ITensors.using_strict_tags()
false

julia> TagSet("abcdefghijklmnopqr")
"abcdefghijklmnop"

julia> ITensors.set_strict_tags!(true)
false

julia> TagSet("abcdefghijklmnopqr")
ERROR: You are trying to make a TagSet from the String "abcdefghijklmnopqr". This has more than the maximum number of allowed tags (4), or has a tag that is longer than the longest allowed tag (16). Either specify fewer or shorter tags, or use `ITensors.set_strict_tags!(false)` to disable this error, in which case extra tags or tag characters will be ignored.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] strict_tags_error(str::String, N::Int64, nchar::Int64)
   @ ITensors ~/.julia/dev/ITensors/src/tagset.jl:73
 [3] TagSet(str::String)
   @ ITensors ~/.julia/dev/ITensors/src/tagset.jl:99
 [4] top-level scope
   @ REPL[20]:1
```

emstoudenmire:
Great!

PR #1151, opened 2023-07-14
[NDTensors] Fix performance issue with NDTensors contract

kmp5VT:
# Description

There was a significant discrepancy between BLAS timings and NDTensors contraction timings. Implement a generic fix that helps the performance of CPU and GPU tensor contractions

### DIFF
--- a/NDTensors/src/dense/tensoralgebra/contract.jl
+++ b/NDTensors/src/dense/tensoralgebra/contract.jl
@@ -331,7 +331,6 @@ function _contract!(
   A = array(AT)
   B = array(BT)
 
-  @show typeof(A)
   return _contract!(C, A, B, props, α, β)
 end
 
--- a/NDTensors/src/dense/tensoralgebra/contract.jl
+++ b/NDTensors/src/dense/tensoralgebra/contract.jl
@@ -327,21 +327,35 @@ function _contract!(
   α::Number=one(El),
   β::Number=zero(El),
 ) where {El,NC,NA,NB}
-  # TODO: directly use Tensor instead of Array
+  C = array(CT)
+  A = array(AT)
+  B = array(BT)
 
+  @show typeof(A)
+  return _contract!(C, A, B, props, α, β)
+end
+
+function _contract!(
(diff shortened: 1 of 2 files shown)


NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -327,21 +327,35 @@ function _contract!(
   α::Number=one(El),
   β::Number=zero(El),
 ) where {El,NC,NA,NB}
-  # TODO: directly use Tensor instead of Array
+  C = array(CT)
+  A = array(AT)
+  B = array(BT)
 
+  @show typeof(A)
```
mtfishman, line 334: Should be removed.

mtfishman:
Thanks for the quick fix @kmp5VT! I'll merge and register a new version asap. @JoeyT1994

mtfishman:
[benchmark]

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -379,7 +392,7 @@ function _contract!(
 
   if props.permuteC
     pC = NTuple{NC,Int}(props.PC)
-    Cr = ReshapedArray(CM.parent, props.newCrange, ())
+    Cr = reshape(CM.parent, props.newCrange)
```
mtfishman, line 395: Looks like `CM.parent` is causing issues since `CM` isn't a wrapped array anymore. Probably can just change to `CM`.

mtfishman: @kmp5VT I tried changing it, let's see if it passes.

mtfishman:
[benchmark]

mtfishman:
@kmp5VT could you also follow the instructions here: https://github.com/ITensor/ITensors.jl/tree/v0.3.37/benchmark#development to update the benchmark manifest to see if we can get the benchmarks running?

emstoudenmire:
This is really great, Karl. Thanks for identifying the issue so quickly and getting it fixed.

PR #1153, opened 2023-07-19
[NDTensors] Use SetParameters in GPU extensions

mtfishman:
Use `SetParameters` module in GPU extensions to simplify setting type parameters of different `AbstractArray` data backends.


mtfishman:
@kmp5VT this made it easier to get more of the NDTensorMetal extension working, but there are still some issues with handling conversion from double to single precision in that case. For example:
```julia
julia> using ITensors

julia> using Metal

julia> NDTensors.mtl(randomITensor(Index(2)))
ERROR: Metal does not support Float64 values, try using Float32 instead
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] MtlVector{Float64, Metal.MTL.MTLResourceStorageModePrivate}(#unused#::UndefInitializer, dims::Tuple{Int64})
    @ Metal ~/.julia/packages/Metal/qeZqc/src/array.jl:45
[...]
```

kmp5VT:
@mtfishman The double precision problem you're quoting is actually rather significant in Metal they simply do not have operations implemented for Float64 vectors so we should not even try to construct them at this point.

mtfishman:
Yeah, seems like we should just aggressively convert to single precision in functions like `adapt` (like the `Metal.mtl` function does).

mtfishman:
Ok, I ended up keeping `adapt` and related functions like `NDTensors.mtl` as-is (i.e. not converting automatically to single precision), but changed the tests to use single precision when the Metal backend is being tested. I now have the NDTensors tests passing locally with and without the Metal backend on my macbook using the new design based on `SetParameters`.

However, there is a strange issue on CI, it's complaining that  the function `get_parameters` isn't defined. This is a new function I introduced in the `NDTensors.SetParameters` module in this PR. My tests pass locally so that function is defined properly. @kmp5VT do you know what might be going on? It seems like it isn't seeing the latest NDTensors version from this PR branch, however I do see in the CI log that it is using the dev'd version of NDTensors:
```julia
  [23ae76d9] NDTensors v0.2.2 `~/.julia/dev/ITensors/NDTensors`
```
Maybe it's using the `main` branch instead of this PR branch?

mtfishman:
@kmp5VT I noticed that the CI log shows this for ITensors:
```julia
  [9136182c] ITensors v0.3.37 `~/work/ITensors.jl/ITensors.jl`
```
but this for NDTensors:
```
  [23ae76d9] NDTensors v0.2.2 `~/.julia/dev/ITensors/NDTensors`
```
so it seems like maybe for ITensors it is using the PR branch while for NDTensors it's using the `main` branch.

mtfishman:
Switching lines like:
```julia
Pkg.develop("NDTensors")
```
to:
```julia
Pkg.develop(path="./NDTensors")
```
seems to fix that issue and get the tests to run based on the version of NDTensors on the PR branch.

However, there's still an issue running the `ITensorGPU` tests on Jenkins, where it's also complaining that the `get_parameters` isn't defined.

mtfishman:
[test ITensors mps]

PR #1155, opened 2023-07-20
[Debug][NDTensors] Unwrap reshapedArray function

kmp5VT:
Currently reshapedarrays with non-CPU backend do not work properly in the mul! function (they are considered CPU arrays and `BLAS.gemm!` is called on them which results in an addressing error) Created a simple function which unwraps the ReshapedArrays computes the gemm and then rewrites the data back to the correct piece of memory. This is especially important in views.

### DIFF
--- a/NDTensors/src/dense/tensoralgebra/contract.jl
+++ b/NDTensors/src/dense/tensoralgebra/contract.jl
@@ -334,6 +334,20 @@ function _contract!(
   return _contract!(C, A, B, props, α, β)
 end
 
+function _contract!(C::ReshapedArray{El}, 
+  A::ReshapedArray{El},
+  B::ReshapedArray{El},
+  props::ContractionProperties,
+  α::Number=one(El),
+  β::Number=zero(El)) where {El}
+  Cp = copy(C)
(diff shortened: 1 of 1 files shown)


NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -334,6 +334,20 @@ function _contract!(
...
+  B::ReshapedArray{El},
+  props::ContractionProperties,
+  α::Number=one(El),
+  β::Number=zero(El)) where {El}
+  Cp = copy(C)
+  _contract!(Cp, copy(A), copy(B), props, α, β)
+  Cp = reshape(Cp, (length(C)))
+  for i in 1:length(C)
+    C.parent[i] = Cp[i]
+  end
```
mtfishman, line 348: Probably can just use `copy!(C, Cp)` for this.

kmp5VT: Makes sense, I replaced this with `copy!`

mtfishman:
Superseded by other PRs.

PR #1158, opened 2023-07-22
[NDTensors] [BUG] Fix bug in in-place contract

kmp5VT:
# Description

Typo in the contract code, it tried to create a tuple of size `NB` instead of `NC` which will fail if `NC != NB`

### DIFF
--- a/NDTensors/src/dense/tensoralgebra/contract.jl
+++ b/NDTensors/src/dense/tensoralgebra/contract.jl
@@ -377,10 +377,16 @@ function _contract!(
   if props.permuteC
     # if we are computing C = α * A B + β * C
     # we need to make sure C is permuted to the same 
-    # ordering as A B
+    # ordering as A B which is the inverse of props.PC
     if β ≠ 0
       pC = NTuple{NC,Int}(props.PC)
-      CM = reshape(permutedims(CT, pC), (props.dleft, props.dright))
+      pC_inv = Vector{Int}(undef, NC)
+      for i in 1:length(pC)
+        pC_inv[pC[i]] = i
+      end
+      pC_inv = NTuple{NC, Int}(pC_inv)
+
+      CM = reshape(permutedims(CT, pC_inv), (props.dleft, props.dright))
     else
       # Need to copy here since we will be permuting
       # into C later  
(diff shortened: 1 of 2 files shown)


NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -377,10 +377,16 @@ function _contract!(
...
+    # ordering as A B which is the inverse of props.PC
     if β ≠ 0
-      pC = NTuple{NB,Int}(props.PC)
-      CM = reshape(permutedims(CT, pC), (props.dleft, props.dright))
+      pC = NTuple{NC,Int}(props.PC)
+      pC_inv = Vector{Int}(undef, NC)
+      for i in 1:length(pC)
+        pC_inv[pC[i]] = i
+      end
+      pC_inv = NTuple{NC,Int}(pC_inv)
```
mtfishman, line 387: Note there is a Julia function `invperm`:
```julia
julia> invperm((3, 1, 2))
(2, 3, 1)
```

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -377,10 +377,16 @@ function _contract!(
   if props.permuteC
     # if we are computing C = α * A B + β * C
     # we need to make sure C is permuted to the same 
-    # ordering as A B
+    # ordering as A B which is the inverse of props.PC
     if β ≠ 0
-      pC = NTuple{NB,Int}(props.PC)
-      CM = reshape(permutedims(CT, pC), (props.dleft, props.dright))
+      pC = NTuple{NC,Int}(props.PC)
```
mtfishman, line 382: The field `PC` is already a Tuple so this conversion shouldn't be necessary: https://github.com/ITensor/ITensors.jl/blob/e4779bdb8c2760028cea727a60ac602ed76930e0/NDTensors/src/tensoralgebra/contraction_logic.jl#L145

NDTensors/src/dense/tensoralgebra/contract.jl
kmp5VT: Great, I tried it and it works!

NDTensors/src/dense/tensoralgebra/contract.jl
kmp5VT: Okay, I just followed what was already in the code. I removed all of the definitions of `p_ = NTuple{N_, Int}(props.P_` where `_` is `A, B, or C`

mtfishman: Thanks. I think it was like that because at some point those fields of `ContractionProperties` had a different type so a conversion was necessary, but the rest of the code wasn't updated accordingly.

PR #1161, opened 2023-07-26
[NDTensors] In the effort to replace EmptyStorage with an empty DataType

kmp5VT:
# Description
In general there can be problems with the EmptyStorage type as it does not act like the other storage types (i.e. store a datatype and a element type). This requires us to often solve corner cases. Replacing EmptyStorage with a Zeros for the datatype with understanding of what kind of datatype is being used (not simply FillArrays::Zeros). Additionally, use `FillArrays::Diag` as a replacement for `uniformDiag` number storage. 

# Checklist:


kmp5VT:
@mtfishman  I am having an issue here where code like `ITensor(Index(2))` is failing because of 
```
ERROR: Setting the type parameter of the type `NDTensors.Zeros{Float64, 1, Vector{Float64}}` at position `NDTensors.SetParameters.Position{1}()` to `Float64` is not currently defined. Either that type parameter position doesn't exist in the type, or `set_parameter` has not been overloaded for this type.
Stacktrace:
[1] error(s::String)
    @ Base ./error.jl:35
  [2] set_parameter(type::Type, position::NDTensors.SetParameters.Position{1}, parameter::Type)
    @ NDTensors.SetParameters ~/.julia/dev/ITensors/NDTensors/src/SetParameters/src/interface.jl:10
  [3] set_parameters(type::Type, position::NDTensors.SetParameters.Position{1}, parameter::Type)
    @ NDTensors.SetParameters ~/.julia/dev/ITensors/NDTensors/src/SetParameters/src/set_parameters.jl:24
  [4] set_eltype(arraytype::Type{NDTensors.Zeros{Float64, 1, Vector{Float64}}}, eltype::Type)
    @ NDTensors ~/.julia/dev/ITensors/NDTensors/src/abstractarray/set_types.jl:8
  [5] ITensor(as::NDTensors.AllowAlias, eltype::Type{Float64}, A::NDTensors.Zeros{Float64, 1, Vector{Float64}}, inds::Tuple{Index{Int64}}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ ITensors ~/.julia/dev/ITensors/src/itensor.jl:352
  [6] ITensor(as::NDTensors.AllowAlias, eltype::Type{Float64}, A::NDTensors.Zeros{Float64, 1, Vector{Float64}}, inds::Tuple{Index{Int64}})
    @ ITensors ~/.julia/dev/ITensors/src/itensor.jl:340
  [7] itensor(::Type, ::Vararg{Any}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ ITensors ~/.julia/dev/ITensors/src/itensor.jl:134
  [8] itensor(::Type, ::NDTensors.Zeros{Float64, 1, Vector{Float64}}, ::Vararg{Any})
    @ ITensors ~/.julia/dev/ITensors/src/itensor.jl:134
  [9] ITensor(ElT::Type{Float64}, is::Tuple{Index{Int64}})
    @ ITensors ~/.julia/dev/ITensors/src/itensor.jl:178
 [10] ITensor(ElT::Type{Float64}, is::Index{Int64})
    @ ITensors ~/.julia/dev/ITensors/src/itensor.jl:181
 [11] ITensor(is::Index{Int64})
    @ ITensors ~/.julia/dev/ITensors/src/itensor.jl:183
 [12] top-level scope
    @ REPL[8]:1
```
The code is ignoring my definition of `set_parameter` for the Zeros type. I am wondering if you can see if I am doing something wrong? Thanks!

src/itensor.jl
```diff
@@ -271,25 +274,25 @@ ITensor(x::RealOrComplex{Int}, is...) = ITensor(float(x), is...)
...
-emptyITensor(is::Indices) = emptyITensor(EmptyNumber, is)
+emptyITensor(is::Indices) = emptyITensor(NDTensors.default_eltype(), is)
 
-emptyITensor(is...) = emptyITensor(EmptyNumber, indices(is...))
+emptyITensor(is...) = emptyITensor(NDTensors.default_eltype(), indices(is...))
 
-function emptyITensor(::Type{ElT}=EmptyNumber) where {ElT<:Number}
-  return itensor(EmptyTensor(ElT, ()))
+function emptyITensor(::Type{ElT}=NDTensors.default_eltype()) where {ElT<:Number}
+  return itensor(NDTensors.Zeros{ElT,1,NDTensors.default_datatype(ElT)}(()), ())
```
mtfishman, line 295: Seems like now would be a good time to deprecate these. Probably we can just remove these definitions and put:
```julia
@deprecate emptyITensor(args...; kwargs...) ITensor(args...; kwargs...)
```
in `deprecated.jl` since it should just be doing the same thing as the default `ITensor(...)` constructor.

NDTensors/src/tensor/tensor.jl
```diff
@@ -184,6 +188,8 @@ setstorage(T, nstore) = tensor(nstore, inds(T))
 
 setinds(T, ninds) = tensor(storage(T), ninds)
 
+isemptystorage(t::Tensor) = datatype(t) isa NDTensors.Zeros
```
mtfishman, line 191: I think it would make sense to rename this, since there won't be an `EmptyStorage` type going forward. Maybe `iszerodata`?

NDTensors/src/zeros/zeros.jl
```diff
@@ -0,0 +1,35 @@
...
+    @assert ndims(DataT) == N
+    z = FillArrays.Zeros(ElT, dim(inds))
+    return new{ElT,N,DataT}(z, inds)
+  end
+end
+
+Base.ndims(::NDTensors.Zeros{ElT,N}) where {ElT,N} = N
+ndims(::NDTensors.Zeros{ElT,N}) where {ElT,N} = N
+Base.eltype(::Zeros{ElT}) where {ElT} = ElT
+datatype(::NDTensors.Zeros{ElT,N,DataT}) where {ElT,N,DataT} = DataT
```
mtfishman, line 15: Maybe a better name would be something like `allocated_type`, `datatype` is pretty generic and has a different meaning in the rest of the library.

NDTensors/src/tensor/tensor.jl
```diff
@@ -107,6 +107,10 @@ function Tensor(datatype::Type{<:AbstractArray}, inds::Tuple)
   return Tensor(generic_zeros(datatype, dim(inds)), inds)
 end
 
+function Tensor()
+  return Tensor(Zeros{default_eltype(),1,default_datatype(default_eltype())}(()), ())
```
mtfishman, line 111: Is the plan to use `default_storagetype` for this kind of thing? Also seems like this could just be defined as: `Tensor() = Tensor(())`.

kmp5VT: This type of definition might be overkill, since I am simply calling the default function, I was using this as a proof of principle to test if the `Zeros` type was working.

mtfishman: Ok, no problem. Definitely understand this is a draft where you are trying different things out and it's not in a final state, I can hold off on reviewing for now and you can let me know when you think it is ready for review.

src/itensor/diagitensor.jl
```diff
@@ -84,12 +65,12 @@ In the case when `x isa Union{Int, Complex{Int}}`, by default it will
 be converted to `float(x)`. Note that this behavior is subject to change
 in the future.
 """
-function diagITensor(as::AliasStyle, eltype::Type{<:Number}, x::Number, is::Indices)
-  return diagITensor(AllowAlias(), eltype, fill(eltype(x), mindim(is)), is...)
+function diagITensor(::AliasStyle, ElT::Type{<:Number}, x::Number, is::Indices)
```
mtfishman, line 68: I actually prefer the style of using `eltype` instead of `ElT`, when it is the argument of a function.

kmp5VT: Is the name `eltype` a typical variable name for this type of thing in julia? My reason for changing the variable name is that eltype is also a Base function name.

mtfishman: I use it when I know I won't use the function `eltype` in the function, or if you want to use the function you can use `Base.eltype` explicitly. But more importantly I prefer the style of using lower case for argument names. So a compromise could be `elt` or something like that.

NDTensors/src/zeros/zeros.jl
```diff
@@ -0,0 +1,46 @@
...
+
+Base.convert(x::Type{T}, z::NDTensors.Zeros) where {T<:Array} = Base.convert(x, z.z)
+
+Base.getindex(a::Zeros, i) = Base.getindex(a.z, i)
+Base.sum(z::Zeros) = sum(z.z)
+LinearAlgebra.norm(z::Zeros) = norm(z.z)
+setindex!(A::NDTensors.Zeros, v, I) = setindex!(A.z, v, I)
+
+Base.iszero(t::Tensor) = iszero(storage(t))
+Base.iszero(st::TensorStorage) = data(st) isa Zeros
```
mtfishman, line 38: I don't think this definition is a good idea, `Base.iszero` is more general than that:
```julia
julia> iszero([0, 0])
true
```

A function to check if the data type is `NDTensors.Zeros` would be useful, but we should come up with a different name for it. Maybe `is_unallocated_zeros`?

src/itensor/diagitensor.jl
```diff
@@ -0,0 +1,109 @@
...
+    diagitensor([elt::Type, ]x::Number, inds...)
+
+Make a sparse ITensor with non-zero elements only along the diagonal.
+In general, the diagonal elements will be set to the value `x` and
+the ITensor will have element type `eltype(x)`, unless specified explicitly
+by `ElT`. The storage will have `NDTensors.Diag` type.
+
+In the case when `x isa Union{Int, Complex{Int}}`, by default it will
+be converted to `float(x)`. Note that this behavior is subject to change
+in the future.
```
mtfishman, line 68: I don't think this is the case if `elt` is specified explicitly, in that case `elt` will be taken literally. The automatic conversion should only happen if only `x` is specified, and is `Int` or `Complex{Int}`.

kmp5VT: @mtfishman I didn't change this comment or behavior of this code, I only moved the `diagITensor` to its own file see https://github.com/ITensor/ITensors.jl/blob/0a5a932eae91b5b8104ba6f19537003c6b17816f/src/itensor.jl#L447

mtfishman: Ah ok! Even so, could you add a part to the end of that sentence like "...if an element type isn't specified explicitly."

NDTensors/src/zeros/zeros.jl
```diff
@@ -0,0 +1,46 @@
+struct Zeros{ElT,N,DataT} <: AbstractArray{ElT,N}
+  z::FillArrays.Zeros
+  is::Tuple
+  function NDTensors.Zeros{ElT,N,DataT}(inds::Tuple) where {ElT,N,DataT}
+    @assert eltype(DataT) == ElT
+    @assert ndims(DataT) == N
+    z = FillArrays.Zeros(ElT, dim(inds))
+    return new{ElT,N,DataT}(z, inds)
+  end
+end
```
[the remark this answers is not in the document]
kmp5VT: Okay I updated with Axes.  Right now I have it possible to make higher order `NDTensor.Zeros` types but the Alloc type does not depend on N. Also, when a `NDTensors.Zeros` is constructed in ITensors, I flatten the indices down and just store a vector (since the true indices will be stored in the `Tensor`).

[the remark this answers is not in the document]
mtfishman: What do you mean by "the Alloc type does not depend on N"? It seems simplest to constrain it so that the `Alloc` type is some `AbstractArray` that is order `N`.

src/deprecated.jl
```diff
@@ -32,6 +32,9 @@
 @deprecate setstore(T::ITensor, st) setstorage(T, st) false
 @deprecate uniqueindex(args...; kwargs...) uniqueind(args...; kwargs...)
 
+# itensor/emptyITensor
+@deprecate emptyITensor(args...; kwargs...) ITensor(args...; kwargs...)
```
mtfishman, line 36: Note that with this, we should be able to just delete all of the code for `emptyITensor` elsewhere, presuming that its functionality is subsumed by the `ITensor` constructor.

kmp5VT: Makes sense, thanks!

NDTensors/src/zeros/zeros.jl
```diff
@@ -0,0 +1,45 @@
+struct Zeros{ElT,N,Axes,Alloc<:AbstractArray} <: AbstractArray{ElT,N}
+  z::FillArrays.Zeros
+  function NDTensors.Zeros{ElT,N,Alloc}(inds::Tuple) where {ElT,N,Alloc}
+    @assert eltype(Alloc) == ElT
```
mtfishman, line 4: This would get constrained automatically if you write the type definition as:
```julia
struct Zeros{ElT,N,Axes,Alloc<:AbstractArray{ElT,N}} <: AbstractArray{ElT,N}
```

kmp5VT: I guess I am wondering if `Alloc` needs to have `ElT` and `N` since its functionally just a place holder?  (this follows from the thread before) We could just store `Zero{Float32, 20, (ind_1, ind_2, ..., ind_20), Vector}` and then process the `Alloc` properly when its actually instantiated later?

mtfishman: I agree that in principle we could design it that way, but it seems better to me to store it in a way where it can be directly used to allocate an array instead of having to process the type first before being able to use it. For example, then the `Alloc` type could only be used properly by some accessor function that filled in the type parameters on the fly (say `allocated_type(::NDTensors.Zeros)` or something like that) but if you access the type from the type parameter directly it wouldn't be complete. Seems better to process the type properly right off the bat.

Something we could do so that it isn't too annoying to construct a `NDTensors.Zeros` with a custom allocated array type is have a constructor:
```julia
function Zeros(alloc::Type{<:AbstractArray}, dims...)
  # 
end

function Zeros{ElT}(alloc::Type{<:AbstractArray}, dims...) where {ElT}
  # 
end
```
which could process the allocated type `alloc` to fill in missing parameters based on `dims` and possibly a default element type if the element type of `alloc` or `Zeros` isn't specified. That way you could be a bit looser about how you define it to begin with, and use `Zeros(Array, 2, 2)`, `Zeros{Float64}(Array, 2, 2)`, `Zeros(Array{Float64}, 2, 2)`, etc. That would be helpful for GPU types like `CuArray` which have the extra buffer type.

src/itensor/oneitensor.jl
```diff
@@ -42,3 +42,39 @@ end
 function emptyITensor(elt::Type{<:Number}=NDTensors.default_eltype())
   return emptyITensor(elt, Index(0))
 end
+
+# To fix ambiguity with QN version
+function randomITensor(::Type{ElT}, is::Tuple{}) where {ElT<:Number}
+  return randomITensor(Random.default_rng(), ElT, Index(0))
```
mtfishman, line 48: I don't think it is right to use `Index(0)`. Can't it just have an empty set of indices?

mtfishman: In that case it seems like you are just missing an explicit definition like:
```julia
ITensor(aliasstyle::AllowAlias, elt::Type{<:Number}, A::AbstractArray, inds::Tuple{})
```
where you can define it to create a tensor with `Dense` storage.

NDTensors/src/dense/densetensor.jl
```diff
@@ -93,6 +93,12 @@ end
 @propagate_inbounds function setindex!(
   T::DenseTensor{<:Number}, x::Number, I::Vararg{Integer}
 )
+  if is_unallocated_zeros(T)
+    ElT = promote_type(eltype(data(T)), typeof(x))
+    d = similartype(alloctype(data(T)), ElT)(undef, dim(to_shape(typeof(data(T)), inds(T))))
+    fill!(d, zero(ElT))
+    T = Tensor(Dense(d), inds(T))
+  end
```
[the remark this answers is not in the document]
mtfishman: Also I'm sure there are many places we will want to allocate a `Zeros` or a `TensorStorage` or `Tensor` wrapping `Zeros` data, probably good to define a generic function (maybe called `allocate`) which allocates `Zeros` data based on `alloctype`, and can do that for `TensorStorage` and `Tensor` types. Seems like `adapt` would be a good way to implement that generically for types like `TensorStorage` and `Tensor`, maybe through a new `AllocAdaptor`.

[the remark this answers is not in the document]
kmp5VT: Yeah I thought I might have put this operation in the wrong location. I can fix that. I definitely agree, I already found that I need to implement an allocate (and type promotion) in the blocksparse setindex!! code as well so it would make sense to have a easily reusable function to simplify the process!

NDTensors/src/tensor/tensor.jl
```diff
@@ -184,6 +188,8 @@ setstorage(T, nstore) = tensor(nstore, inds(T))
 
 setinds(T, ninds) = tensor(storage(T), ninds)
 
+iszerodata(t::Tensor) = data(t) isa NDTensors.Zeros
```
mtfishman, line 191: I think this would be better:
```julia
iszerodata(t::Tensor) = iszerodata(storage(t))
iszerodata(s::TensorStorage) = iszerodata(data(s))
iszerodata(::NDTensors.Zeros) = true
iszerodata(::Any) = false
```

Also I think `iszerodata` doesn't sound quite right for the name of this function, since a `Tensor` might have Zeros data, but it isn't itself Zeros data (or data at all, by our terminology). Also `zero` seems too general since there could be tensors with allocated numerical zeros. Maybe `is_unallocated_zero`?

We could design this around a more generic function:
```julia
data_isa(t::Tensor, datatype::Type) = data_isa(storage(t), datatype)
data_isa(s::TensorStorage, datatype::Type) = data_isa(data(s), datatype)
data_isa(d::AbstractArray, datatype::Type) = d isa datatype # Might have to unwrap if it is reshaped, sliced, etc.
```
and then define:
```julia
is_unallocated_zero(a) = data_isa(a, NDTensors.Zeros)
```
This could generalize to a function `storage_isa` which could be used to define `iscombiner`, `isdense`, etc.

NDTensors/src/unspec_eltype/unspecified_zero.jl
```diff
@@ -4,6 +4,8 @@
 
 struct UnspecifiedZero <: Base.AbstractFloat end
 
+Base.Complex{UnspecifiedZero}() = complex(UnspecifiedZero())
+Base.Complex{UnspecifiedZero}(z::Real) = (z == zero(eltype(z)) ? complex(UnspecifiedZero()) : throw(ErrorException) )
```
mtfishman: Should be able to use `iszero(z)` as a simpler version of `z == zero(eltype(z))`.

kmp5VT: Thanks, I updated it!

NDTensors/src/tensor/tensor.jl
kmp5VT: I really like the `data_isa` function, its again a home brewed version trait function but I think its good when we can so dynamically define what a datatype is in the backend. I have implemented them in the code and made some updates

NDTensors/src/abstractarray/data_isa.jl
```diff
@@ -0,0 +1,4 @@
+data_isa(t::Tensor, datatype::Type) = data_isa(storage(t), datatype)
+data_isa(s::TensorStorage, datatype::Type) = data_isa(data(s), datatype)
+data_isa(d::AbstractArray, datatype::Type) = d isa datatype # Might have to unwrap if it is reshaped, sliced, etc.
+data_isa(n::Number, datatype::Type) = false
```
mtfishman, line 4: What's this definition for?

mtfishman: Seems a bit strange to ask for the data type of a `Number`, since it doesn't have data.

NDTensors/src/tensor/tensor.jl
mtfishman, line 191: Guess this can be removed now?

NDTensors/src/tensor/tensor.jl
```diff
@@ -347,7 +353,10 @@ end
 # Some generic getindex and setindex! functionality
 #
 
-@propagate_inbounds @inline setindex!!(T::Tensor, x, I...) = setindex!(T, x, I...)
+@propagate_inbounds @inline function setindex!!(T::Tensor, x, I...)
+  T = allocate(T, typeof(x))
```
mtfishman, line 357: This relates to the message I sent on slack. What happens if the element type of `T` is already specified (say constructed with `ITensor(Float32, i, j)`)? I think we should preserve the element type if it was already specified, and only set it to `typeof(x)` if it is `UnspecifiedZero`, otherwise it could lead to surprising behavior where a real tensor could get switched to complex, for example.

Maybe it is best to leave that logic out of `allocate`, and instead use the semantics:
```julia
T = allocate(set_eltype_if_unspecified(T, typeof(x)))
```

NDTensors/src/tensoralgebra/generic_tensor_operations.jl
```diff
@@ -124,6 +124,10 @@ function _contract!!(
       β,
     )
   else
+    # Special case when ther are unallocated zeros return nothing do no work.
+    if is_unallocated_zeros(tensor1) || is_unallocated_zeros(tensor2)
+      return output_tensor
```
mtfishman, line 129: One exception I could imagine for this is if the element type of `output_tensor` is `UnspecifiedZero`, while the element types of `tensor1` and/or `tensor2` are specified. Then, I think the element type of `output_tensor` should become specified based on the element types of `tensor1` and `tensor2`. So maybe it would be:
```julia
return set_eltype_if_unspecified(output_tensor, promote_type(eltype(tensor1), eltype(tensor2))
```

I'm not sure if the types of `α` and `β` should play a role, since I think we want that if the element types of `tensor1` and `tensor2` are unspecified, then `output_tensor` would remain unspecified.

NDTensors/src/tensor/tensor.jl
kmp5VT: This has been removed

NDTensors/src/abstractarray/data_isa.jl
kmp5VT: As discussed this is due to diag. I added a comment

NDTensors/src/abstractarray/set_types.jl
```diff
@@ -37,30 +37,34 @@ function set_indstype(arraytype::Type{<:AbstractArray}, dims::Tuple)
   return set_ndims(arraytype, length(dims))
 end
 
-function set_eltype_if_unspecified(
+function specify_eltype(
```
mtfishman, line 40: I was picturing that we would use `SetParameters.specify_parameters` for this (currently `SetParameters.set_unspecified_parameters` but needs to be renamed), i.e.:
```julia
function specify_eltype(  arraytype::Type{<:AbstractArray}, eltype::Type=default_eltype())
  return specify_parameters(arraytype, Position(1), eltype)
end
```
Is that the plan here?

ITensorGPU/src/adapt.jl
```diff
@@ -7,12 +7,16 @@ function NDTensors.cu(eltype::Type{<:Number}, x)
 end
 NDTensors.cu(x) = fmap(x -> adapt(CuArray, x), x)
 
-function NDTensors.set_eltype_if_unspecified(
-  arraytype::Type{CuVector{T}}, eltype::Type
-) where {T}
+function NDTensors.specify_eltype(arraytype::Type{CuVector{T}}, eltype::Type) where {T}
```
mtfishman, line 10: Did you mean to update this? I expected that we would be keeping `ITensorGPU` the same.

NDTensors/src/abstractarray/set_types.jl
```diff
@@ -37,30 +37,34 @@ function set_indstype(arraytype::Type{<:AbstractArray}, dims::Tuple)
...
 
 #TODO transition to set_eltype when working for wrapped types
-function set_eltype_if_unspecified(
-  arraytype::Type{<:AbstractArray}, eltype::Type=default_eltype()
+function specify_eltype(arraytype::Type{<:AbstractArray}, eltype::Type=default_eltype())
+  return similartype(arraytype, eltype)
+end
+
+function specify_eltype(
+  arraytype::Type{<:AbstractArray{UnspecifiedZero}}, eltype::Type=default_eltype()
```
mtfishman, line 52: Ideally we wouldn't need overloads like this, and we could "register" `UnspecifiedZero` as an unspecified parameter like `UnspecifiedParameter` by overloading functionality of `SetParameters` directly.

I think we just need to overload this function: https://github.com/ITensor/ITensors.jl/blob/c49da9e02f7c2465001f3a8446ad49faf23347d0/NDTensors/src/SetParameters/src/set_unspecified_parameters.jl#L7-L11, i.e.
```julia
function SetParameters.replace_unspecified_parameters(
  type::Type, position::Position, current_parameter::Type{<:UnspecifiedZero}, parameter
)
  return set_parameters(type, position, parameter)
end
```

NDTensors/src/blocksparse/blocksparse.jl
```diff
@@ -38,10 +21,17 @@ function BlockSparse(
   )
 end
 
+function BlockSparse(
+  datatype::Type{<:UnallocatedZeros}, blockoffsets::BlockOffsets, dim::Integer; vargs...
+)
+  return BlockSparse(datatype((dim,)), blockoffsets; vargs...)
+end
```
mtfishman, line 28: I guess we need this since the version above that uses `fill!` won't work for `UnallocatedZeros`? That's a bit unfortunate. I wonder if something like this would work generically:
```julia
function BlockSparse(
  datatype::Type{<:AbstractArray}, blockoffsets::BlockOffsets, dim::Integer; vargs...
)
  return BlockSparse(
    generic_zeros(datatype, dim), blockoffsets; vargs...
  )
end
```
and then we would just need to make sure `generic_zeros` works properly for `UnallocatedZeros`.

mtfishman: Also makes me think that we should rename `generic_zeros` to `similar_zeros` but that could be done later.

NDTensors/src/blocksparse/blocksparse.jl
```diff
@@ -51,16 +41,41 @@ end
...
+function BlockSparse(::UndefInitializer, blockoffsets::BlockOffsets, dim::Integer; vargs...)
+  return BlockSparse(NDTensors.default_eltype(), undef, blockoffsets, dim; vargs...)
 end
 
 function BlockSparse(blockoffsets::BlockOffsets, dim::Integer; vargs...)
-  return BlockSparse(Float64, blockoffsets, dim; vargs...)
+  elt = default_eltype()
+  datat = default_datatype(elt)
+  N = ndims(datat)
+  return BlockSparse(UnallocatedZeros{elt,N,datat}, blockoffsets, dim; vargs...)
```
mtfishman, line 55: Probably should not use `UnallocatedZeros` here.

NDTensors/src/blocksparse/blocksparse.jl
```diff
@@ -51,16 +41,41 @@ end
...
+ndims(storagetype::Type{<:BlockSparse{<:Any,<:Any,N}}) where {N} = N
+# TODO: Implement as `fieldtype(storagetype, :blockoffsets)`.
+blockoffsetstype(storagetype::Type{<:BlockSparse}) = BlockOffsets{ndims(storagetype)}
+
+function set_datatype(storagetype::Type{<:BlockSparse}, datatype::Type{<:AbstractVector})
+  return BlockSparse{eltype(datatype),datatype,ndims(storagetype)}
+end
+
+function set_ndims(storagetype::Type{<:BlockSparse}, ndims)
+  return BlockSparse{eltype(storagetype),datatype(storagetype),ndims}
```
mtfishman, line 76: Reminder to use `get_parameters` and `set_parameters` from `SetParameters` here. If we leave that for a future PR, I think we should leave comments about that as a reminder for later.

NDTensors/src/empty/empty.jl
```diff
@@ -5,20 +5,20 @@
 
 struct EmptyOrder end
 
-function similartype(StoreT::Type{<:TensorStorage{EmptyNumber}}, ElT::Type)
+function similartype(StoreT::Type{<:TensorStorage{UnspecifiedZero}}, ElT::Type)
```
mtfishman, line 8: I wonder why these were needed in the first place, seems like they should be covered by some generic `similartype` function and there's nothing special about `UnspecifiedZero` here.

mtfishman: Though I guess these files aren't used any more and will be deleted, so it shouldn't matter.

NDTensors/src/zeros/allocate.jl
```diff
@@ -0,0 +1,36 @@
...
+  return set_datatype(typeof(storage), typeof(alloc))(alloc)
+end
+
+function allocate(storage::BlockSparse, elt::Type=default_eltype())
+  if !is_unallocated_zeros(storage)
+    return storage
+  end
+  alloc = allocate(data(storage), elt)
+
+  #d = adapt(storage, typeof(alloc))(alloc)
```
mtfishman, line 26: Seems like we could use:
```julia
function allocate(T::Tensor)
  return adapt(alloctype(data(T)), T)
end

function allocate(T::TensorStorage)
  return adapt(alloctype(data(T)), T)
end
```
If we define a generic fallback `alloctype(arraytype::Type{<:AbstractArray}) = arraytype`, which I think is reasonable to do since we assume a generic array type is allocated, then this should "just work" whether or not `T` has allocated data.

NDTensors/src/zeros/unallocated_zeros.jl
```diff
@@ -0,0 +1,114 @@
+struct UnallocatedZeros{ElT,N,Axes,Alloc<:AbstractArray{ElT,N}} <: AbstractArray{ElT,N}
```
mtfishman, line 1: Reminder to try making this `<: AbstractFill{ElT,N,Axes}`.

NDTensors/src/zeros/unallocated_zeros.jl
```diff
@@ -0,0 +1,114 @@
...
+  z::FillArrays.Zeros{ElT,N,Axes}
+  function NDTensors.UnallocatedZeros{ElT,N,Alloc}(inds::Tuple) where {ElT,N,Alloc}
```
mtfishman, line 3:
```suggestion
  function UnallocatedZeros{ElT,N,Alloc}(inds::Tuple) where {ElT,N,Alloc}
```
and same thing elsewhere.

kmp5VT: I fixed this.

NDTensors/src/zeros/unallocated_zeros.jl
mtfishman: With that change, it seems like we could just copy any specialized  code for the `Zeros` type from `FillArrays.jl`, for example https://github.com/JuliaArrays/FillArrays.jl/blob/5b27e469b73554aeed0a9205f19a44467033ba78/src/FillArrays.jl#L281-L357 and other functions that are overloaded explicitly for `Zeros` in `FillArrays.jl`, and define them for `UnallocatedZeros` and have `UnallocatedZeros` get all of the same behavior as `FillArrays.Zeros`.

kmp5VT: I updated this to AbstractFill, it looks like its working the same as when I had AbstractArray.

NDTensors/src/zeros/allocate.jl
kmp5VT: This does work, though I think it would be good to verify the performance of tensors that are already allocated. Also I was able to get adapt to work to set the eltype after allocating UnallocatedZeros with UnspecifiedZero type.

mtfishman: Great. Agreed we would want to check it doesn't hurt performance of allocated tensors. Ideally it is just a no-op but that assumes it gets optimized away.

ITensorGPU/src/adapt.jl
kmp5VT: This was a mistake, I thought this was NDTensorCUDA code. I have reverted the changes

NDTensors/src/blocksparse/blocksparse.jl
kmp5VT: Okay, I replaced this with `default_datatype(default_eltype())`

NDTensors/src/blocksparse/blocksparse.jl
kmp5VT: Yeah I have a comment in that code because it is confusing. But yes I can use generic_zeros. My thought here was that in the next PR I will be reworking the blocksparse code and will have to make changes there so I was just getting things to work for now. But I don't see any issue using generic_zeros

NDTensors/src/empty/empty.jl
kmp5VT: Yeah I am not sure, I did delete these files altogether and am working towards having all unit tests pass without them.

NDTensors/src/blocksparse/blocksparse.jl
kmp5VT: This is good to remember. I think I could incorporate them in this PR. I will work on this tomorrow

NDTensors/src/blocksparse/blocksparse.jl
kmp5VT: I have the generic_zeros code hacked together to work. We still need to be able to specify_parameters for the UnallocatedZeros type, but I have started working on that

kmp5VT:
@mtfishman Another comment is I have some diag tests currently failing because I am trying to set the eltype for a `UniformDiagTensor` which should fail since the datatype for this tensor is a `Number`.

kmp5VT:
@mtfishman It looks like one test case that fails is related to the `@strided mul!` call in contract which matches the error on main.  As a reminder per our conversation earlier; I am taking a break from this branch to work on the GPU code and when I return I will try to consolidate the changes here in to a more compact and NDTensors focused manner.

mtfishman:
Oh yeah, sorry about that. Fixing that should align with other things we talked about with making the contract code generic and efficient for both CPU and GPU operations. Thanks, that sounds like a good plan.

kmp5VT:
@mtfishman not a problem, this was mainly just a note to remind both of us where I am in this PR next time we look at it!

mtfishman:
@kmp5VT can we close this? Will be superseded by #1213 and other PRs.

Issue #1164, opened 2023-07-28
[NDTensors] [BUG] Precompilation issue with 1.10-beta1


emstoudenmire:
**Description of bug**

When installing `ITensors` with Julia 1.10-beta1, there seems to be a conflict or issue between the `Base.sort` overload in `tupletools.jl` (line 175) and the `sort.jl` file defining part of the Julia `Base` module.

**Minimal code demonstrating the bug or unexpected behavior**

<details><summary>Minimal runnable code</summary><p>

```julia
julia> ]
pkg> add ITensors
```
</p></details>

**Actual output or behavior**

NDTensors failed to precompile. The error message and Julia versoininfo() are below.

<details><summary>Output of minimal runnable code</summary><p>

```julia
julia> using ITensors
Precompiling ITensors
  ✗ NDTensors
  ✗ ITensors
  0 dependencies successfully precompiled in 6 seconds. 88 already precompiled.

ERROR: The following 1 direct dependency failed to precompile:

ITensors [9136182c-28ba-11e9-034c-db9fb085ebd5]

Failed to precompile ITensors [9136182c-28ba-11e9-034c-db9fb085ebd5] to "/Users/mstoudenmire/.julia/compiled/v1.10/ITensors/jl_BM3z3M".
WARNING: Method definition sort(Tuple) in module Sort at sort.jl:1418 overwritten in module NDTensors at /Users/mstoudenmire/.julia/packages/NDTensors/DkvfK/src/tupletools.jl:175.
ERROR: LoadError: Method overwriting is not permitted during Module precompile.
Stacktrace:
 [1] top-level scope
   @ ~/.julia/packages/NDTensors/DkvfK/src/tupletools.jl:170
 [2] include(mod::Module, _path::String)
   @ Base ./Base.jl:489
 [3] include(x::String)
   @ NDTensors ~/.julia/packages/NDTensors/DkvfK/src/NDTensors.jl:1
 [4] top-level scope
   @ ~/.julia/packages/NDTensors/DkvfK/src/NDTensors.jl:46
 [5] include
   @ Base ./Base.jl:489 [inlined]
 [6] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{…}, dl_load_path::Vector{…}, load_path::Vector{…}, concrete_deps::Vector{…}, source::String)
   @ Base ./loading.jl:2216
 [7] top-level scope
   @ stdin:3
Some type information was truncated. Use `show(err)` to see complete types.
[...]

PR #1169, opened 2023-08-02
[NDTensors] Remove sort(Tuple) in favor of version in Compat

mtfishman:
# Description

`sort(::NTuple)` was introduced in Julia 1.10 (https://github.com/JuliaLang/julia/pull/46104), and can be added to older Julia versions using Compat.jl 4.9 (https://github.com/JuliaLang/Compat.jl/pull/804), so we can now remove our own version of it, which wasn't a good idea to have anyway since it is type piracy (see also #1164).


mtfishman:
Ok, thanks for the heads up.

PR #1171, opened 2023-08-02
[ITensors] Index filtering docstring improvements

emstoudenmire:
Add list of most common keyword arguments to functions for filtering indices (commoninds, uniqueinds, etc.). Currently it's not so obvious to users what keyword arguments could be passed, as shown by a recent [forum post](https://itensor.discourse.group/t/string-correlation-function-of-several-sites/1066/4).

### DIFF
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -1406,6 +1406,15 @@ for find in (:commonind, :noncommonind, :uniqueind, :unionind)
   end
 end
 
+function index_filter_kwargs_docstring()
+  return """
+  Optional keyword arguments:
+  * tags::String - a tag name or comma separated list of tag names that the returned indices must all have
+  * plev::Int - common prime level that the returned indices must all have
+  * inds - Index or collection of indices. Returned indices must come from this set of indices.
+  """
+end
+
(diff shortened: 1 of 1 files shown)


emstoudenmire:
@mtfishman please let me know if these docstrings look accurate to you

mtfishman:
Looks good to me, thanks. One suggestion I would have is defining a function:
```julia
function index_filter_kwargs_docstring()
return """
  Optional keyword arguments:
  * tags::String - a tag name or comma separated list of tag names that the returned indices must all have
  * plev::Int - common prime level that the returned indices must all have
  * inds - Index or collection of indices. Returned indices must come from this set of indices.
  """
end
```
and interpolate that into the other docstrings with `$(index_filter_kwargs_docstring())` so if we want to change or update it we just need to update it in one place.

emstoudenmire:
Great, yes I was thinking that very same thing but couldn't remember quite how to do it. I'll make that change.

emstoudenmire:
@mtfishman ok just made that change

mtfishman:
Thanks!

PR #1174, opened 2023-08-04
[ITensors] Fix some fermionic SVD corner cases

emstoudenmire:
This PR contains the latest code for the fermion system, which I had not submitted due to some technical questions about its correctness. Following recent discussions, it is now more likely that these changes are actually totally correct. At the very least, they are an improvement on the current (main branch) code. 

The specific changes here are:
- SVD now works correctly (in the sense of both norm(U*S*V-T) is small and U, V unitary in the conventional sense) in all cases of T being a matrix (regardless of arrows i.e. for all four possibilities of arrows)
- eigen is working correctly as long as arrows on both sides are same direction as each other and opposite on the two sides. Currently an error is thrown if this is not the case when fermions are being used.
- more unit tests checking the above cases


test/base/test_fermions.jl
```diff
@@ -471,6 +472,155 @@ using ITensors, Test
...
+    psi = productMPS(sites, [2, 2, 1])
+    @test inner(psi, psi) ≈ 1.0
+  end
+
+  @testset "Fermionic AutoMPO Tests" begin
+    @testset "Spinless Fermion Hamiltonian" begin
+      N = 2
+      sites = siteinds("Fermion", N; conserve_qns=true)
+      t1 = 1.0
+      ampo = AutoMPO()
```
mtfishman, line 519: `AutoMPO` -> `OpSum`.

mtfishman:
Looks good! Is this ready to merge?

emstoudenmire:
Yes, ready to go, thanks. It's still an experimental feature, as we discussed, but it is passing many more non-trivial tests now, especially regarding the SVD in all possible matrix cases. I think at this point it implements the complete "algebra" but takes care to use properly.

mtfishman:
Very cool! Sounds like it is ready to test out for more general tensor networks (@JoeyT1994 @b-kloss).

NDTensors/src/blocksparse/linearalgebra.jl
```diff
@@ -165,13 +165,32 @@ function LinearAlgebra.svd(T::BlockSparseMatrix{ElT}; kwargs...) where {ElT}
...
       setdiagindex!(blockviewS, getdiagindex(Sb, i), i)
     end
 
-    blockview(V, blockV) .= Vb
+    #<fermions>
+    sV = left_arrow_sign(vind, blockV[2])
+    # This sign (sVP) accounts for the fact that
+    # V is transposed, i.e. the index connecting to S
+    # is the second index:
+    sVP = block_parity(vind, blockV[2]) == 1 ? -1 : +1
```
mtfishman, line 187: @emstoudenmire I think this line is the issue, I think it should just be:
```julia
    sVP = block_parity(vind, blockV[2])
```
When I change it to that, the ITensorGaussianMPS.jl tests pass.

mtfishman: @emstoudenmire I added a fix to #1178, testing it out there. I want to merge it quickly since I'm worried that the bug I introduced in #1176 (which is now included in the latest registered version of ITensors, v0.3.40) is out there breaking user code, but if I merge the fix in #1178 and register a new version, the bug from this PR (#1174) will also cause issues. So if tests pass in #1178, I'll merge and register that, and then we can add more robust fermion tests in a future PR.

emstoudenmire: Thanks for catching it! Yes, please go ahead with that plan

PR #1192, opened 2023-09-13
[NDTensors] Improve contract performance for wrapped arrays

mtfishman:
# Description

Improve the performance of dense `contract!` for complicated wrapped array types that Julia doesn't forward to BLAS since it doesn't recognize as strided. These can show up in block sparse contractions, which involved sliced, reshaped, and transposed blocks.


mtfishman:
[benchmark]

mtfishman:
[benchmark]

Issue #1193, opened 2023-09-13
[NDTensors] [BUG] Many basic tensor operations fail with CUDA extension due to `LoadError: Scalar indexing is disallowed.`


wbernoudy:
**Description of bug**

I am attempting to use the CUDA extension for NDTensors. I am finding that tensor contraction sometimes fails, along with other functions like `permute`, as well as the `NDTensors.generic_randn` call used in the example of the NDTensors CUDA extension (`NDTensors/ext/examples/NDTensorCUDA.jl`).

They all fail with `ERROR: LoadError: Scalar indexing is disallowed.` (or emit a similar warning in the REPL where scalar indexing is allowed).

**Minimal code demonstrating the bug or unexpected behavior**

<details><summary>Minimal runnable code</summary><p>

```julia
import CUDA
import ITensors
import NDTensors

i, j, k = ITensors.Index(3), ITensors.Index(5), ITensors.Index(2)
A, B = ITensors.randomITensor(i, j), ITensors.randomITensor(j, k)

# works when running script directly, but warns of scalar indexing in REPL
cuA, cuB = NDTensors.cu(A), NDTensors.cu(B)

contracted = cuA * cuB

@show NDTensors.cpu(contracted) ≈ A * B

# without @CUDA.allowscalar, fails with `ERROR: LoadError: Scalar indexing is disallowed.`
permuted = ITensors.permute(cuA, (j, i))
```
</p></details>

**Expected output or behavior**

`NDTensors.cu()` should work without warning in REPL on `ITensors.ITensor`, and `ITensors.permute` should work without the scalar indexing error.

**Actual output or behavior**

<details><summary>Output of minimal runnable code</summary><p>

```julia
NDTensors.cpu(contracted) ≈ A * B = true
ERROR: LoadError: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore are only permitted from the REPL for prototyping purposes.
If you did intend to index this array, annotate the caller with @allowscalar.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] assertscalar(op::String)
    @ GPUArraysCore ~/.julia/packages/GPUArraysCore/uOYfN/src/GPUArraysCore.jl:103
  [3] getindex
[...]

PR #1194, opened 2023-09-13
[NDTensors][NDTensorsCUDAExt] Fill tensors on GPU devices

kmp5VT:
This PR is to address the issue https://github.com/ITensor/ITensors.jl/issues/1193. In order to avoid scalar operations, fill tensors with rand or zeros on device.


NDTensors/ext/NDTensorsCUDAExt/fill.jl
```diff
@@ -0,0 +1,16 @@
+import NDTensors.SetParameters: set_unspecified_parameters, DefaultParameters
+import NDTensors: leaf_parenttype
```
mtfishman, line 2: Should use `using` instead of `import`, `import` only needs to be used when you plan to overload the function.

kmp5VT: Makes sense, I fixed this

NDTensors/src/abstractarray/fill.jl
```diff
@@ -1,13 +1,9 @@
...
   arraytype_specified = set_unspecified_parameters(
     leaf_parenttype(arraytype), DefaultParameters()
   )
   data = similar(arraytype_specified, dim)
-  ElT = eltype(data)
-  for i in 1:length(data)
-    data[i] = randn(ElT)
-  end
-  return data
+  NDTensors.randn!(rng, data)
```
mtfishman, line 6: Good idea. Can we do something similar with `generic_zeros` and call `zeros!`?

mtfishman: Also, I don't think we need `NDTensors.` here, `randn!` is an exported `Base` function.

kmp5VT: Yes I was able to find a more generic zeros initializer

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
```diff
@@ -11,3 +11,8 @@ function NDTensors.svd_catch_error(A::CuMatrix; alg="JacobiAlgorithm")
   end
   return USV
 end
+
+using LinearAlgebra
```
mtfishman, line 15: Let's move `using` statements to `NDTensorsCUDAExt`.

NDTensors/src/linearalgebra/linearalgebra.jl
```diff
@@ -400,6 +400,7 @@ function qx(qx::Function, T::DenseTensor{<:Any,2}; kwargs...)
   Qinds = IndsT((ind(T, 1), q))
   Xinds = IndsT((q, ind(T, 2)))
   QM = convert(typeof(XM), QM)
+  QM = convert(typeof(XM), QM)
```
mtfishman, line 403: Seems to just be a repeat of the line above.

src/itensor.jl
```diff
@@ -1892,7 +1892,7 @@ diag(T::ITensor) = diag(tensor(T))
 
 mul!(C::ITensor, A::ITensor, B::ITensor, args...)::ITensor = contract!(C, A, B, args...)
 
-dot(A::ITensor, B::ITensor) = (dag(A) * B)[]
+dot(A::ITensor, B::ITensor) = NDTensors.cpu(dag(A) * B)[]
```
mtfishman, line 1895: I guess this is to avoid calling `getindex` on GPU? Maybe should be handled in `NDTensors`.

src/itensor.jl
```diff
@@ -1963,8 +1963,9 @@ map(f, x::ITensor) = itensor(map(f, tensor(x)))
 w .+= a .* v
 ```
 """
-axpy!(a::Number, v::ITensor, w::ITensor) = (w .+= a .* v)
-
+function axpy!(a::Number, v::ITensor, w::ITensor)
+  return itensor(data(w) .+= a .* data(v), inds(w))
```
mtfishman, line 1967: This will be wrong for sparse tensors like `BlockSparse` and `Diag`. We'll need to fix it in `NDTensors` directly.

Broadcasting on ITensors eventually calls `permutedims` in NDTensors so seems like that would be the place to fix it.

src/itensor.jl
```diff
@@ -1982,22 +1983,21 @@ Scale the ITensor A by x in-place. May also be written `rmul!`.
...
 
 Scalar multiplication of ITensor B with x, and store the result in A.
 Like `A .= x .* B`.
 """
-mul!(R::ITensor, α::Number, T::ITensor) = (R .= α .* T)
-
-mul!(R::ITensor, T::ITensor, α::Number) = (R .= T .* α)
+mul!(R::ITensor, α::Number, T::ITensor) = itensor(data(R) .= α * data(T), inds(T))
 
+mul!(R::ITensor, T::ITensor, α::Number) = itensor(data(R) .= data(T) * α, inds(T))
```
mtfishman, line 2000: Again, these will all be wrong for sparse tensors so we need to fix it in `NDTensors`.

src/mps/dmrg.jl
```diff
@@ -278,7 +278,7 @@ function dmrg(PH, psi0::MPS, sweeps::Sweeps; kwargs...)
         end
 
         energy = vals[1]
-        phi::ITensor = vecs[1]
+        phi::ITensor = itensor(storagetype(phi)(convert(typeof(data(phi)),data(vecs[1]))), inds(phi))
```
mtfishman, line 281: Note sure why this change is needed.

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -573,7 +573,8 @@ function factorize_eigen(A::ITensor, Linds...; kwargs...)
     # (Lis..., prime(Lis)...)
     delta_A2 = replaceinds(delta_A2, Lis, dag(simLis))
     noprime!(delta_A2)
-    A2 += delta_A2
+    itensor(data(A2) .+= data(delta_A2), inds(A2))
```
mtfishman, line 576: This will be wrong for sparse tensors, needs a fix in `NDTensors`. Like the ones above, I think a change to `permutedims` dispatching on `CuArray` like you did for `mul!` will be sufficient.

NDTensors/src/linearalgebra/linearalgebra.jl
kmp5VT: There seems to be a conversion issue related to `UnifiedMemory` buffer in CUDA. The first `convert` takes the `packedQR` type which is a CPU wrapper to `CuArray{ElT, N, DeviceBuffer}` and the second call converts it to `UnifiedBuffer`. I have tried using `adapt` and calling the `CuArray` constructor but neither seem to create the vector with the correct memory without the second call.

src/mps/dmrg.jl
kmp5VT: Again this is related to the CUDA `UnifiedMemory` issue. The solve above returns `vecs[1]` which is a `CuArray{ElT, N, DeviceBuffer}`.

src/tensor_operations/matrix_decomposition.jl
kmp5VT: Practically all of the code which uses `data` is just a test to see if these operations are correlated with the performance issues with CUDA. I can verify from these tests that the performance issues are caused by these operations and will shift to find a more practical implementation which solves the issues.

mtfishman: Gotcha, makes sense. Glad it seems to be only a few operations that need to be optimized/dispatched properly, I'm sure we can find a simple lower level solution.

src/broadcast.jl
```diff
@@ -281,11 +282,11 @@ end
 #
 
 function fmap(bc::Broadcasted{ITensorStyle,<:Any,typeof(+),<:Tuple{Vararg{ITensor}}})
-  return (r, t) -> bc.f(r, t)
```
mtfishman, line 284: Note that you should be able to just return `+` instead of `(r, t) -> +(r, t)`.

src/broadcast.jl
```diff
@@ -281,11 +282,11 @@ end
...
+  return (r, t) -> +(r, t)
 end
 
 function fmap(bc::Broadcasted{ITensorStyle,<:Any,typeof(-),<:Tuple{Vararg{ITensor}}})
-  return (r, t) -> bc.f(r, t)
+  return (r, t) -> -(r, t)
```
mtfishman, line 289: Same as above, can just use `-` instead of an anonymous function.

src/itensor.jl
```diff
@@ -1892,8 +1892,13 @@ diag(T::ITensor) = diag(tensor(T))
...
 mul!(C::ITensor, A::ITensor, B::ITensor, args...)::ITensor = contract!(C, A, B, args...)
 
-dot(A::ITensor, B::ITensor) = (dag(A) * B)[]
+## TODO this operation does not work with GPU
+dot(A::ITensor, B::ITensor) = _dot(dag(A), B)
 
+function _dot(A::ITensor, B::ITensor)
+  B = permute(B, inds(A))
+  return dot(tensor(A), tensor(B))
+end
```
mtfishman, line 1901: Is the issue the operation `[]`? If so, seems like in `NDTensors` we could write a specialized version of `getindex(A::Tensor)` where `A` is a tensor backed by a `CuArray` (say by first converting to CPU with `NDTensors.cpu`).

src/mps/dmrg.jl
```diff
@@ -346,6 +347,7 @@ function dmrg(PH, psi0::MPS, sweeps::Sweeps; kwargs...)
         )
       end
     end
+    sweepup(typeof(data(psi[1])))
```
mtfishman, line 350: Could you explain this? I'm guessing this is just a temporary thing but I don't think we always want to call this, at least on CPU.

kmp5VT: Of course: DMRG creates a lot of intermediate memory and will often get up to almost 100% of the GPU memory. I was trying to force GC and CUDA.reclaim to run to test and see if doing so influenced performance/memory usage. It has a very slight improvement in performance but I would think not necessary.

src/itensor.jl
mtfishman: I would only propose overloading an empty `getindex` call on a scalar `Tensor`/`ITensor` (i.e. `getindex(T)`, not `getindex(T, I...)`, in which case there is only one element anyway, and also only if that is necessary to make this code work generically without modifying code in ITensors.jl.

src/mps/dmrg.jl
mtfishman: Sounds good. For future reference, you can use an `Observer` to insert custom code into `dmrg` without modifying the source code of `dmrg`, see: https://itensor.discourse.group/t/memory-usage-in-dmrg-with-julia-1-x/1092.

NDTensors/src/abstractarray/fill.jl
```diff
@@ -1,19 +1,17 @@
...
+  return randn!(rng, data)
 end
 
 function generic_zeros(arraytype::Type{<:AbstractArray}, dim::Integer=0)
   arraytype_specified = set_unspecified_parameters(
     leaf_parenttype(arraytype), DefaultParameters()
   )
   ElT = eltype(arraytype_specified)
-  return fill!(similar(arraytype_specified, dim), zero(ElT))
+  return arraytype_specified(zeros(ElT, dim))
```
mtfishman, line 16: This doesn't seem great, since then it might create the data on CPU and then move it to GPU.

Can we use:
```julia
  data = similar(arraytype_specified, dim)
  return zeros!(data)
```
like we do for `generic_randn`?

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -393,15 +397,20 @@ function _contract!(
...
-  @strided mul!(CM, AM, BM, El(α), El(β))
+  mul!!(CM, AM, BM, El(α), El(β))
 
   if props.permuteC
     Cr = reshape(CM, props.newCrange)
     # TODO: use invperm(pC) here?
     #@timeit_debug timer "_contract!: permutedims C" begin
-    @strided CT .= permutedims(Cr, props.PC)
+    #@strided 
+    CT .= permutedims(Cr, props.PC)
```
mtfishman, line 407: Maybe need to refactor out into `permutedims!!` which then dispatches on `Array` or `CuArray`?

NDTensors/src/abstractarray/fill.jl
mtfishman: EDIT: I forgot there isn't a `zeros!` function but you can use `fill!`, which I see is what the previous version of this code is doing, was there an issue with that?

src/mps/dmrg.jl
kmp5VT: Good to know, thanks!

NDTensors/src/abstractarray/fill.jl
mtfishman: I could imagine it might fail in that way for certain wrapped `CuArray`s, but we unwrap with `leaf_parenttype(arraytype)` so that shouldn't be a problem here.

src/broadcast.jl
```diff
@@ -183,9 +183,9 @@ function Base.copyto!(
 )
   T1, T2 = bc.args
   if R === T1
-    map!((t1, t2) -> bc.f(t1, t2), R, T1, T2)
+    map!((t1, t2) -> /(t1, t2), R, T1, T2)
   elseif R === T2
-    map!((t1, t2) -> bc.f(t2, t1), R, T2, T1)
+    map!((t1, t2) -> /(t2, t1), R, T2, T1)
```
mtfishman, line 188: Should be able to use `map!(/, R, T1, T2)` here.

NDTensors/src/linearalgebra/linearalgebra.jl
mtfishman: Reminder to leave a comment about why this is here and that it should be removed once the issue is fixed in CUDA.jl.

kmp5VT: I added a comment here, thanks!

NDTensors/src/linearalgebra/linearalgebra.jl
```diff
@@ -424,7 +424,6 @@ matrix is unique. Returns a tuple (Q,R).
 function qr_positive(M::AbstractMatrix)
   sparseQ, R = qr(M)
   Q = convert(typeof(R), sparseQ)
-  Q = convert(typeof(R), sparseQ)
   nc = size(Q, 2)
   ## TODO issue here for GPU because tying to access indices
```
mtfishman, line 428: Maybe we could add a function `iscu(A::AbstractArray) = leaf_parenttype(A) isa CuArray` and then throw an error at the beginning of these functions based on `iscu`, so users get a nice error message and it reminds us to fix it.

src/itensor.jl
```diff
@@ -1140,15 +1140,16 @@ A[i => 1, i' => 2] # 2.0, same as: A[i' => 2, i => 1]
 @propagate_inbounds (getindex(T::ITensor, ivs::Vararg{Any,N})::Any) where {N} =
   _getindex(tensor(T), ivs...)
 
+## Allowing one to get the first ITensor element if its an order 0 tensor or an order 1 tensor with a dimension of 1. Also convert GPU back to CPU
 @propagate_inbounds function getindex(T::ITensor)::Any
-  if order(T) != 0
-    throw(
+  if order(T) != 0 && dim(T) != 1
+    ITensors.throw(
```
mtfishman, line 1146: Why is this `ITensors.throw`?

src/itensor.jl
```diff
@@ -1140,15 +1140,16 @@ A[i => 1, i' => 2] # 2.0, same as: A[i' => 2, i => 1]
...
       DimensionMismatch(
         "In scalar(T) or T[], ITensor T is not a scalar (it has indices $(inds(T)))."
       ),
     )
   end
-  return tensor(T)[]
+  return NDTensors.cpu(tensor(T))[]
```
mtfishman, line 1152: Can we fix this in NDTensors directly and dispatch `getindex(T::Tensor)` on `leaf_parenttype`, and transfer to CPU if it is `CuArray`? Or just call `cpu` in `getindex(T::Tensor)` in `NDTensors`?

src/mps/dmrg.jl
```diff
@@ -278,7 +278,8 @@ function dmrg(PH, psi0::MPS, sweeps::Sweeps; kwargs...)
         end
 
         energy = vals[1]
-        phi::ITensor = vecs[1]
+        ## Right now this convert problem is prevelent in an eigen solver implemented in KrylovKit
+        phi::ITensor = itensor(convert(typeof(tensor(phi)), tensor(vecs[1])))
```
mtfishman, line 282: I don't understand this comment, and also I don't understand what issue this code is fixing.

src/mps/dmrg.jl
```diff
@@ -278,7 +278,10 @@ function dmrg(PH, psi0::MPS, sweeps::Sweeps; kwargs...)
         end
 
         energy = vals[1]
-        phi::ITensor = vecs[1]
+        ## Right now there is a conversion problem in CUDA.jl where `UnifiedMemory` Arrays are being converted 
+        ## into `DeviceMemory`. This conversion line is here temporarily to fix that problem when it arises
+        phi::ITensor = itensor(convert(typeof(tensor(phi)), tensor(vecs[1])))
```
mtfishman, line 283: An issue with this line is that in principle the type of the storage doesn't have to be exactly the same before and after call `eigsolve`, for example the input could have `Diag` storage but end up with `Dense` storage because of operations performed in `eigsolve`.

Could we use `adapt` here, like `adapt(leaf_parenttype(phi), vec[1])`? That would be more generic and also makes it clearer what it is doing.

Also maybe good to wrap in `iscu(phi) && iscu(vec[1])`, since also it's possible operations in `eigsolve` move the tensor back to CPU (intentionally), say if we have some feature in the future that moves tensors to CPU if GPU memory runs out.

NDTensors/src/abstractarray/similar.jl
```diff
@@ -56,6 +56,9 @@ function similar(arraytype::Type{<:AbstractArray}, dims::Tuple)
   return similartype(arraytype, shape)(undef, NDTensors.to_shape(arraytype, shape))
 end
 
+# For when there are CUArray specific issues inline
+iscu(A::AbstractArray) = iscu(leaf_parenttype(A))
+iscu(::Type{<:AbstractArray}) = false
```
mtfishman, line 61: An issue I see here is that if you pass a type with a nested structure, say `iscu(Adjoint{Float64,CuMatrix{Float64}})`, it looks like this would return `false`.

We could do:
```julia
iscu(A::AbstractArray) = iscu(typeof(A))
iscu(arraytype::Type{<:AbstractArray}) = iscuarray(leaf_parenttype(arraytype))
iscuarray(::Type{<:AbstractArray}) = false
```
and then:
```julia
iscuarray(::Type{<:CuArray}) = true
```

src/mps/dmrg.jl
```diff
@@ -278,7 +278,15 @@ function dmrg(PH, psi0::MPS, sweeps::Sweeps; kwargs...)
         end
 
         energy = vals[1]
-        phi::ITensor = vecs[1]
+        ## Right now there is a conversion problem in CUDA.jl where `UnifiedMemory` Arrays are being converted 
+        ## into `DeviceMemory`. This conversion line is here temporarily to fix that problem when it arises
+        ## Adapt is only called when using CUDA backend. CPU will work as implemented previously.
+        phi::ITensor = if NDTensors.iscu(tensor(phi))
```
mtfishman, line 284: Let's define `iscu(A::ITensor) = iscu(tensor(A))` somewhere since I'm sure this is something we will want to use in other places.

Also I think we should also check `iscu(vecs[1])` as well since this may accidentally move `vecs[1]` to GPU in cases where it was supposed to be on CPU.

src/mps/dmrg.jl
```diff
@@ -278,7 +278,15 @@ function dmrg(PH, psi0::MPS, sweeps::Sweeps; kwargs...)
...
+          itensor(adapt(typeof(data(phi)), tensor(vecs[1])))
```
mtfishman, line 285: I think there is a function `datatype` we can use. It may only be defined for `Tensor`, if so we should define it for `ITensor`.

Also we should probably call `leaf_parenttype(datatype(phi))` to make sure it is adapting according to a leaf type and not a wrapper type.

Finally, we should define `adapt` for ITensor if it's not defined already, so then all of this could be:
```julia
adapt(leaf_parenttype(phi), vecs[1])
```
Another issue I could see happening here is that `vecs[1]` is complex-valued but `phi` is real-valued, say in the case where starting state is real but the Hamiltonian is complex. To handle that we could do:
```julia
adapt(set_eltype(leaf_parenttype(phi), eltype(vecs[1])), vecs[1])
```

src/itensor.jl
```diff
@@ -1892,7 +1893,8 @@ diag(T::ITensor) = diag(tensor(T))
 
 mul!(C::ITensor, A::ITensor, B::ITensor, args...)::ITensor = contract!(C, A, B, args...)
 
-dot(A::ITensor, B::ITensor) = (dag(A) * B)[]
+## TODO this operation does not work with GPU
+dot(A::ITensor, B::ITensor) = NDTensors.cpu(dag(A) * B)[]
```
mtfishman, line 1897: Can this be removed now that there is a new `getindex` definition in NDTensors?

NDTensors/src/linearalgebra/linearalgebra.jl
```diff
@@ -418,9 +422,15 @@ non-negative. Such a QR decomposition of a
 matrix is unique. Returns a tuple (Q,R).
 """
 function qr_positive(M::AbstractMatrix)
+  if iscu(M)
+    println(
+      "WARNING!!: Currently qr positive methods are not efficient for CuArrays because they require scalar operations. Please convert to CPU Array or use generic qr",
+    )
+  end
```
mtfishman, line 429: Should we just convert it ourselves in this case?

NDTensors/src/linearalgebra/linearalgebra.jl
```diff
@@ -442,6 +452,11 @@ non-negative. Such a QL decomposition of a
 matrix is unique. Returns a tuple (Q,L).
 """
 function ql_positive(M::AbstractMatrix)
+  if iscu(M)
+    println(
+      "WARNING!! Currently the ql positive methods are not efficient for CuArrays because they require scalar operations. Please convert to CPU Array or use generic ql",
+    )
+  end
```
mtfishman, line 459: Should we just convert it ourselves in this case?

NDTensors/src/tensor/tensor.jl
```diff
@@ -278,7 +278,7 @@ matrix(T::Tensor{<:Number,2}) = array(T)
 vector(T::Tensor{<:Number,1}) = array(T)
 
 isempty(T::Tensor) = isempty(storage(T))
-
+iscu(T::Tensor) = iscu(data(T))
```
mtfishman, line 281: If we define:
```julia
parenttype(tensortype::Type{<:Tensor}) = storagetype(tensortype)
parenttype(storagetype::Type{<:TensorStorage}) = datatype(tensortype)
```
then this definition wouldn't be necessary.

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -345,7 +345,8 @@ function _contract!(
   tA = 'N'
   if props.permuteA
     #@timeit_debug timer "_contract!: permutedims A" begin
-    @strided Ap = permutedims(AT, props.PA)
+    #@strided 
+    Ap = permutedims(AT, props.PA)
```
mtfishman, line 349: Reminder to make `NDTensors.permutedims` that dispatches on `Array` or `CuArray` to call `@strided` on `Array`.

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -360,7 +361,8 @@ function _contract!(
   tB = 'N'
   if props.permuteB
     #@timeit_debug timer "_contract!: permutedims B" begin
-    @strided Bp = permutedims(BT, props.PB)
+    #@strided 
+    Bp = permutedims(BT, props.PB)
```
mtfishman, line 365: Reminder to make `NDTensors.permutedims` that dispatches on `Array` or `CuArray` to call `@strided` on `Array`.

src/mps/dmrg.jl
mtfishman: Note that if we define `NDTensors.parenttype(t::ITensor) = typeof(tensor(t))` I think `leaf_parenttype(t::ITensor)` should automatically work.

src/set_types.jl
```diff
@@ -0,0 +1,2 @@
+import NDTensors: leaf_parenttype
+NDTensors.leaf_parenttype(T::ITensor) = leaf_parenttype(NDTensors.datatype(T))
```
mtfishman, line 2: If you overload with `NDTensors.leaf_parenttype` you don't need to `import`.

mtfishman: I think you should move `import NDTensors: leaf_parenttype` to https://github.com/ITensor/ITensors.jl/blob/main/src/imports.jl and then define:
```julia
leaf_parenttype(T::ITensor) = leaf_parenttype(tensor(T))
```
i.e. without `NDTensors.`, and also in terms of `tensor`. That will be more general, for example when we move to the new storage type design.

mtfishman: Since in the future, not all tensor storage will have `data`.

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
```diff
@@ -11,3 +11,7 @@ function NDTensors.svd_catch_error(A::CuMatrix; alg="JacobiAlgorithm")
   end
   return USV
 end
+
+function NDTensors.mul!!(::Type{<:CuArray}, CM, AM, BM, α, β)
```
mtfishman, line 15: I think to be more forward-thinking this should be:
```julia
function NDTensors.mul!!(::Type{<:CuArray}, CM, ::Type{<:CuArray}, AM, ::Type{<:CuArray}, BM, α, β)
  # ...
end
```
i.e. dispatch on the `leaf_parenttype` of each of the inputs. Then we can catch more specific situations like if one input is on GPU and one on CPU and do specialized behavior on those cases (for example throw errors or move to CPU or GPU depending on the memory available on the GPU).

NDTensors/src/linearalgebra/linearalgebra.jl
```diff
@@ -489,6 +489,9 @@ end
 # about unpacking Q and L from the A matrix.
 #
 function ql!(A::StridedMatrix{<:LAPACK.BlasFloat})
+  if iscu(A)
```
mtfishman, line 492: Should we also have `ql` and `ql_positive` move back and forth to CPU to make them work on GPU? Seems like we may as well to make them consistent with `qr_positive`.

We can keep this error message for `ql!` since that strategy doesn't work for in-place operations.

NDTensors/src/abstractarray/permutedims.jl
```diff
@@ -0,0 +1,3 @@
+function permutedims!!(::Type{<:AbstractArray}, M, perm)
+  return @strided Mdest = permutedims(M, perm)
+end
```
mtfishman, line 3: I don't think it makes sense to call this `permutedims!!`, since that implies it may or may not mutate, but single-argument `permutedims` in Julia doesn't mutate. So I think for now we should just call this `permutedims`, and make sure to not import `Base.permutedims` so we have our own `NDTensors.permutedims`.

I realize there are ways to do in-place tensor permutation through a series of in-place transpositions which you explained to me at some point, so in the future maybe we will have an `NDTensors.permutedims!` in which case it may make sense to define `NDTensors.permutedims!!`, but let's hold off on that for now.

NDTensors/src/dense/densetensor.jl
```diff
@@ -195,7 +199,7 @@ function permutedims!(
 ) where {N,StoreT<:StridedArray}
   RA = array(R)
   TA = array(T)
-  @strided RA .= permutedims(TA, perm)
+  RA .= permutedims!!(leaf_parenttype(TA), TA, perm)
```
mtfishman, line 202: I think this should be:
```julia
permutedims!!(leaf_parenttype(RA), RA, leaf_parenttype(TA), TA, perm)
```
Otherwise it won't be doing an in-place permutation, what your wrote is doing an out-of-place permutation and then a copy.

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -393,15 +397,30 @@ function _contract!(
...
+    leaf_parenttype(CM), CM, leaf_parenttype(AM), AM, leaf_parenttype(BM), BM, El(α), El(β)
+  )
 
   if props.permuteC
     Cr = reshape(CM, props.newCrange)
     # TODO: use invperm(pC) here?
-    #@timeit_debug timer "_contract!: permutedims C" begin
-    @strided CT .= permutedims(Cr, props.PC)
+    #@timeit_debug timer "_contract!: permutedims C" begin 
+    CT .= permutedims!!(leaf_parenttype(Cr), Cr, props.PC)
```
mtfishman, line 408: Like my above comment, I think this should be `permutedims!!(leaf_parenttype(CT), CT, leaf_parenttype(Cr), Cr, props.PC)`.

There are a few other places you use `permutedims!!` where similar changes should be made.

NDTensors/src/abstractarray/permutedims.jl
```diff
@@ -0,0 +1,24 @@
...
+
+## TODO is it possible to do something like f = identity and remove the permutedims!! funciton?
+function permutedims!(::Type{<:Array}, Mdest, ::Type{<:Array}, M, perm)
+  return @strided Mdest .= Base.permutedims(M, perm)
+end
+
+function permutedims!!(::Type{<:Array}, B, ::Type{<:Array}, A, perm, f)
+  @strided B .= f.(B, Base.permutedims(A, perm))
+end
```
mtfishman, line 12: Should probably move to `NDTensors/src/array/permutedims.jl`.

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -175,14 +175,18 @@ function _contract_scalar_perm!(
     if iszero(α)
       fill!(Rᵃ, 0)
     else
-      @strided Rᵃ .= α .* permutedims(Tᵃ, perm)
+      Rᵃ = permutedims!!(
+        leaf_parenttype(Rᵃ), Rᵃ, leaf_parenttype(Tᵃ), Tᵃ, perm, (r, t) -> *(α, t)
```
mtfishman, line 179: Let's use `(r, t) -> α * t` to match the style below.

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -175,14 +175,18 @@ function _contract_scalar_perm!(
...
+      )
     end
   elseif isone(β)
     if iszero(α)
       # Rᵃ .= Rᵃ
       # No-op
     else
-      @strided Rᵃ .= α .* permutedims(Tᵃ, perm) .+ Rᵃ
+      Rᵃ = permutedims!!(
+        leaf_parenttype(Rᵃ), Rᵃ, leaf_parenttype(Tᵃ), Tᵃ, perm, (r, t) -> r + α * t
```
mtfishman, line 188: To simplify this kind of code, can't we write it as:
```julia
      Rᵃ = permutedims!!(Rᵃ, Tᵃ, perm, (r, t) -> r + α * t)
```
and then define a function:
```julia
permutedims!!(R, T, perm, f) = permutedims!!(leaf_parenttype(R), R, leaf_parenttype(T), T, perm, f)
```
which does the trait dispatch. That would also make it a bit easier to switch to `SimpleTraits.jl` if we want.

NDTensors/src/dense/densetensor.jl
```diff
@@ -195,7 +199,7 @@ function permutedims!(
 ) where {N,StoreT<:StridedArray}
   RA = array(R)
   TA = array(T)
-  @strided RA .= permutedims(TA, perm)
+  RA .= permutedims!(leaf_parenttype(RA), RA, leaf_parenttype(TA), TA, perm)
```
mtfishman, line 202: Shouldn't need `RA .=` since it is already in-place.

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -393,15 +400,34 @@ function _contract!(
...
+    CT .= permutedims(leaf_parenttype(Cr), Cr, props.PC)
     #end # @timeit
   end
 
   return CT
 end
+
+function mul!!(::Type{<:Array}, CM, ::Type{<:Array}, AM, ::Type{<:Array}, BM, α, β)
+  return @strided mul!(CM, AM, BM, α, β)
+end
```
mtfishman, line 420: Maybe move to `src/array/mul.jl`.

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -393,15 +400,34 @@ function _contract!(
...
+  CM,
+  ::Type{<:AbstractArray},
+  AM,
+  ::Type{<:AbstractArray},
+  BM,
+  α,
+  β,
+)
+  return mul!(CM, AM, BM, α, β)
+end
```
mtfishman, line 433: Maybe move to `src/abstractarray/mul.jl`.

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -393,15 +400,34 @@ function _contract!(
 
   #tC = similar(CM)
   #_gemm!(tA, tB, El(α), AM, BM, El(β), CM)
-  @strided mul!(CM, AM, BM, El(α), El(β))
+  CM = mul!!(
+    leaf_parenttype(CM), CM, leaf_parenttype(AM), AM, leaf_parenttype(BM), BM, El(α), El(β)
+  )
```
mtfishman, line 405: Same comment that I made above for `permutedims!!`, let's just use:
```julia
  CM = mul!!(CM, AM, BM, El(α), El(β))
```
and then define `mul!!` with a layer that then does the trait dispatch on `leaf_parenttype`.

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -343,9 +347,10 @@ function _contract!(
   β::Number=zero(El),
 ) where {El,NC,NA,NB}
   tA = 'N'
+  t = TimerOutput()
```
mtfishman, line 350: Reminder to remove this before we merge the PR.

Also, let's consider moving this `_contract!` function to `src/abstractarray/tensoralgebra/contract.jl`. That will help out with future work of removing the `Dense` type.

NDTensors/src/abstractarray/tensoralgebra/contract.jl
```diff
@@ -0,0 +1,179 @@
...
+  AT::AbstractArray{El,NA},
+  BT::AbstractArray{El,NB},
+  props::ContractionProperties,
+  α::Number=one(El),
+  β::Number=zero(El),
+) where {El,NC,NA,NB}
+  tA = 'N'
+  if props.permuteA
+    #@timeit_debug timer "_contract!: permutedims A" begin
+    Ap = permutedims(leaf_parenttype(AT), AT, props.PA)
```
mtfishman, line 117: Another case where we should hide `leaf_parenttype` at a lower level, I see some more down below.

NDTensors/src/dense/densetensor.jl
```diff
@@ -199,7 +199,7 @@ function permutedims!(
 ) where {N,StoreT<:StridedArray}
   RA = array(R)
   TA = array(T)
-  RA .= permutedims!(leaf_parenttype(RA), RA, leaf_parenttype(TA), TA, perm)
+  RA = permutedims!(leaf_parenttype(RA), RA, leaf_parenttype(TA), TA, perm)
```
mtfishman, line 202: Another case where we should call the version without `leaf_parenttype`. Also, we shouldn't assign to `RA` with `RA = ` here since `permutedims!` is just in-place. Unlike the maybe-permute version `permutedims!!`, we shouldn't assume all in-place functions will output the object being modified.

NDTensors/src/abstractarray/permutedims.jl
```diff
@@ -1,15 +1,21 @@
...
 
-function NDTensors.permutedims!(Mdest::AbstractArray, M::AbstractArray, perm)
+# NDTensors.permutedims!
+function permutedims!(Mdest::AbstractArray, M::AbstractArray, perm)
   return permutedims!(leaf_parenttype(Mdest), Mdest, leaf_parenttype(M), M, perm)
 end
 
+# NDTensors.permutedims!
 function permutedims!(::Type{<:AbstractArray}, Mdest, ::Type{<:AbstractArray}, M, perm)
   return Mdest .= Base.permutedims(M, perm)
```
mtfishman, line 20: Should be able to use `Base.permutedims!(Mdest, M, perm)`, which avoids the intermediate allocation created by `Base.permutedims`.

mtfishman: Note that when the `@strided` macro is used it performs the `permutedims` lazily so an intermediate tensor isn't made.

NDTensors/test/arraytensor/array.jl
```diff
@@ -32,7 +32,7 @@ using NDTensors: storage, storagetype
 
   T1r = randn!(similar(T1))
   @test Array(T1r + T1) ≈ Array(T1r) + Array(T1)
-  @test Array(permutedims(T1, (2, 1))) ≈ permutedims(Array(T1), (2, 1))
+  @test Array(NDTensors.permutedims(T1, (2, 1))) ≈ NDTensors.permutedims(Array(T1), (2, 1))
```
mtfishman, line 35: We should probably define `Base.permutedims(t::Tensor, perm) = NDTensors.permutedims(t, perm)` if that is not already defined so that these work without having to qualify with `NDTensors.`.

NDTensors/src/abstractarray/permutedims.jl
```diff
@@ -0,0 +1,29 @@
...
+  return Base.permutedims!(Mdest, M, perm)
+end
+
+function permutedims!!(B::AbstractArray, A::AbstractArray, perm, f)
+  return permutedims!!(leaf_parenttype(B), B, leaf_parenttype(A), A, perm, f)
+end
+
+function permutedims!!(::Type{<:AbstractArray}, B, ::Type{<:AbstractArray}, A, perm, f)
+  return B .= f.(B, Base.permutedims(A, perm))
+end
```
mtfishman, line 29: I think it would be better to organize this into two layers:
```julia
function permutedims!!(Bleaftype::Type{<:AbstractArray}, B, Aleaftype::Type{<:AbstractArray}, A, perm, f)
  permutedims!(Bleaftype, B, Aleaftype, A, f)
end

function permutedims!(::Type{<:AbstractArray}, B, ::Type{<:AbstractArray}, A, perm, f)
  return B .= f.(B, Base.permutedims(A, perm))
end
```
so then types could customize either one. In general, this is how we should organize maybe-mutating functions, where by default they basically just forward to a mutating function but then they could do something else if the type is immutable, like `UnallocatedZeros`.

NDTensors/src/dense/densetensor.jl
```diff
@@ -199,7 +199,7 @@ function permutedims!(
 ) where {N,StoreT<:StridedArray}
   RA = array(R)
   TA = array(T)
-  RA = permutedims!(leaf_parenttype(RA), RA, leaf_parenttype(TA), TA, perm)
+  RA = permutedims!(RA, TA, perm)
```
mtfishman, line 202:
```suggestion
permutedims!(RA, TA, perm)
```

NDTensors/src/array/mul.jl
```diff
@@ -0,0 +1,4 @@
+function mul!!(::Type{<:Array}, CM, ::Type{<:Array}, AM, ::Type{<:Array}, BM, α, β)
+  @strided CM = mul!(CM, AM, BM, α, β)
```
mtfishman, line 2:
```suggestion
  @strided mul!(CM, AM, BM, α, β)
```

NDTensors/src/array/permutedims.jl
```diff
@@ -0,0 +1,14 @@
+# NDTensors.permutedims
+function permutedims(::Type{<:Array}, M, perm)
+  return @strided Mdest = Base.permutedims(M, perm)
```
mtfishman, line 3:
```suggestion
  return @strided Base.permutedims(M, perm)
```

NDTensors/src/array/permutedims.jl
```diff
@@ -0,0 +1,14 @@
...
+end
+
+# NDTensors.permutedims!
+function permutedims!(::Type{<:Array}, Mdest, ::Type{<:Array}, M, perm)
+  return @strided Mdest .= Base.permutedims(M, perm)
```
mtfishman, line 8:
```suggestion
  @strided Mdest .= Base.permutedims(M, perm)
  return Mdest
```
since there's no guarantee in-place `@strided` operations would return `Mdest`.

NDTensors/src/linearalgebra/linearalgebra.jl
```diff
@@ -218,7 +218,7 @@ function LinearAlgebra.eigen(
   use_relative_cutoff::Bool = get(kwargs, :use_relative_cutoff, use_relative_cutoff)
 
   matrixT = matrix(T)
-  if any(!isfinite, matrixT)
+  if any(!isfinite, parent(matrixT))
```
mtfishman, line 221: I'm guessing this was to avoid element-wise operations on GPU? If so, probably good to leave a comment about that since hopefully that won't be necessary with future designs of the storage type.

PR #1197, opened 2023-09-19
[ITensors] Customisable Arrow directions in SVD

JoeyT1994:
# Description

This PR adds customisable arrow directions to the internal bonds on the `S` tensor in the SVD. Moreover the `factorize_svd()` function has been changed when `ortho = "None"` so that it works correctly with QNs and takes a proper square root of `S.` Tests are included for both fermionic and spin QNS. With the fermionic QNS working due to the ability to customise the arrows and set the arrows on `S` to `ITensors.Out` (currently this does not appear to work if the arrow directions are mixed).

# How Has This Been Tested?

Additional tests added to `test/base/test_svd.jl` which check that `svd()` and `factorize_svd()` give correct decompositions of an `ITensor` when custom arrow directions are used


src/tensor_operations/matrix_algebra.jl
```diff
@@ -87,3 +87,25 @@ function exp(A::ITensor; kwargs...)
...
+map_diag(f::Function, t::Tensor) = map_diag!(f, copy(t), t)
+
+# Convenience functions
+sqrt_diag(it::ITensor) = map_diag(sqrt, it)
+inv_diag(it::ITensor) = map_diag(inv, it)
+invsqrt_diag(it::ITensor) = map_diag(inv ∘ sqrt, it)
+pinv_diag(it::ITensor) = map_diag(pinv, it)
+pinvsqrt_diag(it::ITensor) = map_diag(pinv ∘ sqrt, it)
+sqrtabs_diag(it::ITensor) = map_diag(sqrt ∘ abs, it)
+inv_sqrtabs_diag(it::ITensor) = map_diag(inv ∘ sqrt ∘ abs, it)
```
mtfishman, line 111: For the sake of being conservative, I would vote for not defining these and just relying on `map_diag`, which is basically as easy to use.

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -531,7 +529,22 @@ function factorize_qr(A::ITensor, Linds...; kwargs...)
   return L, R
 end
 
-function factorize_svd(A::ITensor, Linds...; kwargs...)
+#
+# Generic function implementing a square root decomposition of a diagonal, order 2 tensor with inds u, v
+#
+function sqrt_decomp(D::ITensor, u::Index, v::Index)
+  (storagetype(D) <: Diag || storagetype(D) <: DiagBlockSparse) ||
```
mtfishman, line 536: A bit more compact: `storage(D) isa Union{Diag,DiagBlockSparse}`.

test/base/test_svd.jl
```diff
@@ -186,6 +186,51 @@ include(joinpath(@__DIR__, "utils", "util.jl"))
...
   end
 
+  @testset "svd arrow directions" begin
+
+    #Test Spins
+    n = 10
+    χ = 5
+    s = siteinds("S=1/2", n; conserve_qns=true)
+    state = [isodd(i) ? "Up" : "Dn" for i in 1:n]
+    ψ = randomMPS(s, state; linkdims=χ)
```
mtfishman, line 196: Can we test this without using `randomMPS` (i.e. without using the MPS type)? We're trying to avoid that in "basic" tests since we plan to move the MPS/MPO code out of the package once `ITensorNetworks` is more developed.

test/base/test_svd.jl
```diff
@@ -186,6 +186,51 @@ include(joinpath(@__DIR__, "utils", "util.jl"))
...
+    A = ψ[3] * ψ[4]
+    linds = [inds(A)[1]]
+
+    L, R, spec = ITensors.factorize_svd(
+      A, linds...; leftdir=ITensors.Out, rightdir=ITensors.Out, ortho="none"
+    )
+    @test norm(L * R - A) <= 1e-14
+
+    ITensors.disable_auto_fermion()
```
mtfishman, line 231: Let's move this over to the other fermion tests. Also could you test without using MPS?

JoeyT1994:
Okay I changed the tests to work for randomITensors instead of using those made from an MPS. Also moved the fermion test to `test_fermions.jl`

emstoudenmire:
Looks like good code, but I have some questions about the philosophy here of allowing the arrows to be set from outside. Joey (and Matt if you want), could we discuss sometime soon to see if this is the best solution for fixing the fermionic case?

My main concern is that allowing custom arrows on S could result in U (or V) having arrows that don't go "through" but instead both go in or both go out (both in the case of SVD'ing a matrix so that U and V have two indices, I mean). That could be ok if the quantum numbers have appropriately flipped signs to keep the flux of U and V to be zero, but it seems much more natural to me to have the arrows just go through so that the QN's are the same across both the in and out indices.

mtfishman:
I agree that hopefully there is a more general/fundamental fix for the fermionic case so that this isn't strictly necessary for that, but I don't see why this shouldn't be allowed more broadly, i.e. I could definitely imagine cases where it would be nice to control the arrow direction.

mtfishman:
I think you can always choose the QNs appropriately so that `U` and `V` have zero flux, no matter what arrow directions are chosen, and we will still keep that convention (@JoeyT1994 have you checked that?).

mtfishman:
An example which we came across is if you are doing a factorization like:
```julia
A, B = U * S, V
A, B = U, S * V
A, B = U * sqrt(S), sqrt(S) * V
```
it's nice to be able to choose whether the arrow ends up pointing from `A` to `B` or vice versa.

emstoudenmire:
I'm fine with it if we mainly view it as an advanced code capability, to be used in rare cases where the user knows what they are doing. It's just that I would hope we could find solutions where you'd never have to resort to using it, because we could provide good defaults (with an exception possibly being when there are mixed arrows on U or V, in which case any kind of automatic arrow determination becomes ambiguous). 

Sticking to the matrix cases where the original tensor T has two indices, one going onto U and the other onto V, the nice thing about having S mimic the original arrow directions of T is that, in addition to U and V having "through" arrows, which is the most natural case for zero-flux tensors and unitary or isometric maps, if one forms (U*S) or (S*V), each of these will have the same arrow structure as T also.

mtfishman:
I agree that the current defaults are good and most of the time customizing the arrow directions should not be used.

JoeyT1994:
Yes happy to discuss! The default on `svd()` is for the directions to be set to nothing and so it will just default the combiner arrows to what they were? And having the arrows `go through' makes the most sense and there should be no need to customise the arrow dirn.

Nonetheless I do think it is nice to have it customisable as an advanced usage case.

JoeyT1994:
@mtfishman I just changed this to restrict to just specifying one arrow direction when calling `factorize_svd()`. I'm not sure what to call the kwarg though? `Dir` is already taken as a deprecated keyword argument.

mtfishman:
I think we can repurpose `dir` for this use case, it has been erroring for a while now so people shouldn't be using that.

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -531,10 +535,28 @@ function factorize_qr(A::ITensor, Linds...; kwargs...)
...
+  δᵤᵥ = copy(D)
+  map_diag!(sign, δᵤᵥ, D)
+  return sqrtDL, prime(δᵤᵥ), sqrtDR
+end
+
+function factorize_svd(A::ITensor, Linds...; (singular_values!)=nothing, kwargs...)
   ortho::String = get(kwargs, :ortho, "left")
   alg::String = get(kwargs, :svd_alg, "divide_and_conquer")
-  USV = svd(A, Linds...; kwargs..., alg=alg)
+  dir = get(kwargs, :dir, ITensors.In)
```
mtfishman, line 556: Let's move this as well as `ortho` and `alg` into the function signature since we are trying to move towards that style in general (as opposed to using `get`).

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -647,11 +676,11 @@ function factorize(A::ITensor, Linds...; kwargs...)
...
-  if haskey(kwargs, :dir)
-    error("""dir keyword in factorize has been replace by ortho.
-    Note that the default is now `left`, meaning for the results
-    L,R = factorize(A), L forms an orthogonal basis.""")
-  end
+  # if haskey(kwargs, :dir)
+  #   error("""dir keyword in factorize has been replace by ortho.
+  #   Note that the default is now `left`, meaning for the results
+  #   L,R = factorize(A), L forms an orthogonal basis.""")
+  # end
```
mtfishman, line 683: Let's just delete these lines (and the one above).

test/base/test_decomp.jl
```diff
@@ -111,7 +111,7 @@ end
     i = Index(2, "i")
     j = Index(2, "j")
     A = randomITensor(i, j)
-    @test_throws ErrorException factorize(A, i; dir="left")
+    #@test_throws ErrorException factorize(A, i; dir="left")
```
mtfishman, line 114: Let's delete this line.

test/base/test_svd.jl
```diff
@@ -186,6 +186,31 @@ include(joinpath(@__DIR__, "utils", "util.jl"))
...
+    r1, r2, r3 = Index(QN("Sz", -2) => 1, QN("Sz", 1) => 1; tags="r1", dir=ITensors.Out),
+    Index(QN("Sz", 2) => 1, QN("Sz", 1) => 1; tags="r2", dir=ITensors.In),
+    Index(QN("Sz", -2) => 1, QN("Sz", 1) => 1; tags="r3", dir=ITensors.In)
+    A = randomITensor(l1, l2, r1, r2, r3)
+
+    for leftdir in [ITensors.Out, ITensors.In]
+      for rightdir in [ITensors.Out, ITensors.In]
+        U, S, V = svd(A, l1, l2; leftdir, rightdir)
+        s1, s2 = inds(S)[1], inds(S)[2]
+        @test ITensors.dir(s1) == leftdir
```
mtfishman, line 201: Can just directly use `dir` instead of `ITensors.dir`.

test/base/test_svd.jl
```diff
@@ -186,6 +186,31 @@ include(joinpath(@__DIR__, "utils", "util.jl"))
...
+    Index(QN("Sz", 2) => 1, QN("Sz", 1) => 1; tags="l2", dir=ITensors.Out)
+    r1, r2, r3 = Index(QN("Sz", -2) => 1, QN("Sz", 1) => 1; tags="r1", dir=ITensors.Out),
+    Index(QN("Sz", 2) => 1, QN("Sz", 1) => 1; tags="r2", dir=ITensors.In),
+    Index(QN("Sz", -2) => 1, QN("Sz", 1) => 1; tags="r3", dir=ITensors.In)
+    A = randomITensor(l1, l2, r1, r2, r3)
+
+    for leftdir in [ITensors.Out, ITensors.In]
+      for rightdir in [ITensors.Out, ITensors.In]
+        U, S, V = svd(A, l1, l2; leftdir, rightdir)
+        s1, s2 = inds(S)[1], inds(S)[2]
```
mtfishman, line 200: Can use `ind(S, 1)`.

mtfishman:
Some tests are failing which check for bad inputs for specifying the `svd_alg` backend, which should be getting forwarded as the `alg` kwarg to `svd` through `factorize_svd`. Maybe the current code logic is not forwarding that keyword argument properly.

JoeyT1994:
Ah I think I found the bug and should be fixed now!

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -106,7 +106,13 @@ Utrunc2, Strunc2, Vtrunc2 = svd(A, i, k; cutoff=1e-10);
 
 See also: [`factorize`](@ref), [`eigen`](@ref)
 """
-function svd(A::ITensor, Linds...; kwargs...)
+function svd(
+  A::ITensor,
+  Linds...;
+  leftdir::Union{Nothing,ITensors.Arrow}=nothing,
+  rightdir::Union{Nothing,ITensors.Arrow}=nothing,
```
mtfishman, line 113: Can just use `Arrow` since we are in the ITensors module.

mtfishman: Better yet, let's just drop the type constraint, it's not really needed.

mtfishman:
Looks good, thanks!

PR #1200, opened 2023-09-21
[ITensors] Tests for general fermionic H

emstoudenmire:
We have not previously had a test for a very general type of fermionic Hamiltonian. This PR adds tests of a rather general H (actually not even assuming it is Hermitian) for both the usual OpSum system and with the auto fermion system enabled.


mtfishman:
@emstoudenmire this seems like an unfortunate duplication of code, and also introduces code that involves MPS/MPO into `test/base` instead of `test/ITensorLegacyMPS/base`.

Can we combine into one testset in `test/ITensorLegacyMPS/base`? As we discussed, I think it is an "anti-pattern" to rely on MPS/MPO tests to test core functionality like auto fermions. If we detect an auto fermion failure through an MPS/MPO test, that reflects a failure of not having enough basic unit tests that work with core NDTensor or ITensor objects like `Tensor`, `ITensor`, etc. Also as discussed this will help with plans for moving the MPS/MPO code to a separate package once ITensorNetworks.jl is more developed.

mtfishman:
Relatedly, it would be nice to move the other fermion tests that make use of MPS/MPO functionality to `test/ITensorLegacyMPS/base`.

emstoudenmire:
Makes sense. I moved all of the MPS, MPO, and OpSum related tests from base/test_fermions.jl into either ITensorLegacyMPS/base/test_autompo.jl or a new file ITensorLegacyMPS/base/test_fermions.jl . It might be better to even move the tests in that new test_fermions.jl file into other existing files, but keeping them all together does help with things like not having to call ITensors.enable_auto_fermions() in multiple places.

I definitely agree overall with the point about not testing AutoFermion too much through higher-level things like MPS algorithms. So probably some of those tests should be deleted eventually. But for now there is a kind of situation where certain issues, like the MPS(sites,state) product MPS constructor not putting in links with arrows in a very certain way worked for every other case and only failed for fermions. However, it wasn't exactly a failure of the AutoFermion system per se, it's more that it was always the mathematically correct thing to do (e.g. SU(2) might need it also), it's just that you can get away with it in every other situation we have right now (e.g. Abelian symmetries or complex tensors).

mtfishman:
Thanks!

PR #1202, opened 2023-09-21
[NDTensors] Add `SmallVectors` submodule

mtfishman:
# Description

This adds a submodule to `NDTensors` called `SmallVectors` which defines the types `SmallVector` and `MSmallVector` which are stack allocated with a maximum compile-time length but have a runtime length that can be less than the maximum  compile-time length. The goal is for this type to be used by the `QN` type and the `TagSet` type, which both have a similar storage requirement. This should also make it easier to allow users to customize the types if they want to.

Here's an example of some functionality:
```julia
julia> using NDTensors.SmallVectors

julia> v = SmallVector{10}([1, 3, 5])
3-element SmallVector{10, Int64}:
 1
 3
 5

julia> v = SmallVectors.push(v, 7)
4-element SmallVector{10, Int64}:
 1
 3
 5
 7

julia> v = SmallVectors.insertsorted(v, 2)
5-element SmallVector{10, Int64}:
 1
 2
 3
 5
 7
```
The test suite has various tests of common operations to test that they have zero allocations.

@emstoudenmire

### DIFF
--- a/NDTensors/src/SmallVectors/test/runtests.jl
+++ b/NDTensors/src/SmallVectors/test/runtests.jl
@@ -47,8 +47,7 @@ function test_smallvectors()
     @test @inferred(length(SmallVectors.buffer(x))) == 10
     @test @allocated(length(SmallVectors.buffer(x))) == 0
 
-    # More allocation are present in older Julia versions
-    nalloc_limit = VERSION < v"1.9" ? 128 : 0
+    nalloc_limit = 128
 
     item = 115
     no_broken = (false, false, false, false)
--- a/NDTensors/src/SmallVectors/test/runtests.jl
+++ b/NDTensors/src/SmallVectors/test/runtests.jl
@@ -1,78 +1,116 @@
 module TestSmallVectors
-  include("SmallVectors/src/SmallVectors.jl")
-  using .SmallVectors
-  using StaticArrays
-  using Test
+include("SmallVectors/src/SmallVectors.jl")
+using .SmallVectors
+using StaticArrays
+using Test
 
-  function test()
-    @testset "SmallVectors" begin
-      x = SmallVector{10}([1, 3, 5])
-      mx = MSmallVector(x)
+function test()
+  @testset "SmallVectors" begin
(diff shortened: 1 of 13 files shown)


mtfishman:
@emstoudenmire one goal was to define an abstract interface `AbstactSmallVector`, so that the `QN` type can both wrap a `SmallVector` as it's storage, and also be a subtype of `AbstactSmallVector` so efficient stack-allocateed operations automatically work (using a few function overloads). You can see the interface requirements for subtypes based on the code for `SmallVector`, `MSmallVector`, `SubSmallVector`, and `SubMSmallVector` (the last two are used to define efficient contiguous slices of SmallVectors), the interface is pretty minimal and mostly involves defining getting and setting elements, some constructors, access to the underlying buffer, etc.

mtfishman:
Unfortunately the design of the mutable types right now relies on a feature that was made available in Julia 1.8 for marking certain fields as `const` (https://github.com/JuliaLang/julia/blob/master/HISTORY.md#new-language-features-2), to have control over which fields or mutable or not. For now I'll just test against Julia 1.8 and later and work out how to make it backwards compatible to Julia 1.6 but still efficient.

emstoudenmire:
Looks good & the interface is clear. More convenient that a plain SArray (or MArray) for sure. When it's merged I will go ahead and use it for the new QN type.

NDTensors/src/SmallVectors/test/runtests.jl
```diff
@@ -0,0 +1,156 @@
...
+    @test @allocated(SmallVector(mx)) == 0
+
+    # length
+    @test @inferred(length(x)) == 3
+    @test @allocated(length(x)) == 0
+    @test @inferred(length(SmallVectors.buffer(x))) == 10
+    @test @allocated(length(SmallVectors.buffer(x))) == 0
+
+    # More allocation are present in older Julia versions
+    nalloc_limit = VERSION < v"1.9" ? 128 : 0
```
mtfishman, line 51:
```suggestion
    nalloc_limit = 128
```

PR #1204, opened 2023-09-26
[NDTensors] Add SortedSets and TagSets

mtfishman:
# Description

This adds `SortedSets` and `TagSets` submodules to NDTensors.

`SortedSets` defines a generic sorted set-like data structure which are more efficient for smaller collections than hash-based sets. Inserting new elements keeps the set sorted (and unique). When combined with `SmallVectors` from #1202 (using a `SmallVector` as a data backend for a `SortedSet`), it should provide very fast data structures for applications like TagSets and QNs. This PR also adds a first draft of a new `TagSet` type design, which is generic but can make use of `SmallVectors` and `SortedSets` to reproduce speed that is similar to the current `ITensors.TagSet` type but with less code and a simpler, more generic design that will allow users to choose their own tag and storage types.

This is still a draft because I need to test the efficiency of `SortedSets` (I know there is some part that doesn't fully take advantage of the speed of using a `SmallVector` but instead uses slower generic code), work out a few more things around the design of the `TagSet` (and relatedly, `QN`) type, and add tests and examples.


mtfishman:
@emstoudenmire I'm going to merge this so we can start testing it out in the wild. You can look at the new `TagSet` type (just a prototype for now, not yet used in `ITensors.jl` to replace the current one) to get an idea for how to design a `QN` type.

Some set operations on `SortedSet` (and the associated `SmallSet`, which is a type alias for a `SortedSet` wrapping a `SmallVector`) still need to be optimized, but that will be relatively simple. Basically each one (`union`, `intersect`, setdiff`, `symdiff`) can take advantage of the data structures being sorted to run in linear time in the length of the sets, but I only implemented that for `union` and `setdiff` so far.

PR #1205, opened 2023-09-26
[NDTensors] `BlockSparseArrays` prototype

mtfishman:
# Description

This is a prototype for a `BlockSparseArray` type based on the [BlockArrays.jl](https://github.com/JuliaArrays/BlockArrays.jl) interface. It is based on wrapping an elementwise sparse array (`SparseArray`, also introduced in this PR) which uses a dictionary-of-keys data storage, similar to the design of [SparseArrayKit.jl](https://github.com/Jutho/SparseArrayKit.jl).

It is fairly functional with just this amount of code, for example you can get and set elements and blocks and some basic linear algebra works (but isn't optimized for sparsity).

An interesting next step would be to define some tensor operations (perhaps using the interface defined in [TensorOperations.jl](https://github.com/Jutho/TensorOperations.jl)), remove the type restriction on `Tensor` constraining the `storage` to `TensorStorage`, and see if we can get it working as an `ITensor` backend.

@kmp5VT I think this could be a good starting point to build off of for reworking the block sparse code. It already allows blocks with general types. The current `BlockSparseArray` type defined here can have blocks that are slices of a contiguous piece of data but also can be totally general `AbstractArray`s. If we add a `DiagArray` type and store them inside the blocks, it would cover our current use cases for the `BlockSparse` and `DiagBlockSparse` storage types. A more fully functional type that works with a contiguous data backend will probably require a slightly more sophisticated type, however, which stores an extra `data` field separate from the block structure, and the block structure stores `SubArray`s which are slices into that data.

### DIFF
--- a/NDTensors/src/BlockSparseArrays/src/sparsearray.jl
+++ b/NDTensors/src/BlockSparseArrays/src/sparsearray.jl
@@ -1,0 +1,30 @@
+struct SparseArray{T,N} <: AbstractArray{T,N}
+  data::Dictionary{CartesianIndex{N},T}
+  dims::NTuple{N,Int64}
+end
+
+Base.size(a::SparseArray) = a.dims
+
+function Base.setindex!(a::SparseArray{T,N}, v, I::CartesianIndex{N}) where {T,N}
+  set!(a.data, I, v)
+  return a
+end
+function Base.setindex!(a::SparseArray{T,N}, v, I::Vararg{Int,N}) where {T,N}
+  return setindex!(a, v, CartesianIndex(I))
+end
+
+function Base.getindex(a::SparseArray{T,N}, I::CartesianIndex{N}) where {T,N}
+  return get(a.data, I, nothing)
+end
+function Base.getindex(a::SparseArray{T,N}, I::Vararg{Int,N}) where {T,N}
+  return getindex(a, CartesianIndex(I))
+end
+
+# `getindex` but uses a default if the value is
+# structurally zero.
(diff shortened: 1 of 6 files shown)


emstoudenmire:
Nice that it's already up and running with so little code.

NDTensors/src/BlockSparseArrays/src/sparsearray.jl
```diff
@@ -0,0 +1,31 @@
+struct SparseArray{T,N} <: AbstractArray{T,N}
+  data::Dictionary{CartesianIndex{N},T}
+  dims::NTuple{N,Int64}
```
mtfishman, line 3: Reminder to add a `zero` field, i.e.:
```suggestion
  data::Dictionary{CartesianIndex{N},T}
  dims::NTuple{N,Int64}
  zero::Zero
```
This would be a custom `zero` function for getting zero elements that aren't just determined from the element type as `zero(T)`, for example when the elements are blocks, in which case the `zero` function would need to know about the block sizes. It should be a function of `T` and the `CartesianIndex` `I`, and could default to `zero(T)`.

mtfishman: Completed.

mtfishman:
@kmp5VT I'm going to merge this so we can start using it.

You can already use it as an `ITensor` storage type, but many operations won't work (so far just getting and setting elements will work, but tensor contractions, additions, and decompositions will fail).

PR #1206, opened 2023-09-28
[NDTensor] Allow general objects are storage backends for Tensor

mtfishman:
# Description

This relaxes the type constraint on the `storage` field of `Tensor` to allow any objects as storage backends.

The idea would be that the `storage` should adhere to a minimal interface requirement, for example in general that it is an `AbstractArray` (though that isn't required) and defines some contraction and permutation functionality (though if we provide a general TTGT backend, we could in general only require permutations and matrix multiplications) as well as factorizations like SVD or QR.

You can see the interface requirements demonstrated in `arraytensor/array.jl`. Some of that can be made more generic, and therefore won't need to be defined for every storage backend.

For now, the design is that non-TensorStorage storage backends are "registered" in the type union `ArrayStorage`, only for the purpose of ensuring that dispatch doesn't get confused between new functionality we write for generic storage backends as opposed to the current `TensorStorage` type which has particular requirements, so that we don't have to change current `TensorStorage` code. Once we replace the current `TensorStorage` types in favor of `AbstractArray` backends (`Array`, `BlockSparseArray`, `DiagonalArray`, etc.) we can just remove the `ArrayStorage` type union (and maybe replace it with `AbstractArray` where appropriate).

@kmp5VT your PR #1161 aligns with this new design since you are getting rid of the `EmptyStorage` type (already in the right direction of removing s`TensorStorage` types) and also the `UnallocatedZeros` type will still be relevant here since it will either be used directly as a `Tensor` storage or used as a data backend for one of the other storage types like `BlockSparseArray` or `DiagonalArray`. Additionally, I don't anticipate getting rid of our `contract` or `permutedims` code, just making sure they both work for the new storage types, so the optimizations you are doing in #1194 are definitely still important.

A strategy for how to proceed with the goal of removing `TensorStorage` types is to "peel off" the storage types one by one. For example, we can start by making sure that using an `Array` as a storage backend for `Tensor` has all of the same functionality as `Dense` (which I already started to do in this PR, as you can see it already works for a number of operations like addition, permutation, contraction, etc.). Then we can just change the default `ITensor` constructor to use `Array` (or `UnallocatedZeros`) storage instead of `Dense` storage when there are dense indices. Then, once we develop external `BlockSparseArray` and `DiagonalArray` AbstractArrays (`BlockSparseArray` already started in #1205), once we make them feature-complete with `BlockSparse` and `Diag` we can use them as defaults as well, and then just remove `BlockSparse` and `Diag`.

The `Combiner` is a slightly awkward corner case, but I think we can basically just leave it as-is for now and hopefully it get replaced by a more principle fusion tree type in the future.

### DIFF
--- a/NDTensors/src/arraytensor/arraytensor.jl
+++ b/NDTensors/src/arraytensor/arraytensor.jl
@@ -22,7 +22,9 @@ const MatrixStorageTensor{T,S,I} = Tensor{T,2,S,I} where {S<:MatrixStorage{T}}
 const MatrixOrArrayStorageTensor{T,S,I} =
   Tensor{T,N,S,I} where {N,S<:MatrixOrArrayStorage{T}}
 
-Tensor(storage::MatrixOrArrayStorageTensor, inds::Tuple) = Tensor(NeverAlias(), storage, inds)
+function Tensor(storage::MatrixOrArrayStorageTensor, inds::Tuple)
+  return Tensor(NeverAlias(), storage, inds)
+end
 
 function Tensor(as::AliasStyle, storage::MatrixOrArrayStorage, inds::Tuple)
   return Tensor{eltype(storage),length(inds),typeof(storage),typeof(inds)}(
--- a/NDTensors/src/arraytensor/arraytensor.jl
+++ b/NDTensors/src/arraytensor/arraytensor.jl
@@ -1,0 +1,90 @@
+# Used for dispatch to distinguish from Tensors wrapping TensorStorage.
+# Remove once TensorStorage is removed.
+const ArrayLike{T,N} = Union{Array{T,N},ReshapedArray{T,N},SubArray{T,N}}
+const MatrixLike{T} = Union{
+  Matrix{T},
+  Transpose{T},
+  Adjoint{T},
+  Symmetric{T},
+  Hermitian{T},
+  UpperTriangular{T},
+  LowerTriangular{T},
+  UnitUpperTriangular{T},
+  UnitLowerTriangular{T},
+  Diagonal{T},
+}
+const MatrixOrArrayLike{T} = Union{MatrixLike{T},ArrayLike{T}}
+
+const ArrayLikeTensor{T,N,S,I} = Tensor{T,N,S,I} where {S<:ArrayLike{T,N}}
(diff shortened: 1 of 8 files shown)


mtfishman:
@emstoudenmire I think as you design the `FusionTensor` type (or whatever it is called), you should add it to the `ArrayStorage` type union instead of making it a `TensorStorage` subtype, and design it as a standalone array-like object independent of the `Tensor` type at first (whether or not is `AbstractArray` shouldn't really matter, but I could see an argument for making at an `AbstractArray` but maybe disallowing `getindex` and `setindex!`).

NDTensors/src/arraytensor/arraytensor.jl
```diff
@@ -0,0 +1,100 @@
...
+  F = svd(storage(tens))
+  U, S, V = F.U, F.S, F.Vt
+  i, j = inds(tens)
+  # TODO: Make this more general with a `similar_ind` function,
+  # so the dimension can be determined from the length of `S`.
+  min_ij = dim(i) ≤ dim(j) ? i : j
+  α = sim(min_ij) # similar_ind(i, space(S))
+  β = sim(min_ij) # similar_ind(i, space(S))
+  Utensor = tensor(U, (i, α))
+  # TODO: Remove conversion to `Matrix` to make more general.
```
mtfishman, line 85:
```suggestion
  # TODO: Remove conversion to `Diagonal` to make more general, or make a generic `Diagonal` concept that works for `BlockSparseArray`.
```

NDTensors/test/arraytensor/arraytensor.jl
```diff
@@ -0,0 +1,53 @@
...
+  D12 = contract(D1, (1, -1), D2, (-1, 2))
+  @test D12 ≈ Array(T12)
+
+  ## IT1 = itensor(T1)
+  ## IT2 = itensor(T2)
+  ## IR = IT1 * IT2
+  ## IX = IT1 + IT1
+
+  ## IU, IS, IV = svd(IT1, i)
+  ## @test IU * IS * IV ≈ IT1
```
mtfishman, line 52:
```suggestion
```

mtfishman:
@emstoudenmire @kmp5VT I'll merge this so we can build on top of it going forward.

emstoudenmire:
Looks really good and thanks for the guidance about the fusion storage type. That's what I needed to know.

PR #1212, opened 2023-10-17
Fix fermion cases in blocksparse SVD

emstoudenmire:
Some previously working cases of the SVD for auto fermion stopped working due to changes in `NDTensors/src/blocksparse/linearalgebra.jl`. This PR changes those lines of code back to a correctly working version and restores the corresponding tests which were marked as broken.

With this PR, all four "matrix" cases of the SVD should again be working. 

The other cases which are still marked as broken are also likely conceptually correct as long as one correctly conjugates certain mixed arrows with a parity tensor before determining whether U or V are unitary. So in a sense it is those last two tests which are broken, not the ITensor code. (This PR does not address those tests; just mentioning it for context.)

### DIFF
--- a/NDTensors/src/blocksparse/linearalgebra.jl
+++ b/NDTensors/src/blocksparse/linearalgebra.jl
@@ -186,7 +186,7 @@ function LinearAlgebra.svd(T::BlockSparseMatrix{ElT}; kwargs...) where {ElT}
     # is the second index:
     sVP = 1
     if using_auto_fermion()
-      sVP = block_parity(vind, blockV[2]) == 1 ? -1 : +1
+      sVP = -block_sign(vind, blockV[2])
     end
 
     if (sV * sVP) == -1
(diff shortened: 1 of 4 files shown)


NDTensors/src/blocksparse/linearalgebra.jl
```diff
@@ -184,9 +184,12 @@ function LinearAlgebra.svd(T::BlockSparseMatrix{ElT}; kwargs...) where {ElT}
     # This sign (sVP) accounts for the fact that
     # V is transposed, i.e. the index connecting to S
     # is the second index:
-    sVP = using_auto_fermion() ? -block_parity(vind, blockV[2]) : 1
+    sVP = 1
+    if using_auto_fermion()
+      sVP = block_parity(vind, blockV[2]) == 1 ? -1 : +1
```
mtfishman, line 189: Maybe could be `sVP = -block_parity(vind, blockV[2])`.

mtfishman:
Thanks for catching that, that was my mistake.

NDTensors/src/blocksparse/linearalgebra.jl
emstoudenmire: The issue with the previous code was that `block_parity` outputs 0 or 1, not +1 or -1. We could define a new function maybe called `block_sign` though.

mtfishman: `block_sign` sounds good.

PR #1215, opened 2023-10-22
[NDTensors] Get more block sparse operations working on GPU

mtfishman:
This is an initial attempt at getting more block sparse operations working on GPU.

The main issue was that block sparse factorizations like QR, eigendecomposition, and SVD were implicitly hardcoded to transfer data to CPU in certain places, which is fixed by calling more general constructors and making use of `leaf_parenttype`.

DMRG now runs on GPU with conserved quantities, though I only test on Metal GPUs so far. There are a few caveats:
1. The performance on Metal GPUs is pretty bad, we'll have to track down which operations are slow.
2. Block sparse SVD is still broken for now, I have to keep investigating that. There are some issues on Metal GPUs with performing permutations of wrapped MtlArrays.
3. On Metal, factorizations like QR, eigendecomposition, and SVD aren't available so they are performed by moving data back and forth to CPU.

This PR is more about getting things working and performance can be tested in future PRs, so my primary goal will be getting block sparse SVD working (i.e. running without errors).


mtfishman:
I fixed block sparse SVD in the latest, so DMRG now runs with QN conservation for any cutoff values. It is still pretty slow, but again the goal here is to get things running and I think analyzing the timings, adding tests, etc. can be left for future PRs.

emstoudenmire:
Great to see how much more can be made to work already. Goes to show the value of good design. Makes sense about the performance not immediately being there.

mtfishman:
> Great to see how much more can be made to work already. Goes to show the value of good design. Makes sense about the performance not immediately being there.

Definitely a testament to the work that Karl and I have been doing developing better generic code patterns across CPU and GPU. Going through these changes to the block sparse code, it was also clear to me how much simpler a lot of it will get when we switch over to using the new `BlockSparseArray` type.

mtfishman:
I'm going to merge this so we can build on top of it in future PRs.

PR #1216, opened 2023-10-23
Kmp5/enhancements/more gpu

kmp5VT:
# Description
As a continued effort to make the ITensors efficient on GPU in this branch I am identifying and fixing areas that were missed in previous PRs. First I am looking at the profiling of GPU based SVD and will also address issues in [1193](https://github.com/ITensor/ITensors.jl/issues/1193)


NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
```diff
@@ -11,3 +11,13 @@ function NDTensors.svd_catch_error(A::CuMatrix; alg="JacobiAlgorithm")
...
+
+## TODO Here truncate does logical operations of the values in P
+## So its more efficient to just make it a CPU vector and 
+## convert back to GPU
+function NDTensors.truncate!(P::CuArray; kwargs...)
+  cpuP = NDTensors.cpu(P)
+  value = NDTensors.truncate!(cpuP; kwargs...)
+  P = adapt(typeof(P), cpuP)
+  return value
+end
```
mtfishman, line 23: I think it is better to use `leaf_parenttype` dispatch here. Then we can do:
```julia
function truncate!!(P::AbstractArray; kwargs...)
  return truncate!!(leaf_parenttype(P), P; kwargs...)
end

# CPU version
function truncate!!(::Type{<:Array}, P::AbstractArray; kwargs...)
  truncate!(P; kwargs...)
  return P
end

# GPU fallback version
function truncate!!(::Type{<:AbstractArray}, P::AbstractArray; kwargs...)
  P_cpu = cpu(P)
  truncate!(P_cpu; kwargs...)
  return adapt(leaf_parenttype(P), P_cpu)
end

# CPU implementation
function truncate!(P::Vector; kwargs...)
  # Implementation
end
```
Note I'm suggesting using `truncate!!` since the CPU version modifies in-place but the GPU version is out-of-place.

mtfishman: This would also generically handle other GPU backends, not just CUDA.

kmp5VT: This makes sense. The original version was a proof of concept and I have updated to use the `truncate!!` system. Will do the same with the subsequent changes!

mtfishman:
A lot of this PR is superseded by #1215.

mtfishman:
@kmp5VT can we close this? Should be superseded by #1215 and #1217. The only thing I didn't add was your change to `copy` but seems like we can add that later as needed (using `leaf_parenttype` dispatch).

kmp5VT:
@mtfishman yes I'll close this PR and open a new one if I find something awry. Thanks!

PR #1218, opened 2023-10-25
[NDTensors][Bug] Fix CPU performance issue

kmp5VT:
# Description

Joey found that contractions slowed down by roughly a factor of 5-10. It turned out that a permutedim was being evaluated lazily resulting in a complicated type that wasn't being parsed properly and as a result generic `mul!` was begin called. Here I create a function to unwrap a strided array to properly show its parenttype and I force evaluate of a permutation to ensure the strided version of mul is called.

# Checklist:

- [ ] Fix Joeys CPU `mul!` slowdown

### DIFF
--- a/NDTensors/src/array/permutedims.jl
+++ b/NDTensors/src/array/permutedims.jl
@@ -1,6 +1,6 @@
 # NDTensors.permutedims
 function permutedims(::Type{<:Array}, M, perm)
-  return @strided Base.permutedims(M, perm)
+  return @strided A = Base.permutedims(M, perm)
 end
 
 # NDTensors.permutedims!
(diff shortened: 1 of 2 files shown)


NDTensors/src/array/permutedims.jl
```diff
@@ -1,6 +1,6 @@
 # NDTensors.permutedims
 function permutedims(::Type{<:Array}, M, perm)
-  return @strided Base.permutedims(M, perm)
+  return @strided A = Base.permutedims(M, perm)
```
mtfishman, line 3: Let's leave a comment here to say we do this to evaluate the permutation avoid returning a `StridedView`.

Also I think this would be clearer for when we look back at the code:
```julia
  @strided Mperm = Base.permutedims(M, perm)
  return Mperm
```

kmp5VT: I updated the code and added a comment. Thanks!

PR #1220, opened 2023-10-26
[NDTensors] `Unwrap` module for dispatching on unwrapped types

### LINKED ISSUE #1193: [NDTensors] [BUG] Many basic tensor operations fail with CUDA extension due to `LoadError: Scalar indexing is disallowed.`

kmp5VT:
# Description
In this PR I attempt to address issues we are having with generic programming. This issue is that wrapped arrays can become too complicated for Julia to parse correctly for dispatch and as a result, incorrect functions are called which in turn slows down the code.  Here I am creating a simple `Expose` type which will determine the base unwrapped type and launch the appropriate kernels based on this type.

This will fix #1193 (we should confirm that before merging).

# Checklist:

- [x] Create Expose type and function to create type.
- [x] Replace NDTensors code where we implicitly unwrap code (through parent_leaftype) and instead dispatch on the `expose` struct
- [x] Ensure that all tests still pass in NDTensors and ITensors
- [x] Fix scalar indexing issues in GPU SVD code
- [x] address merge conflicts


NDTensors/src/Unwrap/src/iswrappedarray.jl
```diff
@@ -0,0 +1,57 @@
...
+
+# For working with instances, not used by
+# `SimpleTraits.jl` traits dispatch.
+is_wrapped_array(array::AbstractArray) = is_wrapped_array(typeof(array))
+
+# By default, the `parentype` of an array type is itself
+parenttype(arraytype::Type{<:AbstractArray}) = arraytype
+
+# TODO: Use `SetParameters` here.
+parenttype(::Type{<:Base.ReshapedArray{<:Any,<:Any,P}}) where {P} = P
```
mtfishman, line 25: Since you used `using Base: ReshapedArray` above you should be able to just use `ReshapedArray` here, instead of `Base.ReshapedArray`.

NDTensors/src/Unwrap/src/Unwrap.jl
```diff
@@ -0,0 +1,16 @@
+module Unwrap
+using SimpleTraits
+using LinearAlgebra
+using Base: ReshapedArray
+using Strided: StridedView
```
mtfishman, line 5: Can use the more minimal [StridedViews.jl](https://github.com/Jutho/StridedViews.jl) which just defines `StridedView` that is used by `Strided.jl`.

kmp5VT: I think I fixed this using `Strided.StridedViews` is that what you mean?

NDTensors/src/Unwrap/src/functions/permutedims.jl
```diff
@@ -0,0 +1,3 @@
+function permutedims(E::Exposed{<:AbstractArray}, perm)
```
mtfishman, line 1: For simplicity we could just use `::Exposed` instead of `::Exposed{<:AbstractArray}`.

NDTensors/ext/NDTensorsMetalExt/mul.jl
```diff
@@ -1,15 +1,11 @@
...
 # TODO: Raise an issue with `Metal.jl`.
-function NDTensors.mul!!(
-  ::Type{<:MtlArray},
-  CM::Transpose,
-  ::Type{<:MtlArray},
-  AM::AbstractMatrix,
-  ::Type{<:MtlArray},
-  BM::AbstractMatrix,
+function LinearAlgebra.mul!(
+  CM::Unwrap.Exposed{<:MtlArray},
```
mtfishman, line 4: I think this should be:
```julia
CM::Unwrap.Exposed{<:MtlArray,<:Transpose},
```
since there was only an issue in that case.

NDTensors/ext/NDTensorsMetalExt/permutedims.jl
```diff
@@ -1,3 +1,10 @@
+function Base.permutedims!(
+  Edest::Unwrap.Exposed{<:MtlArray}, Esrc::Unwraped.Exposed{<:MtlArray}, perm
+)
+  Eperm = expose(permutedims(Esrc), perm)
```
mtfishman, line 4: Looks like `perm` is not in the right location.

NDTensors/ext/NDTensorsMetalExt/permutedims.jl
```diff
@@ -1,3 +1,10 @@
+function Base.permutedims!(
+  Edest::Unwrap.Exposed{<:MtlArray}, Esrc::Unwraped.Exposed{<:MtlArray}, perm
```
mtfishman, line 2: Probably needs a more specific signature here since this shouldn't be used for all `MtlArray` operations, i.e. specialize on what wrapper type was giving trouble (I think `ReshapedArray`)?

NDTensors/src/Unwrap/src/Unwrap.jl
```diff
@@ -0,0 +1,20 @@
...
+include("iswrappedarray.jl")
+include("expose.jl")
+
+include("import.jl")
+## TODO Create functions which take the `Expose` type and launch functions
+## using that type
+## Exposed based functions
+include("functions/permutedims.jl")
+
+export IsWrappedArray, is_wrapped_array, parenttype, unwrap_type, expose, Expose
```
mtfishman, line 16: Should be `Exposed`, not `Expose`.

NDTensors/ext/NDTensorsMetalExt/permutedims.jl
kmp5VT: Makes sense! I knew there would be some issues with the metal implementation I just wasn't able to check on my workstation. I'm on my mac today and will verify this code. Thanks!

kmp5VT: Ahh yes I see in the next line down you have `Adest::Base.ReshapedArray{<:Any,<:Any,<:SubArray}`  so I'll make `Edest::Exposed{<:MtlArray, <:ReshapedArray}` and will run tests to see if that works

NDTensors/ext/NDTensorsMetalExt/mul.jl
```diff
@@ -1,15 +1,11 @@
...
+function LinearAlgebra.mul!(
+  CM::Exposed{<:MtlArray,<:Transpose},
+  AM::Exposed{<:MtlArray},
+  BM::Exposed{<:MtlArray},
   α,
   β,
 )
-  mul!(transpose(CM), transpose(BM), transpose(AM), α, β)
-  return CM
+  return mul!(transpose(CM.object), transpose(BM.object), transpose(AM.object), α, β)
```
mtfishman, line 10: We should explicitly use `return CM` as before.

Again, for mutating/in-place functions (`f!(dest, ...)`) we should call it in a separate line and then return the modified value `dest`, since we can't assume the `f!` will in general return `dest`. For maybe-mutating functions (`f!!(dest, ...)`)  the convention is that they need to return `dest` so then what you wrote here would be ok.

NDTensors/ext/NDTensorsMetalExt/permutedims.jl
```diff
@@ -1,3 +1,10 @@
+function Base.permutedims!(
+  Edest::Exposed{<:MtlArray,<:Base.ReshapedArray}, Esrc::Exposed{<:MtlArray}, perm
+)
+  Eperm = expose(permutedims(Esrc, perm))
```
mtfishman, line 4: The way you have it written now, `expose` isn't doing anything since you just unwrap it in the next line.

Perhaps you want:
```julia
function Base.permutedims!(
  Edest::Exposed{<:MtlArray,<:Base.ReshapedArray}, Esrc::Exposed{<:MtlArray}, perm
)
  Eperm = expose(permutedims(Esrc, perm))
  copyto!(expose(parent(Edest.object)), Eperm)
  return Edest.object
end
```
Note that here I'm trying to ensure `Exposed` objects get passed to `copyto!`, in case there are issues that out own version of `copyto!` needs to fix for `MtlArray`.

I think there are a few things we need to establish in terms of conventions and semantics of `expose`:
1. We should define `unexpose(e::Exposed) = e.object` and prefer to use `unexpose`.
2. What should out-of-place functions that take `Exposed` objects as inputs return? I think they should output an unexposed object, i.e. `identity(e::Exposed) = unexpose(e)`, `permutedims(e::Exposed, perm) = permutedims(unexposed(e), perm)`, etc.
3. What should in-place functions that take `Exposed` objects return? I think they should return the modified unexposed object, i.e. `permutedims!(e::Exposed, perm) = (permutedims!(unexposed(e), perm); return unexposed(e))`.
4. Like I show above, we need to think carefully about when we `expose` or `unexpose` objects to not overuse it but try to anticipate potential issues that `expose` can fix.

NDTensors/ext/NDTensorsMetalExt/mul.jl
mtfishman: See my comment below about conventions of functions that accept `Exposed` objects, as well as other suggestions on semantics of `Exposed` objects.

mtfishman: Not sure why `mul!!` would be removed, that function is still useful for performing maybe-mutating matrix multiplication. So I don't really understand your comment but maybe my comment below will clarify some things.

kmp5VT: Your comment does make sense. I removed the `mul!!(CM, AM, BM, a, b)` function and replaced it with `Base.mul!(CM::Exposed, AM::Exposed, BM::Exposed, a, b)` and dispatch the correct `mul!` using the `Exposed` logic.  `mul!!` is not part of `LinearAlgebra`
```
julia> LinearAlgebra.mul!!
ERROR: UndefVarError: `mul!!` not defined
Stacktrace:
 [1] getproperty(x::Module, f::Symbol)
   @ Base ./Base.jl:31
 [2] top-level scope
   @ REPL[35]:1
   ```
but since it is  useful for the case you discuss I will add the function back into the `abstractarray` folder and define it as 
   ```
 function mul!!(CM::AbstractArray, AM::AbstractArray, BM::AbstractArray, a, b)
    mul!(expose(CM), expose(AM), expose(BM), a, b)
    return unexpose(CM)
```
Let me know if this is okay or if I should do more! Thanks for your comments!

mtfishman: Sure, it's fine to have our own versions of functions that don't exist in `Base`, the goal is just to overload `Base` functions when we can.

I could imagine as an alternative that we define:
```julia
function mul!!(CM::AbstractArray, AM::AbstractArray, BM::AbstractArray, a, b)
 return mul!!(expose(CM), expose(AM), expose(BM), a, b)
end
```
and then have a version of `mul!!` defined on `Exposed` which likely then calls `mul!`. Not sure what is best here or if there is an application for that, I think it depends on how we end up using `mul!!`. As far as I know the main application we have for `mul!!` is handling the logic for allocating `UnallocatedZeros` so not sure how `Exposed` plays a role there.

mtfishman: Maybe a strategy is to try to keep things similar to how we currently do things with `leaf_parenttype` dispatch.

NDTensors/ext/NDTensorsMetalExt/permutedims.jl
mtfishman: So probably it should be `Edest::Exposed{<:MtlArray,ReshapedArray{<:Any,<:Any,<:SubArray}}`, I don't think there was an issue with all `ReshapedArray`, just ones wrapping a `SubArray`.

NDTensors/ext/NDTensorsMetalExt/mul.jl
kmp5VT: Okay I understand! My thinking has been to make the `Exposed` system akin to the `leaf_parenttype` system and I'm just trying to be forward thinking about the extent of the system. In my work so far I am making it as light weight as possible. Thanks again for the direction.

Right now, I am not sure that we need the `mul!!(Exposed)` version right now. With the system I currently have implemented (I will push the changes in a minute) we should be able to define something like 
```
function mul!!(CM::UnallocatedZero, AM::AbstractArray, BM::AbstractArray, a, b)
  CM = allocate(CM)
  CM = mul!(expose(CM), expose(AM), expose(BM), a, b)
  return CM
```
which can cover the `UnallocatedZeros` logic in this case.

NDTensors/src/abstractarray/permutedims.jl
```diff
@@ -1,50 +1,12 @@
...
 
 function permutedims!!(B::AbstractArray, A::AbstractArray, perm)
-  return permutedims!!(leaf_parenttype(B), B, leaf_parenttype(A), A, perm)
-end
-
-function permutedims!!(
-  Bleaftype::Type{<:AbstractArray}, B, Aleaftype::Type{<:AbstractArray}, A, perm
-)
-  permutedims!(leaf_parenttype(B), B, leaf_parenttype(A), A, perm)
+  Base.permutedims!(expose(B), expose(A), perm)
```
mtfishman, line 5: Reminder to drop `Base.`.

NDTensors/src/abstractarray/tensoralgebra/contract.jl
```diff
@@ -97,12 +97,14 @@ function _contract_scalar_perm!(
       # Rᵃ .= β .* Rᵃ
       LinearAlgebra.scal!(length(Rᵃ), β, Rᵃ, 1)
     else
-      Rᵃ .= α .* permutedims(Tᵃ, perm) .+ β .* Rᵃ
+      Rᵃ .= α .* Base.permutedims(expose(Tᵃ), perm) .+ β .* Rᵃ
```
mtfishman, line 100: Reminder to drop `Base.`.

test/base/test_itensor_scalar_contract.jl
```diff
@@ -65,14 +65,14 @@ end
 
     R .= A .* B
     @test !any(isnan, R)
-    @test array(R) ≈ NDTensors.permutedims(array(A), (2, 1, 3)) * array(B)[]
+    @test array(R) ≈ Base.permutedims(array(A), (2, 1, 3)) * array(B)[]
```
mtfishman, line 68: Reminder to drop `Base.`.

NDTensors/src/Unwrap/src/functions/permutedims.jl
```diff
@@ -0,0 +1,14 @@
+function permutedims(E::Exposed, perm)
+  Aperm = permutedims(unexpose(E), perm)
+  return Aperm
```
mtfishman, line 3:
```suggestion
  return permutedims(unexpose(E), perm)
```

NDTensors/ext/NDTensorsMetalExt/imports.jl
```diff
@@ -1,5 +1,6 @@
 import NDTensors: mtl, set_ndims, set_eltype, set_eltype_if_unspecified
 import NDTensors.SetParameters: nparameters, get_parameter, set_parameter, default_parameter
+import NDTensors.Unwrap: Exposed
```
mtfishman, line 3: You should use `using` here instead of `import`, `import` should only be used if you are overloading a function, as opposed to just making it available in the namespace.

NDTensors/ext/NDTensorsMetalExt/linearalgebra.jl
```diff
@@ -1,15 +1,15 @@
-function NDTensors.qr(leaf_parenttype::Type{<:MtlArray}, A::AbstractMatrix)
-  Q, R = NDTensors.qr(NDTensors.cpu(A))
+function qr(A::Exposed{<:MtlArray})
+  Q, R = qr(expose(NDTensors.cpu(unexpose(A))))
```
mtfishman, line 2: We could define `NDTensors.cpu(e::Exposed) = cpu(unexpose(e))` so we don't have to do this unwrapping here.

NDTensors/ext/NDTensorsMetalExt/mul.jl
```diff
@@ -1,15 +1,12 @@
...
+function LinearAlgebra.mul!(
+  CM::Exposed{<:MtlArray,<:Transpose},
+  AM::Exposed{<:MtlArray},
+  BM::Exposed{<:MtlArray},
   α,
   β,
 )
-  mul!(transpose(CM), transpose(BM), transpose(AM), α, β)
-  return CM
+  mul!(transpose(unexpose(CM)), transpose(unexpose(BM)), transpose(unexpose(AM)), α, β)
```
mtfishman, line 10: Could define `transpose(::Exposed)` to make this code simpler.

NDTensors/src/Unwrap/src/expose.jl
```diff
@@ -0,0 +1,9 @@
+struct Exposed{Unwraped,Object}
+  object::Object
+end
+
+expose(object) = Exposed{unwrap_type(object),typeof(object)}(object)
+
+unexpose(E::Exposed) = E.object
+
+Base.parent(E::Exposed) = parent(unexpose(E))
```
mtfishman, line 9: I think this should be moved into a file in `Unwrap/src/functions/`.

NDTensors/src/Unwrap/src/functions/linearalgebra.jl
```diff
@@ -0,0 +1,11 @@
+function qr(E::Exposed)
+  return qr(unexpose(E))
+end
+
+function eigen(E::Exposed)
+  return eigen(expose(E))
```
mtfishman, line 6:
```suggestion
  return eigen(unexpose(E))
```

NDTensors/src/Unwrap/src/functions/linearalgebra.jl
```diff
@@ -0,0 +1,11 @@
...
+end
+
+function svd(E::Exposed)
+  return svd(expose(E))
```
mtfishman, line 10:
```suggestion
  return svd(unexpose(E))
```

NDTensors/src/Unwrap/src/functions/mul.jl
```diff
@@ -0,0 +1,4 @@
+function LinearAlgebra.mul!(CM::Exposed, AM::Exposed, BM::Exposed, α, β)
```
mtfishman, line 1:
```suggestion
function mul!(CM::Exposed, AM::Exposed, BM::Exposed, α, β)
```
since it looks like `mul!` is already imported.

NDTensors/src/arraytensor/array.jl
```diff
@@ -61,8 +61,6 @@ end
 function permutedims!(
   output_array::MatrixOrArrayStorage, array::MatrixOrArrayStorage, perm, f::Function
 )
-  output_array = permutedims!!(
-    leaf_parenttype(output_array), output_array, leaf_parenttype(array), array, perm, f
-  )
+  output_array = permutedims!!(output_array, array, perm, f)
```
mtfishman, line 64: Should this use `expose`?

NDTensors/src/blocksparse/blocksparse.jl
```diff
@@ -104,7 +104,7 @@ end
 function copyto!(D1::BlockSparse, D2::BlockSparse)
   blockoffsets(D1) ≠ blockoffsets(D1) &&
     error("Cannot copy between BlockSparse storages with different offsets")
-  copyto!(data(D1), data(D2))
+  Base.copyto!(expose(data(D1)), expose(data(D2)))
```
mtfishman, line 107: Is there a need for writing it as `Base.copyto!` instead of `copyto!`? I assume we are removing `NDTensors.copyto!`?

NDTensors/test/device_list.jl
```diff
@@ -23,9 +23,9 @@ function devices_list(test_args)
   end
 
   if "metal" in test_args || "all" in test_args
-    push!(devs, NDTensors.mtl)
+    push!(devs, mtl)
     Metal.allowscalar()
-    include(joinpath(pkgdir(NDTensors), "ext", "examples", "NDTensorMetal.jl"))
+    #include(joinpath(pkgdir(NDTensors), "ext", "examples", "NDTensorMetal.jl"))
```
mtfishman, line 28: Why is this dropped?

NDTensors/src/blocksparse/blocksparse.jl
kmp5VT: Thanks for pointing this out. Its the same as before, I implemented the `Base.copyto!` while still having the `NDTensors.copyto!` then removed the NDTensors version to see if/where things failed to replace them with base. But I forgot to go back and remove the `Base.` Ill go through and get rid of them. Same as the `LinearAlgebra.` in the `mul!` call above

NDTensors/src/Unwrap/src/functions/linearalgebra.jl
kmp5VT: Oh I am still working on decompositions (SVD, eigen QR, etc) so this was an accident that it was added here. Thanks though!

NDTensors/test/device_list.jl
kmp5VT: Sorry I commented it out while running tests on my machine, I'll add it back

NDTensors/src/arraytensor/array.jl
kmp5VT: No so `permutedims!!` as I have written will call `permutedims!(expose(output), expose(array), perm, f)` in case the type of `output` or `array` needs to be modified.

NDTensors/src/Unwrap/src/expose.jl
kmp5VT: Okay! Makes sense I can define some of the other small functions like this in the same file

NDTensors/src/arraytensor/array.jl
mtfishman: Thanks, I think I got confused and was looking at an old version of the code, makes sense.

NDTensors/ext/NDTensorsMetalExt/linearalgebra.jl
```diff
@@ -1,16 +1,30 @@
-function NDTensors.qr(leaf_parenttype::Type{<:MtlArray}, A::AbstractMatrix)
-  Q, R = NDTensors.qr(NDTensors.cpu(A))
-  return adapt(leaf_parenttype, Matrix(Q)), adapt(leaf_parenttype, R)
+function LinearAlgebra.qr(A::Exposed{<:MtlMatrix})
+  Q, R = LinearAlgebra.qr(expose(NDTensors.cpu(A)))
```
mtfishman, line 2: Maybe we should use `using LinearAlgebra` so you can just use `qr` instead of `LinearAlgebra.qr`.

NDTensors/src/Unwrap/src/functions/util.jl
```diff
@@ -0,0 +1,5 @@
+Base.parent(E::Exposed) = parent(unexpose(E))
```
mtfishman, line 1: Maybe we could call this file `abstractarray.jl` to indicate it is functionality focused on working with `AbstractArray` objects.

mtfishman:
I left a few more comments about some stylistic changes but beyond that it looks good. Definitely a nice simplification over the previous `leaf_parenttype` dispatch we were using, and nice to make use of more `Base` functions directly. I guess the final step is using this new system to fix the last remaining scalar indexing issues?

kmp5VT:
> I left a few more comments about some stylistic changes but beyond that it looks good. Definitely a nice simplification over the previous `leaf_parenttype` dispatch we were using, and nice to make use of more `Base` functions directly. I guess the final step is using this new system to fix the last remaining scalar indexing issues?

Thank you for the comments! Yes I am looking into the scalar indexing issues this morning!

NDTensors/ext/NDTensorsCUDAExt/indexing.jl
```diff
@@ -1,8 +1,18 @@
...
-  CUDA.@allowscalar data(T)[] = x
-  return T
+function setindex!(E::Exposed{<:CuArray}, x::Number)
+  CUDA.@allowscalar unexpose(E)[] = x
+  return unexpose(E)
+end
+
+function Base.getindex(E::Exposed{<:CuArray,<:Adjoint}, I...)
+  Ep = parent(E)
+  return Base.getindex(Ep, I...)
```
mtfishman, line 12:
```suggestion
  Ap = parent(E)
  return expose(Ap)[I...]
```

mtfishman: I think the way you wrote it, `Ep` would not be `Exposed` since `parent` would unexpose it.

NDTensors/ext/NDTensorsCUDAExt/indexing.jl
```diff
@@ -1,8 +1,18 @@
...
+end
+
+function Base.copy(E::Exposed{<:CuArray,<:Base.ReshapedArray})
+  Ep = parent(E)
+  return copy(Ep)
```
mtfishman, line 17:
```suggestion
  Ap = parent(E)
  return copy(expose(Ap))
```

NDTensors/src/Unwrap/src/functions/abstractarray.jl
```diff
@@ -0,0 +1,18 @@
+parent(E::Exposed) = parent(unexpose(E))
+
+transpose(E::Exposed) = transpose(unexpose(E))
+
+cpu(E::Exposed) = cpu(unexpose(E))
+
+getindex(E::Exposed) = getindex(unexpose(E))
```
mtfishman, line 7:
```suggestion
getindex(E::Exposed) = unexpose(E)[]
```

NDTensors/src/Unwrap/src/functions/abstractarray.jl
```diff
@@ -0,0 +1,18 @@
...
+
+function setindex!(E::Exposed, x::Number)
+  setindex!(unexpose(E), x)
```
mtfishman, line 10:
```suggestion
  unexpose(E)[] = x
```

NDTensors/src/Unwrap/src/functions/abstractarray.jl
```diff
@@ -0,0 +1,18 @@
...
+  return unexpose(E)
+end
+
+getindex(E::Exposed, I...) = getindex(unexpose(E), I...)
```
mtfishman, line 14:
```suggestion
getindex(E::Exposed, I...) = unexpose(E)[I...]
```

NDTensors/src/abstractarray/tensoralgebra/contract.jl
```diff
@@ -162,13 +168,21 @@ function _contract!(
...
-  CM = mul!!(CM, AM, BM, El(α), El(β))
+  ## If C^T = A^T * B^T => C = B * A
+  ## TODO There is an issue in CUDA.jl
+  ## When all are transpose CUDA.mul! isn't being
+  ## Called correctly in `NDTensorsCUDAExt`
+  if AM isa Transpose && BM isa Transpose && CM isa Transpose
+    CM = mul!!(parent(CM), parent(BM), parent(AM), El(α), El(β))
+  else
+    CM = mul!!(CM, AM, BM, El(α), El(β))
+  end
```
mtfishman, line 179: Seems like this should use `expose`.

NDTensors/src/tensorstorage/tensorstorage.jl
```diff
@@ -60,7 +60,9 @@ Base.real(S::TensorStorage) = setdata(S, real(data(S)))
 
 Base.imag(S::TensorStorage) = setdata(S, imag(data(S)))
 
-Base.copyto!(S1::TensorStorage, S2::TensorStorage) = (copyto!(data(S1), data(S2)); S1)
+function copyto!(S1::TensorStorage, S2::TensorStorage)
+  return (copyto!(expose(data(S1)), expose(data(S2))); S1)
```
mtfishman, line 64:
```suggestion
  copyto!(expose(data(S1)), expose(data(S2)))
  return S1
```

NDTensors/src/abstractarray/tensoralgebra/contract.jl
kmp5VT: This is the logic I made before with `permutedims!!` the `mul!!` calls `mul!(expose(CM), expose(AM), expose(BM), a, b)`

mtfishman: So maybe this should be handled with normal dispatch, i.e. define:
```julia
function mul!!(C::Transpose, A::Transpose, B::Transpose, α, β)
  CT = mul!!(transpose(C), transpose(B), transpose(A), α, β)
  return transpose(CT)
end
```
Note that we want to make sure it outputs the correctly transposed version. The point is more that this kind of logic should be handled outside of a high level function like `contract!`.

As a guiding principle, it is usually better to avoid code like `if A isa T ...` and instead use function dispatch.

kmp5VT: This makes a lot of sense, thanks! Also I am not sure that I need to return a transpose of CT because 
```
C(k,i)^T = A(j,i)^T B(k,j)^T => C(k,i) = B(k,j) A(j,i)
```

mtfishman:
@kmp5VT I fixed the merge conflicts.

This looks good to go from my end, is this ready from your end? Any more scalar indexing issues you wanted to address?

kmp5VT:
> Ah looks like some tests are failing, I introduced code into `main` that uses `leaf_parenttype` and `parenttype` which needs to get updated.

Yeah I found those areas and fixed them

kmp5VT:
@mtfishman Okay this PR should be good now. Thanks!

PR #1221, opened 2023-10-27
[ITensors] Optimize `directsum` again

mtfishman:
This should fix a performance issue in `directsum` that was raised in https://itensor.discourse.group/t/directsum-with-qn-seems-very-inefficient-compared-to-the-c-version/1106/8.

In `directsum`, we are constructing projectors to project the tensors being direct summed into the correct subspace. The previous code introduced in #1185 was constructing those in an inefficient way, by creating zero flux tensors by first filling in all blocks consistent with that flux and then setting appropriate elements to 1 to make the projectors. In this PR, in the block sparse case I'm making tensors that initially have no blocks, then those blocks are allocated as needed to make the projectors.

I hacked together some specialized constructors for making QN ITensors without any blocks for this purpose (or in the case where there aren't QNs, it makes a zero tensor). There is probably a simpler way to do that, but @kmp5VT is working on a number of improvements to the ITensor storage types, constructors, and operations on unallocated tensors so I think it is ok to leave it for now and replace that code with better constructors that will be introduced soon.

### DIFF
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -269,6 +269,14 @@ ITensor(x::RealOrComplex{Int}, is...) = ITensor(float(x), is...)
 # EmptyStorage ITensor constructors
 #
 
+# TODO: Replace with a simpler and more generic `zeros` constructor
+# when the new `UnallocatedZeros` type lands.
+# This is only used internally inside the implementation of `directsum`
+# right now.
+function zeros_itensor(elt::Type{<:Number}, inds::Index...)
+  return ITensor(elt, inds...)
+end
+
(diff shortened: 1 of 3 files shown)


mtfishman:
@LHerviou I'm not sure why but this change maybe fixed the issue we were discussing in https://github.com/ITensor/ITensorInfiniteMPS.jl/pull/77? At least the `test_iMPOConversions.jl` tests pass for me locally with this branch.

LHerviou:
From what you mention above, you are no longer calling one(EmptyNumber) because you initialize an empty tensor with type. So that probably solve the issue.

PR #1213, opened 2023-10-18
[NDTensors] `UnallocatedArrays` and `UnspecifiedTypes`

kmp5VT:
# Description

In this PR I add a FieldType folders where I implement a number of abstract, lazy and unspecified immutable fields which will later be implemented into NDTensors.

# Checklist:

- [x] Implement UnallocatedArrays and UnspecifiedTypes
- [x] Create a SetParameter system for the UnallocatedArrays types
- [x] Create rudimentary unittests for these new types

As a recap of our conversation 11/20. I will be working on the following
- [x]  Remove AbstractUnallocatedArray
- [x]  Try to make functions instance off `AbstractFill` and keep `UnallocatedZeros` a `AbstractZero`.
- [x]  Write the discussed constructor `UnallocatedX(f::X, alloc::Type{<:AbstractArray})` where X is fill or zero. And have `set_alloctype` call this function.
- [x] Create arithmetic that preserves the Unallocated type
  - [x] addition and subtraction preserve types
- [x]  create full unittest suite for these types

PR #1222, opened 2023-10-27
[NDTensors] Get more `Array` storage functionality working

mtfishman:
This gets more functionality for `Tensor` directly wrapping an `Array` (as opposed to `Dense`) working, such as tensor contraction, QR, eigen, and SVD.

**EDIT:** This is now working:
```julia
using NDTensors
using ITensors

function heisenberg(n)
  os = OpSum()
  for j in 1:(n - 1)
    os += "Sz", j, "Sz", j + 1
    os += 0.5, "S+", j, "S-", j + 1
    os += 0.5, "S-", j, "S+", j + 1
  end
  return os
end

function main(; n=10, nsweeps=5, conserve_qns=false)
  s = siteinds("S=1/2", n; conserve_qns)
  H = MPO(heisenberg(n), s)
  ψ = randomMPS(s, j -> isodd(j) ? "↑" : "↓"; linkdims=4)
  ψ1, e1 = dmrg(NDTensors.to_arraystorage.((H, ψ))...; nsweeps, maxdim=10, cutoff=1e-12)
  ψ2, e2 = dmrg(H, ψ; nsweeps, maxdim=10, cutoff=1e-12)
  return e1, e2
end

main()
```
where `NDTensors.to_arraystorage` replaces `Dense` storage with `Array`.

This outputs:
```julia
After sweep 1 energy=-4.256581185494193  maxlinkdim=10 maxerr=2.11E-06 time=0.020
After sweep 2 energy=-4.258028428488432  maxlinkdim=10 maxerr=1.25E-06 time=0.020
After sweep 3 energy=-4.258028444854628  maxlinkdim=10 maxerr=8.24E-07 time=0.030
After sweep 4 energy=-4.258028445008366  maxlinkdim=10 maxerr=8.23E-07 time=0.020
After sweep 5 energy=-4.258028445013943  maxlinkdim=10 maxerr=8.23E-07 time=0.021
After sweep 1 energy=-4.2565811854942  maxlinkdim=10 maxerr=2.11E-06 time=0.021
After sweep 2 energy=-4.258028428488428  maxlinkdim=10 maxerr=1.25E-06 time=0.029
After sweep 3 energy=-4.258028444854625  maxlinkdim=10 maxerr=8.24E-07 time=0.021
After sweep 4 energy=-4.258028445008369  maxlinkdim=10 maxerr=8.23E-07 time=0.021
After sweep 5 energy=-4.258028445013948  maxlinkdim=10 maxerr=8.23E-07 time=0.021
```
so DMRG is running just as before, but with the `Dense` type completely removed.


mtfishman:
@kmp5VT @emstoudenmire with this PR we would almost be able to remove the `Dense` type entirely. We might have to add a few more functions for the new `Array` storage type, but DMRG can be run entirely without the `Dense` type.

It may be worth just going ahead and doing that soon in a future PR since it will simplify the code base and make future work on removing `EmptyStorage`, `Diag`, and `BlockSparse` simpler. It would only be breaking for users who are reaching into internals (i.e. working at the storage level), and it would make reaching into internals much simpler since the type becomes simpler. We can think about how to coordinate with work like #1213 (which will lead to us being able to remove the `EmptyStorage` type) but I think they are mostly separate.

mtfishman:
[test ITensors mps]

mtfishman:
[test ITensors mps]

emstoudenmire:
Nice – this is advancing really fast, and I see you're already on to the diagonal case.

PR #1227, opened 2023-11-01
[NDTensors] More `BlockSparseArray` tensor operations

mtfishman:
This adds `permutedims` for `BlockSparseArray`.

I may define `contract` in this PR as well if I have the time.


mtfishman:
@kmp5VT @emstoudenmire I have some more block sparse operations working using `BlockSparseArray` storage, including ITensor addition and reshape using combiner contraction. `BlockSparseArray * BlockSparseArray` tensor contractions and factorizations (`qr`, `eigen`, `svd`) aren't working yet, I'll add those in a future PR.

PR #1229, opened 2023-11-01
[NDTensorsCUDAExt][Debug]  QR based SVD fails for some rectangular matrices

kmp5VT:
# Description

CUDA's QRAlgorithm fails when an `(N x M)` matrix has `N < M`. Here I fix this issue by catching the problem in the SVD dispatch and transposing the matrices and the resulting SVD.

# TODO
- [ ] Fix the QR SVD algorithm
- [ ] Make sure all the unit tests still pass

### DIFF
--- a/NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
+++ b/NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
@@ -1,13 +1,39 @@
-function NDTensors.svd_catch_error(A::CuMatrix; alg="JacobiAlgorithm")
+function NDTensors.svd_catch_error(A::CuMatrix; alg::String="JacobiAlgorithm")
   if alg == "JacobiAlgorithm"
     alg = CUDA.CUSOLVER.JacobiAlgorithm()
   else
     alg = CUDA.CUSOLVER.QRAlgorithm()
   end
+  return NDTensors.svd_catch_error(A, alg)
+end
+
+function NDTensors.svd_catch_error(A::CuMatrix, ::CUDA.CUSOLVER.JacobiAlgorithm)
   USV = try
-    svd(expose(A); alg=alg)
+    svd(A; alg=CUDA.CUSOLVER.JacobiAlgorithm())
   catch
     return nothing
+  end
+  return USV
+end
+
+function NDTensors.svd_catch_error(A::CuMatrix, ::CUDA.CUSOLVER.QRAlgorithm)
+  s = size(A)
+  if s[1] < s[2]
+    At = copy(Adjoint(A))
+    
+    USV = try 
+      svd(At; alg=CUDA.CUSOLVER.QRAlgorithm())
+    catch
+      return nothing
+    end
+    MV ,MS, MU = USV;
+    USV = SVD(copy(MU), MS, Adjoint(MV))
+  else
+    USV = try 
+      svd(A; alg=CUDA.CUSOLVER.QRAlgorithm())
+    catch
+      return nothing
+    end
   end
   return USV
 end


NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
```diff
@@ -1,13 +1,39 @@
-function NDTensors.svd_catch_error(A::CuMatrix; alg="JacobiAlgorithm")
+function NDTensors.svd_catch_error(A::CuMatrix; alg::String="JacobiAlgorithm")
```
mtfishman, line 1: We should rename this to `"jacobi_algorithm"`, following the convention that keyword arguments should be lowercase.

mtfishman: Also, we should explicitly check for an input of `"qr_algorithm"`, and then throw an error if the `alg` is something besides `"jacobi_algorithm"` and `"qr_algorithm"`.

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
```diff
@@ -1,13 +1,39 @@
...
+  if s[1] < s[2]
+    At = copy(Adjoint(A))
+
+    USV = try
+      svd(At; alg=CUDA.CUSOLVER.QRAlgorithm())
+    catch
+      return nothing
+    end
+    MV, MS, MU = USV
+    USV = SVD(copy(MU), MS, Adjoint(MV))
```
mtfishman, line 30: What's the reason for `copy(MU)` here?

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
kmp5VT: I changed it to lowercase. I was thinking since SVD in NDTensors has a way to [catch incorrect keywords](https://github.com/ITensor/ITensors.jl/blob/e1c493d0b0a28148f99ddad30627e74db7322926/NDTensors/src/linearalgebra/linearalgebra.jl#L108C1-L136C6) it wouldn't be necessary here but I will add it anyway

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
kmp5VT: MU here is `typeof(MU) = LinearAlgebra.Adjoint{Float64, CuArray{Float64, 2, CUDA.Mem.DeviceBuffer}}`
In order to form the correct U matrix I need to force MU to evaluate the adjoint so I am doing that using copy.

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
kmp5VT: Looks like throwing an error also causes an issue for tests when we are trying to call CPU svd backends with CU. Question for you (especially in the tests). Should I catch this problem in the code and just run the faster CUDA `qr_algorithm` or in say `NDTensors/tests` should I do something like 
```
if dev == NDTensors.cu
  svd(A; alg = "qr_algorith")
else 
  svd(A; alg="divide_and_conquer")
end
```

mtfishman:
I see, it looks like other code is sending `alg` keyword arguments into this CUDA SVD function which only make sense for the CPU SVD. Maybe it is an issue with picking good defaults for `alg` which are aware of the backend?

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
kmp5VT: @mtfishman I could also make `default_svd_alg(a, ::AbstractArray) = divide_and_conquer` and in `NDTensorsCUDA` make `default_svd_alg(a, ::CuArray) = qr_algorithm` and then just change the line `default_svd_alg(a, unwrap_type(a))`?

mtfishman: I think we should change `default_svd_alg`, since we should pick the best option so users don't have to pick it themselves. Could use `expose` here.

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
mtfishman: I see, so it is an issue with the `SVD` wrapper? Seems like a bad design for the `SVD` wrapper to not allow more general types.

mtfishman: Do we even need the `SVD` wrapper or can we just use something like:
```julia
SVD = (MU, MS, MV')
```
or
```julia
SVD = (; U=MU, S=MS, V=MV')
```

kmp5VT: Good question, I am not sure I will check. Thanks!

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
kmp5VT: Great I have a version coded and am running test sto verify that it works properly

NDTensors/ext/NDTensorsCUDAExt/default_kwargs.jl
```diff
@@ -0,0 +1 @@
+NDTensors.default_svd_alg(a, ::Type{<:CuArray}) = "qr_algorithm"
```
mtfishman, line 1: Let's put the type as the first argument, that's a pretty common convention.

NDTensors/src/arraystorage/arraystorage/tensor/svd.jl
```diff
@@ -51,7 +51,7 @@ function tsvd(
   # Only used by BlockSparse svd
   min_blockdim=nothing,
 )
-  alg = replace_nothing(alg, default_svd_alg(a))
+  alg = replace_nothing(alg, default_svd_alg(a, unwrap_type(a)))
```
mtfishman, line 54: If you define `default_svd_alg(a) = default_svd_alg(unwrap_type(a), a)` then you can just keep this line as it is.

mtfishman:
Looks good, thanks!

PR #1236, opened 2023-11-03
[NDTensorsCUDAExt] Fix bug in slicing Adjoint CuMatrix

kmp5VT:
# Description

CUDA was numerically failing for DMRG when the number of sites was above 8 this is related to a truncation bug found in SVD. This merge should fix this bug and give DMRG equivalent results on GPU and CPU.


mtfishman:
Thanks! Could you add a regression test in NDTensors? Also maybe a good time to add a test for DMRG using NDTensorsCUDAExt as we discussed.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,37 @@
...
+
+  # Create an initial random matrix product state
+  psi0 = dev(randomMPS(elt, sites;))
+
+  # Plan to do 5 DMRG sweeps:
+  nsweeps = 3
+  # Set maximum MPS bond dimensions for each sweep
+  maxdim = [10, 20, 100, 100, 200]
+  # Set maximum truncation error allowed when adapting bond dimensions
+  cutoff = [cut]
```
mtfishman, line 28:
```suggestion
  cutoff = [1e-3, 1e-13]
```
instead of looping over it below.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,37 @@
...
+  # Convert these terms to an MPO tensor network
+  H = dev(MPO(elt, os, sites))
+
+  # Create an initial random matrix product state
+  psi0 = dev(randomMPS(elt, sites;))
+
+  # Plan to do 5 DMRG sweeps:
+  nsweeps = 3
+  # Set maximum MPS bond dimensions for each sweep
+  maxdim = [10, 20, 100, 100, 200]
```
mtfishman, line 26:
```suggestion
  maxdim = 10
```
No need to test such large bond dimensions.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,37 @@
...
+    os += 0.5, "S-", j, "S+", j + 1
+  end
+
+  # Convert these terms to an MPO tensor network
+  H = dev(MPO(elt, os, sites))
+
+  # Create an initial random matrix product state
+  psi0 = dev(randomMPS(elt, sites;))
+
+  # Plan to do 5 DMRG sweeps:
```
mtfishman, line 23:
```suggestion
```
Let's just remove all of these comments.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,37 @@
...
+    os += "Sz", j, "Sz", j + 1
+    os += 0.5, "S+", j, "S-", j + 1
+    os += 0.5, "S-", j, "S+", j + 1
+  end
+
+  # Convert these terms to an MPO tensor network
+  H = dev(MPO(elt, os, sites))
+
+  # Create an initial random matrix product state
+  psi0 = dev(randomMPS(elt, sites;))
```
mtfishman, line 21:
```suggestion
  psi0 = dev(randomMPS(elt, sites; linkdims=4))
```
Good to test constructing an MPS with a nontrivial bond dimension.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,37 @@
+using Test
+using ITensors
+using NDTensors
+
+function test_dmrg(elt, N::Integer, dev::Function, cut::Float64, no::Float64)
+  # Create N spin-one degrees of freedom
+  sites = siteinds("S=1", N)
```
mtfishman, line 7:
```suggestion
  sites = siteinds("S=1/2", N)
```
Good to test against spin 1/2 just to make the tests a bit faster, since the local dimension will be smaller.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,37 @@
...
+  psi0 = dev(randomMPS(elt, sites;))
+
+  # Plan to do 5 DMRG sweeps:
+  nsweeps = 3
+  # Set maximum MPS bond dimensions for each sweep
+  maxdim = [10, 20, 100, 100, 200]
+  # Set maximum truncation error allowed when adapting bond dimensions
+  cutoff = [cut]
+  # Set the noise
+  noise = [no]
```
mtfishman, line 30:
```suggestion
  noise = [1e-12, 0]
```
Instead of looping over it. Also it is more common to set it finite at first and then to zero.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  # Convert these terms to an MPO tensor network
+  H = dev(MPO(elt, os, sites))
+
+  # Create an initial random matrix product state
+  psi0 = dev(randomMPS(elt, sites;))
+
+  # Plan to do 5 DMRG sweeps:
+  nsweeps = 3
```
mtfishman, line 24: May need to do more sweeps to get things converged to the reference energies, we will see.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
kmp5VT: Okay I made this change! Thanks!

NDTensors/test/ITensors/TestITensorDMRG/TestITensorDMRG.jl
```diff
@@ -0,0 +1,92 @@
...
+# cuda_8_cut = Dict([(1e-3, cuda_2_1e3_noise), (1e-13, cuda_2_1e13_noise)])
+
+cpu_sites = Dict([(2, cpu_2_cut), (8, cpu_8_cut)])
+cuda_sites = Dict([(2, cpu_2_cut), (8, cpu_8_cut)])
+
+ref = Dict([(NDTensors.cpu, cpu_sites), (NDTensors.cu, cpu_sites)])
+
+function get_ref_value(device, sites, cutoff, noise, elt)
+  return ref[device][sites][cutoff][noise][elt]
+end
```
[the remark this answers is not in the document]
kmp5VT: Okay this makes sense especially with your comments below. I wasn't sure how dramatically different the different values of `cutoff` and `noise` would make energy. And I realized later that `cuda` energy should be equivalent to `cpu`. I'm in the process of making this much simplier!
Yes I found that four a size of 8 or larger the code in SVD,  i.e. `MV[:, 1:dS]` starts to be used!
Ah makes sense to why the 2-site is such a specific number.
I will add the random seed.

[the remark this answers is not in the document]
mtfishman: In the first post you wrote:

> CUDA was numerically failing for DMRG when the number of sites was above 8 this is related to a truncation bug found in SVD.

so probably we should test systems sizes 4 and 10.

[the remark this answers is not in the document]
kmp5VT: @mtfishman sorry that is a typo, it should say number of sites was above 7. But it was more apparent as N grew larger than 8. So Its no issue to test 10 sites!

kmp5VT:
@mtfishman I am still working on the Unwrap tests. If you would like I can add them to this PR or we can merge this fix and I can open a PR after this one specifically to add those tests (that way we can have the gpu fixed asap)?

mtfishman:
> @mtfishman I am still working on the Unwrap tests. If you would like I can add them to this PR or we can merge this fix and I can open a PR after this one specifically to add those tests (that way we can have the gpu fixed asap)?

I think at the minimum we should add tests to this PR for the things that this PR is fixing.

NDTensors/test/ITensors/TestITensorDMRG/TestITensorDMRG.jl
```diff
@@ -0,0 +1,17 @@
...
+## eigen decomp with CUBLAS to give slightly different answer than BLAS)
+module TestITensorDMRG
+using ITensors
+using NDTensors
+
+reference_energies = Dict([
+  (4, -1.6160254037844384), (8, -3.374932598687889), (10, -4.258035207282885)
+])
+
+function get_ref_value(device, sites, elt)
```
mtfishman, line 12: Seems like this doesn't need `device` or `elt` anymore.

Even simpler would be to just directly access `reference_energies` and not have this function.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,28 @@
+using Test
+using ITensors
+using NDTensors
+using Random
+
+Random.seed!(1234)
```
mtfishman, line 6: I would move this to be right above the line:
```julia
psi0 = dev(randomMPS(elt, sites;))
```
so it is consistent every time you run `test_dmrg`.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,28 @@
...
+  H = dev(MPO(elt, os, sites; linkdims=4))
+
+  psi0 = dev(randomMPS(elt, sites;))
+
+  nsweeps = 7
+  cutoff = [1e-3, 1e-13]
+  noise = [1e-12, 0]
+
+  energy, psi = dmrg(H, psi0; nsweeps, cutoff, noise, outputlevel=0)
+  @test energy ≈ get_ref_value(dev, N, elt)
```
mtfishman, line 27: I think we should add an `rtol` here to make the test a bit more forgiving, it can be tricky comparing floating point numbers in tests. Something like `rtol=10 * eps(real(elt))`.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,28 @@
...
+
+  H = dev(MPO(elt, os, sites; linkdims=4))
+
+  psi0 = dev(randomMPS(elt, sites;))
+
+  nsweeps = 7
+  cutoff = [1e-3, 1e-13]
+  noise = [1e-12, 0]
+
+  energy, psi = dmrg(H, psi0; nsweeps, cutoff, noise, outputlevel=0)
```
mtfishman, line 26: Where did `maxdim` go? We should set that, otherwise the bond dimension can keep increasing uncontrollably.

NDTensors/test/ITensors/TestITensorDMRG/runtest.jl
```diff
@@ -0,0 +1,12 @@
+using Test
+using NDTensors
+include("TestITensorDMRG.jl")
+
+include(joinpath(pkgdir(NDTensors), "test", "device_list.jl"))
+devs = devices_list(ARGS)
+
+@testset "Testing DMRG different backends" begin
+  for dev in devs, N in [4, 10], elt in [Float32, ComplexF32, Float64, ComplexF64]
```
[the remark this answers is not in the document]
mtfishman: Also we should just directly use `devices_list(ARGS)` instead of defining `devs` to not introduce `devs` into the global scope.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
mtfishman: That will probably be particularly important to add if we set a lower `maxdim` and `nsweeps` since it will become less accurate. I think we just want to make sure the energy is accurate to some number of digits, we don't need to test very high precision.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,28 @@
...
+    os += "Sz", j, "Sz", j + 1
+    os += 0.5, "S+", j, "S-", j + 1
+    os += 0.5, "S-", j, "S+", j + 1
+  end
+
+  H = dev(MPO(elt, os, sites; linkdims=4))
+
+  psi0 = dev(randomMPS(elt, sites;))
+
+  nsweeps = 7
```
mtfishman, line 22: We should try to use fewer sweeps. See my comment about setting `rtol`.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,29 @@
...
+
+  nsweeps = 3
+  cutoff = [1e-3, 1e-13]
+  noise = [1e-12, 0]
+  ## running these with nsweeps = 100 and no maxdim
+  ## all problems do not have a maxlinkdim > 32
+  maxdim = 32
+
+  energy, psi = dmrg(H, psi0; nsweeps, cutoff, maxdim, noise, outputlevel=1)
+  @test energy ≈ reference_energies[N] rtol = 1e-12
```
mtfishman, line 28: We need to adjust this for the precision since this won't make sense for single precision. `eps(real(elt))` is a good pattern for that.

mtfishman: I usually use something like `10^3 * eps(real(elt))` and then adjust the prefactor until tests pass consistently.

NDTensors/src/Unwrap/test/runtest.jl
```diff
@@ -0,0 +1,45 @@
+using Test
+using NDTensors.Unwrap
+using LinearAlgebra
+
+## Still working on this
```
mtfishman, line 5: Could you comment once this is ready to review?

NDTensors/test/ITensors/TestITensorDMRG/TestITensorDMRG.jl
```diff
@@ -0,0 +1,14 @@
...
+## Failing for CUDA mostly with eigen (I believe there is some noise in
+## eigen decomp with CUBLAS to give slightly different answer than BLAS)
+module TestITensorDMRG
+using ITensors
+using NDTensors
+
+reference_energies = Dict([
+  (4, -1.6160254037844384), (8, -3.374932598687889), (10, -4.258035207282885)
+])
+test_tolerance = Dict(([Float32, 1e-5], [Float64, 1e-12]))
```
mtfishman, line 11: Looks like this can be removed?

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,29 @@
+using Test
+using ITensors
+using NDTensors
+using Random
```
mtfishman, line 4: Let's move these into `TestITensorDMRG.jl`, following the convention of how modules are generally organized with `using` statements at the top level of the module.

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,29 @@
...
+  os = OpSum()
+  for j in 1:(N - 1)
+    os += "Sz", j, "Sz", j + 1
+    os += 0.5, "S+", j, "S-", j + 1
+    os += 0.5, "S-", j, "S+", j + 1
+  end
+
+  Random.seed!(1234)
+  psi0 = dev(randomMPS(Float64, sites;))
+  H = dev(MPO(elt, os, sites; linkdims=4))
```
mtfishman, line 18:
```suggestion
  psi0 = dev(randomMPS(Float64, sites; linkdims=4))
  H = dev(MPO(elt, os, sites))
```
The link dimensions of the MPO are determined by the model (encoded in the `OpSum`) so it doesn't make sense to try to set them. In principle it should have errored since that isn't a supported keyword argument of `MPO`.

NDTensors/test/ITensors/TestITensorDMRG/TestITensorDMRG.jl
kmp5VT: I wasn't sure if we want to be stricter in the float32 case. the `eps(Float32) * 10^3` is only `0.001`?

mtfishman: Hmm yeah good point. I don't think we need to focus on super high precision tests right now since that is tricky anyway, I think we are just trying to catch if things go very wrong (like what this PR is fixing). Let's keep it simple for now (i.e. the current code) and adjust things as we go along.

mtfishman: Here is a proposal for a formula for `rtol`:
```julia
julia> default_rtol(elt) = 10^(0.75 * log10(eps(real(elt))))
default_rtol (generic function with 1 method)

julia> default_rtol(Float32)
6.415531155005385e-6

julia> default_rtol(Float64)
1.8189894035458617e-12
```
So basically adjust based on the number of digits of precision.

The goal here is to try to find a general formula so we don't have to hardcode for different number types.

In principle `cutoff` should also be adjusted based on the element type but that's a story for another time, since it is something we would want to look at systematically inside the `svd`/`eigen` functions and elsewhere in the library.

mtfishman: Before resolving a conversation could you explain in what way it will be resolved?

kmp5VT: Sure, I created your `default_rtol` function in the `TestITensorDMRG` file and used it

mtfishman: Thanks, it is helpful even if you just say you plan to fix something in your next commit. I use open comments to track what still needs to be done/what I need to check over before merging a PR which becomes difficult if you resolve a conversion without noting down how or if it will be fixed.

NDTensors/src/Unwrap/test/runtest.jl
kmp5VT: @mtfishman I added a number of checks that use both Vector and CuVector. Please let me know if I should add more or a specific kind of check. I have implemented checks for all functions introduced in `Unwrap` though I haven't made one for every kind of wrapping. Thanks!

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -0,0 +1,90 @@
...
+
+  q, r = Unwrap.qr_positive(expose(mp))
+  @test q * r ≈ mp
+
+  square = NDTensors.cu(rand(Float64, (10, 10)))
+  square = (square + square) ./ 2.0
+  ## CUDA only supports Hermitian or Symmetric eigen decompositions
+  ## So I symmetrize square and call symetric here
+  ## TODO I am not sure how to check this because square * U[:,1] != l[1] * U[:,1]
+  l, U = eigen(expose(Symmetric(square)))
```
mtfishman, line 64: Does it work to test:
```julia
@test Symmetric(square) * U ≈ U * Diagonal(l)
```
?

NDTensors/test/runtests.jl
```diff
@@ -34,6 +34,7 @@ end
     "emptystorage.jl",
     "combiner.jl",
     "arraytensor/arraytensor.jl",
+    "ITensors/runtest.jl",
```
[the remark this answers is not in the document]
mtfishman: Also should rename to `"ITensors/runtests.jl"` since top level test files are named `runtests.jl`, not `runtest.jl`.

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Ahh I realize I symmetrized square wrong so if I fix this the test `square * U \approx U * Diagonal(l)` works regardless of the `Symmetric` wrapper.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -0,0 +1,90 @@
...
+  Ea = expose(ma)
+
+  m_type = typeof(m)
+  @test typeof(E) == Exposed{m_type,m_type}
+  @test typeof(Et) == Exposed{m_type,LinearAlgebra.Transpose{e_type,m_type}}
+  @test typeof(Ea) == Exposed{m_type,LinearAlgebra.Adjoint{e_type,m_type}}
+
+  o = dev(Vector{Float32})(undef, 1)
+  expose(o)[] = 2
+  expose(o)[] == 2
```
mtfishman, line 42:
```suggestion
  @test expose(o)[] == 2
```

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -0,0 +1,90 @@
...
+
+  fill!(m, 0)
+  @test any(!Base.isinf, expose(m))
+
+  mp = copy(Ea)
+  mp == ma
```
mtfishman, line 48:
```suggestion
  @test mp == ma
```

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -0,0 +1,90 @@
...
+  fill!(ma, 2.0)
+  copyto!(expose(mp), expose(ma))
+  mp == ma
```
mtfishman, line 51:
```suggestion
  @test mp == ma
```

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -0,0 +1,90 @@
...
+  ## So I symmetrize square and call symetric here
+  ## TODO I am not sure how to check this because square * U[:,1] != l[1] * U[:,1]
+  l, U = eigen(expose(Symmetric(square)))
+
+  U, S, V, = svd(expose(mp))
+  @test U * Diagonal(S) * V' ≈ mp
+
+  cm = dev(fill!(Matrix{Float64}(undef, (2, 2)), 0.0))
+  mul!(expose(cm), expose(mp), expose(mp'), 1.0, 0.0)
+  cm ≈ mp * mp'
```
mtfishman, line 71:
```suggestion
  @test cm ≈ mp * mp'
```

NDTensors/src/Unwrap/test/runtest.jl
mtfishman: Besides the comments I've left, looks good to me, thanks.

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Added to the code thanks!

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Added thanks!

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Fixed, thanks!

NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -0,0 +1,24 @@
...
+
+  os = OpSum()
+  for j in 1:(N - 1)
+    os += "Sz", j, "Sz", j + 1
+    os += 0.5, "S+", j, "S-", j + 1
+    os += 0.5, "S-", j, "S+", j + 1
+  end
+
+  Random.seed!(1234)
+  psi0 = dev(randomMPS(Float64, sites;))
```
mtfishman, line 12:
```suggestion
  psi0 = dev(randomMPS(Float64, sites; linkdims=4))
```
Is there a reason you haven't made that change yet?

mtfishman:
Looks good, thanks for sticking with all of the comments!

PR #1238, opened 2023-11-04
[NDTensors] Add StridedViews dependency and fix using statement

emstoudenmire:
# Description

Before this change, NDTensors was failing to precompile. This was after making sure I had updated various dependencies and taken both NDTensors and ITensors out of 'dev' mode as checks that it wasn't for any of these reasons. 

It seems that StridedViews.jl is now its own separate package, so no longer inside of the latest version of Strided.jl (see [this comment in the What's New section of Strided.jl's README](https://github.com/Jutho/Strided.jl)). This PR adds StridedViews as a dependency and does `using StridedViews` directly instead of the previous code doing `using Strided.StridedViews`.


mtfishman:
That's strange, I don't understand why the current code would have been failing. For example this works:
```julia
julia> using Strided.StridedViews

julia> StridedView(@view randn(4, 4)'[1:2, 1:2])
2×2 StridedView{Float64, 2, Matrix{Float64}, typeof(identity)}:
 -0.501281    -0.0232372
  0.00611145   1.82511
```
i.e. it's fine to load a dependency of a package in that way, even if it is a separate package.

I'm not against this change since it's reasonable to make use of `StridedViews` directly but I have a feeling your precompilation issue is caused by something else.

emstoudenmire:
Helpful to know. I was interpreting the notation `Strided.StridedViews` differently in my head, as if `StridedViews` had to be somehow a module inside of Strided i.e. not a separate package (which it now is). But since that notation can work even if it refers to a separate package, it must have been an issue on my side.

Sure enough, my `Strided.jl` version was out of date.

If you prefer this change, please go ahead and merge or I can but if not we can close this PR.

mtfishman:
Could you add a compat entry of `0.2` into the Project.toml for `StridedViews`? Also we should drop support for older versions of `Strided` and just support `2`, allowing older versions of `Strided` is probably why that issue was allowed to happen on your end, we should only depend on more recent versions of `Strided` which had `StridedViews` separated out.

emstoudenmire:
Great. I modified the following lines of NDTensors Project.toml:
```
Strided = "2"
StridedViews = "0.2"
```
Please let me know if you think it'd be better to put, say, 2.0.4 for Strided or any other change like that.

PR #1244, opened 2023-11-09
[NDTensors] Some fixes for element type promotion and conversion

mtfishman:
This fixes a few issues with element type promotion and conversion, such as mixed element type contractions of block sparse tensors on GPU as well as better element type preservation when the noise term is enabled in DMRG.

I also extended the new DMRG tests on GPU introduced in #1236 to the case of QN conservation/block sparse tensors, which pass locally for me using a Metal backend.

- [x] Add a test for an eigendecomposition of a complex Hermitian matrix, testing that the eigenvalues are real.

### DIFF
--- a/NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
+++ b/NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
@@ -15,7 +15,7 @@ function test_dmrg(elt, N::Integer; dev::Function, conserve_qns)
 
   nsweeps = 3
   cutoff = [1e-3, 1e-13]
-  noise = real(elt)[1e-6, 0]
+  noise = [1e-6, 0]
   ## running these with nsweeps = 100 and no maxdim
   ## all problems do not have a maxlinkdim > 32
   maxdim = 32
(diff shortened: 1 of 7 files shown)


NDTensors/test/ITensors/TestITensorDMRG/dmrg.jl
```diff
@@ -9,12 +9,13 @@ function test_dmrg(elt, N::Integer, dev::Function)
...
   Random.seed!(1234)
-  psi0 = dev(randomMPS(Float64, sites; linkdims=4))
+  init = j -> isodd(j) ? "↑" : "↓"
+  psi0 = dev(randomMPS(elt, sites, init; linkdims=4))
   H = dev(MPO(elt, os, sites))
 
   nsweeps = 3
   cutoff = [1e-3, 1e-13]
-  noise = [1e-12, 0]
+  noise = real(elt)[1e-6, 0]
```
mtfishman, line 18:
```suggestion
  noise = [1e-6, 0]
```

mtfishman:
@kmp5VT so interestingly the new tests I added for block sparse DMRG pass on CPU and Metal (I checked locally) but seem to be failing with CUDA.

kmp5VT:
> @kmp5VT so interestingly the new tests I added for block sparse DMRG pass on CPU and Metal (I checked locally) but seem to be failing with CUDA.

That is interesting, I can look into what is going on later today!

mtfishman:
Thanks! I've marked the tests as broken for now so I can merge this fix.

mtfishman:
@kmp5VT I noticed some of the tests you wrote in #1236 used arrays with uninitialized memory (i.e. `Vector{Float64}(undef, 4)`) and then used those in algebra operations. In general that is not a good idea since then you can randomly get elements with `NaN` or `Inf` which don't play nicely when doing algebra. Better to allocate them to something random, like `randn(Float64, 4)`.

PR #1245, opened 2023-11-09
[NDTensorsCUDAExt] Remove and test for scalar indexing

kmp5VT:
# Description
@mtfishman I thought it was weird that there was a scalar indexing warning in this bug report https://itensor.discourse.group/t/cuda-issue-when-converting-mps-to-mpo/1274/2, so I looked into it. I believe Julia determines which `mul!` to use based on the `C` matrix. When `C` is a `Transpose` the generic matmul `mul!` implementation is called which causes a scalar indexing problem.  Its completely fine if `A` or `B` are a `Transpose`, the CUDA `mul!` kernel will still be called. So I created this exposed `mul!` function to fix the problem


mtfishman:
The issue there turned out to be that one of the tensors was actually on CPU (since the `delta` tensor got allocated there by default), if you look at the stack trace there is a `Matrix` instead of a `CuMatrix`. So are you certain this PR actually fixes a known issue?

kmp5VT:
> The issue there turned out to be that one of the tensors was actually on CPU (since the `delta` tensor got allocated there by default), if you look at the stack trace there is a `Matrix` instead of a `CuMatrix`. So are you certain this PR actually fixes a known issue?

@mtfishman , yes so the direct problem with the issue was related to the delta tensor but there was also a thrown warning about scalar indexing. When I ran the test with `CUDA.allowscalar(false)` it points me to scalar indexing in `mul!` with `A` and `C` being `Transposed` so I made this test 
```
using ITensors, NDTensors
i,j,k = Index.((2,2,2))
A = NDTensors.cu(randomTensor(j,i)); B = NDTensors.cu(randomTensor(j,k)); C = NDTensors.cu(randomTensor(k,i));
contract(C,(3,1), A, (-2,1), B, (-2,3))
```
I also tried with just `Transpose(C) = A * B` also `Transpose(C) = A * Transpose(B)` and all 3 scenarios fail (use scalar indexing) and it looks like its because of the `Transpose(C)` wrapper. But this change fixes all three

mtfishman:
I see, can you add tests with those cases? Ideally it would directly use `CuMatrix` and `mul!` and not test it through `contract!`.

NDTensors/ext/NDTensorsCUDAExt/mul.jl
```diff
@@ -0,0 +1,11 @@
+function LinearAlgebra.mul!(
+  C::Exposed{<:CuArray,<:LinearAlgebra.Transpose},
+  A::Exposed{<:CuArray},
+  B::Exposed{<:CuArray},
+  α,
+  β,
+)
+  return mul!(
+    expose(parent(C)), expose(transpose(unexpose(B))), expose(transpose(unexpose(A))), α, β
+  )
```
mtfishman, line 10:
```suggestion
  mul!(
    expose(parent(C)), expose(transpose(unexpose(B))), expose(transpose(unexpose(A))), α, β
  )
  return unexpose(C)
```
Like in other PRs, for in-place functions we should use the convention that we return the input value that is being modified (in this case the unexposed value since we always unexpose before returning). The way you wrote it, it would output the non-transposed version of `C`. Please try to be mindful about this in the future, I've corrected this a number of times in previous PRs.

kmp5VT:
> I see, can you add tests with those cases? Ideally it would directly use `CuMatrix` and `mul!` and not test it through `contract!`.

Yes I can add the tests, and it does show the same error with
```
A = fill!(CuMatrix{Float64, CUDA.Mem.DeviceBuffer}(undef, (2,3)), randn(Float64))
B = fill!(CuMatrix{Float64, CUDA.Mem.DeviceBuffer}(undef, (3,4)), randn(Float64))
C = fill!(CuMatrix{Float64, CUDA.Mem.DeviceBuffer}(undef, (4,2)), randn(Float64))
CUDA.allowscalar(false)
mul!(transpose(C), A, B, 1.0, 0.0)
```
And the code throws an error.

NDTensors/ext/NDTensorsCUDAExt/mul.jl
kmp5VT: Okay I will make sure to remember this going forward, thanks!

mtfishman:
Also Hermitian transposition:
```julia
julia> using Metal

julia> A, B, C = mtl.((randn(2, 2), randn(2, 2), randn(2, 2)))
(Float32[0.418296 -0.7982497; 2.1851687 0.10532443], Float32[0.028218202 0.96657676; -1.4152029 -0.11327304], Float32[1.0778706 0.3015457; -0.34520924 -0.62605405])

julia> using LinearAlgebra

julia> mul!(C', A, B)
┌ Warning: Performing scalar indexing on task Task (runnable) @0x000000010ad68010.
│ Invocation of getindex resulted in scalar indexing of a GPU array.
│ This is typically caused by calling an iterating implementation of a method.
│ Such implementations *do not* execute on the GPU, but very slowly on the CPU,
│ and therefore are only permitted from the REPL for prototyping purposes.
│ If you did intend to index this array, annotate the caller with @allowscalar.
└ @ GPUArraysCore ~/.julia/packages/GPUArraysCore/uOYfN/src/GPUArraysCore.jl:106
2×2 adjoint(::MtlMatrix{Float32, Metal.MTL.MTLResourceStorageModePrivate}) with eltype Float32:
  1.14149    0.494735
 -0.0873939  2.1002
```

mtfishman:
So please add a similar definition for `Adjoint`.

NDTensors/ext/NDTensorsCUDAExt/mul.jl
kmp5VT: Question regarding this, so in my test I do something like `C(3,1)^T = A(1,2) * B(2,3)`. My understanding here was that C is only being transposed in the contraction so that it can be efficiently added to the result of `A(1,2) * B(2,3)`, i.e. `C(1,3) = 1 .* (A * B) (1,3) + 0 .* C(1,3)` but we are actually interested in `C(3,1)`. So for example in my test below I expect C to have the dimensions `(4x2)` not `(2x4)`, is this correct? 
In the same direction of logic, here `parent(C)` is in the desired index order so I transpose the `(A * B)` portion of the equation such that 
```
(C(3,1))^T = (A * B)(1,3) ==> C(3,1) = (B^T * A^T)(3,1)
```
I just want to be sure I have this right, thanks for the help Matt!

mtfishman: I'm not really following, I think writing a test which I can review would be easier.

NDTensors/ext/NDTensorsCUDAExt/mul.jl
```diff
@@ -0,0 +1,27 @@
+# This was calling generic matrix multiplication.
+# TODO: Raise an issue with `CUDA.jl`.
+function mul!(
```
mtfishman, line 3:
```suggestion
function LinearAlgebra.mul!(
```
Let's use `LinearAlgebra.mul!` for overloading instead of `import LinearAlgebra: mul!`.

I also like that you started listed what we are using explicitly like:
```julia
using LinearAlgebra: LinearAlgebra, Adjoint, Transpose
```
so we should continue with that style of bringing types and functions into the namespace of these submodules/extensions, so then we would add `mul!` there to make use of it.

NDTensors/ext/NDTensorsCUDAExt/imports.jl
```diff
@@ -4,3 +4,4 @@ import NDTensors:
 import NDTensors.SetParameters: nparameters, get_parameter, set_parameter, default_parameter
 
 import .CUDA: CuArrayAdaptor
```
mtfishman, line 6: Is this needed?

NDTensors/ext/NDTensorsCUDAExt/mul.jl
```diff
@@ -0,0 +1,27 @@
...
+  CM::Exposed{<:CuArray,<:LinearAlgebra.Adjoint},
+  AM::Exposed{<:CuArray},
+  BM::Exposed{<:CuArray},
+  α,
+  β,
+)
+  return mul!(parent(CM), BM', AM', α, β)
```
mtfishman, line 24:
```suggestion
  mul!(parent(CM), BM', AM', α, β)
```

mtfishman: Also personally I would prefer:
```julia
return mul!(CM', BM', AM', α, β)
```
and for the `Transpose` version:
```julia
  mul!(transpose(CM), transpose(BM), transpose(AM), α, β)
```
I think it makes it clearer what the math is.

NDTensors/ext/NDTensorsCUDAExt/mul.jl
```diff
@@ -0,0 +1,27 @@
+# This was calling generic matrix multiplication.
+# TODO: Raise an issue with `CUDA.jl`.
+function mul!(
+  CM::Exposed{<:CuArray,<:LinearAlgebra.Transpose},
+  AM::Exposed{<:CuArray},
+  BM::Exposed{<:CuArray},
+  α,
+  β,
+)
+  return mul!(parent(CM), transpose(BM), transpose(AM), α, β)
```
mtfishman, line 10:
```suggestion
  mul!(transpose(CM), transpose(BM), transpose(AM), α, β)
```

NDTensors/ext/NDTensorsCUDAExt/mul.jl
```diff
@@ -0,0 +1,27 @@
...
+
```
mtfishman, line 11:
```suggestion
```

NDTensors/ext/NDTensorsCUDAExt/mul.jl
```diff
@@ -0,0 +1,27 @@
...
+# TODO: Raise an issue with `CUDA.jl`.
+function mul!(
+  CM::Exposed{<:CuArray,<:LinearAlgebra.Adjoint},
+  AM::Exposed{<:CuArray},
+  BM::Exposed{<:CuArray},
+  α,
+  β,
+)
+  return mul!(parent(CM), BM', AM', α, β)
+
```
mtfishman, line 25:
```suggestion
```

NDTensors/ext/NDTensorsMetalExt/mul.jl
```diff
@@ -10,3 +10,13 @@ function LinearAlgebra.mul!(
...
   return unexpose(CM)
 end
+
+# This was calling generic matrix multiplication.
+# TODO: Raise an issue with `Metal.jl`.
+function LinearAlgebra.mul!(
+  CM::Exposed{<:MtlArray,<:Adjoint}, AM::Exposed{<:MtlArray}, BM::Exposed{<:MtlArray}, α, β
+)
+  return mul!(parent(CM), BM', AM', α, β)
+
```
mtfishman, line 20:
```suggestion
  mul!(CM', BM', AM', α, β)
```

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -138,4 +138,33 @@ include("../../../test/device_list.jl")
...
+  #mul!(transpose(C), transpose(A), B, 1.0, 0.0)
+  mul!(C, transpose(B), A, 1.0, 0.0)
+  mul!(expose(transpose(Cp)), expose(transpose(A)), expose(B), 1.0, 0.0)
+  @test C ≈ Cp
+  Cp = fill!(similar(C), 0.0)
+  ## Try calling mul!! with transposes to verify that code works
+  NDTensors.mul!!(transpose(Cp), transpose(A), B, 1.0, 0.0)
+  @test C ≈ Cp
+
+  Cp = fill!(similar(C), 0.0)
```
mtfishman, line 161:
```suggestion
  Cp = fill!(similar(C), false)
```
to be more agnostic about the element type.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -138,4 +138,33 @@ include("../../../test/device_list.jl")
   x = dev(randn(elt, 4, 4))
   permutedims!(expose(y), expose(x), (2, 1))
   @test NDTensors.cpu(y) == transpose(NDTensors.cpu(x))
+
+  ##########################################
+  ### Testing an issue with CUDA&Metal transpose/adjoint mul
+  A = dev(randn(Float64, (3, 2)))
+  B = dev(randn(Float64, (3, 4)))
+  C = dev(randn(Float64, (4, 2)))
```
mtfishman, line 146:
```suggestion
  A = dev(randn(elt, (3, 2)))
  B = dev(randn(elt, (3, 4)))
  C = dev(randn(elt, (4, 2)))
```

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -138,4 +138,33 @@ include("../../../test/device_list.jl")
...
+  Cp = copy(C)
+  if (dev == NDTensors.cu)
+    CUDA.allowscalar(false)
+  end
```
mtfishman, line 150: This seems like an odd place to put this.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -138,4 +138,33 @@ include("../../../test/device_list.jl")
...
+  @test C ≈ Cp
+  Cp = fill!(similar(C), 0.0)
+  ## Try calling mul!! with transposes to verify that code works
+  NDTensors.mul!!(transpose(Cp), transpose(A), B, 1.0, 0.0)
+  @test C ≈ Cp
+
+  Cp = fill!(similar(C), 0.0)
+  ## This fails with scalar indexing 
+  #mul!(C', A', B, 1.0, 0)
+  mul!(C, B', A, 1.0, 0.0)
```
mtfishman, line 164:
```suggestion
  mul!(C, B', A, true, false)
```
Again, to be more agnostic about the element type (same changes below). `1.0` and `0.0` are `Float64`.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -138,4 +138,33 @@ include("../../../test/device_list.jl")
...
+  mul!(expose(Cp'), expose(A'), expose(B), 1.0, 0)
+  @test C ≈ Cp
+  Cp = fill!(similar(C), 0.0)
```
mtfishman, line 167: Should be able to use `Cp = zero(C)` as a shorthand for this.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -138,4 +138,33 @@ include("../../../test/device_list.jl")
...
+  Cp = copy(C)
+  if (dev == NDTensors.cu)
+    CUDA.allowscalar(false)
+  end
+  ## This fails with scalar indexing 
+  #mul!(transpose(C), transpose(A), B, 1.0, 0.0)
+  mul!(C, transpose(B), A, 1.0, 0.0)
+  mul!(expose(transpose(Cp)), expose(transpose(A)), expose(B), 1.0, 0.0)
+  @test C ≈ Cp
+  Cp = fill!(similar(C), 0.0)
```
mtfishman, line 156: Should be able to use `Cp = zero(C)` as a shorthand for this.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -138,4 +138,33 @@ include("../../../test/device_list.jl")
...
+
+  Cp = fill!(similar(C), 0.0)
+  ## This fails with scalar indexing 
+  #mul!(C', A', B, 1.0, 0)
+  mul!(C, B', A, 1.0, 0.0)
+  mul!(expose(Cp'), expose(A'), expose(B), 1.0, 0)
+  @test C ≈ Cp
+  Cp = fill!(similar(C), 0.0)
+  NDTensors.mul!!(Cp', A', B, 1.0, 0.0)
+  @test Cp ≈ C
```
mtfishman, line 169:
```suggestion
  Cpt = NDTensors.mul!!(Cp', A', B, 1.0, 0.0)
  @test transpose(Cpt) ≈ C
```
I think it is better to test the actual API which is that you should capture the output.

NDTensors/ext/NDTensorsCUDAExt/imports.jl
kmp5VT: No its not, they changed the name of this variable in `CUDA.jl` I commented it out in my `UnallocatedArray` PR but I can remove the variable in this one if you'd like

mtfishman: Sure, may as well remove it here.

NDTensors/ext/NDTensorsCUDAExt/mul.jl
kmp5VT: Makes sense to me, I made these changes

NDTensors/ext/NDTensorsCUDAExt/mul.jl
kmp5VT: Made these changes, thanks!

NDTensors/ext/NDTensorsCUDAExt/mul.jl
kmp5VT: Removed

NDTensors/ext/NDTensorsCUDAExt/mul.jl
kmp5VT: removed

NDTensors/ext/NDTensorsMetalExt/mul.jl
kmp5VT: made these changes to match (Sorry was using copy paste and missed the `return mul!` portion. All are fixed now

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: I modified all of the 0's and 1's to be true and false, thanks!

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Replaced Float64 with elt

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: In principle all of the tests will not use scalar but they are at the moment. I removed this from the PR

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: I made this fix

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Right, that makes sense, thanks for catching these things! I'm working on coding more carefully to be more in line with precedent.

kmp5VT: I made this fix

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: I was able to use zero and replaced the code here

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: same as above

NDTensors/src/Unwrap/test/runtests.jl
mtfishman: I see, thanks.

Would it be difficult to go ahead and set `CUDA.allowscalar(false)` globally for tests (i.e. in `device_list.jl`) and then use `@allowscalar` in places we specifically want to allow it or would that lead to too many things to have to fix? Seems like we should do that ASAP or else we will be dealing with a wack-a-mole of scalar indexing issues (i.e. accidentally introducing them and then having to fix them).

kmp5VT: I am not sure but since we are dealing with these situations fairly often I can look into it with this PR. I don't believe it will be difficult but I will let you know if I find otherwise!

NDTensors/ext/NDTensorsCUDAExt/imports.jl
kmp5VT: Okay I removed it!

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -138,4 +138,31 @@ include("../../../test/device_list.jl")
...
+  mul!(expose(transpose(Cp)), expose(transpose(A)), expose(B), true, false)
+  @test C ≈ Cp
+  Cp = zero(C)
+  ## Try calling mul!! with transposes to verify that code works
+  Cpt = NDTensors.mul!!(transpose(Cp), transpose(A), B, true, false)
+  @test transpose(Cpt) ≈ C
+
+  Cp = zero(C)
+  ## This fails with scalar indexing 
+  #mul!(C', A', B, true, false)
```
mtfishman, line 161: It's good to keep track of examples of failures.

If we disable scalar indexing in the tests then we can mark this as broken to keep track of it with:
```julia
@test_broken mul!(C', A', B, true, false)
```

NDTensors/ext/NDTensorsCUDAExt/mul.jl
kmp5VT: I saw you use this in `NDTensorsMetal` and I think that it makes sense, I also like explicitly writing the package function we are overloading in the extensions, I think its nice natural documentation.  I can continue doing this in the future. I have made the modifications to the `mul!` functions in NDTensorsCUDA

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: That makes sense in the case they are fixed by `Metal` or `CUDA` we can remove our case specific `Expose` based code. I made this change for both Adjoint and transpose

NDTensors/ext/NDTensorsCUDAExt/mul.jl
mtfishman: Sounds good, thanks. The importing style is mixed right now in `NDTensors` and `ITensors` because at some point I switched over entirely to using `import Base: *` but I think we should start to move away from that since it can be confusing when reading library code where a function is coming from, so maybe with new code we write we can start using `Base.*` for overloading.

NDTensors/src/Unwrap/test/runtests.jl
mtfishman: Yeah good point. Our overloads of `Exposed` and these tests will serve as a good record of what needs to be fixed upstream in the GPU packages, once the dust settles on our implementation we should definitely raise issues upstream for the cases we have found.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -138,4 +138,31 @@ include("../../../test/device_list.jl")
...
+  mul!(expose(transpose(Cp)), expose(transpose(A)), expose(B), true, false)
+  @test C ≈ Cp
+  Cp = zero(C)
+  ## Try calling mul!! with transposes to verify that code works
+  Cpt = NDTensors.mul!!(transpose(Cp), transpose(A), B, true, false)
+  @test transpose(Cpt) ≈ C
+
+  Cp = zero(C)
+  ## This fails with scalar indexing 
+  @test_broken mul!(C', A', B, true, false)
```
mtfishman, line 161: Shouldn't this only be broken when running on GPU?

NDTensors/ext/NDTensorsCUDAExt/copyto.jl
```diff
@@ -0,0 +1,6 @@
+# Catches a bug in `copyto!` in CuArray backend.
+function Base.copyto!(
+  dest::Exposed{<:CuArray}, src::Exposed{<:CuArray,<:Base.ReshapedArray}
+)
+  return copyto!(dest, expose(parent(src)))
```
mtfishman, line 5: Same thing, please return `unexpose(dest)`.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -40,8 +41,10 @@ include("../../../test/device_list.jl")
   @test typeof(Ea) == Exposed{m_type,LinearAlgebra.Adjoint{e_type,m_type}}
 
   o = dev(randn(elt, 1))
-  expose(o)[] = 2
-  @test expose(o)[] == 2
+  GPUArraysCore.@allowscalar begin
+    expose(o)[] = 2
+    @test expose(o)[] == 2
+  end
```
mtfishman, line 47: What is `@allowscalar` needed here? Shouldn't `expose` make that unnecessary?

mtfishman: Also we should be able to use `@allowscalar` instead of `GPUArraysCore.@allowscalar` since it is exported.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -125,17 +128,57 @@ include("../../../test/device_list.jl")
...
+  A = dev(randn(elt, (3, 2)))
+  B = dev(randn(elt, (3, 4)))
+  C = dev(randn(elt, (4, 2)))
+  Cp = copy(C)
+
+  ## This fails with scalar indexing
+  if dev != NDTensors.cpu
+    @test_broken mul!(transpose(C), transpose(A), B, true, false)
+  else
+    mul!(transpose(C), transpose(A), B, true, false)
```
mtfishman, line 161:
```suggestion
    @test mul!(transpose(C), transpose(A), B, true, false)
```

NDTensors/ext/NDTensorsCUDAExt/copyto.jl
kmp5VT: Okay I will change this in the NDTensorMetal code too

NDTensors/src/Unwrap/test/runtests.jl
mtfishman, line 161: This is a bit funny to have here, it's not really testing anything and also overwrites `C`. Maybe we can just remove this branch.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -125,17 +128,57 @@ include("../../../test/device_list.jl")
...
+  ## Try calling mul!! with transposes to verify that code works
+  Cpt = NDTensors.mul!!(transpose(Cp), transpose(A), B, true, false)
+  @test transpose(Cpt) ≈ C
+
+  Cp = zero(C)
+  ## This fails with scalar indexing 
+  if dev != NDTensors.cpu
+    @test_broken mul!(C', A', B, true, false)
+  else
+    mul!(C', A', B, true, false)
```
mtfishman, line 176: Same comment as above.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -2,6 +2,7 @@ using Test
 using NDTensors.Unwrap
 using NDTensors
 using LinearAlgebra
+using GPUArraysCore
```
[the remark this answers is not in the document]
kmp5VT: Oh I see I tried doing this but now realized I forgot the ampersand. I will change this across the board and remove the `GPUArraysCore` from inline

NDTensors/src/blocksparse/blocksparsetensor.jl
```diff
@@ -335,11 +335,15 @@ view(T::BlockSparseTensor, b::Block) = blockview(T, b)
 # convert to Dense
 function dense(T::TensorT) where {TensorT<:BlockSparseTensor}
   R = zeros(dense(TensorT), inds(T))
+  ## Here this failed with scalar indexing (R[blockindices] = blockview)
+  ## We can fix this by using copyto the arrays
+  r = Array(R)
```
mtfishman, line 340: Does this make a GPU array if `T` was on GPU? If so that seems like a bug since `Array` should only every output a CPU array and we should use something else. It's a bit of a hack but maybe just use `reshape(data(R), size(R))` for now.

mtfishman: Or maybe use `array(R)`?

NDTensors/src/blocksparse/linearalgebra.jl
```diff
@@ -200,7 +200,7 @@ function svd(
     if (sV * sVP) == -1
       Vb *= -1
     end
-    copyto!(expose(blockview(V, blockV)), expose(Vb))
+    copyto!((blockview(V, blockV)), (Vb))
```
mtfishman, line 203: Why was this change made?

Also there are extraneous parentheses in your new code.

kmp5VT: Right I will remove the parenthesis. So here `blockview` and `Vb` are both `Tensors` so they call the `expose` version of `copyto!` which unexposes and calls the `tensor` version of `copyto` which calls the `storage` version of `copyto!` which calls the `exposed` version of `copyto` on the `data` of the storage. I imagine we would want this `copyto!` call to immediately move into the `tensor` version of the function (which it now does). I can revert if that would be better?

NDTensors/src/dense/dense.jl
```diff
@@ -68,7 +68,12 @@ end
 
 function Dense(data::DataT) where {DataT<:AbstractArray{<:Any,N}} where {N}
   #println("Warning: Only vector based datatypes are currenlty supported by Dense. The data structure provided will be vectorized.")
-  return Dense(vec(data))
+  ## Vec has an issue where it makes data a `ReshapedArray` this can 
+  ## cause problems in the dispatch system when there are many wrapper types.
+  ## See test/diag.jl line 47
+  v = set_ndims(unwrap_type(data), 1)(undef, length(data))
+  copyto!(expose(v), expose(data))
+  return Dense(v)
```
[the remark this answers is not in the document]
kmp5VT: Ah I hadn't considered the no copy aspect. Let me try this, I believe that it will work. Thanks!

NDTensors/test/blocksparse.jl
```diff
@@ -1,6 +1,7 @@
 using NDTensors
 using LinearAlgebra
 using Test
+using GPUArraysCore
```
mtfishman, line 4:
```suggestion
using GPUArraysCore: @allowscalar
```

NDTensors/test/blocksparse.jl
```diff
@@ -45,26 +46,29 @@ using Test
     @test nnzblocks(A) == 2
     @test nnzblocks(B) == 2
 
-    A[1, 5] = 15
-    A[2, 5] = 25
-
-    @test A[1, 1] == 0
-    @test A[1, 5] == 15
-    @test A[2, 5] == 25
+    GPUArraysCore.@allowscalar begin
```
mtfishman, line 49: Same comment about 
```suggestion
    @allowscalar begin
```

NDTensors/test/dense.jl
```diff
@@ -25,83 +26,91 @@ using Test
...
-      @test A[1, 1] == 11
+      B = dev(Tensor(undef, (3, 4)))
+      randn!(B)
+      C = copy(A)
+      C = permutedims!!(C, B, (1, 2), .+)
 
-      Aview = A[2:3, 2:3]
+      Ap = permutedims(A, (2, 1))
+      A
+      Ap
```
mtfishman, line 43: ?

NDTensors/test/blocksparse.jl
kmp5VT: I don't think that we need these lines in the tests since `using GPUArraysCore: @allowscalar` is now (I just added it) in the `device_lists.jl` file and this file is included in all of the necessary tests. So I removed it from the header of all tests files except `runtests.jl` and `device_list.jl`

NDTensors/test/blocksparse.jl
kmp5VT: I have removed this from all files now they all read `@allowscalar`

NDTensors/test/runtests.jl
```diff
@@ -1,5 +1,6 @@
 using Test
 using SafeTestsets
+using GPUArraysCore
```
mtfishman, line 3: I think we don't need this at the top level and can just rely on importing it properly where it is needed.

NDTensors/test/dense.jl
kmp5VT: These were a relic of debugging, I have removed them thanks!

NDTensors/test/runtests.jl
kmp5VT: Makes sense, since its in the `device_lists.jl` file. I have removed it from the top level

NDTensors/ext/NDTensorsCUDAExt/copyto.jl
kmp5VT: I have updated the code!

NDTensors/src/blocksparse/linearalgebra.jl
mtfishman: The new code is fine as long as it works.

NDTensors/test/device_list.jl
```diff
@@ -12,7 +13,7 @@ function devices_list(test_args)
   end
 
   if "cuda" in test_args || "all" in test_args
-    CUDA.allowscalar()
+    #CUDA.allowscalar()
```
mtfishman, line 16: Reminder to delete this once we switch to using `@allowscalar`.

NDTensors/test/device_list.jl
```diff
@@ -1,3 +1,4 @@
+using GPUArraysCore
```
mtfishman, line 1:
```suggestion
using GPUArraysCore: @allowscalar

```

kmp5VT: I have added this to the code just haven't made the push to the branch yet. Thanks!

mtfishman: Though actually I prefer not having this here and instead putting it at the top of each file that uses `@allowscalar`.

I.e. in test files, put `using Package: f` in files where the file itself actually uses the functionality `f` from `Package`.

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: I was missing the `Base.` in `setindex` in `NDTensorsCUDAExt` I added it back in and this code works. Sorry I realized something weird was here but forgot that the `expose` version should work. Thanks!

NDTensors/src/Unwrap/test/runtests.jl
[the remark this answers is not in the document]
kmp5VT: I have updated all the code now to match your suggestion here and am resolving this comment

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Yeah I wasn't sure if we needed to "check" the cpu version but I deleted the else branch from this and the Adjoint version

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Same as transpose comment, thanks!

NDTensors/test/device_list.jl
kmp5VT: I have removed this line, thanks!

NDTensors/src/dense/dense.jl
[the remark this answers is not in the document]
kmp5VT: Both of these codeblocks seem to work fine with the problem case that I ran into in `test/blocksparse.jl` I also did a quick profile and they both use minimal allocations. I even tried changing the block sizes to be significantly larger and they still had a very small number of allocations (about 100 regardless of the size). I am going with the second block because it is more compact

NDTensors/src/blocksparse/blocksparsetensor.jl
kmp5VT: Array does make a GPU array if T has a GPU backend. Yes I do feel like this code is a hack. Really we need a way to create an unallocated array of the correct size and shape, then we need to loop through every block. If the block is zero we fill the view with 0's, if the block is non-zero we use `copyto!`. After thats finished we can package it into `Tensor` and `Dense`

mtfishman: I think it's not worth doing anything too sophisticated right now since I'm rewriting a lot of this anyway using the `BlockSparseArray` type, but I also just don't want to use incorrect code and I consider `r = Array(R)` to be incorrect (i.e. it is a bug that it outputs a GPU array).

So does it work to use `r = array(R)` here?

mtfishman: Though for the record I disagree with your suggestion, I think it is best to write a conversion from a block sparse array to a dense array by first allocating a zero array of the full size, then looping through just the nonzero blocks and slicing them over. If the block sparse array is very sparse that will be a lot less work than looping through every block (imagine the limit where the block sizes are really small and there are many blocks). That's the strategy here and will also be how the new code is designed. It is pretty fast to allocate all of the zeros up front.

mtfishman: (There may be some crossover where your suggestion is faster for a small number of large blocks since you're not allocating nonzero blocks twice, perhaps worth investigating but this operation is likely not a bottleneck anywhere anyway.)

NDTensors/test/device_list.jl
kmp5VT: Okay I will  change to have this at the top of all the files that use `@allowscalar` working on that now so I unresolved this conversation

NDTensors/src/blocksparse/blocksparsetensor.jl
kmp5VT: I agree, I don't think that this operation is the bottleneck for any of our algorithms at the moment but it definitely would be something interesting to look into!

kmp5VT: I completely understand your reasoning for seeing `r = Array(R)` as incorrect and agree. Also using `r = array(R)` does work so I will change the code to use `array`

NDTensors/src/dense/dense.jl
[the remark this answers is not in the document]
kmp5VT: It looks like there may be an issue with this code and can be found in the `test_itensor.jl` test. I am working on it now (I also left a comment in the code) I unresolved the conversation to remind

NDTensors/test/device_list.jl
kmp5VT: Okay this code is updated! Thanks

NDTensors/ext/NDTensorsCUDAExt/mul.jl
```diff
@@ -0,0 +1,45 @@
...
+  BM::Exposed{<:CuArray},
+  α,
+  β,
+)
+  mul!(CM', BM', AM', α, β)
+  return unexpose(CM)
+end
+
+## TODO I wasn't sure the best route to go here, if there is a better route than
+## copy please let me know!
```
mtfishman, line 28: Copying seems reasonable, maybe there is something more clever but with complicated wrapper types that may be the best we can do (without flattening the wrappers using `StridedView`, but as you know that will require more effort to get working for GPU backends). Is there a test for this case?

kmp5VT: @mtfishman Yes there is a test case for this its in the `diag.jl` test - these lines https://github.com/ITensor/ITensors.jl/blob/d9265a22773e4948c38cec28a39e1489bf89b90c/NDTensors/test/diag.jl#L41C1-L42C48

mtfishman: I think it would be good to add a test directly for `mul!` so it isn't just being tested indirectly through `contract`.

kmp5VT: Do you have a preference on where the tests should live? I.e. is it fine to add lines to `diag.jl` which derive a `mul!` test from the existing code? Something like 
```
S1M = dev(Matrix{elt}(undef, (2,12)))
SM = array(S)
VM = transpose(reshape(array(V), (12,2)))
NDTensors.mul!!(S1M, SM, VM, 1.0, 0.0)
@test @allowscalar reshape(S1M, (2,3,4)) ≈ array(S1)
```

NDTensors/src/dense/dense.jl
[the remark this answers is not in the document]
kmp5VT: Closing this since we have another conversation below

NDTensors/ext/NDTensorsCUDAExt/mul.jl
mtfishman: I think it is better to make a more minimal self-contained test in [Unwrap/test/runtests.jl](https://github.com/ITensor/ITensors.jl/pull/1245/files#diff-41118dc0c84433e17fb9c7eae2e5734daff5e01501d8d182050f108233600f6b) that is independent of `Tensor`/`TensorStorage` objects. For example, you could use something like `transpose(reshape(randn(2, 2)', 2, 2))` to make a wrapper with that type. That way we can better track what issues there are in different backends, and they won't be sensitive to changes we make to the design of the `Tensor` type.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -174,4 +174,13 @@ include("../../../test/device_list.jl")
   Cp = zero(C)
   Cpt = NDTensors.mul!!(Cp', A', B, true, false)
   @test Cpt' ≈ C
+
+  ##################################
+  ### Add test for transpose(reshape(adjoint )) failure in CUDA
+  A = dev(transpose(reshape(randn(elt, 2, 12)', (12, 2))))
+  B = dev(randn(elt, 2, 2))
+  C = dev(zeros(elt, (2, 12)))
```
mtfishman, line 182: Should use consistent parentheses style, i.e. `C = dev(zeros(elt, 2, 12))`.

NDTensors/test/combiner.jl
```diff
@@ -22,12 +23,12 @@ using ITensors: QN, Index
       @test output_tensor isa DenseTensor
       @test dims(output_tensor) == output_tensor_inds
       for i in 1:length(input_tensor)
-        @test input_tensor[i] == output_tensor[i]
+        @test @allowscalar input_tensor[i] == output_tensor[i]
```
mtfishman, line 26: Maybe better to annotate the for loop with `@allowscalar` instead, I notice you use that style in other places.

In general, it is good to try to choose a style or convention to follow for things like that and stick to it, to keep the code consistent and make it clearer what to do when adding new code.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -125,17 +126,61 @@ include("../../../test/device_list.jl")
...
+  @test Cpt' ≈ C
+
+  ##################################
+  ### Add test for transpose(reshape(adjoint )) failure in CUDA
+  A = dev(transpose(reshape(randn(elt, 2, 12)', (12, 2))))
+  B = dev(randn(elt, 2, 2))
+  C = dev(zeros(elt, (2, 12)))
+  NDTensors.mul!!(C, B, A, true, false)
+  Cp = B * copy(A)
+  @test @allowscalar C ≈ Cp
```
mtfishman, line 185: Looks like the `copy` is causing scalar indexing issues, probably best to transfer to CPU.

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Okay I modified this line, thanks!

NDTensors/test/combiner.jl
kmp5VT: I went through and looked at all the allowscalar to try and be consistent with this, thanks!

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Ahh, I only checked this on `Metal` last night. I changed the code to do the work on CPU but will check on my workstation this morning that it does properly run on CUDA

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -125,17 +126,64 @@ include("../../../test/device_list.jl")
...
+  ### Add test for transpose(reshape(adjoint )) failure in CUDA
+  ## TODO I have only tested this with Metal so far and will be running on 
+  ## workstation to verify that it works for CUDA
+  A = dev(transpose(reshape(randn(elt, 2, 12)', (12, 2))))
+  B = dev(randn(elt, 2, 2))
+  C = dev(zeros(elt, 2, 12))
+  NDTensors.mul!!(C, B, A, true, false)
+  Cp = NDTensors.cpu(similar(C))
+  NDTensors.mul!!(Cp, NDTensors.cpu(B), NDTensors.cpu(A), true, false)
+  @test NDTensors.cpu(C) ≈ Cp
```
mtfishman, line 188: I think it is better to test `mul!` directly, not `mul!!`.

Also a reminder to clean up the comments.

NDTensors/test/dense.jl
```diff
@@ -25,83 +26,89 @@ using Test
...
 
-      A[1, 1] = 11
+      Aview = A[2:3, 2:3]
+      @test dims(Aview) == (2, 2)
 
-      @test A[1, 1] == 11
+      B = dev(Tensor(undef, (3, 4)))
+      randn!(B)
+      C = copy(A)
+      C = permutedims!!(C, B, (1, 2), .+)
```
mtfishman, line 39: Why is this `.+` instead of `+`?

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Okay I changed this to test `mul!` instead and removed the comment after verifying this works on CUDA. Thanks!

kmp5VT: Should I replace the other calls to `mul!!` in this file with `mul!`?

test/base/test_itensor.jl
```diff
@@ -872,6 +872,9 @@ end
       3 4 0
       0 0 0
     ]
+    ## TODO the current definition of Dense is causing an issue here
+    ## Because the data is being copied somewhere and is not a pointer
+    ## To M. Still working here will get to it tomorrow
```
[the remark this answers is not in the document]
kmp5VT: Yes this can be removed now. Thanks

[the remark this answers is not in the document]
kmp5VT: I have removed it

NDTensors/test/dense.jl
kmp5VT: Oh let me modify some of these lines. I was trying to debug an issue that had to do with an order of operation in my `allowscalar` portion of the code. This bug has been fixed. I will change this to just + and I will also comment out the `C = A + B` in the `allowscalar` portion of the code. I also added the reason why the code fails.

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: The code has been modified and works on cuda

NDTensors/src/Unwrap/test/runtests.jl
kmp5VT: Okay after looking though again it seems like we are mostly testing the `mul!` functions and do a second verification using the `mul!!` function. I will add a `mul!!` to this new test too so we test both `mul!!` and `mul!`

NDTensors/Project.toml
```diff
@@ -23,6 +23,7 @@ Strided = "5e0ebb24-38b0-5f93-81fe-25c709ecae67"
 StridedViews = "4db3bf67-4bd7-4b4e-b153-31dc3fb37143"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 TupleTools = "9d95972d-f1c8-5527-a6e0-b4b365fa01f6"
+GPUArraysCore="46192b85-c4d5-4398-a991-12ede77f4527"
```
mtfishman, line 26: Could you add an entry to `[compat]` below (preserving the alphabetical order)? Looks like `0.1` is the latest version: https://github.com/JuliaGPU/GPUArrays.jl/blob/master/lib/GPUArraysCore/Project.toml.

NDTensors/ext/NDTensorsCUDAExt/copyto.jl
```diff
@@ -0,0 +1,28 @@
+## IT looks like CuArray suffers from the same issues as MtlArray.
+## To fix this subarray copyto problem I copied same code from MetalExt
+## This means we can probably write a generic implmenetation for GPUArrays
```
mtfishman, line 3:
```suggestion
# Same definition as `MtlArray`.
```

NDTensors/ext/NDTensorsCUDAExt/mul.jl
mtfishman, line 28:
```suggestion
```

NDTensors/Project.toml
kmp5VT: I alphabetized all instances and added a compat of `0.1` to this file. Thanks!

mtfishman:
Thanks @kmp5VT, this will be very useful for ensuring performance of GPU backends going forward. Is this ready to merge from your end once tests pass?

kmp5VT:
> Thanks @kmp5VT, this will be very useful for ensuring performance of GPU backends going forward. Is this ready to merge from your end once tests pass?

@mtfishman not a problem. Yes this is ready to merge, I have no additional changes this round. Thanks!

PR #1247, opened 2023-11-10
[NDTensors] `BlockSparseArray` contract, QR, and Hermitian eigendecomposition

mtfishman:
This defines some more tensor and matrix algebra for `BlockSparseArray`, such as contraction, QR, and Hermitian eigendecomposition.


emstoudenmire:
I see what you mean about the decompositions. The QR code here is very compact – much of it by line number seems to mainly be the various small helper functions, then the core code actually doing the algorithm is quite short.

PR #1249, opened 2023-11-15
Add check for zero norm to expect

emstoudenmire:
# Description

Currently if you pass an MPS to `expect` whose norm is zero, the code will compute the norm and later divide by it, siletly producing NaN's in the output. A [user recently ran into this issue](https://itensor.discourse.group/t/applying-down-projection-operator-to-up-spin-site-gives-psi-with-sz-nan-instead-of-0/1284). This PR adds a simple error check of the norm.

```julia
using ITensors

function ITensors.op!(Op::ITensor,
                      ::OpName"Pdn",
                      ::SiteType"S=1/2",
                      s::Index)
  Op[s'=>2,s=>2] = 1.0
end

let
  N = 10
  s = siteinds("S=1/2",N)

  psi0 = MPS(s,"Up")
  psi1 = apply(op("Pdn",s[5]),psi0)

  @show expect(psi1,"Sz")

  return
end
```
</p></details>

### DIFF
--- a/src/mps/mps.jl
+++ b/src/mps/mps.jl
@@ -989,6 +989,7 @@ function expect(psi::MPS, ops; sites=1:length(psi), site_range=nothing)
 
   orthogonalize!(psi, start_site)
   norm2_psi = norm(psi)^2
+  iszero(norm2_psi) && error("MPS has zero norm in function `expect`")
 
   ex = map((o, el_t) -> zeros(el_t, Ns), ops, el_types)
   for (entry, j) in enumerate(site_range)


mtfishman:
Makes sense, thanks! Could you add a test?

test/ITensorLegacyMPS/base/test_mps.jl
```diff
@@ -856,6 +856,13 @@ end
     @test res[2, 1] ≈ eSx[3:7]
     @test res[1, 2] ≈ eSx[3:7]
     @test res[2, 2] ≈ eSz[3:7]
+
+    # Test that passing zero-norm MPS leads to an error
+    # (expect not well-defined in that case)
+    psi0 = copy(psi)
+    psi0[1] *= 0.0
```
mtfishman, line 863:
```suggestion
    psi0[1] *= false
```
Generally we are trying to be more agnostic about data types when possible, `false` is the lowest zero type so promotes up to any element type `psi0` would have. It wouldn't change this test but I think it's a good habit to get into.

emstoudenmire: Just made that change. I think it makes sense for library code to be able to be generic, but I'd feel nervous to do this too much in code that users might read, because if I didn't know the reason from the above, I would have not understood that `false` was behaving as a number type here.

mtfishman: Sounds good. Also there are probably better patterns for this kind of thing like `zero(scalartype(psi0))`. In general I agree we should think carefully about the readability of generic code.

PR #1264, opened 2023-11-16
[NDTensorsMetalExt] Fix more scalar indexing issues on Metal

kmp5VT:
# Description

Trying to get `Metal` up to speed as CUDA is and have it be testable in the NDTensors/test suite.

Fixes #(issue)

Remove scalar indexing but also running into issues with `Float64` typing in NDTensors.

- [x] Changing the file names in NDTensors/test to test_filename.jl and writing the loop to go over test_*.jl
- [x] In each test file I am going to add a @safetestset to the beginning of the file and put the using statements in the safetestset section.
- [x] I'll explicitly write which functions are included from each of the modules in all of the test files.
- [x] In files where metal breaks because of eltype issues I will migrate Tensor(...) constructors to Tensor{elt}(...) and loop over eltypes.


NDTensors/test/blocksparse.jl
```diff
@@ -9,7 +9,12 @@ using GPUArraysCore: @allowscalar
   devs = devices_list(copy(ARGS))
 
   @testset "test device: $dev" for dev in devs
-    elt = (dev == NDTensors.mtl ? Float32 : Float64)
+    if dev == NDTensors.mtl
+      elt = Float32
+      NDTensors.default_eltype() = Float32
```
mtfishman, line 14: This is not a good idea.

NDTensors/test/blocksparse.jl
```diff
@@ -258,7 +270,7 @@ using GPUArraysCore: @allowscalar
       randn!(A)
       U, S, V = svd(A)
       @test @allowscalar isapprox(
-        norm(array(U) * array(S) * array(V)' - array(A)), 0.0; atol=1e-13
+        norm(array(U) * array(S) * array(V)' - array(A)), 0.0; atol=atol
```
mtfishman, line 273: I think it is nicer to do:
```julia
      @test @allowscalar array(U) * array(S) * array(V)' ≈ array(A)) atol=atol
```

NDTensors/test/blocksparse.jl
mtfishman: A pattern I used for this can be seen here:
https://github.com/ITensor/ITensors.jl/blob/408516dce2e0182e04a6fb27b501b04a31b630fc/NDTensors/test/ITensors/TestITensorDMRG/runtests.jl

NDTensors/test/blocksparse.jl
```diff
@@ -226,12 +234,16 @@ using GPUArraysCore: @allowscalar
   end
 
   @testset "svd on $dev" for dev in devs
+    if dev == NDTensors.mtl
+      NDTensors.default_eltype() = Float32
+    end
+    atol = 10^(0.75 * log10(eps(real(elt))))
```
mtfishman, line 240: It seems like there is a lot of logic in:
https://github.com/ITensor/ITensors.jl/blob/408516dce2e0182e04a6fb27b501b04a31b630fc/NDTensors/test/ITensors/TestITensorDMRG/TestITensorDMRG.jl
that is helpful in other tests. Probably best to make a module `TestNDTensorsUtils` in `NDTensors/test/TestNDTensorsUtils/TestNDTensorsUtils.jl` that defines some general helper functions for tests.

That could be merged with `NDTensors/test/devices_list.jl`, probably that should go in `TestNDTensorsUtils` as well.

NDTensors/test/blocksparse.jl
kmp5VT: This makes sense should I change this across the board?

NDTensors/test/blocksparse.jl
kmp5VT: So I was going to do something like this but one problem I was running into is that constructors which do not specify the eltype like `Tensor(undef, (2,4))`, make a Float64 tensor and then when I call `NDTensors.mtl` the function to convert a Float64 tensor which fails. My thought was that, since we have our constructors initialize Tensors' eltype to `NDTensors.default_eltype()` we could modify the default eltype for the time being. 
To keep the future in mind, I think this problem will be mitigated when we convert the `default_eltype` to `UnspecifiedType`

mtfishman: It seems like a better solution is to call more general constructors `Tensor{elt}(undef, (2, 4))`, `randomTensor(elt, (2, 4))`, etc. , and if those don't exist then they shouldn't be hard to define. I think most constructors support passing the element type in some way at this point.

Otherwise we can use `adapt(elt, a)` after the tensor is created already.

I really don't think we should be controlling this in a global way, even in tests. That could cause some very confusing behavior for people, for example if they run the tests and then they run some other code and suddenly their element types are all changed.

NDTensors/test/blocksparse.jl
mtfishman: Sure, thanks!

NDTensors/test/blocksparse.jl
kmp5VT: I had written this up previously using adapt but I was worried about the same problem (i.e. changing types without users knowing whats going on or why we are using adapt) so I think using `Tensor{elt}` makes the most sense and is the most transparent for users. I can start making the changes. Thanks for the input!

kmp5VT:
@mtfishman To recap our conversation and as a refresher for me. The notes for changes are as follows
1. Changing the file names in `NDTensors/test` to `test_filename.jl` and writing the loop to go over `test_*.jl` 
2. In each test file I am going to add a `@safetestset` to the beginning of the file and put the `using` statements in the `safetestset` section.
3. I'll explicitly write which functions are included from each of the modules in all of the test files.
4. In files where metal breaks because of eltype issues I will migrate `Tensor(...)` constructors to `Tensor{elt}(...)` and loop over eltypes.

NDTensors/test/test_combiner.jl
```diff
@@ -1,24 +1,27 @@
...
+using .NDTensorsTestUtils: NDTensorsTestUtils
 
 # Testing generic block indices
 using ITensors: QN, Index
 
 @testset "CombinerTensor basic functionality" begin
-  devs = devices_list(copy(ARGS))
-  @testset "test device: $dev" for dev in devs
+  @testset "test device: $dev" for dev in NDTensorsTestUtils.devices_list(copy(ARGS)),
+    elt in [Float64, Float32]
```
mtfishman, line 13:
```suggestion
    elt in (Float32, Float64)
```
When putting objects with different types in a collection it is better to use a `Tuple`.

NDTensors/src/blocksparse/blocksparsetensor.jl
```diff
@@ -65,15 +65,15 @@ function BlockSparseTensor(eltype::Type{<:Number}, blockoffsets::BlockOffsets, i
...
+  return BlockSparseTensor(NDTensors.default_eltype(), blockoffsets, inds)
 end
 
 """
     BlockSparseTensor(inds)
 
 Construct a block sparse tensor with no blocks.
 """
-BlockSparseTensor(inds) = BlockSparseTensor(Float64, inds)
+BlockSparseTensor(inds) = BlockSparseTensor(FloNDTensors.default_eltype()at64, inds)
```
mtfishman, line 76: Typo.

NDTensors/src/Unwrap/test/runtests.jl
```diff
@@ -1,11 +1,13 @@
-using Test
+using Test: @testset, @test, @test_broken
 using NDTensors.Unwrap
-using NDTensors
-using LinearAlgebra
+using NDTensors: NDTensors, mul!!
+using LinearAlgebra:
+  LinearAlgebra, Transpose, qr, Symmetric, eigen, Hermitian, Diagonal, svd, mul!
```
mtfishman, line 5: Could you alphabetize?

In general I have used the convention that I first list types in alphabetical order, and then I list functions in alphabetical order.

kmp5VT: Yes, I'll use that convention as well. Thank you!

NDTensors/ext/NDTensorsMetalExt/mul.jl
```diff
@@ -19,3 +19,21 @@ function LinearAlgebra.mul!(
...
+  BM::Exposed{
+    <:MtlArray,
+    <:LinearAlgebra.Transpose{
+      <:Any,<:Base.ReshapedArray{<:Any,<:Any,<:LinearAlgebra.Adjoint}
+    },
+  },
+  α,
+  β,
+)
+  mul!(CM, AM, expose(transpose(copy(expose(parent(BM))))), α, β)
```
kmp5VT, line 37: @mtfishman would this be better split into two lines
```
B = copy(expose(parent(BM))
mul!(CM, AM, expose(transpose(B)), \alpha, \beta)
```

mtfishman: Yeah, I think that looks better.

NDTensors/src/tensorstorage/tensorstorage.jl
```diff
@@ -28,6 +28,7 @@ Base.@propagate_inbounds function Base.setindex!(S::TensorStorage, v, i::Integer
   return (setindex!(data(S), v, i); S)
 end
 
+## Missing a check or conversion when calling number * Tensor. This causes Metal to fail numerically because it tries to convert it to Float64. Preserve S eltype. ## TODO this could probably be handled differently/better?
```
kmp5VT, line 31: This line can fail on Metal GPU when x is a Float64 because I believe there is an internal type promotion. I am not currently on my mac and will send code when I am. Could this type of operation cause the code to fail when, for example, we are using multi-float?

NDTensors/ext/NDTensorsMetalExt/mul.jl
kmp5VT: Okay I made the change, thanks!

NDTensors/src/blocksparse/blocksparsetensor.jl
kmp5VT: Fixed this, thanks!

NDTensors/src/tensorstorage/tensorstorage.jl
mtfishman: That's just how Julia's type promotion works:
```julia
julia> one(Float64) * rand(Float32, 2)
2-element Vector{Float64}:
 0.9090054631233215
 0.7104539275169373
```
So whatever issue you are seeing, the primary problem is that a `Float64` is being sent down to this code, this code looks ok to me.

kmp5VT: @mtfishman a quick example on CUDA shows this 
```
julia> v = CuVector{Float32}(undef, 10)
10-element CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}:

julia> v * 2.0
10-element CuArray{Float64, 1, CUDA.Mem.DeviceBuffer}:
```

kmp5VT: Right that all makes sense, thanks!
I guess the issue I am finding is on line 77 in `NDTensors/test/dense.jl` 
```
data(A * 2.0) == data(2.0 * A)
```
since metal uses a `Float32`, this line tries to promote the type up to  `Float64` but for the most part in ITensors/NDTensors we aren't multiplying a our tensors by random number without considering the eltype of `A` and the scalar.  So in this test I will wrap the number in `elt(2.0)` to ensure that it preserves eltype.

NDTensors/test/test_combiner.jl
kmp5VT: Thanks, I updated this across all the tests files in NDTensors

NDTensors/test/Project.toml
```diff
@@ -16,6 +16,4 @@ TBLIS = "48530278-0828-4a49-9772-0f3830dfa1e9"
 TensorOperations = "6aa20fa7-93e2-5fca-9bc0-fbd0db3c71a2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
-
-[extras]
```
mtfishman, line 20: Was this meant to be removed?

kmp5VT: I made this change to ask you if we should  allow users to run the command `Pkg.test("NDTensors"; test_args=["metal"])".
When `Metal` is under the `[extras]` and you run this test command the code fails with 
```
NDTensors: Error During Test at /Users/kpierce/.julia/packages/SafeTestsets/raUNr/src/SafeTestsets.jl:30
  LoadError: ArgumentError: Package Metal not found in current path.
  - Run `import Pkg; Pkg.add("Metal")` to install the Metal package.
  ```
  However, it looks like there is an issue in `compat` so I will remove it.

Comment (mtfishman) [reply NDTensors/test/Project.toml:20]: Yeah, I guess ideally `Pkg.test("NDTensors"; test_args=["metal"])` would work (that seems like the only way to run the tests on Metal right now anyway?). If this is needed to allow that, I think it could make sense to do this, as long as it doesn't cause other issues like Metal not installing properly on certain systems.

File: NDTensors/test/arraytensor/runtests.jl
Comment (kmp5VT) [inline NDTensors/test/arraytensor/runtests.jl:1]: I renamed this to runtests to follow the other convention. Let me know if I should change it back to arraytensor.jl

File: NDTensors/test/readwrite.jl
Comment (kmp5VT) [inline NDTensors/test/readwrite.jl:1]: I am not sure what to do with this file. It doesn't appear to work and was not included in the NDTensor test suite prior to my changes. Also commenting to delete todo when resolved

File: NDTensors/test/test_blocksparse.jl [hunk 10]
```diff
@@ -1,22 +1,29 @@
...
-  devs = devices_list(copy(ARGS))
 
-  @testset "test device: $dev" for dev in devs
-    elt = (dev == NDTensors.mtl ? Float32 : Float64)
+  @testset "test device: $dev, eltype: $elt" for dev in NDTensorsTestUtils.devices_list(
+      copy(ARGS)
+    ),
+    elt in (Float32, Float64)
+
+    if dev == NDTensors.mtl && elt == Float64
```
kmp5VT, line 17: TODO leaving a comment here to use the, is_supported_eltype function here and in other locations in testing

NDTensors/ext/NDTensorsMetalExt/permutedims.jl
```diff
@@ -1,7 +1,26 @@
...
+## Theres an issue in metal that `ReshapedArray' wrapped arrays cannot be permuted
+## To get around this copy and permute Esrc, reshape to the size of Edest's parent
+## and broadcast into the parent.
+function Base.permutedims!(
+  Edest::Exposed{<:MtlArray,<:Base.ReshapedArray},
+  Esrc::Exposed{<:MtlArray,<:Base.ReshapedArray},
+  perm,
+  f,
+)
+  Aperm = unwrap_type(Esrc)(reshape(permutedims(Esrc, perm), size(parent(Edest))))
```
mtfishman, line 23: Why is `unwrap_type(Esrc)` needed here?

NDTensors/ext/examples/NDTensorCUDA.jl
```diff
@@ -1,10 +1,10 @@
-using CUDA
 using NDTensors
-
-using ITensors
-using Test
-
-using Zygote
+using CUDA: cu, CUDA, CuVector
```
mtfishman, line 2:
```suggestion
using CUDA: CUDA, CuVector, @allowscalar, cu
```
In the examples we it's better to use `@allowscalar` from `CUDA` instead of from `GPUArraysCore`, that's what most users would be doing. Internally it makes sense to try to use `GPUArraysCore` since it is a lighter weight dependency, but for end users they will be using packages like `CUDA`, `Metal`, etc. that ship with `@allowscalar`, so there isn't a need to load `GPUArraysCore` separately.

NDTensors/ext/examples/NDTensorCUDA.jl
```diff
@@ -1,10 +1,10 @@
...
+using ITensors:
+  Index, inner, ITensor, orthogonalize, qr, randomMPO, randomMPS, siteinds, storage, svd
+using Test: @test
+using Zygote: gradient
+using GPUArraysCore: @allowscalar
```
mtfishman, line 7:
```suggestion
```
See comment above.

NDTensors/ext/examples/NDTensorCUDA.jl
```diff
@@ -61,20 +62,17 @@ function main()
...
   @allowscalar @test (cB * cC * cD) ≈ grad[1]
   # Create a tuple of indices
-  decomp = (
-    dim(NDTensors.ind(grad[1], 1)),
-    dim(NDTensors.ind(grad[1], 2)) * dim(NDTensors.ind(grad[1], 3)),
-  )
+  decomp = (dim(ind(grad[1], 1)), dim(ind(grad[1], 2)) * dim(ind(grad[1], 3)))
   # Reshape the CuVector of data into a matrix
-  cuTensor_data = CUDA.reshape(NDTensors.data(storage(grad[1])), decomp)
+  cuTensor_data = CUDA.reshape(data(storage(grad[1])), decomp)
```
mtfishman, line 71:
```suggestion
  cuTensor_data = reshape(data(storage(grad[1])), decomp)
```
and same below.

Also, this code seems pretty sketchy, generally we should be calling `data(...)` in external code since that is an internal detail. Can we use `array` here or some other higher level function? We don't really want to encourage users to use `data(...)` by putting it in an example.

NDTensors/src/blocksparse/blocksparsetensor.jl
```diff
@@ -49,7 +49,7 @@ Construct a block sparse tensor with uninitialized memory
 from indices and locations of non-zero blocks.
 """
 function BlockSparseTensor(::UndefInitializer, blockoffsets, inds)
-  return BlockSparseTensor(Float64, undef, blockoffsets, inds)
+  return BlockSparseTensor(NDTensors.default_eltype(), undef, blockoffsets, inds)
```
mtfishman, line 52:
```suggestion
  return BlockSparseTensor(default_eltype(), undef, blockoffsets, inds)
```
and same below. If we are in the `NDTensors` library there shouldn't be a need to qualify a function with `NDTensors.`, unless I'm missing something here.

NDTensors/src/lib/Unwrap/test/runtests.jl
```diff
@@ -1,11 +1,27 @@
...
+  norm,
+  qr,
+  svd
 using GPUArraysCore: @allowscalar
-
-include(joinpath(pkgdir(NDTensors), "test", "device_list.jl"))
-@testset "Testing Unwrap $dev, $elt" for dev in devices_list(ARGS),
+include(
+  joinpath(
+    pathof(NDTensors)[1:(end - 16)], "test", "NDTensorsTestUtils", "NDTensorsTestUtils.jl"
```
mtfishman, line 19: The code `pathof(NDTensors)[1:(end - 16)]` looks pretty sketchy, could you explain that?

NDTensors/src/lib/Unwrap/test/runtests.jl
```diff
@@ -53,7 +69,7 @@ include(joinpath(pkgdir(NDTensors), "test", "device_list.jl"))
   copyto!(expose(mp), expose(ma))
   @test mp == ma
 
-  q, r = qr(expose(mp))
+  q, r = LinearAlgebra.qr(expose(mp))
```
mtfishman, line 72: Why does `qr` need to be qualified with `LinearAlgebra.` if you did `using LinearAlgebra: qr` above?

mtfishman: Same comment for `eigen`, `Symmetric`, etc. below.

NDTensors/ext/examples/NDTensorCUDA.jl
kmp5VT: Yeah that makes sense, we also are sure that CUDA and Metal are around in their examples which we aren't sure of in the generic code.  I made these changes and also changed the next line to be variables first then functions, thanks for that reminder too.

NDTensors/ext/examples/NDTensorCUDA.jl
kmp5VT: This line has been removed, thanks. I am still working on changing the Metal example.

NDTensors/ext/examples/NDTensorCUDA.jl
kmp5VT: Yeah I was worried about these lines especially with your changes to array tensor. I have modified these lines to use array and also remove the `storage` function.  Thanks!

NDTensors/src/blocksparse/blocksparsetensor.jl
kmp5VT: Makes sense, I removed all instances in this file. Thanks!

NDTensors/src/lib/Unwrap/test/runtests.jl
kmp5VT: I was using the wrong function; I meant to be using `pkgdir(NDTensors)` I've changed this. Thanks!

NDTensors/src/lib/Unwrap/test/runtests.jl
kmp5VT: I meant to remove the `LinearAlgebra` throughout this file. Thanks

NDTensors/test/runtests.jl
```diff
@@ -1,43 +1,25 @@
...
-    println("Running $filename")
-    include(filename)
+  using Test: @testset
+  @testset "$(@__DIR__)" begin
+    filenames = filter(readdir(@__DIR__)) do f
+      startswith("test_")(f) && endswith(".jl")(f)
+    end
+    for dir in ["lib/", "arraytensor/", "ITensors/"]
+      push!(filenames, dir * "runtests.jl")
+    end
```
mtfishman, line 11:
```suggestion
    for dir in ["lib", "arraytensor", "ITensors"]
      push!(filenames, joinpath(dir, "runtests.jl"))
    end
```
`joinpath("a", "b")` output `"a/b"`, but also handles creating paths cross-platform (i.e. handles Windows conventions for how paths are defined).

NDTensors/test/runtests.jl
```diff
@@ -1,43 +1,25 @@
...
+    @testset "Test $(@__DIR__)/$filename" for filename in filenames
+      println("Running $(@__DIR__)/$filename")
+      include(filename)
+    end
   end
   if "cuda" in ARGS || "all" in ARGS
+    using NDTensors
```
mtfishman, line 18:
```suggestion
    using NDTensors: NDTensors
```
to only bring the module name `NDTensors` into scope.

NDTensors/test/runtests.jl
```diff
@@ -1,43 +1,25 @@
...
     include(joinpath(pkgdir(NDTensors), "ext", "examples", "NDTensorCUDA.jl"))
   end
   if "metal" in ARGS || "all" in ARGS
+    using NDTensors
```
mtfishman, line 22:
```suggestion
    using NDTensors: NDTensors
```
to only bring the module name `NDTensors` into scope.

NDTensors/ext/examples/NDTensorCUDA.jl
```diff
@@ -61,20 +61,17 @@ function main()
...
   @allowscalar @test (cB * cC * cD) ≈ grad[1]
   # Create a tuple of indices
-  decomp = (
-    dim(NDTensors.ind(grad[1], 1)),
-    dim(NDTensors.ind(grad[1], 2)) * dim(NDTensors.ind(grad[1], 3)),
-  )
+  decomp = (dim(ind(grad[1], 1)), dim(ind(grad[1], 2)) * dim(ind(grad[1], 3)))
   # Reshape the CuVector of data into a matrix
-  cuTensor_data = CUDA.reshape(NDTensors.data(storage(grad[1])), decomp)
+  cuTensor_data = CUDA.reshape(array(grad[1]), decomp)
```
mtfishman, line 70: I still see a few instances of `CUDA.reshape` instead of just `reshape`.

NDTensors/test/runtests.jl
kmp5VT: This makes sense. I fixed both of these, thanks!

NDTensors/ext/examples/NDTensorCUDA.jl
kmp5VT: I saw this and wasn't sure why I used it over Base.reshape (i.e. reshape) but didn't look into it. I removed `CUDA.` in all instances now. Thanks

NDTensors/ext/NDTensorsMetalExt/permutedims.jl
kmp5VT: I can't remember my rational in the moment. I went through it again today and I don't think its necessary. I added a tests in the `Unwrap/test` file and am verifying that all tests pass on all infrastructures now.

NDTensors/test/runtests.jl
kmp5VT: Ah yes that is very important, sorry and thank you for the reminder!

NDTensors/test/runtests.jl
kmp5VT: I also made this modification

NDTensors/test/readwrite.jl
mtfishman: Ok interesting. Maybe we could move it to a directory `NDTensors/test/broken/` for now.

kmp5VT:
@mtfishman I believe this is ready for review

NDTensors/src/lib/Unwrap/test/runtests.jl
```diff
@@ -1,10 +1,22 @@
-using Test
+using Test: @testset, @test, @test_broken
```
mtfishman, line 1: Should probably wrap all of this in:
```julia
@eval module $(gensym())
# ...
end
```
Same for any other test files that don't use `@safetestset` or aren't wrapped in a module already.

NDTensors/ext/examples/NDTensorCUDA.jl
```diff
@@ -61,20 +58,17 @@ function main()
...
-  @allowscalar @test (cB * cC * cD) ≈ grad[1]
+  grad = gradient(f, cA, cB, cC, cD)
+  @test cpu(cB * cC * cD) ≈ cpu(grad[1])
+  @test (cB * cC * cD) ≈ grad[1]
   # Create a tuple of indices
-  decomp = (
-    dim(NDTensors.ind(grad[1], 1)),
-    dim(NDTensors.ind(grad[1], 2)) * dim(NDTensors.ind(grad[1], 3)),
-  )
+  decomp = (dim(ind(grad[1], 1)), dim(ind(grad[1], 2)) * dim(ind(grad[1], 3)))
```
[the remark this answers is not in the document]
kmp5VT: Thanks, I missed this one. Its updated to match your suggestion

NDTensors/src/lib/Unwrap/test/runtests.jl
kmp5VT: Okay, and I added this to all the other `runtests.jl` in the lib folder that didn't have it

NDTensors/src/lib/SmallVectors/test/runtests.jl
```diff
@@ -1,5 +1,6 @@
+@eval module $(gensym())
```
kmp5VT, line 1: This seems to fail with `f!_impl_broken` not defined

mtfishman: Probably an issue with having nested `@eval`, maybe you could try removing the inner one. If that doesn't work, you can just remove this module wrapper and we can figure it out later, not a priority for this PR.

NDTensors/src/lib/Unwrap/test/runtests.jl
mtfishman: Great, thanks!

NDTensors/src/lib/SmallVectors/test/runtests.jl
kmp5VT: I tried removing the inner `eval` and that didn't work so I removed it from the file for now

kmp5VT:
@mtfishman I am not sure why these tests are failing. When I test with Julia 1.9 all pass but it looks like theres an issue with `sort` in julia 1.6 and 1.8. I didn't make any modifications to the code that is failing. To verify I grabbed the last passing version of this branch (tag b56f8a421000e56c25fac910fc4c38d6400e4e8c) and ran `]test NDTensors` in julia 1.8 and it also fails in the `sort` function in blocksparse contraction with the combiner. Do you have any idea what is happening?

mtfishman:
May be related to https://github.com/JuliaLang/Compat.jl/pull/812 and https://github.com/JuliaLang/julia/pull/52010, sorting tuples was temporarily supporting in Base Julia but then removed.

In places where `sort(::Tuple)` is failing we can try to change the call to `TupleTools.sort(...)`.

mtfishman:
Thanks for the quick fix to the tuple sorting issue. I'll merge this so we can register a new version.

Issue #1250, opened 2023-11-16
[NDTensors] [ENHANCEMENT] Roadmap to removing `TensorStorage` types


mtfishman:
Here is a roadmap to removing `TensorStorage` types (`EmptyStorage`, `Dense`, `Diag`, `BlockSparse`, `DiagBlockSparse`, `Combiner`).

### UnallocatedZeros

- [ ] Define `UnallocatedZeros` (in progress in #1213).
- [ ] Use `UnallocatedZeros` as default data type instead of `EmptyStorage` in ITensor constructors.

### DiagonalArray

- [ ] Define `DiagonalArray`.
- [ ] Tensor contraction, addition, QR, eigendecomposition, SVD.
- [ ] Use `DiagonalArray` as default data type instead of `Diag` in ITensor constructors.

### BlockSparse

- [ ] Define `BlockSparseArray`.
- [ ] Tensor contraction, addition, QR, eigendecomposition, SVD.
- [ ] Use `BlockSparseArray` as default data type instead of `BlockSparse` in ITensor QN constructors.

### DiagBlockSparse

- [ ] Use `BlockSparseArray` with blocks storing `DiagonalArray`, make sure all tensor operations work.
- [ ] Replace `DiagBlockSparse` in ITensor QN constructors.

### Combiner

- [ ] Not sure what to do with this, but a lot of functionality will be replaced by the new `fusedims`/`matricize` functionality in `TensorAlgebra`/`BlockSparseArrays` and also by the new `FusionTensor` type.

### Simplify ITensor and Tensor constructors

- [ ] Make ITensor constructors more uniform by using a style `tensor(storage::AbstractArray, inds::Tuple)`, avoid constructors like `DenseTensor`, `DiagTensor`, `BlockSparseTensor`, etc.
- [ ] Use `rand(i, j, k)`, `randn(i, j, k)`, `zeros(i, j, k)`, fill(1.2, i, j, k)`, `diagonal(i, j, k)`, etc. instead of `randomITensor(i, j, k)`, `ITensor(i, j, k)`, `ITensor(1.2, i, j, k)`, `diagonITensor(i, j, k)`.
- [ ] Consider `randn(2, 2)(i, j)` as a shorthand for creating an ITensor with indices `(i, j)` wrapping an array. Also could use `setinds(randn(2, 2), i, j)`.
- [ ] Remove automatic conversion to floating point in ITensor constructor.

### Define `TensorAlgebra` submodule

[...]

PR #1265, opened 2023-11-17
[NDTensors] Start `TensorAlgebra` module

mtfishman:
This is initial work on the `TensorAlgebra` module mentioned in #1250.

So far, it defines a simplified version of TTGT tensor contraction (https://arxiv.org/abs/1607.00145), which can be called with `TensorAlgebra.contract(randn(2, 3), (:a, :b), randn(3, 4), (:b, :c); alg="matricize")`.

The code design is inspired by the one in [TensorOperations.jl](https://github.com/Jutho/TensorOperations.jl), in particular it makes use of what I call a `BipartitionedPermutation` which is a permutation of the form `((1, 3), (2, 4))`, which defines a matricization of a tensor. The code logic revolves around analyzing the input labels and defining bipartitioned permutations/matricizations of the input tensors which define how they will be mapped to matrix multiplications.

Right now the logic misses some optimizations that are captured by the current `ContractionProperties` logic, namely it doesn't avoid certain permutations in cases which can be mapped to matrix multiplication through reshapes and transpositions, for example `TensorAlgebra.contract(a, (:a, :b), b, (:c, :b))` can be mapped to a matrix multiplication `a * transpose(b)` but currently the code will eagerly permute `b`. This will be handled in future work, hopefully in a simpler way than the current `ContractionProperties` code (perhaps by simply analyzing the `BipartitionedPermutation` and swapping the two partitions and seeing if that corresponds to a trivial permutation, in which case it will be marked as a transposition).

Some of the code overlaps with code logic in #1246, for example both have `matricize`/`fusedims` functionality and some quasi-generic functionality for allocating the output of tensor operations. #1246 focused more on block sparse arrays but also has some more generic code. The goal will be to make the code in `TensorAlgebra` more generic so then a lot of the logic for matricization, along with matricized linear algebra (tensor contraction, tensor decompositions, etc.), can live just in `TensorAlgebra` and submodules like `BlockSparseArrays` can mostly make use of generic code and it will "just work" when certain overloads are defined, like information about how to fuse blocked axes/indices.

@kmp5VT @emstoudenmire


emstoudenmire:
Great to see this. I think it will lead to a lot shorter code that can do more things (e.g. plug in more backends which can be simpler themselves).

PR #1266, opened 2023-11-17
[TensorAlgebra] Matricized QR tensor decomposition

mtfishman:
This defines a function for performing a matricized QR decomposition of a tensor.

The high level API is:
```julia
julia> using NDTensors.TensorAlgebra

julia> using LinearAlgebra

julia> a = randn(2, 2, 2, 2);

julia> q, r = qr(a, (:a, :b, :c, :d), (:a, :c), (:b, :d));

julia> q
2×2×4 Array{Float64, 3}:
[:, :, 1] =
 -0.320888   0.370723
 -0.306767  -0.815775

[:, :, 2] =
  0.47314   -0.653369
 -0.515355  -0.289234

[:, :, 3] =
 -0.369393  -0.595887
  0.616975  -0.357504

[:, :, 4] =
 -0.732611  -0.283888
 -0.509553   0.350778

julia> r
4×2×2 Array{Float64, 3}:
[:, :, 1] =
 2.9428  -0.54139
 0.0      2.80448
 0.0      0.0
 0.0      0.0

[:, :, 2] =
 -0.225018  -0.0430216
  1.07535    0.0289476
 -1.58676    0.550383
  0.0        0.0227928

julia> a′ = TensorAlgebra.contract((:a, :b, :c, :d), q, (:a, :c, :qr), r, (:qr, :b, :d));

julia> norm(a - a′)
1.21523787027731e-15
```
We can also provide a permutation-based API, for example `qr(a, (1, 3), (2, 4))` which would be equivalent to the example above.

There is currently a bug in `TensorAlgebra.contract` which is marked with a broken test so some label combinations can't be properly checked with `TensorAlgebra.contract`, I'll investigate that in a future PR. But the main point here is to get started with the general code patterns involving `matricize` and `unmatricize` for performing matricized tensor decompositions, i.e. these patterns and a lot of code can be shared with SVD and other decompositions.


emstoudenmire:
Nice – the interface looks really intuitive.

PR #1277, opened 2023-12-04
[NDTensors] `Sectors` library

emstoudenmire:
Add `Sectors` library for NDTensors. Works similarly as the current `QN` system with the following main differences:

1. Each `Label` within a `Sector` holds a `Category` object which can be extended by overloading various methods based on the category's name. For instance, one can define mappings from string inputs such as "τ" or "1" to the numerical values used internally in a custom way for just that Category. Fusion rules are defined by simple functions such as:
    ```julia
    fusion_rule(C::CategoryName"Z", n1, n2) = [(n1 + n2) % groupdim(C)]
    ```
    for the case of the group Z(N) (i.e. U(1) mod N).
    For SU(2) it is basically
    ```julia 
    fusion_rule(C::CategoryName"SU", j1, j2) = collect(abs(j1 - j2):(j1 + j2))
    ```
    So a non-Abelian group is just one where `fusion_rule` returns a vector of length greater than 1.

2. Fusing Sectors with the `⊗` (or equivalently `*`) operator can return a direct sum of `Sector` objects (when the Category is non-Abelian). Direct sums are represented as `Vector{Sector}`. The `⊕` operator combines Sectors into such Vectors for convenience.

3. Internally, the `Sector` type uses a sorted set storage to make the code implementing it very compact.

4. Label values are allowed to be integers or half integers (using the HalfIntegers `Half{Int}` type internally, for convenient conversion from floats such as 1/2 to the Half{Int} type etc.).

5. As a proof of concept, Labels can now hold either one or two values, making it possible to have "quantum number" pairs such as (j,m).

We should be able to easily add many other custom method overloads for additional group and category concepts just by overloading methods on the category name (as already done for fusion rules).

Future improvements could include storing an arbitrary number of values, more than two, using a 'small vector' pattern (i.e. union of a fixed size array and vector).

To do list:

- [ ] Change how defaults are handled (instead of const variables and such)


NDTensors/src/lib/Sectors/src/Sectors.jl
```diff
@@ -0,0 +1,19 @@
...
+import ..SortedSets: AbstractSmallSet
+using InlineStrings
+using HalfIntegers
+
+import LinearAlgebra: norm
+import Base: *, ==, getindex, intersect, union, setdiff, symdiff, iterate, length, show
+#import NDTensors: ⊗
+#import NDTensors.SortedSets: AbstractSmallSet
+
+export ⊗, ⊕, U, SU, SUd, SUz, Z, Fib, Ising, Sector, nactive, show
```
mtfishman, line 13: I have been going with the strategy of not exporting functions defined in submodule libraries at first and requiring end users to do `using NDTensors.Sectors: ⊗`, to be conservative at first. We can decide on exports later once things are more developed.

emstoudenmire: Sounds good – let's make that change when appropriate.

NDTensors/src/lib/Sectors/src/Sectors.jl
```diff
@@ -0,0 +1,19 @@
+module Sectors
+
+using ..SortedSets
+import ..SortedSets: AbstractSmallSet
+using InlineStrings
+using HalfIntegers
+
+import LinearAlgebra: norm
+import Base: *, ==, getindex, intersect, union, setdiff, symdiff, iterate, length, show
```
mtfishman, line 9: Recently we have been using the style of explicitly listing functions that are being used in `using` statements, such as:
```julia
using InlineStrings: String7
```
Additionally, we have been listing those at the tops of the files where those functions are actually being used, instead of in the top level `module` definition.

Also, instead of `import Base: getindex`, we have been moving towards the style of overloading with `Base.getindex`.

I think it makes thing much more explicit, i.e. you can read the file locally and see where the functions were initially defined.

emstoudenmire: Great

NDTensors/src/lib/Sectors/src/sector.jl
```diff
@@ -0,0 +1,109 @@
+const default_sector_size = 4
+default_sector_data() = Label
+const SStorage = SmallSet{default_sector_size,default_sector_data()}
+const MSStorage = MSmallSet{default_sector_size,default_sector_data()}
```
mtfishman, line 4: We should discuss this code, I think there is a better code pattern for accomplishing the same thing without defining globals like this.

NDTensors/src/lib/Sectors/src/sector.jl
```diff
@@ -0,0 +1,109 @@
...
+  common_labels_match = all(t -> (t[1] == t[2]), common_labels)
+  unique_labels_zero = all(l -> iszero(val(l)), symdiff(A, B))
+  return common_labels_match && unique_labels_zero
+end
+
+# Treat [q] as q for comparison purposes
+# so we can check things like a ⊗ b = c
+# when a ⊗ b returns a Vector
+==(q::Sector, v::Vector{<:Sector}) = (length(v) == 1 && q == first(v))
+==(v::Vector{<:Sector}, q::Sector) = (q == v)
```
mtfishman, line 84: I don't think this kind of definition is a good idea.

NDTensors/src/lib/Sectors/test/runtests.jl
```diff
@@ -0,0 +1,186 @@
...
+
+    @test q0 ⊗ q0 == q0
+    @test q0 ⊗ q1 == q1
+    @test q1 ⊗ q0 == q1
+    @test q0 ⊗ q2 == q2
+    @test q1 ⊗ q2 == q3
+  end
+
+  @testset "Ƶ_2" begin
+    z0 = Sector(0, Z(2))
```
mtfishman, line 19: What about `Sector(Z(2), 0)` instead?

NDTensors/src/lib/Sectors/src/categories.jl
```diff
@@ -0,0 +1,192 @@
...
+Convenience macro for value dispatch on names 
+of Category objects. Writing (::CategoryName"Name", ...)
+in function arguments allows Val-based dispatch based on
+the name value.
+"""
+macro CategoryName_str(s)
+  return :(CategoryName{$(Expr(:quote, String7(s)))})
+end
+
+Category(C::CategoryName) = Category(basename(C), groupdim(C), level(C))
```
mtfishman, line 77: I would move this to a file `categoryname.jl` (I like to put new types in their own files).

NDTensors/src/lib/Sectors/src/categories.jl
```diff
@@ -0,0 +1,192 @@
...
+fusion_rule(::CategoryName"Ising", a1, a2) = fusion_rule(SU(2, 2), a1, a2)
+
+val_to_str(::CategoryName"Ising", a) = ("1", "σ", "ψ")[Int(2 * a + 1)]
+
+function string_to_val(::CategoryName"Ising", s::AbstractString)
+  for (a, v) in enumerate(("1", "σ", "ψ"))
+    (v == s) && return (a - 1)//2
+  end
+  return error("Unrecognized string \"$s\" for Ising category")
+end
```
mtfishman, line 192: I would put this in a separate file like `category_definitions.jl`, or maybe in a folder `category_definitions` and put each definition into a it's own file.

NDTensors/src/lib/Sectors/src/sector.jl
mtfishman: Definitely we should avoid `const default_sector_size = 4` since that is then being used as a type parameter which is not good for type stability.

NDTensors/src/lib/Sectors/test/runtests.jl
mtfishman: Also what about a shorthand `Z(2)(0)` or something like that?

NDTensors/src/lib/Sectors/src/sector.jl
mtfishman: I see you define it so that you can have code like `@test d1 ⊗ d1 == d1` but I think it is better to be explicit and use `@test d1 ⊗ d1 == [d1]`, instead of having "magic" like this where a collection of sectors of length 1 act like a single sector, but only for the sake of equality.

NDTensors/src/lib/Sectors/src/categories.jl
```diff
@@ -0,0 +1,192 @@
...
+  return fusion_rule(CategoryName(C), a, b)
+end
+
+#
+# Version of Category type but with
+# basename held statically as a 
+# type parameter. Mostly for internal use.
+#
+struct CategoryName{N}
+  N::Int
```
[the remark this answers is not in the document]
mtfishman: Also, another design to consider is:
```julia
struct Category{Name}
  name::Name
  groupdim::Int
  level::Int
end
```
`Category("SU", 2)` would make the current `Category` object, and `Category(CategoryName"SU", 2)` would make the current `CategoryName` object. The new `CategoryName` would be a singleton type with no fields, like the `OpName` object in `ITensors.jl`.

My motivation for that change is that the current design has two slightly strange aspects that don't feel right to me:
1. `CategoryName` and `Category` are almost the same type except the name is static, but in the current design they are two different types.
2. `CategoryName` isn't a great name for the current type, since it isn't just the name but also has the group dimension and level.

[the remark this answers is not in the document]
mtfishman: It seems like the current overloading syntax for `fusion_rule` could still be used, but instead:
```julia
fusion_rule(::Category"U", n1, n2) = [(n1 + n2)]
```
and `Category"U"` could be the type of a `Category` with a static name of type `CategoryName"U"`.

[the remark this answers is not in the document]
mtfishman: We could have an interface for converting back and forth between `Category` with dynamic and static names, such as:
```julia
static(Category("SU", 2)) == Category(CategoryName("SU"), 2)
dynamic(Category(CategoryName("SU"), 2)) == Category("SU", 2)
```

emstoudenmire:
Thanks for the comments. I'll make all of the smaller changes first then we can discuss the more structural and deeper changes.

NDTensors/src/lib/Sectors/src/sector.jl
emstoudenmire: Removed & changed

NDTensors/src/lib/Sectors/src/category.jl
```diff
@@ -0,0 +1,118 @@
...
+# TODO:
+# This method makes `basename` return the same type (and value) 
+# of the dynamic and static versions of a Category
+# Is that the behavior we want? (Otherwise basename(D) != basename(static(D)).)
+# May be enough to just make CategoryName's print whatever Name they hold.
+basename(C::Category{CName}) where {CName<:CategoryName} = basename(CName())
+
+function static(C::Category)
+  CName = CategoryName{basename(C)}
+  return Category(CName(), groupdim(C), level(C); name_type=CName)
```
mtfishman, line 64: Maybe a good use case for using `_Category`, since then it doesn't require passing `name_type`?

NDTensors/src/lib/Sectors/src/category.jl
```diff
@@ -0,0 +1,118 @@
...
+end
+
+dynamic(C::Category) = Category(basename(C), groupdim(C), level(C))
+
+is_static(C::Category) = (name_type(C) <: CategoryName)
```
mtfishman, line 69: I would have a slight preference for using `basename(C) isa CategoryName` here.

emstoudenmire: Done

NDTensors/src/lib/Sectors/src/category.jl
emstoudenmire: Good point, done

NDTensors/src/lib/Sectors/src/category.jl
```diff
@@ -0,0 +1,118 @@
...
+
+is_dynamic(C::Category) = !is_static(C)
+
+"""
+Convenience macro for value dispatch on names 
+of Category objects. Writing (::CategoryType"Name", ...)
+in function arguments allows Val-based dispatch based on
+the name value.
+"""
+# TODO: how to avoid using default_category_name_type() here?
```
mtfishman, line 79: Good question, definitely a bit subtle. I think one way to guide the design would be based on where this constructor is being used.

NDTensors/src/lib/Sectors/src/category.jl
```diff
@@ -0,0 +1,118 @@
...
+
+function fusion_rule(C::Category, a, b)
+  return error("fusion_rule not implemented for category $(basename(C))")
+end
+function fusion_rule(C::DynamicCategory, a, b)
+  # TODO: improve... maybe don't use 
+  # nvals and just pass a and b?
+  if nvals(C) == 1
+    return fusion_rule(static(C), a[1], b[1])
+  end
```
mtfishman, line 106: Agreed this doesn't feel quite right.

emstoudenmire: Yes, this is more in the zone of some things I'm still iterating on, to make everything more generic to the value type stored inside of `Label`.

NDTensors/src/lib/Sectors/src/category.jl
```diff
@@ -0,0 +1,118 @@
...
+
+"""
+Convenience macro for value dispatch on names 
+of Category objects. Writing (::CategoryType"Name", ...)
+in function arguments allows Val-based dispatch based on
+the name value.
+"""
+# TODO: how to avoid using default_category_name_type() here?
+# Maybe always convert to Symbol?
+macro CategoryType_str(s)
```
mtfishman, line 81: Something doesn't feel quite right to me about this `CategoryType` macro, I expected there to be a type `CategoryType`. Can't think of anything better, I see why you ended up with this design.

NDTensors/src/lib/Sectors/src/category_definitions.jl
```diff
@@ -0,0 +1,115 @@
+
+#
+# 2D rotation group U(1)
+#
+
+U(N::Int) = Category("U", N)
+
+fusion_rule(::CategoryType"U", n1, n2) = [(n1 + n2)]
```
mtfishman, line 8: I feel like we already discussed this kind of design, but what makes sense to me is something like:
```julia
struct U
  N::Int
end
CategoryName(c::U) = CategoryName{String7("U")}()
Category(c::U) = Category(CategoryName(c), c.N)
category(l::Label{<:Any,CategoryName"U"}) = U(only(values(l))) # Not sure if this is right, and ideally the syntax would be simplified

fusion_rule(c1::U, c2::U) = [(c1.N + c2.N)]
```
i.e. the user defines a symmetry/category object and fusion rules for that object, and then defines conversion back and forth between their type and the "canonical" `Category`/`Label` types.

mtfishman: One reason for going in that direction would be that `U` could in principle just be used directly as a sector of a `GradedAxes`, instead the `Label`/`Sector` type. So we can think of the job of the `Sectors` module as providing canonical `Label`/`Sector` types which you can convert a custom symmetry type to in order to get fancier behavior in terms of mixing symmetry types together.

emstoudenmire: Yes, we can definitely add back having the separate types for each group or category. (So an actual U type, SU type, etc.) 

One thing I'm missing about the above, though, is that in your example it seems the `N` field is storing the actual value of the label e.g. the number of particles. Whereas I was envisioning the Category types as just specifying which group the values go with. We could of course make a type that holds the group name and values together, but that's what `Label` already was supposed to be. 

But I'll think about it some more, because it could be a good direction. So like we could have something like:
```julia
struct U1 <: AbstractSector
  n::int
end

struct SU{N} where {N} <: AbstractSector
  J::NTuple{Half{Int},N-1}
end
```
i.e. distinct types that both specify a group and the particular irrep (or "sector value") of that group.  That would certainly be nice from the perspective of defining fusion rules etc. directly as dispatching on these types and returning arrays of such objects.

Then we could reimagine `Label` as being just a means to store the above structs in a common, dynamic, format with an optional label name. And then Sector as a sorted set of those.

emstoudenmire: Probably should schedule a chat about this since it's more fundamental about the whole design. It could be better, but it's a big change so would need to be on the same page about it.

mtfishman: Yeah it's a good question if these objects should be just the group or an irrep of the group. It seems better both for defining fusion rules and conversions for them to be irreps but I haven't thought through the implications of the design beyond that. Sure, let's set up a chat about it.

emstoudenmire: It really could be a better design, since for one thing it would push off the issue of the `Label` type having to hold various value types (e.g. tuple of different lengths) inside that could then "collide" when they are to be stored in a single `Sector`. So maybe a design where the categories and values are more designed together could clarify how to do these conversions/promotions more nicely. (Like finding a common type when two of these custom types "meet" each other.)

mtfishman:
Another design proposal would be to remove the `name` field from the `Label` type. Instead, if we want "named labels", we can use a dictionary as the storage of `Sector` where the keys are the names and the values are `Label`. Then, named labels or position labels can be used inside of `Sector` by using either a dictionary or vector of `Label`.

It would be easy enough to make a `SmallDictionary` type based on the current `SmallSet` type. The main reason for this new design would be to simplify the `Label` type and make it clearer what it is, i.e. basically it is just an irrep, and whether it is named or not is an orthogonal concept which really is only relevant when there are multiple labels in a sector and you want to decide how they mix with each other.

PR #1278, opened 2023-12-05
[BlockSparseArrays] Sparse and block sparse matrix multiplication

mtfishman:
This adds support for sparse and block sparse matrix multiplication.

TODO:

- [x] Test matrix multiplication of `DiagonalArray` and `SparseArrayDOK`.


emstoudenmire:
Great. Does this already use the sparse tensor type in the way you were saying, with the 'elements' being dense tensors/matrices?

mtfishman:
Yeah, that's been the design of the `BlockSparseArray` type in the [BlockSparseArrays library](https://github.com/ITensor/ITensors.jl/tree/main/NDTensors/src/lib/BlockSparseArrays).

PR #1281, opened 2023-12-06
[ITensors] Update CI testing and package versions

kmp5VT:
# Description

CI uses out of data packages which can lead to error in testing suite. 

# Checklist:

- [x] Update Jenkins to Julia 1.9
- [x] Make sure all tests pass with up to date Zygote
- [x] Chainrules test fail for lower Julia versions, understand what Zygote version causes break

with regard to the last point. Zygote@0.6.60 works in Julia versions below v"1.8" it would appear the the breaking change in Zygote is in the 0.6.61 update


test/ITensorChainRules/test_chainrules_ops.jl
```diff
@@ -157,21 +157,23 @@ using Zygote: ZygoteRuleConfig, gradient
     atol=1.0e-7,
   )
 
-  f = function (x)
-    y = ITensor(exp(-x * Op("X", 1) * Op("X", 2)), s)
-    return norm(y)
-  end
-  args = (x,)
-  test_rrule(ZygoteRuleConfig(), f, args...; rrule_f=rrule_via_ad, check_inferred=false)
+  if VERSION > v"1.7"
```
mtfishman, line 160:
```suggestion
  if VERSION > v"1.6"
```
Is this what you meant here (and below)? I believe you said it works for Julia 1.7 and later.

kmp5VT: For reference: With the current version of Zygote (0.6.64) the codes here fail on Julia versions below 1.8. The error is in a library JLL file and the error causes julia to crash. I have updated to skip these tests when julia version is 1.7 and below.

kmp5VT:
> Thanks! Ready to merge?

Ah yes, thanks!

PR #1286, opened 2023-12-14
GPU functionality in NamedDimsArray

kmp5VT:
# Description

@mtfishman This is branch to see if its possible to use different array backends in the NamedDimsArray object.

# Checklist:

- [ ] Add `device` loop to tests
- [ ] If things break create minimal changes to allow code to work with different backends
- [ ] Tests should run with `allowscalar = false`


kmp5VT:
@mtfishman I realize now this is a relatively confusing PR because I built it off of the `UnallocatedArrays` branch. I am closing this and will move it to a new branch from `Main`

PR #1288, opened 2023-12-15
[NDTensors] Optimize `permutedims`

mtfishman:
This fixes a few performance issues in `permutedims`.

Inspired by [this discourse question](https://itensor.discourse.group/t/is-there-a-better-way-to-use-the-combiner-function/1358).

@kmp5VT


kmp5VT:
It looks like there is an error with `Strided` and `Base.ReshpaedArray`. Heres the example 
```
y = reshape(randn(elt, 8)', 2, 4)
x = Base.ReshapedArray(randn(elt, 8, 8)'[1:8], (2, 4), ())
y .= y .* x
# works 
@strided x .= x .* x

# stack overflow
@strided y .= y .* x
@strided x .= x .* y
@strided y .= y .* y
```
so I suggest we make a fix like 
```
function permutedims(Edest::Exposed{<:Array, <:Base.ReshpaedArray}, Esrc::Exposed{<:Array}, perm, f) 
  adest = unexpose(Edest)
  asrc = unexpose(Esrc)
  adest .= f.(adest, permutedims(asrc, perm))
  return adest
end
```
and I can make an issue in `Strided` with my example above

mtfishman:
That issue should be fixed by https://github.com/Jutho/StridedViews.jl/pull/5.

PR #1290, opened 2023-12-17
[TensorAlgebra] Refactor `contract`

mtfishman:
This refactors the `contract` function in `TensorAlgebra` based on refactored `fusedims` and `splitdims` functions for matricizing and unmatricizing.

It is based around a `BlockedPermutation` object which represents a permutation such as `(2, 1, 4, 3)` which is blocked or partitioned into parts, such as `((2, 1), (4, 3))`. That is used to represent a matricization of an n-dimensional array.

Basically ready, I still need to update definitions of `fusedims` and `splitdims` for `NamedDimsArray` where dimension names are used to specify matricization/dimension fusion.


mtfishman:
[test ITensors mps]

Issue #1294, opened 2023-12-19
[ITensors] [BUG] AD of blocksparse ITensors failing in v0.3.52


ntausend:
**Description of bug**

Hey, upon updating my ITensor version from `v0.3.43` to `v0.3.52` I experienced the AD system of Zygote to be broken with blocksparse matrices. By trying to do the derivative of the Frobenius scalar product $Tr{A^\dagger B}$ Zygote reports to miss an Adjoint function for the constructor of `NDTensors.BlockSparse{Float64, Vector{Float64}, 0}` (full message below).

**Minimal code demonstrating the bug or unexpected behavior**

<details><summary>Minimal runnable code</summary><p>

```julia
import Pkg
Pkg.activate(;temp = true)
Pkg.add(name="ITensors", version = "0.3.52")
Pkg.add(name="Zygote", version="0.6.67")

using ITensors
using Zygote: pullback

N = 2	
s = siteinds("SpinHalf", N; conserve_szparity = true)
A = randomITensor(s)
B = randomITensor(s)

f(A,B) = dot(A,B)
val,grad = pullback(f,A,B)
grad(1)
```

</p></details>

**Expected output or behavior**
The expected output would be the Gradient of the Frobenius product $(A,B)$. Instead under `v0.3.53` an error is returned.

<details><summary>Output of minimal runnable code</summary><p>

```julia
ERROR: LoadError: Need an adjoint for constructor NDTensors.BlockSparse{Float64, Vector{Float64}, 0}. Gradient is of type NDTensors.Dense{Float64, Vector{Float64}}
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] (::Zygote.Jnew{NDTensors.BlockSparse{Float64, Vector{Float64}, 0}, Nothing, false})(Δ::NDTensors.Dense{Float64, Vector{Float64}})
    @ Zygote ~/.julia/packages/Zygote/YYT6v/src/lib/lib.jl:330
  [3] (::Zygote.var"#2214#back#313"{Zygote.Jnew{NDTensors.BlockSparse{Float64, Vector{Float64}, 0}, Nothing, false}})(Δ::NDTensors.Dense{Float64, Vector{Float64}})
    @ Zygote ~/.julia/packages/ZygoteRules/4nXuu/src/adjoint.jl:71
  [4] Pullback
    @ ~/.julia/packages/NDTensors/GwjuH/src/blocksparse/blocksparse.jl:11 [inlined]
[...]

PR #1295, opened 2023-12-21
[ITensors] Remove `NDTensors.cpu` from `dot`

kmp5VT:
To address the AD issue found in https://github.com/ITensor/ITensors.jl/issues/1294, remove the unnecessary call to `NDTensors.cpu`. In a previous PR we have fixed the `[]` function for order-0 tensors on GPU by using the `@allowscalar` function.


mtfishman:
Thanks, could you add a simple test for running AD on `dot(...)`?

kmp5VT:
> Thanks, could you add a simple test for running AD on `dot(...)`?

Yes, should I add the test to `ITensors` and should I loop over different devices and or element types?

mtfishman:
Just adding a simple test to https://github.com/ITensor/ITensors.jl/blob/v0.3.52/test/ITensorChainRules/test_chainrules.jl that matches the style of the tests there is sufficient, no need to loop over element types and devices since we don't do that there anyway.

test/ITensorChainRules/test_chainrules.jl
```diff
@@ -352,6 +352,17 @@ Random.seed!(1234)
     args = (A, B)
     test_rrule(ZygoteRuleConfig(), f, args...; rrule_f=rrule_via_ad, check_inferred=false)
   end
+
+  @testset "issue 1294" begin
```
kmp5VT, line 356: @mtfishman I added this test but it fails after 3 tests in `test_rrule` with the error 
```
test_rrule: f on ITensor,ITensor: Error During Test at /Users/kpierce/.julia/packages/ChainRulesTestUtils/by7Dj/src/testers.jl:202
  Fluxes not all equal
  ```
  I am not too familiar with this `test_rrule` function is this something I should look into fixing or is there a way to modify the test to not run the one that is expected to fail for the electron case

mtfishman:
Looks good, thanks!

PR #1296, opened 2023-12-28
[NDTensors] `Sectors` library take 2

emstoudenmire:
Redesign of Sector system as well as group irreps and category labels.

Group irreps are now standalone types that can be customized as needed and just overload methods such as `dimension` (return the dimension of an irrep), `trivial` (return the trivial irrep), and `fusion_rule` (fuse two irreps, with two possible interfaces). They are subtypes of either `AbstractGroup` or `AbstractCategory`. These types can be used directly as "quantum numbers" in that they can be fused to give other irreps / labels.

The `Sector` type is modeled on a dictionary, offers a set-like interface, and uses a NamedTuple as storage. It can hold arbitrary types as values, labeled by symbols.

The `OrderedSector` is modeled on a vector, and represents the concept of a Cartesian product of different group irreps, and can be formed using the `×` (\times) operator. I used a specific type here rather than a plain Vector because Vector already stood for direct sum and so that `OrderedSector` can subtype `AbstractCategory`. Unlike `Sector`, an `OrderedSector` requires all of the irrep labels to be explicitly present. (Otherwise there would be no way to distinguish repeated U1's, for example.)

Examples:
```julia
using BenchmarkTools
using Test

@test U1(1) ⊗ U1(2) == [U1(3)]

@test Fib("τ") ⊗ Fib("τ") == Fib("1") ⊕ Fib("τ")

@test SU2(1//2) ⊗ SU2(1//2) == SU2(0) ⊕ SU2(1)
@test SU2(1) ⊗ SU2(1) == SU2(0) ⊕ SU2(1) ⊕ SU2(2)

# Initial support for SU{N} irreps, using GT labels:
@test dimension(SU{3}((4, 2, 0))) == 27


#
# Sector type
# 

s1 = Sector(A=U1(1), B=SU2(1//2))
s2 = Sector(B=SU2(1))

@test s1 ⊗ s2 == Sector(A=U1(1),B=SU2(1//2)) ⊕ Sector(A=U1(1),B=SU2(3//2))

@btime s1 ⊗ s2 
#gives 5.118 μs

# Can make Sectors from pairs too
s = Sector("A" => U1(3), "J" => SU2(1//2))


#
# OrderedSector type
# 

s11 = U1(1) × U1(1) 
@test s11 ⊗ s11 == [U1(2) × U1(2)] 

phh = SU2(1//2) × SU2(1//2)
@test phh ⊗ phh ==
      (SU2(0) × SU2(0)) ⊕ (SU2(1) × SU2(0)) ⊕ (SU2(0) × SU2(1)) ⊕ (SU2(1) × SU2(1))


```

The `N=2` case of the `SU{N}` irrep type uses irrep dimensions as labels. In contrast, the `SU2` irrep type uses conventional J values such as 0,1/2,1, 3/2.
* `SU{2}(1)` is equivalent of `SU2(0)`
* `SU{2}(2)` is equivalent of `SU2(1//2)`

Possible downsides of this design / future considerations:
* All entries of a Sector are now required to have names
* The Julia formatter automatically changes syntax like `Sector(J=SU2(2))` into `Sector(; J=SU2(1//2))`which is correct but looks strange.
* Need to think through how "outer degeneracy" will be passed along by `fusion_rule`. Possibilities include returning integer information (degeneracy counts) alongside the values in the returned vector of irreps, or the irreps themselves storing their own degeneracy counts.
* Related to this point about outer degeneracy, we already store "degeneracy" or "multiplicity" in graded spaces. So maybe `fusion_rule` should already be returning something more like a graded space so that outer degeneracies can be handled. (One conceptual question is, are outer degeneracies really any different from other types of degeneracy or multiplicity, other than how they originate?)


mtfishman:
Thanks, I'll review and leave comments.

We could consider:
```julia
(A=U1(1),) × (B=U1(1),)
```
as an alternative syntax for constructing a named sector.

Also, we could have a function `sector`:
```julia
sector(U1(1), U1(1))
sector(A=U1(1), B=U1(1))
```
that constructs named or positional sectors, depending on the input.

NDTensors/src/lib/Sectors/src/abstractcategory.jl
```diff
@@ -0,0 +1,41 @@
+abstract type AbstractCategory end
+
+abstract type AbstractGroup <: AbstractCategory end
```
mtfishman, line 3: The `AbstractGroup` doesn't seem very useful, it doesn't look like there are any functions overloaded for `AbstractGroup`.

NDTensors/src/lib/Sectors/src/abstractcategory.jl
```diff
@@ -0,0 +1,41 @@
...
+
+label(c::AbstractCategory) = error("method `label` not defined for type $(typeof(c))")
+
+function dimension(c::AbstractCategory)
+  return error("method `dimension` not defined for category or group $(typeof(c))")
+end
+
+function fusion_rule(::Type{C}, l1, l2) where {C<:AbstractCategory}
+  return error("fusion_rule not defined for type $(C)")
+end
```
mtfishman, line 13: Is this version needed? I was just picturing that types would overload `fusion_rule(c1::AbstractCategory, c2::AbstractCategory)`.

NDTensors/src/lib/Sectors/src/abstractcategory.jl
```diff
@@ -0,0 +1,41 @@
...
+function Base.show(io::IO, q::Vector{<:AbstractCategory})
+  isempty(q) && return nothing
+  symbol = ""
+  for l in q
+    print(io, symbol, l)
+    symbol = " ⊕ "
+  end
+end
+
+function trivial(::Type{C}) where {C<:AbstractCategory}
```
mtfishman, line 37: Can this just be defined on instances, i.e. `trivial(::AbstractCategory)`?

NDTensors/src/lib/Sectors/src/abstractnamedset.jl
```diff
@@ -0,0 +1,31 @@
...
+end
+
+function Base.symdiff(ns1::AbstractNamedSet, ns2::AbstractNamedSet)
+  ndata = merge(
+    Base.structdiff(data(ns1), data(ns2)), Base.structdiff(data(ns2), data(ns1))
+  )
+  return typeof(ns1)(ndata)
+end
+
+Base.length(ns::AbstractNamedSet) = length(data(ns))
```
mtfishman, line 31: Can these just be defined on `NamedTuple`, with names like `nt_union`? `AbstractNamedSet` could be useful but I would prefer to avoid introducing extra types like that for now.

NDTensors/src/lib/Sectors/src/abstractcategory.jl
mtfishman: Never mind, I think this makes sense, I see you forward `istrivial(::AbstractCategory)` to this one so either one can be used.

mtfishman: However, I have preferred the style:
```julia
function trivial(categorytype::Type{<:AbstractCategory})
  return error("method `trivial` not defined for type $categorytype")
end
```

NDTensors/src/lib/Sectors/src/category_definitions.jl
```diff
@@ -0,0 +1,183 @@
...
+  n::Half{Int}
+end
+
+label(u::U1) = u.n
+
+dimension(::U1) = 1
+
+trivial(::Type{U1}) = U1(0)
+
+fusion_rule(::Type{U1}, n1, n2) = (n1 + n2,)
```
mtfishman, line 17: I think it would be better to split group/category definitions into different files.

NDTensors/src/lib/Sectors/src/ordered_sector.jl
```diff
@@ -0,0 +1,42 @@
+struct OrderedSector <: AbstractCategory
+  data::Vector{AbstractCategory}
+end
```
mtfishman, line 3: I was picturing a design like:
```julia
struct Sector{Data} <: AbstractCategory
  data::Data
end
```
where then it is positional if `Data <: Tuple` and named if `Data <: NamedTuple`. Not sure if that is too naive or has some downsides.

The idea would be that most functions are defined that forward to `data`, and then there are some inner functions defined on either `Tuple` or `NamedTuple` that implement some functionality like fusion.

mtfishman: So then, there would be some sort of minimal required interface for what could be used as `data` inside a `Sector`, and precocious users could stick in anything they want as `data` as long as it implements some minimal set of functions.

emstoudenmire:
Helpful comments. I was wondering many of the same things about the design. Let me review each one and I'll probably go ahead with many, then we can discuss a few which might not work or how the overall design is looking after.

NDTensors/src/lib/Sectors/src/abstractcategory.jl
emstoudenmire: Agreed, I put it in there mostly as a discussion point. While I can envision that some things may only be defined for groups, and not for categories, my sense is that these finer differences can just be kind of 'emergent' and more of like a 'duck typing' thing for each category.

mtfishman: I would vote for removing it for now just to keep the code simpler, and adding it back if needed.

NDTensors/src/lib/Sectors/src/abstractcategory.jl
emstoudenmire: This one would be good to discuss a bit. The reason for having this (Type, value1, value2) sort of interface is so that for the simpler fusion rules, like U1, there can be a very 'clean' and minimal implementation like e.g. for U1 and SU2:
```
fusion_rule(::Type{U1}, n1, n2) = (n1 + n2,)

fusion_rule(::Type{SU2}, j1, j2) = abs(j1 - j2):(j1 + j2)
```
where the reason for the Tuple in the first example is that the return value is that it needs to be something iterable (like a Tuple, UnitRange, etc.).

Otherwise the returned value needs to be a Vector of the same category type. So the above examples would become:
```
fusion_rule(a::U1, b::U1) = [U1(label(a)+label(b))]

function fusion_rule(s1::SU2, s2::SU2) 
  j1, j2 = label(s1), label(s2)
  return [SU2(J) for J in abs(j1-j2):(j1+j2)]
end
```

To be clear I am fine with just having the second type! An advantage is it would make it clearer what is needed to overload `fusion_rule` because each type would do it the same way. Also it might be slightly faster in some cases, which could matter when these get called thousands of times.

emstoudenmire: One other very important thing we'll need to think through more is whether *either* of the above is actually enough, when "outer degeneracies" can occur. I can currently see three ways of dealing with those:

* Just repeat irreps that occur more than once. Then the outer degeneracy is just how many times an irrep repeats.
* Return the outer degeneracy along with the new irreps (fusion results) in some way, like by adding a degeneracy count inside the irrep (probably not) or returning pairs of `irrep => int` (I doubt this is a good idea though)
* Maybe the best way is to embrace the idea of fusion as acting not on irreps, but on graded spaces where in the math literature the spaces are usually assumed to have a degeneracy of 1. Then the fusion could return another graded space where now some of the blocks have degeneracies that can be larger than 1, i.e. outer degeneracies.

mtfishman: Gotcha, thanks for the explanation.

Another possibility would be something in between, like:
```julia
label_fusion_rule(a::U1, b::U1) = label(a) + label(b)

function label_fusion_rule(s1::SU2, s2::SU2)
  j1, j2 = label(s1), label(s2)
  return abs(j1-j2):(j1+j2)
end
```
But I see why you designed it the way you did, since then at that point you may as well extract the label for the user (and I guess also the categories should always be the same if they are being fused, if so maybe that should be checked?). I would vote for calling the inner function defined on the labels something like `label_fusion_rule` to make it clear that it is fusing the labels.

Also interesting point about outer degeneracy. Maybe returning pairs isn't the worst idea? Though I agree at that point it may be better just to go fully with graded spaces. Perhaps the code could be designed such that if you return something that isn't a pair or graded space it converts it to a graded space with degeneracy of 1 automatically but I don't have a good picture in mind for how well that would work in practice.

emstoudenmire: I like the `label_fusion_rule` idea. Looking back at the regular `fusion_rule` implementations, I guess they aren't so bad i.e. still pretty readable.

About the outer degeneracies, I'd be fine to sort of "punt" on it a bit and do something like return pairs for now which might end up being the best design. But then if it's not, I wouldn't yet view this PR, even if merged, as somehow fully committing us to the final overall design. We can always see how the whole fusion arrays system looks and revisit parts of it if needed in a future PR.

mtfishman: Makes sense to me. I would focus on keeping the design here both general but also simple, so as long as we hit on both of those (and getting the code style right) I'm fine leaving some more advanced features like outer degeneracy for future PRs. Some of this will require trying things out in the wild, having Olivier test them out, etc. and iterating based on that.

NDTensors/src/lib/Sectors/src/abstractcategory.jl
emstoudenmire: I think that's a better style too, and originally had it, but I put back the placeholder name `C` just so I could print it out as part of the error message.

NDTensors/src/lib/Sectors/src/abstractnamedset.jl
emstoudenmire: Done

NDTensors/src/lib/Sectors/src/category_definitions.jl
emstoudenmire: Done

NDTensors/src/lib/Sectors/src/ordered_sector.jl
emstoudenmire: I did this merging of types and got it to work well. Two good things happened:
1. It's no more code than before, and probably slightly less
2. Now both modes of Sector are a subtype of AbstractCategory.

But so far it's been a bit of a failure in terms of really forwarding the behavior to the data (except for some simple things like `length`). One issue preventing this is that iteration over NamedTuples is both kind of strange by default (doesn't give the name info, so I had to work around that slightly) but more fundamentally, I think there has to be some logic about propagating the names forward in the named case for the fusion_rule. So then it prevents such code from working generically for the ordered case where there are no names.

Possibly this could be improved, though, if we introduce a kind of intermediate type or concept like a `Pair{Symbol,<:AbstractCategory}` and also define `fusion_rule` on this and have that handle the name-peeling-off-and-back-on logic at a different level.

NDTensors/src/lib/Sectors/src/abstractcategory.jl
emstoudenmire: Great

emstoudenmire:
I implemented the `×` interface for making named `Sector`.  I also made `Sector` constructors that can work like:
```julia
Sector(A=U1(1), B=U1(2))
Sector(U1(1), U1(2))
```
making named and ordered Sectors, respectively.

I can still make lowercase `sector` functions you described, but would these be different from the `Sector` constructors shown just above?

Also I still need to think about if the implementation of things like `fusion_rule` for Sector can be done more generically by forwarding some things to the data or the type one gets when iterating the data. There are some tensions there such as about whether set logic and ordered logic can be made compatible too, but perhaps if the right set logic is written with the arguments in the correct order it will work for the ordered case too? I'll have to see.

NDTensors/src/lib/Sectors/src/sector.jl
```diff
@@ -0,0 +1,128 @@
+
+struct Sector{Data} <: AbstractCategory
+  data::Data
+end
+
+Sector(nt::NamedTuple) = Sector{NamedTuple}(nt_sort(nt))
```
mtfishman, line 6: Using `Sector{NamedTuple}` means the compiler doesn't have all of the type information about `data` anymore, so accessing `data` is type unstable and can hurt performance. Better to do something like:
```julia
function Sector(nt::NamedTuple)
  data = nt_sort(nt)
  return Sector{typeof(data)}(data)
end
```
or define an inner constructor `_Sector` that you call.

NDTensors/src/lib/Sectors/src/sector.jl
```diff
@@ -0,0 +1,128 @@
...
+
+Sector(; kws...) = Sector((; kws...))
+
+function Sector(pairs::Pair...)
+  keys = ntuple(n -> Symbol(pairs[n][1]), length(pairs))
+  vals = ntuple(n -> pairs[n][2], length(pairs))
+  return Sector(NamedTuple{keys}(vals))
+end
+
+Sector(cats::AbstractCategory...) = Sector([cats...])
```
mtfishman, line 16: I was picturing we would use a tuple instead of a `Vector` for performance, especially because the `cats` may not all be the same type.

NDTensors/src/lib/Sectors/test/test_sector.jl
```diff
@@ -0,0 +1,147 @@
...
+    @test s[:A] == U1(1)
+    @test s[:B] == SU2(2)
+
+    s = s × (C=Ising("ψ"),)
+    @test length(s) == 3
+    @test s[:C] == Ising("ψ")
+  end
+
+  @testset "Construct from × of Pairs" begin
+    s = ("A" => U1(1)) × ("B" => SU2(2))
```
mtfishman, line 17: Is the version using `Pair` needed if we have the `NamedTuple` syntax? Just trying to be conservative about the interface for now.

NDTensors/src/lib/Sectors/src/ordered_sector.jl
mtfishman: Right, so basically you were hoping to share more code across the definitions of `fusion_rule(A::NamedSector, B::NamedSector)` and `fusion_rule(o1::OrderedSector, o2::OrderedSector)`? I wasn't expecting there would be much code sharing there because of the very different semantics of `Tuple`/`Vector` and `NamedTuple`.

A slight side note, but one reason I like [Dictionaries.jl](https://github.com/andyferris/Dictionaries.jl) so much is that it defines a dictionary interface which is more closely related to a vector interface in terms of iteration and other operations, so it makes it easier to write code that is generic for both `Vector` and `Dictionary`. It would be nice if there was a `NamedTuple` version of that so it would be easier to right generic code that works for both `Tuple` and `NamedTuple`.

One possibility would be to have a design like:
```julia
fusion_rule(s1::Sector, s2::Sector) = data_fusion_rule(data(s1), data(s2))

data_fusion_rule(s1::Tuple, s2::Tuple) = positional_fusion_rule(s1, s2)
function positional_fusion_rule(s1, s2)
  # Implementation for positional data
end

data_fusion_rule(s1::NamedTuple, s2::NamedTuple) = named_fusion_rule(s1, s2)
function named_fusion_rule(s1, s2)
  # Implementation for named data
end
```
This is kind of a trait-like design/interface.

We could go all-in on a trait-based design with something like this:
```julia
fusion_rule(s1::Sector, s2::Sector) = data_fusion_rule(data(s1), data(s2))

data_fusion_rule(d1, d2) = data_fusion_rule(fusion_style(d1, d2), d1, d2)

abstract type AbstractFusionStyle end
struct PositionalFusion <: AbstractFusionStyle end
struct NamedFusion <: AbstractFusionStyle end

fusion_style(::Tuple) = PositionalFusion()
fusion_style(::NamedTuple) = NamedFusion()

fusion_style(d1, d2) = promote_fusion_style(fusion_style(d1), fusion_style(d2))
promote_fusion_style(s1::AbstractFusionStyle, s2::AbstractFusionStyle) = error("Not implemented")
promote_fusion_style(s1::S, s2::S) where {S <: AbstractFusionStyle} = s1

function data_fusion_rule(::PositionalFusion, d1, d2)
  # Implementation of fusion for positional data
end

function data_fusion_rule(::NamedFusion, d1, d2)
  # Implementation of fusion for named data
end
```
This could allow for some customization of fusion rules when mixing different data types, and then help extend to custom data types like `Dict`/`Dictionary`, etc. It would also allow us to not hard-code definitions of `NamedSector` and `OrderedSector`. It could also be used in other places to provide more fine-grained control of definitions which need to be customized based on the data type, again without using `NamedSector` and `OrderedSector` explicitly, but with the advantage that code can be shared across data structures that do have common interfaces, like `Vector` and `Tuple`.

mtfishman: I'm going back and forth in my head about that trait design or a design like this:
```julia
struct PositionalSector{Data} <: AbstractSector
  data::Data
end

struct NamedSector{Data} <: AbstractSector
  data::Data
end
```
and then functionality can just be customized on the `AbstractSector` type, but there could still be flexibility about the `Data` type, where as long as the `data` defined certain functionality, it should "just work", and additionally common functionality like `length` could be defined on `AbstractSector`. However I think it may still be helpful to have some kind of trait system to customize functionality on the `data` since not all `data` will "just work" as a backend for `PositionalSector` or `NamedSector` anyway, so it's not clear to me that design has a big advantage over the trait design I proposed above.

Another design would be:
```julia
struct TupleSector{Data<:Tuple} <: AbstractSector
  data::Data
end

struct NamedTupleSector{Data<:NamedTuple} <: AbstractSector
  data::Data
end

struct VectorSector{Data<:AbstractVector} <: AbstractSector
  data::Data
end

struct DictSector{Data<:AbstractDict} <: AbstractSector
  data::Data
end
```
so then you get even more fine-grained control, but the issue with that for me is that some sector types might share implementations for fusion (like `TupleSector` and `VectorSector`), so then you are left either with making definition on unions of `AbstractSector` to share code or you are back to a trait-like system anyway, and the different `AbstractSector` subtypes aren't so useful since they are really just reflecting the type of the `data` which can be accessed through a trait.

mtfishman:
> I can still make lowercase `sector` functions you described, but would these be different from the `Sector` constructors shown just above?

I only suggested that in the case where we went with a design where there were multiple sector types, such as `NamedSector` and `PositionalSector`, and we wanted a common constructor for either one depending on the inputs. With the current design it doesn't seem necessary.

NDTensors/src/lib/Sectors/test/test_sector.jl
emstoudenmire: No problem to remove this –

emstoudenmire:
Got it. Yes after merging the `Sector` types, it's pretty clear that they are just semantically different, so most lkely need to be different types. However, I can see your point about the actual data type inside still being parameterized and generic (i.e. could be reasons down the road to use a Vector instead of a Tuple in the ordered one).

mtfishman:
> Got it. Yes after merging the `Sector` types, it's pretty clear that they are just semantically different, so most lkely need to be different types. However, I can see your point about the actual data type inside still being parameterized and generic (i.e. could be reasons down the road to use a Vector instead of a Tuple in the ordered one).

I'm not so sure about needing different sector types, what about the trait system I proposed above? One way to think about that design is that `Sector` is just a wrapper around some collection of `Category` (`Sector` is just a wrapper telling us that the collection is a collection of `Category`), and we use traits as a way to decide how to work with those collections if they need some special treatment.

One issue I have with having multiple wrappers is that it seems like we need many different wrappers for each possible collection we might store anyway (unless they have the same implementations of fusion and other operations), and then some system for how those different sector types may interact with each other, and we get back to some kind of trait system anyway. I don't think the choice is totally obvious, but the trait approach feels right to me in this case.

emstoudenmire:
Yes, thinking about it some more, I was initially gravitating toward the separate types because it seemed "simpler" and more true to the reality that they function mostly differently. 

But I can see the advantages of a trait design, mainly since it would result in fewer (wrapper) types overall (i.e. just one parameterized type).

I think this plan is easy enough to implement, especially if I start with the simpler version that doesn't involve trait type promotions (do you think that's important in this case?). 

The main awkward part I can anticipate is that I had defined iteration over the named Sector type as giving Pair{Symbol,AbstractCategory} values throughout the iteration, whereas the ordered Sector will just give category values. Probably that's ok for the iteration to behave differently depending on the storage but it does seem slightly strange.

mtfishman:
I don't think the choice of traits vs. wrappers is obvious, but I think it's worth thinking through how either one might look.

We can definitely keep the trait implementation simpler for now and not implement a promotion system, the code I showed above just handles the case where the data is the same but anticipates a design where they might be different.

I think we'll have to go through each operation one a time and try to come up with definitions that are reasonable across various data backends, and then specialize them as needed (we would have to do that with traits vs. wrappers anyway, traits vs. wrappers would just determine how we dispatch for the specialization).

For iteration, I have a few thoughts on that:

1. When is iteration being used? That can help determine what it should output.
2. I think [Dictionaries.jl](https://github.com/andyferris/Dictionaries.jl) made a nice choice that iteration over a `Dictionary` iterates over just the values, not key-value pairs like `Base.Dict` does. To iterate over the keys, you can use `keys` or `eachindex`, which matches more closely with vector-like collections, and `pairs` is used to iterate over key-value pairs. We could follow the same style here. In fact, that's the choice that was made for `NamedTuple` already:
```julia
julia> eachindex((; a=1, b=2))
(:a, :b)

julia> keys((; a=1, b=2))
(:a, :b)

julia> values((; a=1, b=2))
(1, 2)

julia> [x for x in (; a=1, b=2)]
2-element Vector{Int64}:
 1
 2
```

`Dict` seems to be the only odd one where iteration iterates over pairs of keys and values, which I think was a mistake since it makes it harder to write generic code that could accept either `Dict` or other data structures:
```julia
julia> eachindex(Dict([:a => 1, :b => 2]))
KeySet for a Dict{Symbol, Int64} with 2 entries. Keys:
  :a
  :b

julia> keys(Dict([:a => 1, :b => 2]))
KeySet for a Dict{Symbol, Int64} with 2 entries. Keys:
  :a
  :b

julia> values(Dict([:a => 1, :b => 2]))
ValueIterator for a Dict{Symbol, Int64} with 2 entries. Values:
  1
  2

julia> [x for x in Dict([:a => 1, :b => 2])]
2-element Vector{Pair{Symbol, Int64}}:
 :a => 1
 :b => 2
```

This is the behavior for `Dictionary`:
```julia
julia> using Dictionaries

julia> eachindex(Dictionary([:a, :b], [1, 2]))
2-element Indices{Symbol}
 :a
 :b

julia> keys(Dictionary([:a, :b], [1, 2]))
2-element Indices{Symbol}
 :a
 :b

julia> values(Dictionary([:a, :b], [1, 2]))
2-element Dictionary{Symbol, Int64}
 :a │ 1
 :b │ 2

julia> [x for x in Dictionary([:a, :b], [1, 2])]
2-element Vector{Int64}:
 1
 2
```
Note that `values` returns the `Dictionary` itself, but that already acts as a collection of the values of the dictionary for the sake of iteration as exemplified by the last example.

mtfishman:
If we don't want to use traits just yet, one suggestion I would have for the current design is to separate the implementation of fusion into two layers, and not define type aliases `NamedSector` and `OrderedSector`. For example:
```julia
function fusion_rule(s1::Sector, s2::Sector)
  return Sector(data_fusion_rule(data(s1), data(s2)))
end

data_fusion_rule(d1::Tuple, d2::Tuple)
  # Implementation
end

data_fusion_rule(d1::NamedTuple, d2::NamedTuple)
  # Implementation
end
```
and same for other functions that are specialized right now on `NamedSector` or `OrderedSector`, like `==`. This will help anticipate other data backends, and I also think is better code design (it makes it clearer that `Sector` is just a wrapper, and implementations for certain data backends are performed in a separate layer of code).

Then, you can remove the definitions:
```julia
Base.intersect(s1::NamedSector, s2::NamedSector) = Sector(nt_intersect(data(s1), data(s2)))
Base.symdiff(s1::NamedSector, s2::NamedSector) = Sector(nt_symdiff(data(s1), data(s2)))
Base.union(s1::NamedSector, s2::NamedSector) = Sector(nt_union(data(s1), data(s2)))
```
which I think then are no longer necessary.

We could also choose something more descriptive than `data` for the data of a `Sector`, like `categories`, which implies it should be some collection of `Category`. Then the inner functions could be `categories_fusion_rule` instead of `data_fusion_rule`. Curious about your thoughts on that or maybe you have another suggestion for that. Additionally, it could be called `category_product` or something like that (in fact, `CategoryProduct` may be a more descriptive name than `Sector`, and we could hide that name behind construct functions like `sector` or `×`).

mtfishman:
In `TensorKit`, they call this `ProductSector`, which is a subtype of an abstract type `Sector`:
```julia
julia> using TensorKit

julia> typeof(Irrep[U₁](0) ⊠ Irrep[U₁](1))
TensorKit.ProductSector{Tuple{U1Irrep, U1Irrep}}

julia> Irrep[U₁](0) ⊠ Irrep[U₁](1) isa Sector
true
```

emstoudenmire:
Interesting about TensorKit. Let's discuss the names some more.

The new code makes the following changes:
- rename `Sector` to `CategoryProduct` (I like that name) though still using `Sector` as an alias sometimes for now
- rename data inside to `labels::Labels`. Though I also use labels to refer to the value inside of a category (like n or j) so maybe it's a confusing double usage?
- reimplement `CategoryProduct` as a wrapper type with most functionality through dispatch on the `Labels` types
- improve some printing things, like of `CategoryProduct`'s or direct sums of these containing zero or one elements

NDTensors/src/lib/Sectors/src/category_product.jl
```diff
@@ -0,0 +1,107 @@
+
+struct CategoryProduct{Labels} <: AbstractCategory
+  labels::Labels
```
mtfishman, line 3: What about `categories::Categories`?

NDTensors/src/lib/Sectors/src/namedtuple_operations.jl
```diff
@@ -0,0 +1,16 @@
...
+  return :(NamedTuple{$(Tuple(intersect(N1, N2)))}(merge(nt2, nt1)))
+end
+
+nt_union(ns1::NamedTuple, ns2::NamedTuple) = Base.merge(ns2, ns1)
+
+nt_setdiff(ns1::NamedTuple, ns2::NamedTuple) = Base.structdiff(ns1, ns2)
+
+function nt_symdiff(ns1::NamedTuple, ns2::NamedTuple)
+  return merge(Base.structdiff(ns1, ns2), Base.structdiff(ns2, ns1))
+end
```
mtfishman, line 16: I think a better name for these would be `sort_keys`, `intersect_keys`, `union_keys`, etc. (then they would also generalize to `Dict` or `Dictionary`).

NDTensors/src/lib/Sectors/src/category_product.jl
```diff
@@ -0,0 +1,107 @@
...
+Base.length(S::CategoryProduct) = length(labels(S))
+Base.getindex(S::CategoryProduct, args...) = getindex(labels(S), args...)
+
+function fusion_rule(s1::CategoryProduct, s2::CategoryProduct)
+  return [CategoryProduct(l) for l in labels_fusion_rule(labels(s1), labels(s2))]
+end
+
+Base.:(==)(A::CategoryProduct, B::CategoryProduct) = labels_equal(labels(A), labels(B))
+
+const Sector = CategoryProduct
```
mtfishman, line 21: I would vote for defining a `sector` function and using that instead of `Sector`.

NDTensors/src/lib/Sectors/src/category_product.jl
emstoudenmire: I'm good with that

NDTensors/src/lib/Sectors/src/namedtuple_operations.jl
emstoudenmire: Yes I had wanted to discuss these

NDTensors/src/lib/Sectors/src/category_product.jl
```diff
@@ -0,0 +1,107 @@
...
+
+CategoryProduct(; kws...) = CategoryProduct((; kws...))
+
+function CategoryProduct(pairs::Pair...)
+  keys = ntuple(n -> Symbol(pairs[n][1]), length(pairs))
+  vals = ntuple(n -> pairs[n][2], length(pairs))
+  return CategoryProduct(NamedTuple{keys}(vals))
+end
+
+function labels_fusion_rule(A::NamedTuple, B::NamedTuple)
```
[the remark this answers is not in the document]
emstoudenmire: I went with `categories_fusion_rule`. Does that sound good to you?

emstoudenmire:
One thing we will need to eventually consider (probably in a separate PR) is the interplay between this system and the existing `QN` system. Perhaps the way to go is to, at some point, redefine the `QN` methods to just be functions that construct `CategoryProduct`?

NDTensors/src/lib/Sectors/src/category_product.jl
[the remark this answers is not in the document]
mtfishman: Sounds good to me.

mtfishman:
> One thing we will need to eventually consider (probably in a separate PR) is the interplay between this system and the existing `QN` system. Perhaps the way to go is to, at some point, redefine the `QN` methods to just be functions that construct `CategoryProduct`?

Yeah I've been wondering about that too. One thing I've been considering is to at first only start using this new system inside of ITensorNetworks, and eventually just deprecate the old MPS code using the old QN type, but we'd have to think through the details of that. But I think we're pretty far from that point anyway, since for now this will only be used inside the new NamedDimsArray/BlockSparseArray design and in the new fusion tensor code, which won't get exposed to users for a while.

mtfishman:
This looks good to go from my end, ready to merge once tests pass?

emstoudenmire:
Sounds like a plan. Yes I'm good to merge this.

It's remarkable how short all this code became in the end! Compared to QN and the earlier versions of this PR, shows the power of iterating code design. (Well and also it's fortunate that `NamedTuple` and `Tuple` are simultaneously fast and can handle heterogeneous types.)

PR #1299, opened 2024-01-03
[NDTensors][ITensors] Fix issues caused by Julia 1.10

kmp5VT:
# Description

Minor number of errors thrown as a result of julia's update to release v1.10. This PR attempts to fix the problems.

# Checklist:

- [x] Number of allocations changed in `SmallVectors` code so loosen the threshold 
- [x] Some broadcast functions now return a result but the result is wrong. Check for operation and throw an error


NDTensors/src/lib/SmallVectors/src/abstractsmallvector/deque.jl
```diff
@@ -311,7 +311,7 @@ function prepend(vec::AbstractSmallVector, item::AbstractVector)
 end
 
 # Don't @inline, makes it slower.
-function Base.vcat(vec1::AbstractSmallVector, vec2::AbstractVector)
+function Base.vcat(vec1::AbstractSmallVector{<:Number}, vec2::AbstractVector{<:Number})
```
mtfishman, line 314: This doesn't seem great since we generally will not only have `Number`s inside of a `SmallVector`.

Something we can do is:
```julia
function smallvector_vcat(vec1::AbstractSmallVector, vec2::AbstractVector)
  # Current definition
end

Base.vcat(vec1::AbstractSmallVector{<:Number}, vec2::AbstractVector{<:Number}) = smallvector_vcat(vec1, vec2)

Base.vcat(vec1::AbstractSmallVector, vec2::AbstractVector) = smallvector_vcat(vec1, vec2)
```

mtfishman:
Looks good, thanks for the quick fix!

PR #1301, opened 2024-01-05
[TensorAlgebra] `BlockSparseArray` tensor contraction

mtfishman:
This is work in progress towards getting `BlockSparseArray` tensor contraction working (for now, independent of symmetries).

The design right now is that `BlockSparseArray` contraction is implemented through the generic matricized contract code in `TensorAlgebra`. This relies on having proper implementations of permutation, generalized reshape (which I call `fusedims` and `splitdims` in `TensorAlgebra`), and matrix multiplication, so the actual implementation of block sparse array contraction mostly relies on the generic one in `TensorAlgebra` and relies on those operations being defined.

The current matricized `contract` code already works for other array types like `DiagonalArray` and `SparseArrayDOK` since they implement the required permutation, reshape, and matrix multiplication functionality.

This will generalize to symmetric tensors with even more specalized implementations of `fusedims` and `splitdims` that actually fuse and split the blocks according to symmetry sector fusion, but that will be left for future work and will rely on the work in #1296 (though we could start testing it without that by making simple abelian group irrep types and using `GradedAxes` as the axes of a `BlockSparseArray`).


emstoudenmire:
I'll be really interested to see how the performance turns out when using the generic matricized contract. It would be great if it's already quite good in many cases. What do you expect or do you already have some timings?

mtfishman:
> I'll be really interested to see how the performance turns out when using the generic matricized contract. It would be great if it's already quite good in many cases. What do you expect or do you already have some timings?

Yeah I'm curious about that too. I think I could argue that if written efficiently, matricized contraction is either about the same as what we are doing currently if no QN merging is done (it requires the same tensor permutations of each block of the input tensors, and then once those permutations are done it requires the same matrix multiplication), and then could be faster if QN merging is done (depending on the particulars of the symmetries, sparsity pattern, tensor order, etc.). Currently I'm focusing more on code organization, generic interfaces, and correctness, but once the code structure is in place and working the next step would be to do benchmarking/profiling and optimization.

NDTensors/src/lib/BlockSparseArrays/src/blocksparsearrayinterface/blocksparsearrayinterface.jl
```diff
@@ -45,10 +45,20 @@ function block_nstored(a::AbstractArray)
...
+
+## function blocksparse_copyto!(a_dest::AbstractArray, a_src::AbstractArray)
+##   blocksparse_map!(identity, a_dest, a_src)
+##   return a_dest
+## end
+## 
+## function blocksparse_permutedims!(a_dest, a_src::AbstractArray, perm)
+##   blocksparse_map!(identity, a_dest, PermutedDimsArray(a_src, perm))
+##   return a_dest
+## end
```
mtfishman, line 61:
```suggestion
```

NDTensors/src/lib/TensorAlgebra/src/fusedims.jl
```diff
@@ -1,21 +1,5 @@
...
-  a_dest[] = a_src[]
-  return a_dest
-end
-function _permutedims(a::AbstractArray{<:Any,N}, perm::Tuple{Vararg{Int,N}}) where {N}
-  return permutedims(a, perm)
-end
-function _permutedims(a::AbstractArray{<:Any,0}, perm::Tuple{})
-  return copy(a)
-end
+# include("BaseExtensions/BaseExtensions.jl")
```
mtfishman, line 1:
```suggestion
```

NDTensors/src/lib/SparseArrayInterface/src/sparsearrayinterface/interface.jl
```diff
@@ -3,8 +3,10 @@
 
 # Minimal interface
 # Data structure of the stored (generally nonzero) values
-# nonzeros(a::AbstractArray) = a
-sparse_storage(a::AbstractArray) = a
+# nonzeros(a::AbstractArray) = error("Not implemented")
```
mtfishman, line 6:
```suggestion
```

PR #1214, opened 2023-10-18
Fix behavior of factorize mindim

ryanlevy:
# Description

Fixes #1207 

<details><summary>Minimal demonstration of previous behavior</summary><p>

> Quick update, a more minimal example. Say we have a two-site tensor that we want to factorize
> 
> ```julia
> s  = Index(2)
> s′ = Index(2)
> l = Index(64)
> r = Index(16)
> T = randomITensor(l,s,s′,r)
> ```
> 
> The inconsistent behavior manifests as
> 
> ```julia
> > L,R,spec = factorize(T,l,s; cutoff=0,mindim=64)
> > @show inds(L)
> inds(L) = ((dim=64|id=719), (dim=2|id=499), (dim=32|id=847|"Link,fact"))
> ```
> 
> where we get the 64->32->16 stepdown. Or with cutoff
> 
> ```julia
> > L,R,spec = factorize(T,l,s; cutoff=1e-10,mindim=64)
> > @show inds(L)
> inds(L) = ((dim=64|id=444), (dim=2|id=941), (dim=64|id=961|"Link,fact"))
> ```
> 
> Where the factorize respects the mindim by padding the link dimension. I think the `cutoff=0` behavior is best here.

</p></details>

<details><summary>Minimal demonstration of new behavior</summary><p>

```julia
using ITensors

s  = Index(2)
s′ = Index(2)
l = Index(64)
r = Index(16)
T = randomITensor(l,s,s′,r)

L,R,spec = factorize(T,l,s; cutoff=1e-10,maxdim=64,mindim=64)
@show inds(L)
@show inds(R)
```

```julia
> inds(L) = ((dim=64|id=96), (dim=2|id=968), (dim=32|id=502|"Link,fact"))
> inds(R) = ((dim=32|id=502|"Link,fact"), (dim=2|id=814), (dim=16|id=189))

```
</p></details>

# How Has This Been Tested?

No new tests

# Checklist:

- [x] My code follows the style guidelines of this project. Please run `using JuliaFormatter; format(".")` in the base directory of the repository (`~/.julia/dev/ITensors`) to format your code according to our style guidelines.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have added tests that verify the behavior of the changes I made.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] Any dependent changes have been merged and published in downstream modules.

PR #1302, opened 2024-01-10
[ITensors] Fix `eigen_perturbation` rank

mtfishman:
It appears that #1214 may have caused an issue when non-trivial `eigen_perturbation` is passed to `factorize`, where it is too aggressive about setting an upper bound on the rank of the factorization. This appears to cause issues for the noise term in DMRG, see https://itensor.discourse.group/t/revert-back-to-older-version-of-itensors-jl/1366.

Really this logic should not be in `factorize` since it adds this extra complexity and breaks the abstraction of `factorize` that it is doing a factorization of the input tensor, but disentangling it will be left for future work.

I need to add a minimal test once we confirm this fixes the issue with the noise term.


mtfishman:
[test ITensors mps]

PR #1306, opened 2024-01-12
[ITensorGPU] Loosen test rtol

kmp5VT:
# Description

@mtfishman This is a push to reduce the tightness of the rtol that we saw fail on the `UnallocatedArrays` branch.  In general the `isapprox` function has a default `rtol` of `sqrt(eps(typeof(real(x-y))))` which I assume here is `sqrt(eps(Float64)) = 1.4901161193847656e-8` So I am not sure why the tests were failing in that previous PR because the absolute error was only `1e-10`. I loosened the rtol more by multiplying the default by 10. Let me know if you have a better solution, thanks!


ITensorGPU/test/test_dmrg.jl
```diff
@@ -93,7 +93,7 @@ end
     dmrg_params = (; nsweeps=4, maxdim=10, cutoff=1e-8, noise=1e-8, outputlevel=0)
     observer = DMRGObserver(["Z"], s; energy_tol=1e-4, minsweeps=10)
     E, ψ = dmrg(H, ψ0; observer=observer, dmrg_params...)
-    @test expect(ψ, "Z") ≈ observer.measurements["Z"][end]
+    @test expect(ψ, "Z") ≈ observer.measurements["Z"][end] rtol = sqrt(eps(eltype(ψ0[1])))
```
mtfishman, line 96: You can use `ITensors.scalartype(ψ0)` to get the scalar type of an MPS, a bit more general than just basing it on the first ITensor. Also it would be more general to take the real part, in case we start testing complex element types here.

kmp5VT: Ahh, thanks I didn't know about that function. I changed the code to `real(scalartype(psi0))` and I added the `10 *` back, somehow that was removed before I pushed to the branch.  Thanks!

mtfishman:
Thanks!

PR #1307, opened 2024-01-12
[ITensors] Add `checkflux` debug check to `factorize`

emstoudenmire:
Adds a call to `checkflux` at the beginning of `factorize` when debug checks are enabled.

Related to [this forum discussion](https://itensor.discourse.group/t/bug-in-factorize-qr-leading-to-very-slow-calls-to-orthogonalize/1412/9).

### DIFF
--- a/src/tensor_operations/matrix_decomposition.jl
+++ b/src/tensor_operations/matrix_decomposition.jl
@@ -763,6 +763,9 @@ function factorize(
   (singular_values!)=nothing,
   dir=nothing,
 )
+  @debug_check begin
+    checkflux(A)
+  end
   if !isnothing(eigen_perturbation)
     if !(isnothing(which_decomp) || which_decomp == "eigen")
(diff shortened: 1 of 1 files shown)


emstoudenmire:
@mtfishman please let me know if you think this is worth adding a unit test for.

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -763,6 +763,9 @@ function factorize(
   (singular_values!)=nothing,
   dir=nothing,
 )
+  @debug_check begin
+    checkflux(A)
+  end
```
mtfishman, line 768:
```suggestion
  @debug_check checkflux(A)
```

mtfishman:
Thanks! I think it would be good to add a simple unit test.

mtfishman:
I'll just merge this, since `checkflux` is tested I don't think this needs a specialized test.

PR #1308, opened 2024-01-16
[SetParameters] Introduce `[un]specify_parameters`, `set_eltype`, etc.

kmp5VT:
# Description

I am working on refactorizing the SetParameters module. This includes renaming some the `set_x_if_unspecified` functions, adding `unspecify_parameters` functions, creating `set_x` functions and defining them on some set of types. In general I want to make sure we are utilizing these functions and there is extraneous code around the `NDTensors` library since we have rapidly gone through a few iterations of code in the last few months

# Checklist:

- [x] Rename the `set_x_if_unspecified` functions
- [x] Create `unspeicify_paramters` functions
- [x] Create tests for unspecify paramters
- [x] Ensure we are using SetParameters appropriately throughout the `NDTensors` library
- [x] Create `set_ndim` and `set_eltype` functions in `SetParameters`


mtfishman:
Looks like a good start, will be nice to have all of these improvements.

kmp5VT:
@mtfishman Should the `UnspecifiedParameter` in `SetParameters` be moved the `UnspecifiedTypes`?

mtfishman:
> @mtfishman Should the `UnspecifiedParameter` in `SetParameters` be moved the `UnspecifiedTypes`?

Let's keep it in `SetParameters` for now, it has a pretty narrow usage in `SetParameters`. Maybe we can reassess once `UnspecifiedTypes` is developed more. I wouldn't really picture most of `UnspecifiedTypes` being useful for `SetParameters` so my intuition is that it doesn't make sense to make `UnspecifiedTypes` a dependency of `SetParameters`.

NDTensors/ext/NDTensorsMetalExt/adapt.jl
```diff
@@ -7,14 +7,11 @@ end
 # More general than the version in Metal.jl
 function Adapt.adapt_storage(arraytype::Type{<:MtlArray}, xs::AbstractArray)
   # TODO: Do this in one call.
-  arraytype_specified_1 = set_unspecified_parameters(
-    arraytype, Position(1), get_parameter(xs, Position(1))
+  arraytype_specified = specify_parameters(
+    arraytype,
+    get_parameter(xs, Position(1)),
+    get_parameter(xs, Position(2)),
+    get_parameter(xs, Position(3)),
```
mtfishman, line 14: Can you use `get_parameters(xs)` here?

NDTensors/src/dense/set_types.jl
```diff
@@ -31,8 +8,9 @@ function set_datatype(storagetype::Type{<:Dense}, datatype::Type{<:AbstractArray
...
-  storagetype::Type{<:Dense{ElT,DataT}}, ElR::Type
-) where {ElT,DataT<:AbstractArray}
-  return Dense{ElR,similartype(DataT, ElR)}
-end
+## TODO move to UnspecifiedArrays
+# function set_eltype(
+#   storagetype::Type{<:Dense{ElT,DataT}}, ElR::Type
+# ) where {ElT,DataT<:AbstractArray}
+#   return Dense{ElR,similartype(DataT, ElR)}
+# end
```
mtfishman, line 16: Reminder to delete this if it is no longer needed.

NDTensors/src/lib/SetParameters/src/Base/abstractarray.jl
```diff
@@ -0,0 +1,7 @@
+function set_eltype(type::Type{<:AbstractArray}, elt::Type)
+  return set_parameter(type, Position(1), elt)
+end
+
+function set_ndims(type::Type{<:AbstractArray}, ndim::Int)
+  return set_parameter(type, Position(2), ndim)
+end
```
mtfishman, line 7: I think these are reasonable defaults, though we need to remember that if it is a wrapped array type then the wrapped array also will need to be adjusted. So we will need specialized definitions for wrapped array types like `SubArray`.

NDTensors/src/dense/set_types.jl
mtfishman: Scratch that, I think we will need a specialized definition for `Dense` since it is a wrapper type, see my comment below about the `set_eltype` and `set_ndims` definitions for `AbstractArray`. Though it seems like this definition could be simplified by making use of `set_eltype`/`set_parameters`.

NDTensors/src/lib/SetParameters/src/Base/abstractarray.jl
mtfishman: One way to handle wrapper types like `SubArray` could be to also introduce a function `set_parenttype`. Then we could have definitions like:
```julia
function set_eltype(type::Type{<:SubArray}, elt::Type)
  new_parenttype = set_eltype(parenttype(type), elt)
  return set_parenttype(set_parameter(type, Position(1), elt), new_parenttype)
end
```
which would work generically for many (maybe all) wrapped array types.

That probably requires making `Unwrap` a dependency of `SetParameters`, since that is where `parenttype` is defined and we don't want to repeat that definition here. In the end, it might make sense to move `parenttype` from `Unwrap` to `SetParameters` but to keep this PR simpler let's leave that for future work.

Probably the best/most general design of `set_eltype` would be to make it a trait function based around the trait `IsWrappedArray` from `Unwrap`, which then has one simpler definition of `set_eltype` when `!IsWrappedArray` (the current definition in this PR), and another when `IsWrappedArray` which is the definition using `parenttype` that I wrote above.

kmp5VT: @mtfishman Thank you for these ideas! I am still working/thinking about the best way to implement `set_eltype` and `set_ndims`. and I will work to include these suggestions today

NDTensors/src/blocksparse/blocksparsetensor.jl
```diff
@@ -242,7 +242,8 @@ end
 
 @propagate_inbounds function getindex(T::BlockSparseTensor{ElT,0}) where {ElT}
   nnzblocks(T) == 0 && return zero(ElT)
-  return storage(T)[]
+  return getindex(expose(storage(T)))
```
kmp5VT, line 245: @mtfishman This line was making `Metal` testing fail for NDTensors because calling `getindex(storage)` does scalar indexing. Would this line be okay to include in this PR?

mtfishman: Sure, though I would write it as `expose(storage(T))[]`.

NDTensors/src/lib/SetParameters/src/Base/abstractarray.jl
kmp5VT: @mtfishman  I have made some changes to the `set_eltype` and `set_ndims` system and mostly have deleted the functions from NDTensors. I haven't yet made a `Wrapped` based type system though I do plan to.  Right now I have simply created multiple `set_x` functions based on the type input. I am pushing my changes so I can test on CUDA (so far I have only tested using `Metal` and CPU) and will continue to work on this PR. Thanks!

NDTensors/ext/examples/Readme.md
```diff
@@ -2,6 +2,7 @@ In this file I will layout how to construct a new backend NDTensor datatype stor
 
 One should start by constructing a folder designated `NDTensor___` where ___ is replaced with the name of the datatype. In this folder one creates an `NDTensor___.jl` file. This is the only file which should link `NDTensors` to your new datatype.  In this file, one can import all necessary packages for your datatype. These packages should NOT be added to the `Project.toml`.
 
+## TODO This is no longer necessary, should it be deleted?
```
kmp5VT, line 5: As I point out in this comment, I don't think we will need to have users define the `set_eltype` and `set_ndims` functions for different backends (though I guess they will need to define the `SetParameters` functions from `interface.jl`, correct? In principle, since we have the `unset_parameters` function now across our known types, we could start implementing a compiled system to make `set_parameter` functions for different values of `nparameters` too right? After a generic system is implemented that would mean users just need to define a few functions like `nparameters`, `default_parameter` for each index value, and `unset_parameters`.

NDTensors/ext/NDTensorsMetalExt/adapt.jl
kmp5VT: I modified this to use `set_parameters(xs)` and it looks like it works properly. Thanks!

NDTensors/ext/examples/Readme.md
mtfishman: Good point, I was thinking the same thing. It does seem like we can make a generic version of `set_parameters` based on `nparameters` and `unset_parameters`.

NDTensors/src/lib/SetParameters/src/Base/abstractarray.jl
mtfishman: Sounds good, makes sense as an initial step.

NDTensors/ext/examples/Readme.md
kmp5VT: To summerize, I will make a minimal system using `parenttype_position` for the small number of types and will create a generic `set_ndim` and `set_eltype` function that uses this `parenttype_position` function. I will also revert the push `601dc652244bf06be59dd7bc0ee8cf5649640a85`

mtfishman:
@kmp5VT I think it makes sense to also define `eltype_position` and `ndims_position` and use those in `set_eltype` and `set_ndims`, to be consistent with the planned design of writing `set_parenttype` in terms of `parenttype_position`.

NDTensors/src/dense/set_types.jl
```diff
@@ -31,8 +11,18 @@ function set_datatype(storagetype::Type{<:Dense}, datatype::Type{<:AbstractArray
...
+SetParameters.get_parameter(::Type{<:Dense{P1}}, ::Position{1}) where {P1} = P1
+SetParameters.get_parameter(::Type{<:Dense{<:Any,P2}}, ::Position{2}) where {P2} = P2
+
+function SetParameters.set_parameter(t::Type{<:Dense}, ::Position{1}, P1)
+  return specify_parameters(unspecify_parameters(t){P1}, get_parameters(t))
+end
+
+function SetParameters.set_parameter(t::Type{<:Dense}, ::Position{2}, P2)
+  P1 = get_parameter(t, Position(1))
+  return specify_parameters(unspecify_parameters(t){P1,P2}, get_parameters(t))
```
mtfishman, line 27: Will these fail if any of the type parameters of `t` were unspecified to begin with, i.e. if you passed `Dense`?

If so, I'm ok with leaving it like that for now to keep things simpler, I just want to confirm my understanding. Also if that's true we should leave comments as reminders.

mtfishman:
Looks good, definitely a nice simplification. I left a few more comments but besides that I think it is ready to merge.

I think if we come up with a good way to automatically generate definitions of `get_parameter` and `set_parameter` for new types this could be a good package to make public, I could see this being used a lot.

NDTensors/src/dense/set_types.jl
kmp5VT: This was my attempt to start thinking about a generic version of the `set_parameter` function. These do fail but for two different reasons.
This kind of code `set_parameters(Dense, Position(1), Float32)` fails because of a type error 
```
ERROR: TypeError: in Dense, in DataT, expected DataT<:AbstractArray, got Type{NDTensors.SetParameters.UnspecifiedParameter}
``` 
I think this could be fixed by using the `UnspecifiedArray` type. But the problem eludes to an issue of the generic implementation such that parameters might be constrained.
The second function `set_parameters(Dense, Position(2), Vector{Float32})` fails because of a stack overflow issue. The `specify_parameter` function always starts at index `1` so if `P1 = UnspecifiedParameter` calling `specify_parameter` will create an infinite loop. I think this problem can be solved by allowing one to modify the start position of the `specify_parameters` function like this 
```
function specify_parameters(type::Type, parameters::Tuple, start_position::Int = 1)
  new_type = type
  for i in start_position:(length(parameters))
    new_type = specify_parameters(new_type, Position(i), parameters[i])
  end
  return new_type
end
```
I will add a note about these problems in the code. Thanks!

kmp5VT: Do you think it's useful to modify the `specify_parameters` function to allow the for loop to start at a value other than 1, i.e. the function I have written above?

mtfishman: That sounds like it makes sense.

Though it should be written using recursion and have a similar interface to `set_parameters` as defined here: https://github.com/ITensor/ITensors.jl/blob/v0.3.54/NDTensors/src/lib/SetParameters/src/set_parameters_generic.jl. The for-loop version you wrote probably won't be type-stable.

kmp5VT: Sounds good! I was actually able to just use the `set_parameters` function since it takes a function as its first argument so now the `specify_parameters` function looks like this 
```
function specify_parameters(
  type::Type, parameters::Tuple, starting_position::Position=Position(1)
)
  return set_parameters(specify_parameters, type, starting_position, parameters)
end
```

NDTensors/src/lib/SetParameters/src/specify_parameters.jl
```diff
@@ -28,12 +28,10 @@ function specify_parameters(type::Type, parameter...)
...
-function specify_parameters(type::Type, parameters::Tuple)
-  new_type = type
-  for i in 1:(length(parameters))
-    new_type = specify_parameters(new_type, Position(i), parameters[i])
-  end
-  return new_type
+function specify_parameters(
+  type::Type, parameters::Tuple, starting_position::Position=Position(1)
+)
+  return set_parameters(specify_parameters, type, starting_position, parameters)
```
mtfishman, line 34: I'm confused, isn't this the same functionality as the one defined on line 20 of this file?

NDTensors/src/lib/SetParameters/src/deprecated/set_unspecified_default_parameters.jl
mtfishman, line 1: Is there a reason to keep this file around?

NDTensors/src/lib/SetParameters/src/specify_parameters.jl
kmp5VT: Okay sorry let me look at this file again.

NDTensors/src/lib/SetParameters/src/deprecated/set_unspecified_default_parameters.jl
kmp5VT: It does not look like we need this file anymore, I have removed it

NDTensors/src/lib/SetParameters/src/specify_parameters.jl
kmp5VT: So it looks like this function is not necessary. I think I added it for the `Dense` `set_parameter` functions and since we have changed their definitions the function is not used/necessary. I have removed it from the PR

PR #1309, opened 2024-01-17
[NDTensors] Replace `axpby` with broadcast when β ≡ 0

kmp5VT:
axpby computes `y[i] += a * x[i] + b * y[i]` which can fail if `y[i]` is `#undef` for example in the case of `BigFloat`

 Since `iszero(\beta)` is true just compute `y[i] .= a * x[i]` which is slightly
 more efficient too.

# Description

This pr addresses the issue found in https://itensor.discourse.group/t/using-itensor-with-big-data-types/1386. The main thing is that undefined vectors of `BigFloat` don't actually contain `BigFloat` numbers so computing `axpby` on them fails.

@mtfishman Would you like me to make some kind of unittest for this example?
# Checklist:

- [x] The minimal code from the issue runs properly
- [x] All previous tests in ITensors and NDTensors still pass


mtfishman:
Thanks @kmp5VT. Should we change over the other BLAS calls as well? Seems like those could have the same issue.

I think the reason I switched over to using BLAS was that then these calls could take advantage of BLAS multithreading, but I'm not sure if it's worth it if we lose out on using more general element types. We could always add in some dispatch later where we dispatch to BLAS for `LinearAlgebra.BlasFloat` types, but we don't need to do that here.

kmp5VT:
@mtfishman One question I would ask, do we want an error to be thrown if someone tries to do something like this
```
using ITensors
i,j = Index.((3,3))
X = tensor(randomITensor(BigFloat, (i,j)))
Y = similar(X)

NDTensors._contract_scalar_noperm!(Y, X, 1., 2.0)
```
?
Right now this fails for `BigFloat` but works fine for other kind of element types. Another issue I found is that the `LinearAlgebra.scal!` function is not defined for `Vector{BigFloat}` so this operation `NDTensors._contract_scalar_noperm!(Y, X, 0., 2.0)` also throws an error.

mtfishman:
I don't see why we wouldn't want those to work with `BigFloat` as well, though we should be careful about focusing too much on the behavior of an internal function like `NDTensors._contract_scalar_noperm!` that users shouldn't be calling anyway, and instead focus on testing functionality that is part of the public API, like tensor addition and contraction.

Can't those be fixed by also switching to broadcast calls? Functions like `LinearAlgebra.axpy!`, `LinearAlgebra.scal!`, `LinearAlgebra.axpby!`, etc. are Julia's API for the corresponding BLAS functions, and BLAS backends only work for arrays with basic float types (those contained in `LinearAlgebra.BlasFloat`), so it is expected that those would fail for element types like `BigFloat`.

We should add tests for whatever functionality we change in this PR.

kmp5VT:
@mtfishman I completely agree with not focusing on the `_contract_scalar_noperm!` function to directly I'm am only doing so here to make things easier for myself.  I also agree that it is necessary for the aforementioned functions to also work on `BigFloat`. Further, I tried replacing the `LinearAlgebra` functions with broadcast calls and for the following cases the functions still fail because `Y` has values of `#undef`
```
using ITensors
i,j = Index.((3,3))
X = tensor(randomITensor(BigFloat, (i,j)))
Y = similar(X)

NDTensors._contract_scalar_noperm!(Y, X, 0, 1.0)
NDTensors._contract_scalar_noperm!(Y, X, 1.0, 1.0)
NDTensors._contract_scalar_noperm!(Y, X, 1.0, 2.0)
```
I believe these failures to be 'expected' because all of these functions attempt to operate on `Y` which is constructed with `Undefined`

kmp5VT:
I will add tests using `onehot` it looks like this is where it fails in `randomMPS` from the original issue. 
```
i, j, k = Index.((1,2,1))
M = randomITensor(BigFloat, j, i, k)
M *= onehot(elt, i => 1)
```
Also I tried to use the `permutedim!!` function `Rᵃ = permutedims!!(Rᵃ, Tᵃ, perm, (r, t) -> α * t)` by forcing the `M * onehot` to call `_contract_scalar_perm!` and this code also fails with the same reason as in my previous post, i.e. the values in the tensor constructed using similar is `#undef` i.e. `ERROR: UndefRefError: access to undefined reference`

mtfishman:
Right, if `β != 0` and `y` has undefined data, `y .= α .* x .+ β .* y` isn't well defined anyway so it wouldn't make sense to call that (it should be considered unsupported behavior or a bug). The fact that it doesn't error for simpler element types like `Float64` is just for efficiency reasons, and it's fine if that case errors for `BigFloat`.

I didn't realize that you were asking about that case in your previous comment, my comment only referred to inputs of `α` and `β` that would give well defined results for simpler element types like `Float64` (i.e. only overwriting undefined data, but not having that undefined data play a role in computing the result).

mtfishman:
So I've lost track, does the current version of `NDTensors._contract_scalar_noperm!` in this PR now work with `BigFloat` for reasonable/well defined inputs?

mtfishman:
If so, I would be fine just keeping the code as-is, and just keep the change that you've already made.

kmp5VT:
@mtfishman Yes the code now does work for `BigFloat` for reasonable and well defined inputs and specifically all the cases you have there. I added a check to `test_linearalgebra` I am not sure thats where the test should go and I can move it to a new place if you'd like. I called it `Contract with exotic types` because I can image we might want to add checks for some `MultiFloat` operations or perhaps tropic numbers.  Let me know if I should move this somewhere else and if I should add more tests

mtfishman:
I see similar tests here: https://github.com/ITensor/ITensors.jl/blob/v0.3.54/NDTensors/test/test_dense.jl#L188-L207, seems like it should go next to that test.

NDTensors/test/test_dense.jl
```diff
@@ -232,6 +232,22 @@ using .NDTensorsTestUtils: devices_list
   end
 
   # Only CPU backend testing
+  @testset "Contract with exotic types" begin
+    # BigFloat is not supported on GPU
+    M = Tensor(BigFloat, (1, 2, 3))
+    fill!(M, BigFloat(randn(Float64)))
+    O = Tensor(M, (2, 3))
+    N = Tensor([BigFloat(1.0)], (1,))
```
mtfishman, line 240:
```suggestion
    a = randn(BigFloat, 2, 3)
    t = Tensor(a, (1, 2, 3))
    m = Tensor(a, (2, 3))
    v = Tensor([one(BigFloat)], (1,))
```
I think this is a bit cleaner/easier to understand.

NDTensors/test/test_dense.jl
```diff
@@ -232,6 +232,22 @@ using .NDTensorsTestUtils: devices_list
...
+  
+    @test O ≈ contract(M, (-1, 2, 3), N, (-1,))
+    O = similar(M)
+    ## TODO find a way to test the next line, its not possible to access an undefined
+    ## reference
```
mtfishman, line 245: I don't understand this comment.

NDTensors/test/test_dense.jl
```diff
@@ -232,6 +232,22 @@ using .NDTensorsTestUtils: devices_list
...
+    NDTensors.contract!!(O, (2, 3), M, (-1, 2, 3), N, (-1,), 0.0, 1.0)
+    NDTensors.contract!!(O, (2, 3), M, (-1, 2, 3), N, (-1,), 0.0, 0.0)
```
mtfishman, line 247: Can we use `contract!` here?

NDTensors/test/test_dense.jl
mtfishman: From what I can tell, it doesn't make sense to test `!iszero(β)` and `O` initialized with `similar`, since that wouldn't give a well defined result (whether you are using `BigFloat` or a simpler element type). If you want to test that case it seems like you should initialize the data, for example with `randn!`.

kmp5VT: Sorry this was a comment for myself. I replaced the test with this
```
tp = similar(t)
NDTensors.contract!(tp, (1, 2, 3), t, (-1, 2, 3), v, (-1,), 0.0, 1.0)
@test inds(tp) == inds(t)
    ``` 
    But its not possible to do a test like `@test tp == t` because accessing elements of the array throws an UndefRefError
    ```
    
Error During Test at /Users/kpierce/.julia/dev/ITensors/NDTensors/test/test_dense.jl:247
  Test threw exception
  Expression: m == t
  UndefRefError: access to undefined reference
  ```

Comment (mtfishman) [reply NDTensors/test/test_dense.jl:245]: My point was that I don't think we should expect `contract!` to work when `!iszero(β)` and `tp` has undefined data (for example initialized with `similar`) so that doesn't make sense to test anyway.

File: NDTensors/test/test_dense.jl (hunk 3 above)
Comment (kmp5VT) [reply NDTensors/test/test_dense.jl:247]: Yes I replaced this with `contract!`

File: NDTensors/test/test_dense.jl (hunk 1 above)
Comment (kmp5VT) [reply NDTensors/test/test_dense.jl:240]: The function `randn(BigFloat, ...)` is not defined in Julia 1.6 so I wrote this as `a = BigFloat.(randn(Float64, 2, 3))`

File: NDTensors/test/test_dense.jl (hunk 2 above)
Comment (kmp5VT) [reply NDTensors/test/test_dense.jl]: Makes sense, I removed the test. Thanks!

Comment (mtfishman) [reply NDTensors/test/test_dense.jl]: Calling `contract!` in that way just isn't well defined and users shouldn't do that, just like users shouldn't use call `LinearAlgebar.mul!(c, a, b, α, β)` if `c` has undefined data and `!iszero(β)`. It may or may not error depending on the element type, but either way it isn't a good idea to do that since the answer would be undefined (you could just get complete junk data as the result of `c`).

Even if `iszero(α)`, the operation you are doing is `c .= β .* c`, so if `c` is filled with junk (undefined data) then the answer you get would be junk.

File: NDTensors/test/test_dense.jl [hunk 4]
```diff
@@ -234,18 +234,16 @@ using .NDTensorsTestUtils: devices_list
...
-    @test iszero(O)
+    ## randn(BigFloat, ...) is not defined in Julia 1.6
+    a = BigFloat.(randn(Float64, 2, 3))
+    t = Tensor(a, (1, 2, 3))
+    m = Tensor(a, (2, 3))
+    v = Tensor([one(BigFloat)], (1,))
+
+    @test m ≈ contract(t, (-1, 2, 3), v, (-1,))
+    tp = similar(t)
+    NDTensors.contract!(tp, (1, 2, 3), t, (-1, 2, 3), v, (-1,), 0.0, 0.0)
```
mtfishman, line 245: Let's use `false` or `zero(Bool)` instead of `0.0` to be more agnostic about the element type (`0.0` is implicitly `Float64` which is pretty opinionated).

mtfishman: We might want to also test a case where `iszero(α)` and `!iszero(β)`, in which case you can construct `tp` with random data, for example with `tp = similar(t); randn!(tp)`.

kmp5VT: Makes sense, I replaced alpha and beta with `true` and `false` instead of `0.0` and `1.0` and I added the case `NDTensors.contract!(tp, (1, 2, 3), t, (-1, 2, 3), v, (-1,), false, BigFloat(randn(Float64)))` to check the scaling option. the `LinearAlgebra` function `scal` is not actually defined on `BigFloat` so I converted to using Broadcast.

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -69,7 +69,7 @@ function _contract_scalar_noperm!(
       fill!(Rᵈ, 0)
     else
       # Rᵈ .= α .* T₂ᵈ
```
mtfishman, line 71:
```suggestion
```

NDTensors/src/dense/tensoralgebra/contract.jl
```diff
@@ -81,8 +81,8 @@ function _contract_scalar_noperm!(
     end
   else
     if iszero(α)
-      # Rᵈ .= β .* Rᵈ
-      LinearAlgebra.scal!(length(Rᵈ), β, Rᵈ, 1)
+      Rᵈ .= β .* Rᵈ
+      #LinearAlgebra.scal!(length(Rᵈ), β, Rᵈ, 1)
```
mtfishman, line 85:
```suggestion
```

PR #1310, opened 2024-01-17
[BlockSparseArrays] Naive symmetric tensor fusion and contraction

mtfishman:
Naive symmetric tensor fusion and contraction, which doesn't fuse blocks yet. That will be implemented in a future PR by sorting the sectors.


emstoudenmire:
Good changes to Sectors.

If I understand correctly, does this PR enable things like U(1) symmetric tensors to already work with the new system, just not in a totally ideal way yet?

mtfishman:
> If I understand correctly, does this PR enable things like U(1) symmetric tensors to already work with the new system, just not in a totally ideal way yet?

Yeah, for example you can call functions like `fusedims` and `splitdims` but they just do naive reshapes and don't permute and fuse sectors yet, that will be the next step.

In terms of `contract`, the existing `BlockSparse` `contract` code for symmetric tensors doesn't permute and fuse sectors either, so this implementation should in principle have essentially the same performance since it does the same set of operations (though I haven't benchmarked it yet), besides not having block sparse multithreading yet.

PR #1318, opened 2024-01-25
[NDTensors] Fix `BlockSparse` map-permutedims bug involving mismatched blocks

mtfishman:
This fixes a bug in the `BlockSparse` map-permutedims function (`permutedims![!]` with a map function passed to it).

Before, it was only looping over blocks in the source tensor (the tensor being permuted), so if there were blocks in the destination tensor that weren't in the source tensor it wasn't performing the map operation on those blocks which was leading to incorrect results in some cases. This rewrites that logic to be simpler and handle that case correctly.

This issue was first pointed out by @b-kloss, which he hit in some TTN DMRG/TDVP code when certain initial states led to contractions that resulted in tensors with no blocks, which were then being scaled incorrectly inside the Krylov solver.

- [x] Fix some remaining failing tests.
- [x] Add a minimal regression test.
- [x] Bring back GPU support using `generic_zeros` and `unwrap_type` (though this PR breaks `Metal` support).

**EDIT:** Unfortunately this PR breaks `Metal` support for certain block sparse operations since it relies on `append!` to add on new block data, which is broken since `resize!` isn't implemented for `MtlVector`, see https://github.com/JuliaGPU/Metal.jl/issues/277. Block sparse operations on GPU haven't been efficient yet anyway, so this isn't much of a loss, and I would rather keep the code simpler and break `Metal` support rather than make it more complicated so it can work on `Metal` but be slow anyway.

### DIFF
--- a/NDTensors/ext/NDTensorsMetalExt/NDTensorsMetalExt.jl
+++ b/NDTensors/ext/NDTensorsMetalExt/NDTensorsMetalExt.jl
@@ -18,6 +18,5 @@ include("copyto.jl")
 include("append.jl")
 include("permutedims.jl")
 include("mul.jl")
-include("splice.jl")
 
 end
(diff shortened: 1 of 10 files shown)


mtfishman:
[test ITensors mps]

mtfishman:
[test ITensors mps]

b-kloss:
It's also working in my tests as far as I can tell.

mtfishman:
[test ITensors mps]

PR #1321, opened 2024-01-31
[CI] Add pre-commit hooks for formatting

ogauthe:
Set-up `pre-commit` to automatically apply formatters. Once `pre-commit` is installed, the hooks are run automatically before any commit. Only the diff are checked, no the full code basis. If at least one hook fails, the commit is rejected. The files are left in the state imposed by the formatter, so that one just needs to `git add` the changes and run again `git commit`.

ref : https://pre-commit.com/

There are 3 commits here with different hooks added. I ran `pre-commit` on all files after adding a hook.
- the first one install hooks to check `.toml`, `.yaml`, `merge-conflicts` and `julia-format`. No file is modified outside of .pre-commit-config.yaml
- the second one adds `trailing-whitespace` hook. Many files are modified.
- the third one adds `end of files` hook. Again, many files are modified.

As many files are modified, this will appear as a large commit in history, but it only needs to be done once. We may also discard some of these hooks to avoid adding too many constraints.


mtfishman:
Looks good, thanks. That sounds like it will make it easier for contributors to apply the formatter. My main concern would be if it confuses novice git users, but it sounds better than the current system where it only tells you about a formatting issue after the fact. Another concern I have is if the formatter has a bug (which it occasionally does, though I haven't seen one in a while) it could block a commit, but you can tell the formatter to ignore pieces of code (https://domluna.github.io/JuliaFormatter.jl/stable/skipping_formatting/), we would just need to tell people how to do that if there are issues with the formatter. But let's try this out and see how it works in practice.

mtfishman:
@ogauthe, I see this removed whitespace from some unicode visualization files stored in `.txt` files which are being used as reference images to test that functionality. I think we should have skipped `.txt` here, could you fix that and revert those reference images?

PR #1323, opened 2024-01-31
[CI] Exclude references/*.txt from formatting hooks

ogauthe:
exclude 'references/*.txt' from hooks and recover previous file state.

I can also exclude all '.txt' files, but I think it is not required.


mtfishman:
[test ITensorUnicodePlots]

mtfishman:
Thanks for the quick fix!

PR #1324, opened 2024-01-31
[ITensors] Fix occasional test failure in `test_dmrg.jl`

kmp5VT:
# Description

Computing the linkdim instead of assuming the value is 4 because occasionally the value is smaller from `randomMPS`.  I added a comment in the test to remind us of the issue


test/ITensorLegacyMPS/base/test_dmrg.jl
```diff
@@ -110,7 +110,10 @@ using ITensors: nsite, set_nsite!
     @test length(PHdisk) == N
     @test ITensors.site_range(PH) == n:(n + 1)
     @test eltype(PH) == Float64
-    @test size(PH) == (3^2 * 4^2, 3^2 * 4^2)
+    ## TODO sometimes randomMPS gives a linkdim value of 3 
+    ## which causes an error in `size = 3^2 * 4^2`
+    size = linkdim(psi, n)^2 * dim(siteind(psi, n))^2
+    @test size(PH) == (size, size)
```
mtfishman, line 116: This isn't quite right. `PH` is the MPO projected onto sites `n` and `n + 1`, so it should be:
```julia
size = linkdim(psi, n - 1) * linkdim(psi, n + 1) * dim(siteind(psi, n)) * dim(siteind(psi, n + 1))
```

kmp5VT: Ahh I see, I updated the code. Thanks!!

mtfishman:
[test ITensors mps]

test/ITensorLegacyMPS/base/test_dmrg.jl
```diff
@@ -110,7 +110,14 @@ using ITensors: nsite, set_nsite!
     @test length(PHdisk) == N
     @test ITensors.site_range(PH) == n:(n + 1)
     @test eltype(PH) == Float64
-    @test size(PH) == (3^2 * 4^2, 3^2 * 4^2)
+    ## TODO sometimes randomMPS gives a linkdim value of 3 
+    ## which causes an error in `size = 3^2 * 4^2`
+    size =
```
mtfishman, line 115: You should use a variable name other than `size` since that overwrite the function`size(...)`.

kmp5VT: Make sense, thanks!

mtfishman:
Thanks! Will be nice to have the tests pass more consistently.

PR #1325, opened 2024-02-02
[WIP] [NDTensors] Add `AMDGPU.jl` (ROCm) based extension for NDTensors

wbernoudy:
# Description

This PR adds an AMDGPU.jl extension for NDTensors following the same structure as the CUDA extension.

It is still a WIP, but I am hoping to get feedback on a few points before finishing it up.

1. Dot product on `ITensor`s sometimes fails or falls back on slow default methods. If I implement a simple method that makes a copy and permutes the indices if they are different, I can then just call `dot` on the underlying `ROCArray`s, like so:
```julia
function dot(A::ITensor, B::ITensor)
  pB = B
  if (inds(A) != inds(B))
    pB = permute(B, inds(A))
  end
  return dot(NDTensors.data(tensor(A)), NDTensors.data(tensor(pB)))
end
```

I'm having trouble understanding which low level functions (presumably in `LinearAlgebra`) `ITensors.dot` actually calls. Trying to trace the functions in the Julia REPL brings me to `NDTensors.contract` but I can't figure out what it calls past that.

This brings me to a file in `SimpleTraits.jl`
```julia
julia> i, j = Index(10, "i"), Index(5, "j")
((dim=10|id=395|"i"), (dim=5|id=574|"j"))

julia> A = ITensor(randomTensor(ROCArray, (i, j)))
ITensor ord=2 (dim=10|id=395|"i") (dim=5|id=574|"j")
Dense{Float64, ROCArray{Float64, 1, AMDGPU.Runtime.Mem.HIPBuffer}}

julia> B = ITensor(randomTensor(ROCArray, (i, j)))
ITensor ord=2 (dim=10|id=395|"i") (dim=5|id=574|"j")
Dense{Float64, ROCArray{Float64, 1, AMDGPU.Runtime.Mem.HIPBuffer}}

julia> labelsA, labelsB = ITensors.compute_contraction_labels(inds(A), inds(B))
((-1, -2), (-1, -2))

julia> @edit NDTensors.contract(A.tensor, labelsA, B.tensor, labelsB)
```

I expect it's very similar to an issue I was having with other slow tensor contraction with `ROCArray` backed `ITensor`s which I was fixed by exposing the right parameters for `LinearAlgebra.mul!` in `AMDGPU.jl` (https://github.com/JuliaGPU/AMDGPU.jl/pull/585).

2. Though this PR does implement the rocSOLVER Jacobi method for SVD, I found it to be significantly slower than divide-and-conquer on CPU for the TDVP simulations I'm doing. To make using the CPU SVD work I added a `roc_to_cpu_svd` option for `svd_alg` which simply copies the matrix to host, calls SVD, and then converts USV back to `ROCArray`s. However, I'm guessing there may be a cleaner or more general way to do this (maybe to allow the other `NDTensors` extensions like the CUDA extension to also easily use CPU for SVD).

3. When calling the rocSOLVER SVD algorithm, the function may try to allocate extra memory on the GPU (even though I can't find any documentation for why in the rocSOLVER documentation from AMD), so if the memory pool has allocated the entire GPU's memory (i.e. `AMDGPU.Mem.free()` returns 0) the function may fail. This is why I am trimming the pool if there is less than 1GB free before calling `gesvdj!`. I will try to investigate this more to see if there is a proper way to handle it.

4. There is a small bug in the rocSOLVER SVD algorithms in `AMDGPU.jl` which has been fixed but not released yet (https://github.com/JuliaGPU/AMDGPU.jl/pull/588).

# How Has This Been Tested?

I added the new `NDTensors.roc` function to the devices list in the `NDTensors` tests. However, there are several tests failing due to:

1. The aforementioned `dot` issue
2. Tests that call `svd` directly on `ROCArray`s which seems to return junk https://github.com/ITensor/ITensors.jl/blob/98b95a20f7b3c31324a2f23c730af4fc173b61f3/NDTensors/src/lib/Unwrap/test/runtests.jl#L94 I believe this should be fixed by `AMDGPU.jl`.
3. There are many tests that are calling `ql` on `ROCArray`s which is not handled by `AMDGPU.jl` (similar to `CUDA.jl`). There seems to be an exception made for CUDA arrays in various QL decomposition functions, e.g. https://github.com/ITensor/ITensors.jl/blob/98b95a20f7b3c31324a2f23c730af4fc173b61f3/NDTensors/src/linearalgebra/linearalgebra.jl#L392 Not sure if I should add similar code for `ROCArray`s here.

I also added an example script similar to the CUDA one.

# Checklist:


NDTensors/ext/NDTensorsROCmExt/NDTensorsROCmExt.jl
```diff
@@ -0,0 +1,26 @@
+module NDTensorsROCmExt
```
mtfishman, line 1: This should be called `NDTensorsAMDGPUExt`, following the convention that extensions are named based on the names of the two packages involved.

wbernoudy: Ah ok, I will change that!

mtfishman:
At first glance this looks good, thanks! This is a nice payoff for a lot of work @kmp5VT and I have been doing to make the overload requirements for new GPU backends as minimal as possible, I'm glad to see this doesn't seem to require very much code to get working.

The SVD code definitely looks like it is the most complicated part that we will have to look at carefully, but besides that it all looks similar to what we needed to overload for the `CUDA` and `Metal` backends.

In terms of `dot`, it just calls tensor contraction: https://github.com/ITensor/ITensors.jl/blob/v0.3.55/src/itensor.jl#L1905 so maybe you are hitting a corner case of tensor contraction that is going wrong for the `AMDGPU` backend? You would have to share a minimal example for us to have more information.

mtfishman:
Looks like there is some package compatibility issue in the tests.

mtfishman:
One suggestion I would have to make this first PR simpler to review would be to just support an SVD implementation that transfers to CPU, and then add in other SVD backends in a future PR.

I think it is a subtle question to figure out how to select different SVD backends for different devices (say you want to specify that if you run on CUDA you want to use a certain cuLAPACK implementation, but then on AMD you want to transfer to CPU to perform SVD, then on Metal you want to...). That will be a longer discussion and may require us to think about the higher level SVD interface.

It may by that we should add support to the `svd` function that the `alg` keyword argument could in general be a function which takes in information about the backend and returns a string or `Algorithm` object depending on that backend, i.e.:
```julia
svd_alg(::Type{<:CuArray}) = "cpu"
svd_alg(::Type{<:ROCArray}) = "jacobi_algorithm"
svd_alg(::Type{<:MtlArray}) = "cpu"

u, s, v = svd(t, (i, j); alg=svd_alg)
```

We would also like to add automated testing for AMD GPUs, @kmp5VT will start to look into that.

kmp5VT:
@wbernoudy I am currently working on finding a way to access an AMD GPU device. Would it be possible to provide me access to your branch so I can help accelerate your development? I am not that familiar with the `AMDGPU.jl` package but it does look like there is a `rocblas` and `rocsolver` directory which links the julia code to the C libraries in `ROC`. We should be able to access these functions through `linearalgebra.jl` , if we can't thats something we could bring up to `AMDGPU.jl` and should be a relatively easy implementation. regarding the `dot` function I believe we are using `LinearAlgebra.mul!` to launch a BLAS `dot` function.

wbernoudy:
> Would it be possible to provide me access to your branch so I can help accelerate your development?
Of course, I added you as a collaborator on my fork.

To be clear, this extension is already using the accelerated BLAS functions provided by `AMDGPU.jl` for tensor operations, and it is using the SVD provided in rocSOLVER (and even eigen methods).

`AMDGPU.jl` now provides both the three and five argument versions of `LinearAlgebra.mul!`. But I'm guessing there is a certain transformation on one or more of the matrices used somewhere in the call stack of `dot` that is not yet handled by `AMDGPU.jl`. I found another of these cases that I fixed here https://github.com/wbernoudy/ITensors.jl/blob/01f6c71844114eb92a9a50c4b23343161b9b028f/NDTensors/ext/NDTensorsROCmExt/mul.jl#L27

wbernoudy:
@mtfishman I tried the idea of disabling scalar indexing to see if some generic matmul function is using that in the `dot` code path, but no luck. I disabled the indexing simply by commenting out the relevant functions in `NDTensors/ext/NDTensorsAMDGPUExt/indexing.jl`, which I believe was the right thing to do to test this (and I do see it it reaching a `ERROR: Scalar indexing is disallowed.` on some calls). Is there another way it's allowing scalar indexing that I might be missing?

Here you can see the standard `dot` function does not hit any scalar indexing (other than the very final reading of the zero dimensional tensor) but is ~500x slower than calling `dot` on the underlying arrays.

```julia
julia> using AMDGPU

julia> using ITensors

julia> using NDTensors

julia> using LinearAlgebra

julia> i, j = Index(6152, "i"), Index(1324, "j")
((dim=6152|id=857|"i"), (dim=1324|id=611|"j"))

julia> A = ITensor(randomTensor(ROCArray, (i, j)))
ITensor ord=2 (dim=6152|id=857|"i") (dim=1324|id=611|"j")
Dense{Float64, ROCArray{Float64, 1, AMDGPU.Runtime.Mem.HIPBuffer}}

julia> B = ITensor(randomTensor(ROCArray, (j, i)))
ITensor ord=2 (dim=1324|id=611|"j") (dim=6152|id=857|"i")
Dense{Float64, ROCArray{Float64, 1, AMDGPU.Runtime.Mem.HIPBuffer}}

julia> dot(A, B)
ERROR: Scalar indexing is disallowed.
(rest of stack trace)

julia> dag(A) * B
ITensor ord=0
Dense{Float64, ROCArray{Float64, 1, AMDGPU.Runtime.Mem.HIPBuffer}}

julia> AMDGPU.@allowscalar (dag(A) * B)[]
-3066.643034152186

julia> AMDGPU.@elapsed dag(A) * B
0.29850277f0

julia> function LinearAlgebra.dot(A::ITensor, B::ITensor)
         pB = B
         if (inds(A) != inds(B))
           pB = permute(B, inds(A))
         end
         return dot(data(tensor(A)), data(tensor(pB)))
       end

julia> dot(A, B)
-3066.643034153356

julia> AMDGPU.@elapsed dot(A, B)
0.000623521f0
```

wbernoudy:
Actually I think I have figured out the slowdown with `dot` (so my previous comment can be ignored). The issue is actually that it _is_ calling the AMDGPU gemm method, but this method itself is quite slow when doing the dot product on large vectors. Perhaps other GPU libraries (e.g. CUDA) are able to handle that case better, but I'm guessing the `rocblas_{T}gemm` functions are doing the naive thing and having threads write to the same output value atomically.

Calling `dot` directly on the arrays calls the method defined in the `GPUArrays` library, which uses `mapreduce` and seems to be far faster. This makes me think there should be definition of `dot` in `ITensors.jl` that does call `dot` on the underlying arrays instead of calling general matmul implicitly by contracting as normal. I'd be happy to make a separate PR for this.

wbernoudy:
> One suggestion I would have to make this first PR simpler to review would be to just support an SVD implementation that transfers to CPU, and then add in other SVD backends in a future PR.

This is exactly what I added the `"roc_to_cpu_svd"` alg type for. I would be happy to make this the default in this PR (instead of `"jacobi_algorithm"`).

@mtfishman Would it be better if I removed the Jacobi implementation from this PR?

NDTensors/ext/examples/NDTensorAMDGPU.jl
```diff
@@ -0,0 +1,85 @@
...
+  @test inner(cm', cH, cm) ≈ inner(m', H, m)
+
+  m = orthogonalize(m, 1)
+  cm = gpu(orthogonalize(cm, 1))
+  @test inner(m', m) ≈ inner(cm', cm)
+
+  H = orthogonalize(H, 1)
+  cH = gpu(cH)
+
+  @test inner(cm', cH, cm) ≈ inner(m', H, m)
```
mtfishman, line 81: I would prefer to remove these tests that rely on MPO/MPS functionality. We are planning to disentangle code based on MPO/MPS types from the rest of the ITensors.jl codebase and move it into a separate repository, and having tests that intermingle basic ITensor functionality from MPO/MPS functionality will make that more challenging.

mtfishman: Oh I see these are examples, not tests.

Anyway, I'm not really a fan of having these extension-specific example files, let's just remove this file for now.

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
```diff
@@ -0,0 +1,66 @@
...
+      AMDGPU.HIP.used_memory(pool) / 1e9,
+      "GB",
+    )
+  end
+
+  float_type = real(eltype(A))
+  U, S, V, residual, n_sweeps, info = try
+    # second parameter is the absolute tolerance, setting to <= 0 means
+    # setting it to machine precision
+    AMDGPU.rocSOLVER.gesvdj!(A, float_type(-1.0), Int32(JACOBI_MAX_ITERATIONS))
```
mtfishman, line 50:
```suggestion
    AMDGPU.rocSOLVER.gesvdj!(A, -one(float_type), Int32(JACOBI_MAX_ITERATIONS))
```

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
```diff
@@ -0,0 +1,66 @@
...
+struct QRAlgorithm end
+struct ROCtoCPUSVD end
+
+JACOBI_MAX_ITERATIONS = 500
+
+function NDTensors.svd_catch_error(A::ROCMatrix; alg::String="jacobi_algorithm")
+  if alg == "jacobi_algorithm"
+    alg = JacobiAlgorithm()
+  elseif alg == "qr_algorithm"
+    alg = QRAlgorithm()
```
mtfishman, line 12: Is `QRAlgorithm` supported for `ROCMatrix`? If not this should be removed.

mtfishman:
> > One suggestion I would have to make this first PR simpler to review would be to just support an SVD implementation that transfers to CPU, and then add in other SVD backends in a future PR.
> 
> This is exactly what I added the `"roc_to_cpu_svd"` alg type for. I would be happy to make this the default in this PR (instead of `"jacobi_algorithm"`).

Yes, I understand that. I just don't like that name. How would you choose the CPU SVD backend? Would we need to define a new `"*_to_cpu_svd"` for every GPU backend? That doesn't seem like a sustainable forward-looking API. I think we will have to think more carefully about how to allow users to control device transfers in low level operations more generally, perhaps with an API that is orthogonal to the `alg` keyword argument.

> 
> @mtfishman Would it be better if I removed the Jacobi implementation from this PR?

Yes, I think so. The code you wrote seems more appropriate to add as an `svd` function to `AMDGPU.jl`, and given that it doesn't appear to help with performance it doesn't seem very compelling to add it here and have to support it. It would also simplify the code a lot and let us push off a decision about what to name the `alg` arguments.

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
wbernoudy: It is supported by `rocSOLVER`, and I have written the wrapper for it in `AMDGPU.jl` which I will put in a future PR. But I'm happy to take this out for now of course.

wbernoudy:
> Yes, I understand that. I just don't like that interface. How would you choose the CPU SVD backend? Would we need to define a new `"*_to_cpu_svd"` for every GPU backend? That doesn't seem like a sustainable forward-looking API.

Very much agreed. An orthogonal argument that allows you to choose the device (in addition to the SVD algorithm) sounds great to me.

Either way, I'm happy to either wait until after a more precise API has been figured out to finalize this PR, or keep the awkward `roc_to_cpu_svd` for now. Or perhaps there is a cleaner way to do this now without the API change? My understanding is that we need _something_ for the name of the SVD alg type for ROCMatrix, as it must be used here: https://github.com/ITensor/ITensors.jl/blob/2fd3e0a2658831903317b753934d04f37fcdcb46/NDTensors/src/linearalgebra/linearalgebra.jl#L86

> The code you wrote seems more appropriate to add as an svd function to AMDGPU.jl

I figured this might be the case, and it certainly makes sense to handle the out-of-memory properly in `AMDGPU.jl` instead. I will take this out :+1:

mtfishman:
> > Yes, I understand that. I just don't like that interface. How would you choose the CPU SVD backend? Would we need to define a new `"*_to_cpu_svd"` for every GPU backend? That doesn't seem like a sustainable forward-looking API.
> 
> Very much agreed. An orthogonal argument that allows you to choose the device (in addition to the SVD algorithm) sounds great to me.
> 
> Either way, I'm happy to either wait until after a more precise API has been figured out to finalize this PR, or keep the awkward `roc_to_cpu_svd` for now. Or perhaps there is a cleaner way to do this now without the API change? My understanding is that we need _something_ for the name of the SVD alg type for ROCMatrix, as it must be used here:

Glad we're in agreement about that. I think since there will be only one SVD backend for `ROCMatrix` with the current plan (one where we transfer to CPU), we wouldn't need to involve the `alg` keyword argument in this PR, see for example the factorizations defined for `Metal`: https://github.com/ITensor/ITensors.jl/blob/main/NDTensors/ext/NDTensorsMetalExt/linearalgebra.jl. So we could still finish this PR in a simpler form and then add back other SVD backends later as they become available, with a better API for choosing between those backends on GPU or transferring to CPU.

mtfishman:
Sorry, I didn't see you referring to `default_svd_alg(T)`. I forget how that is handled for `Metal` but it could just be handled the same way, presumably it chooses the default that would be used for CPU and then forwards that choice to the CPU SVD backend.

wbernoudy:
Oh, I didn't think about just overloading `LinearAlgebra.svd` directly! That's much more straightforward for this. I will make that change.

kmp5VT:
@wbernoudy Hi, just giving you an update. I now have access to an AMD GPU and have made a few changes (nothing significant). One thing is that the changes I am making in another [PR](https://github.com/ITensor/ITensors.jl/pull/1331) also affects the work here.  What I am doing is working to finish this other PR, pull the changes in here and make those modifications here as well so that we don't have to write this code twice. I am hoping to finish 1331 this week and put more time into your PR next week.

wbernoudy:
As far as I can tell, the `RUN NDTensors tests` CI check is failing because latest version of `AMDGPU.jl` that supports Julia 1.6 is 0.2.17. 

> ─AMDGPU [21141c5a] log:
...
     ├─restricted by julia compatibility requirements to versions: 0.1.0-0.2.17 or uninstalled, leaving only versions: 0.1.0-0.2.17

I'm guessing upgrading the CI check to Julia 1.9 may not be easy/feasible, but I'm not sure how else we can handle this.

kmp5VT:
@wbernoudy We use both `Julia v1.6` and `Julia v1.10` in our CI tests. I completely understand the problem you are running into and can definitely help you fix it. I think we are running into the same issue with Metal.jl and both will need to be addressed in order to create CI unittesting for these backends.
To give you an update this PR is number 3 on my list of things to do. The first two things are related to the module `TypeParameterAccessors`. I am currently in the works of updating the system across the ITensors library. To limit the number of rewrites of this PR I want to finish the integration first. I would guess that it will take me about a week to finish this integration and next week I can get to this PR.

NDTensors/ext/NDTensorsAMDGPUExt/imports.jl
```diff
@@ -0,0 +1,2 @@
+import NDTensors:
+  ContractionProperties, _contract!, GemmBackend, auto_select_backend, _gemm!
```
kmp5VT, line 2: Are these used in this extension? If so move to appropriate file

NDTensors/src/adapt.jl
```diff
@@ -3,6 +3,7 @@ adapt_structure(to, x::TensorStorage) = setdata(x, adapt(to, data(x)))
 adapt_structure(to, x::Tensor) = setstorage(x, adapt(to, storage(x)))
 
 ## use unwrap cpu here because Expose is included before NDTensors
+## TODO make a AMDGPUExtensions library with `AMDGPUExtensions.roc` definied.
```
kmp5VT, line 6: Reminder to delete this comment, the library is made

NDTensors/src/abstractarray/isroc.jl
```diff
@@ -0,0 +1,5 @@
+isroc(A::AbstractArray) = isroc(typeof(A))
+
```
kmp5VT, line 2: Reminder to determine if `iscu` is necessary and delete the functions if we can use `expose` instead

NDTensors/test/Project.toml
```diff
@@ -1,4 +1,5 @@
 [deps]
+AMDGPU = "21141c5a-9bdb-4563-92ae-f87d6854732e"
```
kmp5VT, line 2: Reminder to move this to `Extras` for now with `Metal` while we determine how to test .

kmp5VT: Moved for now.

NDTensors/src/adapt.jl
kmp5VT: This comment is removed

NDTensors/ext/NDTensorsAMDGPUExt/imports.jl
kmp5VT: these are not necessary and I removed them

NDTensors/src/abstractarray/isroc.jl
kmp5VT: I found where iscu is necessary (in a place where adapt was failing due to an issue in the CUDA.jl library) and I made a comment. I'll keep iscu in for now but I removed isroc in favor of using `Expose` system

kmp5VT:
@mtfishman If you have time to look through this PR I have made some comments in the code about why I have programmed things up. For the most part its exactly the same as the other extensions. I found two issues. One is that AMDGPU doesn't have an `append!` function which means that `Blocksparse` dmrg fails. The other is that there is something happening in `qr` when `singular=true` and `real(elt) = Float32`. I suspect this is in the ROCSolver library and I have disabled these tests.  Otherwise all NDTensors test pass using AMDGPU.

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
kmp5VT: This code has been removed.

NDTensors/ext/examples/NDTensorAMDGPU.jl
kmp5VT: Note for myself. I haven't looked at the example file yet.

kmp5VT:
@wbernoudy Does this PR look good for you? Also do you know why I am seeing the F32 and ComplexF32 QR failing? Thanks!

NDTensors/ext/NDTensorsAMDGPUExt/NDTensorsAMDGPUExt.jl
```diff
@@ -0,0 +1,19 @@
+module NDTensorsAMDGPUExt
+
+using NDTensors
+using Adapt
+using Functors
+using AMDGPU
+using AMDGPU.Runtime.Mem
+using AMDGPU.rocBLAS
+using AMDGPU.rocSOLVER
```
mtfishman, line 9: Let's explicitly list the functions we are using from each module, i.e. `using Adapt: adapt, ...`.

NDTensors/ext/NDTensorsAMDGPUExt/adapt.jl
```diff
@@ -0,0 +1,19 @@
...
+using NDTensors.GPUArraysCoreExtensions: storagemode
+function AMDGPUExtensions.roc(xs)
+  return fmap(x -> adapt(ROCArrayAdaptor{AMDGPU.Runtime.Mem.HIPBuffer}(), x), xs)
+end
+
+function Adapt.adapt_storage(adaptor::ROCArrayAdaptor, xs::AbstractArray)
+  ElT = eltype(xs)
+  BufT = storagemode(adaptor)
+  N = ndims(xs)
+  return isbits(xs) ? xs : adapt(ROCArray{ElT,N,BufT}, xs)
```
mtfishman, line 11: @kmp5VT this seems like a good place to use the new `TypeParameterAccessors` functionality.

NDTensors/ext/NDTensorsAMDGPUExt/indexing.jl
```diff
@@ -0,0 +1,20 @@
+using NDTensors.Expose: Exposed, expose, parent, unexpose
+using NDTensors.GPUArraysCoreExtensions: cpu
+function Base.getindex(E::Exposed{<:ROCArray})
+  return AMDGPU.@allowscalar unexpose(E)[]
+end
+
+function Base.setindex!(E::Exposed{<:ROCArray}, x::Number)
+  AMDGPU.@allowscalar unexpose(E)[] = x
```
mtfishman, line 8: I would prefer to call this from where it is defined, i.e. `GPUArraysCore.@allowscalar`. Probably we should put `using GPUArraysCore: @allowscalar` at the top of the file and then just use `@allowscalar` in the code.

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
```diff
@@ -0,0 +1,20 @@
...
+using LinearAlgebra: svd
+
+function LinearAlgebra.svd(A::Exposed{<:ROCMatrix}; kwargs...)
+  U, S, V = svd(cpu(A))
+  return roc.((U, S, V))
+end
+
+## TODO currently AMDGPU doesn't have ql so make a ql function
+function NDTensors.Expose.ql(A::Exposed{<:ROCMatrix})
+  Q, L = ql(expose(NDTensors.cpu(A)))
```
mtfishman, line 14:
```suggestion
  Q, L = ql(expose(cpu(A)))
```

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
```diff
@@ -41,3 +41,15 @@ function NDTensors.svd_catch_error(A::CuMatrix, ::CUDA.CUSOLVER.QRAlgorithm)
   end
   return USV
 end
+
+using NDTensors.Expose: expose, ql, ql_positive
```
mtfishman, line 45:
```suggestion
using NDTensors.Expose: Expose, expose, ql, ql_positive
```

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
```diff
@@ -41,3 +41,15 @@ function NDTensors.svd_catch_error(A::CuMatrix, ::CUDA.CUSOLVER.QRAlgorithm)
...
+using NDTensors.TypeParameterAccessors: unwrap_array_type
+## TODO currently AMDGPU doesn't have ql so make a ql function
+function NDTensors.Expose.ql(A::Exposed{<:CuMatrix})
+  Q, L = ql(expose(NDTensors.cpu(A)))
+  return adapt(unwrap_array_type(A), Matrix(Q)), adapt(unwrap_array_type(A), L)
+end
+function NDTensors.Expose.ql_positive(A::Exposed{<:CuMatrix})
```
mtfishman, line 52:
```suggestion
function Expose.ql_positive(A::Exposed{<:CuMatrix})
```

NDTensors/ext/examples/NDTensorAMDGPU.jl
mtfishman, line 1: I would prefer to not have this example file and just rely on the tests for now.

NDTensors/src/lib/AMDGPUExtensions/src/roc.jl
```diff
@@ -0,0 +1,15 @@
...
+# Implemented in NDTensorsAMDGPUExt
+function roc end
+
+## Here we need an ROCArrayAdaptor to prevent conversion of 64 bit numbers to 32 bit.  
+## We cannot write `adapt(CuVector, x)` because this
+## will not allow us to properly utilize the buffer preference without changing the value of
+## default_buffertype. Also `adapt(CuVector{<:Any, <:Any, Buffertype})` fails to work properly
+struct ROCArrayAdaptor{B} end
+
+## TODO remove TypeParameterAccessors when SetParameters is removed
```
mtfishman, line 12: Seems like there is something to do here.

NDTensors/src/lib/AMDGPUExtensions/test/runtests.jl
```diff
@@ -0,0 +1,9 @@
+@eval module $(gensym())
+using Test: @testset, @test
+using NDTensors.AMDGPUExtensions: roc, ROCArrayAdaptor
+using NDTensors.GPUArraysCoreExtensions: storagemode
+@testset "cu function exists" begin
```
mtfishman, line 5:
```suggestion
@testset "roc and ROCArrayAdaptor" begin
```

NDTensors/test/runtests.jl
```diff
@@ -18,6 +18,10 @@ using SafeTestsets: @safetestset
     using NDTensors: NDTensors
     include(joinpath(pkgdir(NDTensors), "ext", "examples", "NDTensorCUDA.jl"))
   end
+  if "rocm" in ARGS || "all" in ARGS
+    using NDTensors: NDTensors
```
mtfishman, line 22: It seems better to just have one call to `using NDTensors: NDTensors` at the top of the file.

NDTensors/src/lib/AMDGPUExtensions/test/runtests.jl
kmp5VT: This is updated. Thanks!

NDTensors/ext/NDTensorsAMDGPUExt/indexing.jl
kmp5VT: Makes sense, I updated that code. thanks!

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
kmp5VT: Updated, thanks!

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
```diff
@@ -0,0 +1,20 @@
...
+  return adapt(unwrap_array_type(A), Matrix(Q)), adapt(unwrap_array_type(A), L)
+end
+function NDTensors.Expose.ql_positive(A::Exposed{<:ROCMatrix})
+  Q, L = ql_positive(expose(NDTensors.cpu(A)))
```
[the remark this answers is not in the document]
kmp5VT: This is updated, thanks!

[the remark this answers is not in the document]
kmp5VT: For the line 
```
NDTensors.Expose.ql_positive(A::Exposed{<:ROCMatrix})
```
Would you rather I have 
```
using NDTesnors.Expose: Expose
Expose.ql_positive(A::Exposed{<:ROCMatrix})
```
?

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
kmp5VT: This is added, thanks!

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
```diff
@@ -41,3 +41,15 @@ function NDTensors.svd_catch_error(A::CuMatrix, ::CUDA.CUSOLVER.QRAlgorithm)
   end
   return USV
 end
+
+using NDTensors.Expose: expose, ql, ql_positive
+using NDTensors.TypeParameterAccessors: unwrap_array_type
+## TODO currently AMDGPU doesn't have ql so make a ql function
+function NDTensors.Expose.ql(A::Exposed{<:CuMatrix})
```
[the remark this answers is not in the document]
kmp5VT: Updated, thanks!

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
```diff
@@ -41,3 +41,15 @@ function NDTensors.svd_catch_error(A::CuMatrix, ::CUDA.CUSOLVER.QRAlgorithm)
...
+  Q, L = ql(expose(NDTensors.cpu(A)))
+  return adapt(unwrap_array_type(A), Matrix(Q)), adapt(unwrap_array_type(A), L)
+end
+function NDTensors.Expose.ql_positive(A::Exposed{<:CuMatrix})
+  Q, L = ql_positive(expose(NDTensors.cpu(A)))
```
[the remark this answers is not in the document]
kmp5VT: Updated, thanks!

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
```diff
@@ -0,0 +1,20 @@
...
+  U, S, V = svd(cpu(A))
+  return roc.((U, S, V))
+end
+
+## TODO currently AMDGPU doesn't have ql so make a ql function
+function NDTensors.Expose.ql(A::Exposed{<:ROCMatrix})
+  Q, L = ql(expose(NDTensors.cpu(A)))
+  return adapt(unwrap_array_type(A), Matrix(Q)), adapt(unwrap_array_type(A), L)
+end
+function NDTensors.Expose.ql_positive(A::Exposed{<:ROCMatrix})
```
mtfishman, line 17: Should be changed to `Expose.ql_positive`.

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
```diff
@@ -0,0 +1,20 @@
...
+using NDTensors.TypeParameterAccessors: unwrap_array_type
+using LinearAlgebra: svd
+
+function LinearAlgebra.svd(A::Exposed{<:ROCMatrix}; kwargs...)
+  U, S, V = svd(cpu(A))
+  return roc.((U, S, V))
+end
+
+## TODO currently AMDGPU doesn't have ql so make a ql function
+function NDTensors.Expose.ql(A::Exposed{<:ROCMatrix})
```
mtfishman, line 13: Should be changed to `Expose.ql`.

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
```diff
@@ -0,0 +1,20 @@
+using NDTensors.AMDGPUExtensions: roc
+using NDTensors.Expose: Exposed, ql, ql_positive
```
mtfishman, line 2:
```suggestion
using NDTensors.Expose: Expose, Exposed, ql, ql_positive
```

NDTensors/ext/NDTensorsCUDAExt/linearalgebra.jl
kmp5VT: This is updated, thanks!

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
kmp5VT: Updated, thanks!

NDTensors/ext/NDTensorsAMDGPUExt/linearalgebra.jl
kmp5VT: Updated, thanks!

NDTensors/src/lib/AMDGPUExtensions/src/roc.jl
kmp5VT: Yes, I added the `using Position` and removed the `TypeParameterAccessors.` from this function. Thanks!

NDTensors/ext/NDTensorsAMDGPUExt/adapt.jl
kmp5VT: I updated this function and wrote an implementation with `TypeParameterAccessors` let me know if the way I wrote it is good or if you had something else in mind

NDTensors/ext/NDTensorsAMDGPUExt/NDTensorsAMDGPUExt.jl
kmp5VT: Okay, I have done this and moved all the usings to appropriate folders

NDTensors/ext/examples/NDTensorAMDGPU.jl
mtfishman: Same for the other extensions, I don't think they need examples right now (though in the future we should make documentation for them with examples).

NDTensors/test/runtests.jl
kmp5VT: Okay I have updated this!

NDTensors/ext/examples/NDTensorAMDGPU.jl
kmp5VT: Okay I am going to leave them in this PR and I will take them out in the next PR when I update the `using` calls. I'll add it to the google doc

NDTensors/ext/NDTensorsAMDGPUExt/NDTensorsAMDGPUExt.jl
```diff
@@ -0,0 +1,12 @@
+module NDTensorsAMDGPUExt
+using Functors
```
mtfishman, line 2: This should be removed from here as well.

NDTensors/ext/NDTensorsAMDGPUExt/adapt.jl
```diff
@@ -0,0 +1,29 @@
...
+
+function AMDGPUExtensions.roc(xs)
+  return fmap(x -> adapt(ROCArrayAdaptor{AMDGPU.Runtime.Mem.HIPBuffer}(), x), xs)
+end
+
+function Adapt.adapt_storage(adaptor::ROCArrayAdaptor, xs::AbstractArray)
+  roctype = set_type_parameters(
+    ROCArray, (eltype, ndims), type_parameters(xs, (eltype, ndims))
+  )
+  roctype = set_type_parameter(roctype, storagemode, storagemode(adaptor))
```
mtfishman, line 17: Seems like this can be combined into one function call, for example:
```julia
  new_parameters = (type_parameters(xs, (eltype, ndims)..., storagemode(adaptor))
  roctype = set_type_parameters(
    ROCArray, (eltype, ndims, storagemode), new_parameters)
  )
```

NDTensors/ext/NDTensorsAMDGPUExt/NDTensorsAMDGPUExt.jl
mtfishman: I think the only thing we are using from `Functors` is `fmap`.

NDTensors/ext/NDTensorsAMDGPUExt/adapt.jl
```diff
@@ -0,0 +1,29 @@
+using NDTensors: NDTensors, EmptyStorage, adapt_storagetype, emptytype
+using NDTensors.AMDGPUExtensions: AMDGPUExtensions, ROCArrayAdaptor
+using NDTensors.GPUArraysCoreExtensions: storagemode
+using NDTensors.TypeParameterAccessors:
+  set_type_parameter, set_type_parameters, type_parameter, type_parameters
+using Adapt: Adapt, adapt
+using AMDGPU: AMDGPU, ROCArray, ROCVector
+
+function AMDGPUExtensions.roc(xs)
+  return fmap(x -> adapt(ROCArrayAdaptor{AMDGPU.Runtime.Mem.HIPBuffer}(), x), xs)
```
mtfishman, line 10: Maybe this should use `default_type_parameter(ROCArray, storagemode)`.

NDTensors/ext/NDTensorsAMDGPUExt/NDTensorsAMDGPUExt.jl
kmp5VT: Sorry I missed this, I ran the unittests and forgot about looking into this. Thank you

NDTensors/ext/NDTensorsAMDGPUExt/adapt.jl
kmp5VT: In the other instances of this function we have the ability to modify the storagemode type so I am rewriting this as 
```
function AMDGPUExtensions.roc(xs, buffer=default_type_parameter(ROCArray, storagemode))
  return fmap(x -> adapt(ROCArrayAdaptor{buffer}(), x), xs)
end
```

NDTensors/ext/NDTensorsAMDGPUExt/indexing.jl
```diff
@@ -0,0 +1,23 @@
...
+using NDTensors.GPUArraysCoreExtensions: cpu
+using AMDGPU: AMDGPU, ROCArray
+using GPUArraysCore: @allowscalar
+
+function Base.getindex(E::Exposed{<:ROCArray})
+  return @allowscalar unexpose(E)[]
+end
+
+function Base.setindex!(E::Exposed{<:ROCArray}, x::Number)
+  AMDGPU.@allowscalar unexpose(E)[] = x
```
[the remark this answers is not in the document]
kmp5VT: I missed this, thank you!

NDTensors/ext/NDTensorsAMDGPUExt/adapt.jl
kmp5VT: This works and I have updated the code

NDTensors/ext/NDTensorsAMDGPUExt/adapt.jl
mtfishman: Should we call the keyword argument `storagemode` or `buffer`? What does `AMDGPU` call that?

kmp5VT: @mtfishman in AMDGPU they call it a buffer 
```
function ROCArray{T, N, B}(
        ::UndefInitializer, dims::Dims{N},
    ) where {T, N, B <: Union{Mem.HIPBuffer, Mem.HostBuffer}}
        @assert isbitstype(T) "ROCArray only supports bits types"
        buf = B(prod(dims) * sizeof(T); stream=stream())
        xs = new{T, N, B}(DataRef(_free_buf, buf), dims, 0)
        finalizer(unsafe_finalize!, xs)
        return xs
    end
    ```

Comment (mtfishman) [reply NDTensors/ext/NDTensorsAMDGPUExt/adapt.jl]: Right, but `CUDA.jl` also refers to it internally as a "buffer", but we've decided to use `storagemode` as a GPU-agnostic term for that, and the `AMDGPU.roc` function doesn't have a name for it: https://github.com/JuliaGPU/AMDGPU.jl/blob/v0.8.11/src/array.jl#L238. So perhaps we should go ahead and make our own decision to call that `storagemode` for any of the GPU backends.

Comment (kmp5VT) [reply NDTensors/ext/NDTensorsAMDGPUExt/adapt.jl]: I can definitely change this name of this variable to `storagemode`.
Just a question in NDTensorsCUDAExt this function looks like this
```
function CUDAExtensions.cu(xs; unified::Bool=false)
  return fmap(
    x -> adapt(CuArrayAdaptor{unified ? Mem.UnifiedBuffer : Mem.DeviceBuffer}(), x), xs
  )
end
```
I know we discussed changing this to accept the CUDA Mem types but I think that was on a PR that was rewritten. So I can make an entry in our googledoc to change this in the next PR when I update the GPU backend files. And when I do I will change the variable to `storagemode`

Comment (kmp5VT) [reply NDTensors/ext/NDTensorsAMDGPUExt/adapt.jl]: The boolean system was something they previously had implemented until they created the `Mem.HostBuffer` type. But I have made a note of this and will work on the other GPUs in the next PR. For now, this code is updated to be `storagemode`

File: NDTensors/ext/examples/NDTensorAMDGPU.jl
Comment (mtfishman) [inline NDTensors/ext/examples/NDTensorAMDGPU.jl:1]: Let's delete this example file.

File: NDTensors/ext/NDTensorsAMDGPUExt/adapt.jl [hunk 28]
```diff
@@ -0,0 +1,31 @@
...
+  default_type_parameter,
+  set_type_parameter,
+  set_type_parameters,
+  type_parameter,
+  type_parameters
+using Adapt: Adapt, adapt
+using AMDGPU: AMDGPU, ROCArray, ROCVector
+using Functors: fmap
+
+function AMDGPUExtensions.roc(xs, storagemode=default_type_parameter(ROCArray, storagemode))
```
mtfishman, line 14:
```suggestion
function AMDGPUExtensions.roc(xs; storagemode=default_type_parameter(ROCArray, storagemode))
```

NDTensors/ext/examples/NDTensorAMDGPU.jl
kmp5VT: @mtfishman I'm deleting this now, thanks I wasn't sure what to do with this file.

NDTensors/ext/examples/NDTensorAMDGPU.jl
kmp5VT: The file is deleted, thanks!

mtfishman:
Looks good! Thanks @wbernoudy and @kmp5VT.

PR #1329, opened 2024-02-05
Define checkflux for ProjMPOSum (actually AbstractSum) and test

emstoudenmire:
Define `checkflux` for `AbstractSum` which includes `ProjMPOSum` as a special case. Added test. Meant to address the issue raised in [this forum post](https://itensor.discourse.group/t/bug-in-factorize-qr-leading-to-very-slow-calls-to-orthogonalize/1412/11).

### DIFF
--- a/src/mps/projmposum.jl
+++ b/src/mps/projmposum.jl
@@ -111,6 +111,12 @@ function disk(sum::AbstractSum; disk_kwargs...)
   return set_terms(sum, map(t -> disk(t; disk_kwargs...), terms(sum)))
 end
 
+function checkflux(sum::AbstractSum)
+  for t in terms(sum)
+    checkflux(t)
+  end
+end
+
 #
 # Definition of concrete, generic SequentialSum type
 #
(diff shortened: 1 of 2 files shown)


mtfishman:
Thanks @emstoudenmire, I should have let you know but I'm working on a PR which also include `checkflux` definitions for the other types as well.

emstoudenmire:
I thought we had that auto formatting thing turned on now, but it didn't work in this case, at least not for the file I changed in the tests. Maybe it needs to be extended to cover those? @ogauthe

mtfishman:
Yeah, I also haven't noticed the changes to my git workflow that I was expecting to see from #1321.

mtfishman:
Closed in favor of #1330.

ogauthe:
@emstoudenmire I added the pre-commit config file, then one needs to setup pre-commit locally. Sorry if I was not clear.

1. install `pre-commit` on your computer with e.g. `pip install pre-commit`.
2. In the project directory, run `pre-commit install`
in order for pre-commit to run automatically before `git commit`.

Maybe I should make another PR to add this in the README.md?

mtfishman:
Thanks for the instructions @ogauthe. It would be good to add those instructions somewhere. Maybe it could be added here: https://itensor.github.io/ITensors.jl/dev/faq/Development.html.

PR #1334, opened 2024-02-15
[CI] Add pre-commit guidelines

ogauthe:
add pre-commit guidelines as mentioned in #1329


mtfishman:
Thanks @ogauthe. I think it would be good to also mention the alternative of using `using JuliaFormatter; format(".")` in the directory `~/.julia/dev/ITensors`. In fact, I would mention that first, and then describe the pre-commit hook as an alternative.

mtfishman:
Thanks @ogauthe.

PR #1342, opened 2024-02-25
[Lib][WIP]Create the `TypeParameterAccessors` library which will replace `SetParameters`

kmp5VT:
# Description

A generic library for setting and unsetting type parameters. 
Also includes code which develops the idea of a `wrappertype` and `parenttype` which then trait functions which modify parameters can be written for WrappedArrays and parenttypes

PR #1331, opened 2024-02-05
[WIP][SetParameters] Generalize to work for any type, rename to `TypeParameterAccessors`

kmp5VT:
# Description
Rename the SetParameters package and utilize the type information provided by `DataType` and modify the package to use Base functions to change/access the parameters.

# Checklist:

- [ ] Rework the code to use base functions
- [ ] Verify that the code works with equivalent speed as previous implementation
- [ ] Verify type stability of the code
- [x] Rename the module per https://github.com/ITensor/ITensors.jl/issues/1327

PR #1343, opened 2024-02-25
[NDTensors] `TypeParameterAccessors` submodule

mtfishman:
This is a rewrite of the `TypeParameterAccessors` submodule started in #1331 and #1342.

TODO:
- [x] Investigate type inference issues, which are marked in the tests (the remaining ones are mostly corner cases and will be left for future work).
- [x] Make an `interface.jl` file with functions that are expected to be overloaded.
- [x] Fix `set_eltype` for certain wrapped arrays.
- [x] Define `set_eltype(array::AbstractArray, param) = convert(set_eltype(typeof(array), param), array)`.
- [x] Define `parenttype` and related functionality for handling wrapped array types (like the `is_wrapped_array` trait function and associated trait), like in #1342.
- [x] More tests for custom types that implement different subsets of the interface, like named type parameter positions and default type parameter values.


NDTensors/src/lib/TypeParameterAccessors/src/parameter.jl
```diff
@@ -0,0 +1,4 @@
+parameter(type::Type, pos) = parameter(type, position(type, pos))
+parameter(type::Type, pos::Position) = parameters(type)[Int(pos)]
+parameter(type::Type, pos::Int) = parameter(type, Position(pos))
```
kmp5VT, line 3: I don't think this function is necessary since you have this defined 
```
position(type::Type, pos::Int) = Position(pos)
```

mtfishman: Fixed, thanks.

NDTensors/src/lib/TypeParameterAccessors/src/base/array.jl
```diff
@@ -0,0 +1,12 @@
+position(::Type{<:Array}, ::typeof(eltype)) = Position(1)
+position(::Type{<:Array}, ::typeof(ndims)) = Position(2)
+
+default_type_parameters(::Type{<:Array}) = (Float64, 1)
+
+function set_eltype(type::Type{<:Array}, param)
```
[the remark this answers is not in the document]
mtfishman: Good point, I'll remove this one.

NDTensors/src/lib/TypeParameterAccessors/src/base/array.jl
```diff
@@ -0,0 +1,12 @@
...
+  return set_parameter(type, eltype, param)
+end
+
```
mtfishman, line 9:
```suggestion
```

NDTensors/src/lib/TypeParameterAccessors/src/unspecify_parameters.jl
```diff
@@ -0,0 +1,45 @@
...
+
+# Like `set_parameters` but less strict, i.e. it allows
+# setting with `TypeVar` while `set_parameters` would error.
+function new_parameters(type::Type, params)
+  return to_unionall(unspecify_parameters(type){params...})
+end
+
+function _unspecify_parameter(type::Type, pos::Int)
+  !is_parameter_specified(type, pos) && return type
+  unspecified_param = parameter(unspecify_parameters(type), pos)
```
kmp5VT, line 17: Does it matter that you get the `TypeVar` for specifically the position `pos`, I could imagine just grabbing the first one is sufficient. I.e. does the "name" of the typevar get written into the Type. like 
```
v = unspecify_parameter(Vector{Float32}, 2) #= Array{Float32}
v.body.parameters = svec(Float32, N)
```
I think it should say N regardless of which TypeVar you take

NDTensors/src/lib/TypeParameterAccessors/src/unspecify_parameters.jl
```diff
@@ -0,0 +1,45 @@
...
+  params = Base.setindex(parameters(type), unspecified_param, pos)
+  return new_parameters(type, params)
+end
+@generated function unspecify_parameter(type_type::Type, pos_type::Position)
+  type = parameter(type_type)
+  pos = parameter(pos_type)
```
kmp5VT, line 23: In places like this, would it be useful to use `Int` since its defined?

NDTensors/src/lib/TypeParameterAccessors/src/unspecify_parameters.jl
```diff
@@ -0,0 +1,45 @@
...
+  for pos in positions
+    type = unspecify_parameter(type, pos)
+  end
+  return type
+end
+@generated function unspecify_parameters(
+  type_type::Type, positions_type::Tuple{Vararg{Position}}
+)
+  type = parameter(type_type)
+  positions = parameter.(parameters(positions_type))
```
kmp5VT, line 40: Same here `Int.(parameters(positions_type))`?

NDTensors/src/lib/TypeParameterAccessors/test/test_basics.jl
```diff
@@ -0,0 +1,77 @@
...
+    @test_throws ErrorException type_parameter(Array, 1)
+    @test_inferred type_parameter(Array{Float64}, eltype) == Float64
+    @test_inferred type_parameter(Matrix{Float64}, ndims) == 2
+    @test_throws ErrorException type_parameter(Array{Float64}, ndims) == 2
+    @test_inferred type_parameters(Matrix{Float64}, (2, eltype)) == (2, Float64) wrapped =
+      true
+    @test_inferred type_parameters(Matrix{Float64}, (Position(2), eltype)) == (2, Float64)
+  end
+  @testset "Set parameters" begin
+    @test_inferred set_parameter(Array, 1, Float64) == Array{Float64} wrapped = true
```
kmp5VT, line 30: does this fail if `wrapped = false`

NDTensors/src/lib/TypeParameterAccessors/test/test_wrappers.jl
```diff
@@ -0,0 +1,112 @@
...
+    for wrapped_array in (
+      Base.ReshapedArray(array, (2, 2), ()),
+      SubArray(randn(2, 2), (:, :)),
+      PermutedDimsArray(randn(2, 2), (1, 2)),
+    )
+      wrapped_array_type = typeof(wrapped_array)
+      @test parenttype(wrapped_array) == Matrix{Float64}
+      @test is_wrapped_array(wrapped_array)
+      @test unwrap_array_type(wrapped_array_type) == Matrix{Float64}
+      # TODO: Test `set_eltype`.
```
kmp5VT, line 48: Reminder to add this test in potentially.

NDTensors/src/lib/TypeParameterAccessors/src/unspecify_parameters.jl
mtfishman: I think it is important to grab the right `TypeVar`. `TypeVar`s also encode type constraints, for example:
```julia
julia> struct MyType{T<:Real,N} end

julia> MyType.body.body.parameters
svec(T<:Real, N)

julia> Base.typename(MyType).wrapper{TypeVar(:X)}.body.parameters
svec(X, N)
```
so you can see if you don't grab the right one you can lose information.

NDTensors/src/lib/TypeParameterAccessors/src/unspecify_parameters.jl
mtfishman: We could use that, though I was trying to have these generated functions follow a generic structure, where it just preprocesses the types of the inputs into the correct form needed by the inner function in the most general way possible. Here, in principle the `Position` could be wrapping something besides an `Int`, so I think it is best to just leave it more generic/general.

NDTensors/src/lib/TypeParameterAccessors/src/unspecify_parameters.jl
mtfishman: Same comment above, I think it is a matter of style mostly, but I think here it is best to leave it more generic/general.

NDTensors/src/lib/TypeParameterAccessors/test/test_basics.jl
mtfishman: Yes, type inference for most of these functions with runtime inputs (like the `1` here) fail without wrapping into a function call. I think that is just how Julia's compiler works:
```julia
julia> @code_warntype (Float32, Float64, 3)[2]
MethodInstance for getindex(::Tuple{DataType, DataType, Int64}, ::Int64)
  from getindex(t::Tuple, i::Int64) @ Base tuple.jl:31
Arguments
  #self#::Core.Const(getindex)
  t::Tuple{DataType, DataType, Int64}
  i::Int64
Body::Union{Int64, DataType}
1 ─      nothing
│   %2 = Base.getfield(t, i, $(Expr(:boundscheck)))::Union{Int64, DataType}
└──      return %2


julia> @code_warntype (() -> (Float32, Float64, 3)[2])()
MethodInstance for (::var"#6#7")()
  from (::var"#6#7")() @ Main REPL[16]:1
Arguments
  #self#::Core.Const(var"#6#7"())
Body::Type{Float64}
1 ─ %1 = Core.tuple(Main.Float32, Main.Float64, 3)::Core.Const((Float32, Float64, 3))
│   %2 = Base.getindex(%1, 2)::Core.Const(Float64)
└──      return %2
```
I believe that technically the `1` is a global variable, so then it can't be turned into a `Const` by the compiler.

NDTensors/src/lib/TypeParameterAccessors/test/test_wrappers.jl
mtfishman: I added these missing tests.

PR #1345, opened 2024-02-27
[NDTensors] Remove FLoops.jl

MasonProtter:
This package doesn't appear to be used at all, I think it seems to have been replaced by Folds.jl.


mtfishman:
Thanks @MasonProtter, I think I was trying it out in our block sparse multithreading code but ended up going with `Folds.jl`.

Is `FLoops.jl` going to be supported going forward?

MasonProtter:
Yes I would like to continue to support it, but I haven't found the time yet to do the initial push. The main thing is I need to port it over to JuliaFolds2, make its test suite work in a more standard way, and then update the BangBang dependency.

PR #1348, opened 2024-02-29
[TypeParameterAccessors] Update function names from `X_parameter` to `X_type_parameter`

kmp5VT:
# Description

Update names

### DIFF
--- a/NDTensors/src/lib/TypeParameterAccessors/src/base/abstractarray.jl
+++ b/NDTensors/src/lib/TypeParameterAccessors/src/base/abstractarray.jl
@@ -1,7 +1,7 @@
 struct Self end
 
 parameter(type::Type, pos::Self) = type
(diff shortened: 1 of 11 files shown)


kmp5VT:
> I think we should also change `unspecify_parameter[s]` to `unspecify_type_parameter[s]`.

I will change that now, I wasn't sure. 
Do you think its fine to keep internal functions like `_specify_parameter` as they are or should I rename those as well?

mtfishman:
> > I think we should also change `unspecify_parameter[s]` to `unspecify_type_parameter[s]`.
> 
> I will change that now, I wasn't sure. Do you think its fine to keep internal functions like `_specify_parameter` as they are or should I rename those as well?

I think those should be updated too.

mtfishman:
Thanks!

PR #1350, opened 2024-03-08
[NDTensors] Remove `unwrap_array` from `Unwrap` and rename

kmp5VT:
This is my first refactor PR where I am only dealing with `unwrap_array` and renaming `Unwrap` to something that makes more sense.

# Checklist:

- [x] Use `TypeParameterAccessors.unwrap_array_type` over `Unwrap.unwrap_array`
- [x] Define parenttype for `Tensors`, `StorageTypes`, and `ITensors`
- [x] All NDTensors and ITensors test pass.
- [x] Fix `NDtensors.mtl` it was not converting objects properly.
- [x] Rename `Unwrap` module


NDTensors/src/empty/EmptyTensor.jl
```diff
@@ -96,6 +96,9 @@ fill!!(T::EmptyTensor{EmptyNumber}, α::Number) = _fill!!(eltype(α), T, α)
 
 isempty(::EmptyTensor) = true
 
+zero(type::Type{<:EmptyTensor}) = type
+zero(empty::EmptyTensor) = empty
```
mtfishman, line 100: Could you explain these definitions?

The definition `zero(::EmptyTensor)` makes sense to me, but the definition for types (`zero(::Type{<:EmptyTensor})`) doesn't make sense to me. In general, `zero` should output an object which is the additive identity of the input, i.e. `x + zero(x) == x`.

We could follow the behavior of `FillArrays.Zeros`:
```julia
julia> using FillArrays

julia> zero(Zeros(2, 2))
2×2 Zeros{Float64}

julia> zero(Zeros{Float64})
ERROR: MethodError: no method matching zero(::Type{Zeros{Float64}})
...
```

NDTensors/src/empty/empty.jl
```diff
@@ -91,6 +91,10 @@ function Base.show(io::IO, mime::MIME"text/plain", S::EmptyStorage)
   return println(io, typeof(S))
 end
 
+using .TypeParameterAccessors: TypeParameterAccessors
+TypeParameterAccessors.parenttype(empty::Type{<:EmptyStorage}) = storagetype(empty)
+zero(type::Type{<:EmptyStorage}) = type
+zero(empty::EmptyStorage) = empty
```
mtfishman, line 97: Same comment as above.

src/set_types.jl
```diff
@@ -1 +1,2 @@
-NDTensors.Unwrap.unwrap_type(T::ITensor) = unwrap_type(tensor(T))
+#NDTensors.Unwrap.unwrap_type(T::ITensor) = unwrap_type(tensor(T))
```
mtfishman, line 1: Reminder to delete this.

NDTensors/ext/NDTensorsMetalExt/adapt.jl
```diff
@@ -6,6 +6,7 @@ end
 
 # More general than the version in Metal.jl
 function Adapt.adapt_storage(arraytype::Type{<:MtlArray}, xs::AbstractArray)
-  arraytype_specified = specify_parameters(arraytype, get_parameters(xs))
+  params = get_parameters(xs)
+  arraytype_specified = specify_parameters(arraytype, params[1], params[2], params[3:end])
```
mtfishman, line 10: I'm confused by this rewrite. What's the advantage of the new way you wrote it?

Also, there seems to be a few issues here:
1. `get_parameters` is the syntax from `SetParameters`, I think you should be using `type_parameters`.
2. It seems like this isn't general enough. For example, `xs` may be an array type that doesn't have type parameters that match those of `MtlArray` (for example, maybe it is a CPU array so doesn't have a storage mode/buffer type, maybe it is a wrapper array that doesn't have an `ndims` or even an `eltype` field, etc. One way I could imagine handling that is a pattern like this:
```julia
param_names = (eltype, ndims, storagemode)
params = type_parameters(xs, param_names)
arraytype_specified = specify_parameters(arraytype, param_names, params)
```
This may not be quite right, since we would need to check if `xs` in fact has a storage mode, and if not, we should use `UnspecifiedTypeParameter`. I think right now `type_parameters` will error if there is no `storagemode` defined. I forget if we have a more lenient alternative to `type_parameters` that either outputs the type parameters or `UnspecifiedTypeParameter`, but if not we should add that `TypeParameterAccessors`.

mtfishman: Another option would be to define `parameter_names(xs)` which would output just the parameter names that are valid for `xs`, for example it could output `(eltype, ndims)` if `xs isa Array`. Then it could be written:
```julia
param_names = parameter_names(xs)
params = type_parameters(xs, param_names)
arraytype_specified = specify_parameters(arraytype, param_names, params)
```
However, that may not be the best approach. For example, `xs` may be a matrix wrapper type like `Hermitian` which doesn't have an `ndims` parameter field, but `ndims` is implicitly `2`:
```julia
julia> using LinearAlgebra

julia> typeof(Hermitian(randn(2, 2)))
Hermitian{Float64, Matrix{Float64}}

julia> ndims(typeof(Hermitian(randn(2, 2))))
2
```
so that wouldn't quite work, since `parameter_names(xs::Hermitian)` would be `(eltype, parenttype)`, and not have an `ndims` name. But that kind of idea could be used to make the code from the previous post a bit more general:
```julia
param_names = parameter_names(arraytype)
params = type_parameters(xs, param_names)
arraytype_specified = specify_parameters(arraytype, param_names, params)
```

mtfishman: Anyway, maybe we don't need to address this issue in full generality in this PR, and just focus on making the code work as well as it did before the PR, and fix this in a future PR.

mtfishman: Another thing, isn't this definition type piracy?

NDTensors/src/empty/EmptyTensor.jl
kmp5VT: Right so for some reason when I migrated to `TypeParameterAccessors.unwrap_array_type` the function `Zero(::EmptyStorage)` had some issue with preserving the type of the storagetype in these tests https://github.com/ITensor/ITensors.jl/blob/9413240cc703a17abf6218951ba1171f3096802e/NDTensors/test/test_emptystorage.jl#L28C1-L32C32
This was a quick fix to this problem but it doesn't look like the `Type` function is necessary and doesn't make sense based on your logic so I will remove it. Its a quick fix because we are ultimately removing  these objects, especially the `EmptyStorage` type

NDTensors/src/empty/empty.jl
kmp5VT: I removed the `Type` function

src/set_types.jl
kmp5VT: Deleted, thanks!

NDTensors/ext/NDTensorsMetalExt/adapt.jl
kmp5VT: Thank you for your messages. Currently in the `main` branch of ITensors  this function`NDTensor.mtl` is **broken**. This is an effort to patch the problem without significantly modifying how the code is written because in subsequent PRs I will be modifying the code in general.
In this PR I am not migrating the code to use `TypeParameterAccessors` beyond `unwrap_array_type` and `parenttype` to limit the scope. Because of this these changes are simply a short term patch to get the code working again. In the Next PR I will be changing this code since I will be removing `SetParameters`and I can integrate the changes in `TypeParameterAccessors` to make this code more generic and robust. But again for now this is a short term change just to get the metal code working enough to run the unittests.

mtfishman: I see, thanks for the explanation. Makes sense to do this in stages as you outline. Though I'm still confused by why the change you made is better than the previous version.

kmp5VT: The issue that I was seeing is that the function 
```
arraytype_specified = specify_parameters(arraytype, get_parameters(xs))
```
Which ultimately calls this 
```
function set_parameters(
  set_parameter_function::Function,
  type::Type,
  position::Position,
  parameter1,
  parameters_tail...,
)
  new_type = set_parameter_function(type, position, parameter1)
  new_position = Position(get_parameter(position) + 1)
  return set_parameters(set_parameter_function, new_type, new_position, parameters_tail...)
end
```
The problem was `get_parameters(...)` was returning a tuple so for some reason `parameter1` was being interpreted as a tuple
```
using NDTensors.SetParameters
f(param1, param2...) = println("$(param1)")
f(get_parameters(Vector{Float32})) # = (Float32, 1)
```
and this was incorrect. This shouldn't be a problem in our new system.

mtfishman: I see, so can you just write it as:
```julia
  params = get_parameters(xs)
  arraytype_specified = specify_parameters(arraytype, params...)
```
?

kmp5VT: Yes this works so I updated the code to match your suggestion. Thanks!

NDTensors/src/dense/set_types.jl
```diff
@@ -31,3 +32,7 @@ SetParameters.set_parameter(::Type{<:Dense}, ::Position{2}, P2) = Dense{<:Any,P2
 function SetParameters.set_parameter(::Type{<:Dense{P1}}, ::Position{2}, P2) where {P1}
   return Dense{P1,P2}
 end
+function TypeParameterAccessors.position(::Type{<:Dense}, ::typeof(parenttype))
```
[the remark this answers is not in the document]
mtfishman: Sure, we should remove anything that isn't being used.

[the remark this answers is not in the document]
kmp5VT: Okay I removed it. Thanks!

kmp5VT:
@mtfishman This should be ready for you to review when you have time.  If it is merged I will begin working on the 2nd PR note from the Drive document. Thank you!

mtfishman:
Looks good, thanks!

PR #1332, opened 2024-02-09
[BlockSparseArrays] More general broadcasting and slicing

mtfishman:
This is work in progress for defining `map!` for `BlockSparseArray`s with mismatched blocking.

This is functionality that is required to perform block fusion, as a follow-up to #1326.

So far, it performs some unnecessary copying, and also only works with one input array, which both should be relatively easy fixes.

- [x] Add tests.
- [x] For out-of-place expressions like `a + b`, currently it outputs a block sparse array with `axes(a)`. If the axes/blocking of `a` and `b` are different, it should combine the axes (i.e. output a block sparse array with a finer blocking structure compatible with the blocking of `a` and `b`).
- [x] `map(+, a, b)` outputs a dense block array, not a block sparse array.
- [x] Fix issues with using `GradedAxes` as axes of block sparse arrays and performing mapping/broadcasting operations. There are some lingering bugs with that, and we need to make sure we preserve the sectors of the graded axes properly.
- [x] Generalize new `map!` functionality to arbitrary number of input arrays.
- [x] Avoid extraneous copying of source blocks (requires fixing a bug when using views of a `BlockSparseArray`).
- [x] Investigate some other slicing errors I came across during this PR, either fix or add broken tests.

A demonstration of new functionality:
```julia
using BlockArrays: Block, blockedrange, blocksize
using NDTensors.BlockSparseArrays: BlockSparseArray
using Random: randn!

function blockdiagonal(f!, elt::Type, axes::Tuple)
  a = BlockSparseArray{elt}(axes)
  for i in 1:minimum(blocksize(a))
    b = Block(ntuple(Returns(i), ndims(a)))
    a[b] = f!(a[b])
  end
  return a
end

elt = Float64
d_src = blockedrange([4, 4])
d_dest = blockedrange([2, 4, 2])
a_src = blockdiagonal(randn!, elt, (d_src, d_src))
a_dest = blockdiagonal(randn!, elt, (d_dest, d_dest))

a_dest .= 10 .* a_src
```
which outputs:
```julia
julia> a_src
[...]

PR #1351, opened 2024-03-12
[NDTensors] Introduce `LabelledNumbers` and `GradedAxesNext`

mtfishman:
This is a rewrite and redesign of `GradedAxes`, for now called `GradedAxesNext` but the goal will be to remove `GradedAxes` and replace it with the code in `GradedAxesNext` (either here or in a follow-up PR). The goal is to make it easier to implement slicing operations of graded unit ranges which preserve the sector labels (which is needed for finishing #1332), which I have found is a bit more cumbersome than I think is necessary with the current `GradedAxes` design.

@emstoudenmire @ogauthe


mtfishman:
@ogauthe this doesn't have all of the features of `GradedAxes`, but it has a lot more support for slicing (see the tests). I will merge this as-is and we can start adding more features to it, like fusion, and plan to get ride of `GradedAxes` asap.

ogauthe:
Great. I paused my work on `GradedAxes` to wait for your updates.

PR #1352, opened 2024-03-12
[NDTensors] Reorganize some GPU functions

kmp5VT:
# Description
Create a `extensions` folder in NDTensors and define some functions for GPU extensions

# Checklist:

- [x] Decide on storagemode/buffertype
- [x] AbstractGPUArray extensions
- [ ] Generic adapt_storagetype
- [x] Default parameters for GPUs
- [x] Create extensions folder
- [x] All tests pass


NDTensors/src/extensions/cuda.jl
mtfishman, line 1: We could move this to a new library `NDTensors/src/lib/CUDAExtensions`, where we put anything that is extending the behavior of CUDA.jl and doesn't rely on NDTensors.

NDTensors/src/extensions/metal.jl
mtfishman, line 1: Same as above, maybe move to a new library `NDTensors/src/lib/MetalExtensions.jl`.

kmp5VT: Do you think this should this be a file or folder like the `CUDAExtensions`

NDTensors/src/extensions/gpuarrayscore.jl
```diff
@@ -0,0 +1,12 @@
+using .TypeParameterAccessors: TypeParameterAccessors, type_parameter
+using GPUArraysCore: AbstractGPUArray
+using Adapt: Adapt
+
+TypeParameterAccessors.default_type_parameters(::Type{<:AbstractGPUArray}) = (Float32, 1)
```
mtfishman, line 5: I'm not sure if this is a good idea since I don't think there is any guarantee that subtypes will have that type parameter convention.

NDTensors/src/extensions/gpuarrayscore.jl
```diff
@@ -0,0 +1,12 @@
+using .TypeParameterAccessors: TypeParameterAccessors, type_parameter
+using GPUArraysCore: AbstractGPUArray
+using Adapt: Adapt
```
mtfishman, line 3: I don't see `Adapt` being used here.

NDTensors/src/extensions/gpuarrayscore.jl
mtfishman, line 1: Maybe move to `NDTensors/src/lib/GPUArraysCoreExtensions`.

NDTensors/src/extensions/gpuarrayscore.jl
mtfishman: That's a downside of the current `default_type_parameters`, it relies on type parameter positions instead of names, so it only works for concrete types, so we can't really make definitions for all subtypes of abstract types.

NDTensors/src/extensions/metal.jl
mtfishman: I meant it to be a folder, that was a typo.

kmp5VT: Great thanks!

NDTensors/src/extensions/cuda.jl
kmp5VT: Okay I added this folder and moved cuda.jl into it

NDTensors/src/extensions/metal.jl
kmp5VT: Same as cuda

NDTensors/src/extensions/gpuarrayscore.jl
kmp5VT: Made this folder and moved the file into it

NDTensors/src/extensions/gpuarrayscore.jl
kmp5VT: Ahh yes this makes sense. So I should define this for `CuArrays` and `MetalArrays`?

NDTensors/ext/NDTensorsMetalExt/adapt.jl
```diff
@@ -6,7 +7,10 @@ end
 
 # More general than the version in Metal.jl
 function Adapt.adapt_storage(arraytype::Type{<:MtlArray}, xs::AbstractArray)
-  params = get_parameters(xs)
-  arraytype_specified = specify_parameters(arraytype, params...)
+  arraytype_specified = TypeParameterAccessors.specify_type_parameters(
+    arraytype,
+    (eltype, Base.ndims),
+    TypeParameterAccessors.type_parameters(xs, (eltype, Base.ndims)),
```
kmp5VT, line 13: @mtfishman I am not sure if this exactly is what you had in mind in with respect to this comment 
```
More general Adapt.adapt_storage definitions for GPU array types based on TypeParameterAccessors (see discussion in https://github.com/ITensor/ITensors.jl/pull/1350#discussion_r1517835891). This may involve defining a type_parameter_names function in TypeParameterAccessors as well as a less stringent version of type_parameters which outputs UnspecifiedTypeParameter if a parameter doesn’t exist, as opposed to erroring, which I think is the current behavior. It could be called something like maybe_unspecified_type_parameters.
```

mtfishman:
The new `NDTensors/src/lib/*Extensions` libraries should be organized like the other libraries in `NDTensors/src/lib`, with `src` and `test` directories and source code organized into a module `*Extensions`.

NDTensors/src/extensions/gpuarrayscore.jl
mtfishman: Yeah.

kmp5VT: Got it, I made a defaults for both of them separately. Thanks!

NDTensors/ext/NDTensorsMetalExt/adapt.jl
kmp5VT: I am reverting these changes as we just discussed. Thanks

NDTensors/ext/NDTensorsCUDAExt/adapt.jl
```diff
@@ -1,3 +1,5 @@
+using NDTensors.TypeParameterAccessors: TypeParameterAccessors
+using NDTensors: storagemode
```
mtfishman, line 2:
```suggestion
using NDTensors.GPUArraysCoreExtensions: storagemode
```

NDTensors/ext/NDTensorsCUDAExt/adapt.jl
```diff
@@ -11,18 +13,22 @@ function cu(xs; unified::Bool=false)
   )
 end
 
-buffertype(::NDTensorCuArrayAdaptor{B}) where {B} = B
+function TypeParameterAccessors.position(
+  ::Type{<:NDTensorCuArrayAdaptor}, ::typeof(NDTensors.storagemode)
```
mtfishman, line 17: We should rename this from `NDTensorCuArrayAdaptor` to `CuArrayAdaptor` and move it to `CUDAExtensions`.

NDTensors/ext/NDTensorsCUDAExt/set_types.jl
```diff
@@ -38,3 +38,18 @@ default_parameter(::Type{<:CuArray}, ::Position{3}) = Mem.DeviceBuffer
...
 SetParameters.unspecify_parameters(::Type{<:CuArray}) = CuArray
+
+using NDTensors.TypeParameterAccessors: TypeParameterAccessors
+function TypeParameterAccessors.position(::Type{<:CuArray}, ::typeof(eltype))
+  return TypeParameterAccessors.Position(1)
+end
+function TypeParameterAccessors.position(::Type{<:CuArray}, ::typeof(Base.ndims))
+  return TypeParameterAccessors.Position(2)
+end
+function TypeParameterAccessors.position(::Type{<:CuArray}, ::typeof(NDTensors.storagemode))
```
mtfishman, line 49:
```suggestion
function TypeParameterAccessors.position(::Type{<:CuArray}, ::typeof(storagemode))
```

NDTensors/ext/NDTensorsCUDAExt/set_types.jl
```diff
@@ -38,3 +38,18 @@ default_parameter(::Type{<:CuArray}, ::Position{3}) = Mem.DeviceBuffer
 nparameters(::Type{<:CuArray}) = Val(3)
 
 SetParameters.unspecify_parameters(::Type{<:CuArray}) = CuArray
+
+using NDTensors.TypeParameterAccessors: TypeParameterAccessors
+function TypeParameterAccessors.position(::Type{<:CuArray}, ::typeof(eltype))
+  return TypeParameterAccessors.Position(1)
```
mtfishman, line 44:
```suggestion
  return Position(1)
```
and add `Position` to the `using` statement.

NDTensors/ext/NDTensorsMetalExt/adapt.jl
```diff
@@ -1,3 +1,4 @@
+using NDTensors.TypeParameterAccessors: TypeParameterAccessors
 NDTensors.cpu(e::Exposed{<:MtlArray}) = adapt(Array, e)
```
mtfishman, line 2: Probably we should move `cpu` to `GPUArraysCoreExtensions`.

NDTensors/ext/NDTensorsMetalExt/set_types.jl
```diff
@@ -39,9 +39,26 @@ default_parameter(::Type{<:MtlArray}, ::Position{3}) = Metal.DefaultStorageMode
...
-function set_storagemode(arraytype::Type{<:MtlArray}, storagemode)
-  return set_parameter(arraytype, Position(3), storagemode)
+function set_storagemode(arraytype::Type{<:MtlArray}, param)
+  return TypeParameterAccessors.set_type_parameter(arraytype, NDTensors.storagemode, param)
 end
 
 SetParameters.unspecify_parameters(::Type{<:MtlArray}) = MtlArray
+
+function TypeParameterAccessors.position(::Type{<:MtlArray}, ::typeof(eltype))
+  return TypeParameterAccessors.Position(1)
```
mtfishman, line 51:
```suggestion
  return Position(1)
```
and same thing elsewhere.

NDTensors/ext/NDTensorsMetalExt/set_types.jl
```diff
@@ -39,9 +39,26 @@ default_parameter(::Type{<:MtlArray}, ::Position{3}) = Metal.DefaultStorageMode
...
+end
+function TypeParameterAccessors.position(::Type{<:MtlArray}, ::typeof(Base.ndims))
+  return TypeParameterAccessors.Position(2)
+end
+function TypeParameterAccessors.position(
+  ::Type{<:MtlArray}, ::typeof(NDTensors.storagemode)
```
mtfishman, line 57:
```suggestion
  ::Type{<:MtlArray}, ::typeof(storagemode)
```
and same thing elsewhere.

NDTensors/src/imports.jl
```diff
@@ -5,6 +5,10 @@
 # https://github.com/oschulz/HeterogeneousComputing.jl
 using .Expose: cpu
 
+import .CUDAExtensions: cu
+import .GPUArraysCoreExtensions: storagemode
+import .MetalExtensions: mtl
```
mtfishman, line 10: Let's not import these, and overload them with the syntax `CUDAExtensions.cu(...) = ...` as needed.

NDTensors/src/lib/GPUArraysCoreExtensions/test/runtests.jl
```diff
@@ -0,0 +1,7 @@
+@eval module $(gensym())
+using Test: @testset, @test
+using NDTensors.GPUArraysCoreExtensions
```
mtfishman, line 3:
```suggestion
using NDTensors.GPUArraysCoreExtensions: storagemode
```
and same thing elsewhere. We should use that style of explicitly listing functions being brought into the namespace in general.

kmp5VT: Okay i have updated this across the library, thanks

NDTensors/src/lib/CUDAExtensions/src/CUDAExtensions.jl
```diff
@@ -0,0 +1,5 @@
+module CUDAExtensions
+include("cuda.jl")
+
+export cu
```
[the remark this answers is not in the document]
kmp5VT: So that would mean externally to use this function you would now need to write `NDTensors.CUDAExtensions.cu` is that okay?

NDTensors/src/imports.jl
kmp5VT: Okay I removed these imports. I only did this so that `mtl` and `cu` could live in `NDTensors` namespace, see the comment in `CUDAExtensions`

NDTensors/ext/NDTensorsMetalExt/set_types.jl
kmp5VT: This is updated across the board thanks!

NDTensors/ext/NDTensorsCUDAExt/set_types.jl
kmp5VT: Fixed now, thanks!

NDTensors/ext/NDTensorsCUDAExt/adapt.jl
kmp5VT: This is updated, thanks!

NDTensors/ext/NDTensorsMetalExt/adapt.jl
kmp5VT: Also I think that maybe `cpu` is a bad name for this operation because it doesn't account for the fact that there are many types of CPU Arrays like `Array`, `BlockSparseArray`, `DistributedArray` ... Potentially we want to make a `cpu(device_array, host_type = Array)` function in `GPUArraysCoreExtensions` to be more generic? 
This brings me to the thought that someone might try to do this 
```
cpu(Array, DistributedArray)
```
Which is still just a wrapper for `Adapt` but then the function wouldn't specifically match in `GPUArraysCoreExtensions` so we could do something like this instead `cpu(device_array::AbstractGPUArray, host_type = Array)`   Do you have any thoughts on this?

NDTensors/ext/NDTensorsMetalExt/set_types.jl
kmp5VT: Same as before, I am waiting until the next PR to make this switch. Thanks!

NDTensors/src/lib/CUDAExtensions/src/CUDAExtensions.jl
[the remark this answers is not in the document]
mtfishman: If the concern is that it is a breaking change since users have been using `NDTensors.cu`, we can put `using .CUDAExtensions: cu` somewhere in the `NDTensors` module to still allow that syntax.

kmp5VT:
As discussed I am moving all of the `using` statements from `NDTensors.jl` to `import.jl` and defining things like `using .CUDAExtensions: cu` so operations like `NDTensors.cu` still work.  
I am also going to move `cpu` from `Expose` to `GPUArraysCoreExtension`

kmp5VT:
@mtfishman I pushed my changes here, I am stepping out this afternoon and will look through the changes later this afternoon to verify that everything is inline. Thanks!

NDTensors/test/ITensors/TestITensorDMRG/TestITensorDMRG.jl
```diff
@@ -4,14 +4,15 @@
 module TestITensorDMRG
 using ITensors
 using NDTensors
+using NDTensors.CUDAExtensions: cu
```
kmp5VT, line 7: @mtfishman I have now in `NDTensors/imports.jl`, `using .CUDAExtensions: cu` so I could have this (and other statements) written as `NDTensors: cu` Do you have a preference on which one to prefer?

mtfishman: I think it is better the way you currently wrote it. In general, it is best to load a function into a namespace from where it is actually defined, as you did here.

NDTensors/ext/NDTensorsCUDAExt/set_types.jl
kmp5VT: Okay I'll make sure to do that next time, I also added comments in the code

NDTensors/src/lib/CUDAExtensions/src/CUDAExtensions.jl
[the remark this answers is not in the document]
kmp5VT: I believe this is resolved, we added `using NDTensors.CUDAExtensions.cu` to `imports.jl`

NDTensors/ext/NDTensorsMetalExt/adapt.jl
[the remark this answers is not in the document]
kmp5VT: Okay I added a comment there, thanks!

NDTensors/ext/NDTensorsMetalExt/adapt.jl
mtfishman: I think `cpu` serves a clear purpose for single-processor calculations, to generically move data from the device to the host (the inverse of the functions `cu`, `mtl`, etc.). The same argument you make about it not being general enough for multi-processor environments can also be applied to the functions `cu`, `mtl`, etc. In terms of multi-processor environments, we will have to think about how all of those functions will need to be extended, but that is beyond the scope of this PR.

PR #1353, opened 2024-03-14
[NDTensors] Remove SetParameters and replace with TypeParameterAccessors

kmp5VT:
# Description

In this PR I am trying to remove the `SetParameters` module and replace it with the `TypeParameterAccessors` version.  The goal here is not to completely rewrite the systems which exist for `TensorStorage` and `Tensor` types because these will eventually be removed. For these objects I will try to apply a patch which minimizes the number of changes.  I will also try to move `ndims` from `NDTensors` to `TypeParameterAccessors`. In this PR I will not try to rewrite functions with `TypeParameterAccessors` in the future we can try to simplify functions with the new functionality of this module

# Checklist:

- [x] Set Parameters is removed 
- [x] All unittests pass with `TypeParameterAccessors` definitions 
- [x] ndims function is moved to `TypeParameterAccessors`


NDTensors/ext/NDTensorsCUDAExt/set_types.jl
```diff
@@ -1,57 +1,21 @@
...
 end
 function TypeParameterAccessors.position(::Type{<:CuArray}, ::typeof(storagemode))
-  return TypeParameterAccessors.Position(3)
+  return Position(3)
 end
 
 function TypeParameterAccessors.default_type_parameters(::Type{<:CuArray})
   return (Float64, 1, CUDA.Mem.DeviceBuffer)
 end
+
```
kmp5VT, line 18: Because there is no `set_ndims(::Type{<:AbstractArray}` I defined one for `CuArray` and `MtlArray`. Potentially that is not right? I also have not yet moved `NDTensors.ndims` to `TypeParameterAccessors`

NDTensors/ext/NDTensorsMetalExt/adapt.jl
```diff
@@ -10,7 +11,7 @@ end
 # More general than the version in Metal.jl
 ## TODO Rewrite this using a custom `MtlArrayAdaptor` which will be written in  `MetalExtensions`.
 function Adapt.adapt_storage(arraytype::Type{<:MtlArray}, xs::AbstractArray)
-  params = get_parameters(xs)
-  arraytype_specified = specify_parameters(arraytype, params...)
+  params = type_parameters(xs)
+  arraytype_specified = specify_type_parameters(arraytype, params)
```
kmp5VT, line 15: TODO I have to check if metal is working. I have only verified CUDA

NDTensors/src/abstractarray/set_types.jl
```diff
@@ -11,7 +11,7 @@ TODO: Use `Accessors.jl` notation:
 # TODO: Delete this when we change to using a
 # `FillArray` instead. This is a stand-in
 # to make things work with the current design.
-function SetParameters.set_ndims(numbertype::Type{<:Number}, ndims)
+function TypeParameterAccessors.set_ndims(numbertype::Type{<:Number}, ndims)
```
kmp5VT, line 14: Potentially this should be in `TypeParametersAccessors`

NDTensors/src/blocksparse/blocksparse.jl
```diff
@@ -23,7 +23,7 @@ function set_datatype(storagetype::Type{<:BlockSparse}, datatype::Type{<:Abstrac
   return BlockSparse{eltype(datatype),datatype,ndims(storagetype)}
 end
 
-function SetParameters.set_ndims(storagetype::Type{<:BlockSparse}, ndims::Int)
+function TypeParameterAccessors.set_ndims(storagetype::Type{<:BlockSparse}, ndims::Int)
```
kmp5VT, line 26: This kind of code could be updated with `TypeParametersAccessors`. Should I make the update here?

mtfishman: Let's just leave it as-is.

NDTensors/src/abstractarray/set_types.jl
mtfishman: I would lean towards leaving this here since it will get removed later anyway, since it only exists to serve the quirky design of the `Diag` type.

NDTensors/ext/NDTensorsCUDAExt/set_types.jl
mtfishman: I think this can be turned into a generic `AbstractArray` function based on the way it is written here.

NDTensors/ext/NDTensorsMetalExt/adapt.jl
kmp5VT: I verified that metal passes

NDTensors/ext/NDTensorsCUDAExt/set_types.jl
kmp5VT: I guess one thing to think about with this is what happens when you want to do 
```
set_ndims(Transpose{Float32, Matrix{Float32}}, 3)
```
Should ndims be based on `IsWrappedArray`? But then an issue I have had before is that `TensorStorage` is a wrapped array so `set_ndims` of a tensor storage will try to set the dimension of the inner data which right now is required to be a vector.

kmp5VT:
I have this now working on all GPUs and CPU. I just need to look into moving `ndims` to `TypeParameterAccessors`

NDTensors/ext/NDTensorsCUDAExt/set_types.jl
mtfishman: `set_ndims` for `Transpose` will fail since it doesn't have an `ndims` type parameter. I don't think we have to worry about `set_ndims` for wrapped array types right now.

kmp5VT: Oh good point, I forgot `Transpose` doesn't have `position(::, typeof(ndims))` defined. Thanks!

mtfishman: It's a good question to think about though, my guess is that `set_ndims` just won't really make sense for most wrapper array types, so we might consider constraining the generic definition to `!IsWrappedArray`.

NDTensors/ext/NDTensorsMetalExt/linearalgebra.jl
```diff
@@ -19,17 +19,24 @@ function NDTensors.Expose.ql_positive(A::Exposed{<:MtlMatrix})
   return adapt(unwrap_array_type(A), Matrix(Q)), adapt(unwrap_array_type(A), L)
 end
 
+using NDTensors.TypeParameterAccessors: set_type_parameters
 function LinearAlgebra.eigen(A::Exposed{<:MtlMatrix})
   Dcpu, Ucpu = eigen(expose(NDTensors.cpu(A)))
-  D = adapt(set_ndims(set_eltype(unwrap_array_type(A), eltype(Dcpu)), ndims(Dcpu)), Dcpu)
+  D = adapt(
+    set_type_parameters(unwrap_array_type(A), (eltype, ndims), (eltype(Dcpu), ndims(Dcpu))),
```
kmp5VT, line 26: This is implicitly `Base.ndims` because its in an extension. Should I be more explicit?

NDTensors/src/imports.jl
```diff
@@ -63,6 +62,8 @@ using Base.Threads: @spawn
 using .CUDAExtensions: cu
 using .MetalExtensions: mtl
 using .GPUArraysCoreExtensions: cpu
+## Adding this here for now so that ndims in `NDTensors`` uses `TypeParameterAccessors`
+using .TypeParameterAccessors: ndims
```
kmp5VT, line 66: Potentially I shouldn't have this here and I should be more explicit? This is to allow `NDTensors.ndims` to still be functional.

NDTensors/src/lib/TypeParameterAccessors/src/base/ndims.jl
```diff
@@ -0,0 +1,17 @@
+## NDTensors.ndims (not imported from Base)
+
+## TODO So here I try to use the new type_parameters system for `NDTensors.ndims`
+## But if `ndims` is not defined for a type, I revert to using Base.ndims
+function TypeParameterAccessors.ndims(array)
```
kmp5VT, line 5: I am not sure this is the best way to implement this function. I haven't checked lower julia version so this might break on julia 1.6

NDTensors/src/tensor/tensor.jl
```diff
@@ -130,7 +130,8 @@ function randomTensor(StoreT::Type{<:TensorStorage}, inds::Tuple)
 end
 ## End Random Tensor
 
-ndims(::Type{<:Tensor{<:Any,N}}) where {N} = N
+## Potentially it would be better to define `TypeParameterAccessors.positoin(::Type{<:Tensor}, ::typeof(ndims)) = Position(2)` ?
```
kmp5VT, line 133: reminder to address tis comment

NDTensors/ext/NDTensorsMetalExt/linearalgebra.jl
mtfishman: I think it is fine as it is. Though we should keep that question in mind as we continue developing this code and be more explicit if it starts to get confusing and we can't keep track of which `ndims` we are using.

NDTensors/src/imports.jl
mtfishman: I would say let's remove this, if users want to use `TypeParameterAccessors.ndims` they should be explicit about that. `NDTensors.cu` is a different situation because we have recommended that to users in discourse posts.

NDTensors/src/lib/TypeParameterAccessors/src/base/ndims.jl
mtfishman: No matter how we implement it, let's definitely not use `try ... catch ...` for this (I very rarely come across a good use for that, that can't be implemented in a better way). What about just `type_parameter(array, ndims)`? Also, it seems better to me to base the definition of `TypeParameterAccessors.ndims` off of accessing the type parameter position name `TypeParameterAccessors.ndims`.

Additionally, we should make sure that either `TypeParameterAccessors.ndims` or `Base.ndims` can be used interchangeably as type parameter position names in any appropriate `TypeParameterAccessors` function.

NDTensors/src/imports.jl
mtfishman: Perhaps we should establish some rules about using `ndims`, since it seems like it could get quite complicated otherwise. One proposal would be the following:
1. Bias towards using `Base.ndims` throughout the code.
2. Only use `TypeParameterAccessors.ndims` if needed to circumvent an issue with `Base.ndims`.
3. When using `Base.ndims`, write it in the code as `ndims`.
4. When using `TypeParameterAccessors.ndims`, type it explicitly as `TypeParameterAccessors.ndims`, not `ndims`.

So we should think about `TypeParameterAccessors.ndims` as a workaround for issues with `Base.ndims` that we use sparingly. Otherwise, I think it will get chaotic trying to decide when to use `TypeParameterAccessors.ndims` or `Base.ndims`, what `ndims` is supposed to mean, etc.

mtfishman: An exception to the rules above may be that within the `TypeParameterAccessors` module itself, we may need to use `Base.ndims` explicitly to avoid having Julia interpret `ndims` as `TypeParameterAccessors.ndims`.

NDTensors/src/lib/TypeParameterAccessors/src/base/ndims.jl
mtfishman: See a related comment https://github.com/ITensor/ITensors.jl/pull/1353/files#r1527215937.

NDTensors/src/imports.jl
mtfishman: Also note that we don't have to update the entire codebase with these new rules/conventions in this PR, maybe just make sure that the code that you are changing otherwise in this PR follows those rules. It's not really worth updating all of `NDTensors`, and instead focus on using those rules consistently in the new library submodules.

kmp5VT: It looks like we actually don't even need our own version of `ndims`. I removed it from the codebase and all tests pass in ITensors and NDTensors. I guess the question is, do we still want a version in `TypeParameterAccessors` even if its not used?

mtfishman: If I recall correctly it was defined to circumvent this issue in older versions of Julia:
```julia
julia> ndims(Array{<:Any,3})
ERROR: MethodError: no method matching ndims(::Type{Array{var"#s63", 3} where var"#s63"})
Closest candidates are:
  ndims(::CartesianIndices) at multidimensional.jl:362
  ndims(::Base.Iterators.ProductIterator) at iterators.jl:967
  ndims(::AbstractArray{T, N}) where {T, N} at abstractarray.jl:212
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1

julia> versioninfo()
Julia Version 1.6.7
Commit 3b76b25b64 (2022-07-19 15:11 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.4.0)
  CPU: Apple M1 Max
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, westmere)
```
while in more recent versions:
```julia
julia> ndims(Array{<:Any,3})
3

julia> versioninfo()
Julia Version 1.10.2
Commit bd47eca2c8a (2024-03-01 10:14 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 10 × Apple M1 Max
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, apple-m1)
Threads: 1 default, 0 interactive, 1 GC (on 8 virtual cores)
```

kmp5VT: I completely agree that is why we had implemented it in the past, but I think the new design we circumvent times when we would have previously called ndims using the new `set_ndims` function and using the function like `type_parameters(type, (eltype, ndims))`. So now all unittests pass using the current Julia version and V1.6. I double checked on my machine and found all 1.6 test pass

mtfishman: I see, thanks for the explanation. So if I understand correctly, effectively instead of having `TypeParameterAccessors.ndims`, if `Base.ndims(type)` has an issue in an older version of Julia, we can use `type_parameter(type, ndims)`. That sounds reasonable, let's get rid of `NDTensors.ndims`/`TypeParameterAccessors.ndims` and go with that design.

kmp5VT: Yes that is correct and seems safer/more explicit.  The `NDTensors` and `TypeParameterAccessors` `ndims` function is already removed and this PR should be ready for review.

NDTensors/src/lib/TypeParameterAccessors/src/base/ndims.jl
kmp5VT: As discussed above, we removed the custom `ndims` functions and rely on `TypeParameterAccessors.type_paramter(type, ndims)`

kmp5VT:
@mtfishman This branch should be ready now to review. Thanks!

mtfishman:
Looks good, thanks!

emstoudenmire:
Nice PR!

PR #1356, opened 2024-03-18
[ITensors] `ITensorMPS` submodule

emstoudenmire:
# Description

Creates a new `ITensorMPS` module inside of the `ITensors` module. The user experience should be unchanged, since all exports are kept the same in `ITensors` i.e. re-exporting the `ITensorMPS` exports. Most of the changes in this PR besides moving files and renaming the `mps/` folder to the `ITensorMPS/` folder are:
* introduce `ITensorMPS` module as discussed above
* add `imports.jl` and `exports.jl` for this module (most of the work was getting the correct imports figured out)
* moved all of `physics/autompo/` into the `ITensorMPS` folder and module, since it is really a system for creating `MPO`s from `OpSum` data
* minor modifications to some tests which used unexported types and methods to get them from the `ITensors.ITensorMPS` module now

# How Has This Been Tested?

Checked that `ITensorLegacyMPS/base` tests are passing and `ITensors` base tests are passing on my machine.

# Checklist:


emstoudenmire:
One remaining thing to be discussed is whether the precompile related code in `ITensors` that involves types like `MPS` and `MPO` should be moved into the `ITensorMPS` module or not.

src/ITensorMPS/autompo/opsum_to_mpo_generic.jl
```diff
@@ -299,8 +299,8 @@ function MPO(os::OpSum, sites::Vector{<:Index}; splitblocks=true, kwargs...)::MP
   return M
 end
 
-function MPO(eltype::Type{<:Number}, os::OpSum, sites::Vector{<:Index}; kwargs...)
-  return NDTensors.convert_scalartype(eltype, MPO(os, sites; kwargs...))
+function MPO(elt::Type{<:Number}, os::OpSum, sites::Vector{<:Index}; kwargs...)
```
emstoudenmire, line 302: This was a leftover change since I wantd to use the actual `eltype` function here as a test at one point. I can change it back but it's kind of a trivial change.

test/ITensorLegacyMPS/base/runtests.jl
```diff
@@ -14,3 +14,5 @@ ITensors.disable_threaded_blocksparse()
     @time include(filename)
   end
 end
+
+nothing
```
emstoudenmire, line 18: I like to put `nothing` at the end of testing files so that they don't print a very large output when I run them locally.

src/ITensorMPS/autompo/opsum_to_mpo_generic.jl
mtfishman: Fine with me.

test/ITensorLegacyMPS/base/runtests.jl
mtfishman: Can't you just run it as `include("runtests.jl");`?

mtfishman:
Thanks @emstoudenmire. Perhaps we should change `test/ITensorLegacyMPS` to `test/ITensorMPS`.

src/ITensorMPS/ITensorMPS.jl
```diff
@@ -0,0 +1,37 @@
+
```
mtfishman, line 1:
```suggestion
```

src/ITensorMPS/ITensorMPS.jl
```diff
@@ -0,0 +1,37 @@
...
+using Adapt
+using HDF5
+using ..ITensors
+using IsApprox
+using KrylovKit
+using Printf
+using Random
+using TupleTools
+
+import ..ITensors.Ops: params
```
mtfishman, line 13: Seems to make sense to move this to `imports.jl`.

mtfishman:
Thanks @emstoudenmire, this is a nice change to make sure we keep the MPS code self-contained, and also more clearly separated from things like operator definitions.

Something else we might want to do is move `ITensorMPS` and submodules like `src/ContractionSequenceOptimization`, `src/ITensorChainRules`, `src/Ops`, etc. into a file `src/lib`. That is the organization we have in NDTensors (https://github.com/ITensor/ITensors.jl/tree/v0.3.57/NDTensors/src/lib) which I think has helped to keep things organized.

I even went with organizing those files as projects themselves, i.e. in this case we could have a file organization:
```
src/lib/ITensorMPS/src/ITensorMPS.jl
src/lib/ITensorMPS/src/# Files making up the `ITensorMPS` module
src/lib/ITensorMPS/test/runtest.jl
src/lib/ITensorMPS/test/# Test files, the ones currently in `test/ITensorLegacyMPS`
```
Then, the directory `src/lib/ITensorMPS` will already be organized in a way that it can be moved to a separate repository, once we are ready to do that.

Maybe we don't need to do that all in this PR, let's discuss the logistics.

mtfishman:
> One remaining thing to be discussed is whether the precompile related code in `ITensors` that involves types like `MPS` and `MPO` should be moved into the `ITensorMPS` module or not.

Do you mean code in [src/packagecompile/](https://github.com/ITensor/ITensors.jl/blob/v0.3.57/src/packagecompile/precompile_itensors.jl) or in [precompile/](https://github.com/ITensor/ITensors.jl/tree/v0.3.57/precompile)?

src/ITensorMPS/ITensorMPS.jl
emstoudenmire: Yes

test/ITensorLegacyMPS/base/runtests.jl
emstoudenmire: I did run it that way, but on my machine it prints a huge output at the end. Most of the output seems related to broken tests but I'm not sure if that's the main reason. I think it's because the Test module returns some kind of data structure that gets displayed in the REPL afterward.

emstoudenmire:
I made the above changes such as renaming ITensorLegacyMPS/ to ITensorMPS/ and the review comments on the code.
I'd be happy to move the `src/packagecompile/` code and remove the `precompile/` too but maybe better in a different PR? What do you think?

Also I like the idea of going ahead with the lib folder and making ITensorMPS/ have a package structure (or even be an actual package already). Should we go for it in this PR? Maybe in another to make each PR more "atomic"?

src/ITensorMPS/exports.jl
```diff
@@ -0,0 +1,134 @@
...
+
+  # abstractmps.jl
+  # Macros
+  @preserve_ortho,
+  # Methods
+  AbstractMPS,
+  add,
+  common_siteind,
+  common_siteinds,
+  contract,
```
mtfishman, line 13: Isn't this defined and exported by `ITensors.jl`?

src/ITensorMPS/exports.jl
```diff
@@ -0,0 +1,134 @@
...
+  convert_eltype,
+  convert_leaf_eltype,
+  dag!,
+  eltype,
```
mtfishman, line 17: Seems like a strange export since it is defined and exported by `Base`.

src/ITensorMPS/exports.jl
```diff
@@ -0,0 +1,134 @@
...
+  # abstractmps.jl
+  # Macros
+  @preserve_ortho,
+  # Methods
+  AbstractMPS,
+  add,
+  common_siteind,
+  common_siteinds,
+  contract,
+  convert_eltype,
```
mtfishman, line 14: This is defined in `ITensors.jl` so I'm not sure why it is being exported from here.

src/ITensorMPS/exports.jl
```diff
@@ -0,0 +1,134 @@
...
+  convert_leaf_eltype,
```
mtfishman, line 15: Same as my comment for `convert_eltype`.

src/ITensorMPS/exports.jl
```diff
@@ -0,0 +1,134 @@
...
+  dag!,
```
mtfishman, line 16: I can't even find this defined anywhere.

src/ITensorMPS/exports.jl
```diff
@@ -0,0 +1,134 @@
...
+  findsite,
+  findsites,
+  firstsiteind,
+  firstsiteinds,
+  logdot,
+  loginner,
+  lognorm,
+  movesite,
+  movesites,
+  ndims,
```
mtfishman, line 29: This is a `Base` function, seems strange to export it.

src/exports.jl
```diff
@@ -247,6 +248,7 @@ export
   lognorm,
   movesite,
   movesites,
+  ndims,
```
mtfishman, line 251: Seems strange to export this since it is a `Base` function.

src/imports.jl
```diff
@@ -88,6 +88,8 @@ import Adapt: adapt_structure, adapt_storage
 
 import HDF5: read, write
 
+import KrylovKit: orthogonalize, orthogonalize!
```
mtfishman, line 91: I'm not sure we want to be overloading these functions.

src/exports.jl
```diff
@@ -358,6 +360,7 @@ export
   # physics/fermions.jl
   fparity,
   isfermionic,
+  parity_sign,
```
mtfishman, line 363: What's the reason behind adding this as a new export?

src/iterativesolvers.jl
```diff
@@ -18,7 +18,7 @@ function get_vecs!((phi, q), M, V, AV, ni)
   return lambda
 end
 
-function orthogonalize!(q::ITensor, V, ni)
+function orthogonalize!(q::ITensor, V, ni::Integer)
```
mtfishman, line 21: Is the reasoning behind this change because we are now overloading the `KrylovKit` version?

src/exports.jl
mtfishman, line 1: I wonder if we can simplify this file using [Reexport.jl](https://github.com/simonster/Reexport.jl), i.e. automatically reexport the functions from `ITensorMPS` instead of re-listing all of them here.

mtfishman:
> I'd be happy to move the `src/packagecompile/` code and remove the `precompile/` too but maybe better in a different PR? What do you think?

Yeah, let's do that in a separate PR to keep this simpler to review.

> Also I like the idea of going ahead with the lib folder and making ITensorMPS/ have a package structure (or even be an actual package already). Should we go for it in this PR? Maybe in another to make each PR more "atomic"?

Let's do that in a separate PR.

src/exports.jl
```diff
@@ -230,6 +230,7 @@ export
   # Macros
   @preserve_ortho,
   # Methods
+  AbstractMPS,
```
mtfishman, line 233: Was this meant to get added to the exports of `ITensors`?

test/ITensorLegacyMPS/base/runtests.jl
mtfishman: The `;` at the end should suppress output to the REPL.

emstoudenmire: Oh I see. Then yes no problem to run it that way.

src/exports.jl
emstoudenmire: Agreed, don't need this one.

src/iterativesolvers.jl
emstoudenmire: Yes. We could discuss other ways to go but without this change, in the context of overloading `KrylovKit.orthogonalize!`, the existing code was leading to an ambiguity error.

src/imports.jl
emstoudenmire: Good catch. Even if we need the one inside `ITensorMPS` (which I believe we do?) then we don't need to import it here, probably.

src/iterativesolvers.jl
emstoudenmire: I'll change it back if I can successfully get rid of the ITensors-level KrylovKit imports. Better to keep these changes minimal.

src/exports.jl
emstoudenmire: I'll try removing it.

emstoudenmire:
Just for context, these imports, especially, and to some extent exports were very tricky to sort out because Julia does not give helpful error messages for issues with imports (instead it would just call some wrong function and cause a seemingly unrelated error). You've probably experienced it. So I probably ended up with some extra imports and exports that were unneeded in the end while trying to find the right ones. I'm glad to remove any extraneous ones you catch.

Agreed about trying `Reexport.jl` instead of having `ITensorMPS` things in the `ITensors` exports (which is a not-nice kind of dependency between the modules). I'll give that a go.

src/exports.jl
emstoudenmire: Yes that worked. Much nicer.

src/exports.jl
emstoudenmire: Agreed – I'll just use `ITensors.parity_sign` for the one place where it gets called from `ITensorMPS`.

src/ITensorMPS/exports.jl
emstoudenmire: Removed

test/ITensorLegacyMPS/base/runtests.jl
emstoudenmire: I removed those `nothing` statements from all the tests and `runtests.jl` in `ITensorMPS`.

src/ITensorMPS/exports.jl
emstoudenmire: When I remove it, I get an error `UndefVarError: `contract` not defined` from line 25 of `test_abstractprojmpo.jl`. I thought maybe just importing it might be enough since ITensor exports `contract` but I guess the export is needed to also export the `contract` overloads defined by ITensorMPS?

src/imports.jl
emstoudenmire: I tried removing it and I get a warning `WARNING: using ITensorMPS.orthogonalize! in module ITensors conflicts with an existing identifier.`. Keeping this import is the only way I've managed to successfully get rid of this warning. But if you know a better way to fix it I am definitely open to it!

src/ITensorMPS/exports.jl
mtfishman: Maybe in that case `contract` isn't being brought into the namespace from `ITensors`. You could try adding `using ITensors: contract` at the top of that test file.

src/ITensorMPS/ITensorMPS.jl
```diff
@@ -0,0 +1,34 @@
+module ITensorMPS
+
+using Adapt
+using HDF5
+using ..ITensors
+using IsApprox
+using KrylovKit
+using Printf
+using Random
+using TupleTools
```
mtfishman, line 10: We could move these to a `usings.jl` file, to match the style of how we handle imports and exports.

test/ITensorMPS/base/test_dmrg.jl
```diff
@@ -1,6 +1,5 @@
 using ITensors, Test, Random
-
-using ITensors: nsite, set_nsite!
+import ITensors.ITensorMPS: nsite, set_nsite!, site_range
```
mtfishman, line 2:
```suggestion
using ITensors.ITensorMPS: nsite, set_nsite!, site_range
```
Generally I use `using` if I am just using a function, and not overloading it.

emstoudenmire:
Good feedback. I’ll try all those.

   -=Miles Stoudenmire=-
   ***@***.***
   ***@***.***
   http://itensor.org/miles/

mtfishman:
@emstoudenmire as a general comment on code style, some rules I have started to follow around `using`, `import`, and `export` are:
1. Be very sparing about what we decide to `export`. With new libraries, I'm starting with no exports, and will add them slowly as we determine what is most used in practice.
2. I have gone with the style of _never_ using `import`. If I want to bring a function or type into a namespace, I use `using M: f`, and if I want to overload a function, I use the more explicit `function M.f(...) ... end` syntax (to overload function `f` from module `M`).
3. In library code, including in source code, examples, and tests, I have started to never use a bare call of `using`, such as `using M`, which brings every exported function or type into the current namespace. Instead, I list the functions and types I am using explicitly, with `using M: f`. I've also started to put those calls at the top of files where those functions are types are actually being used, rather than in one long global list for the entire module.
4. For test files, along with that change in style to `using`, I've also started wrapping test files in modules with unique names, see for example: https://github.com/ITensor/ITensors.jl/blob/v0.3.57/NDTensors/src/lib/BlockSparseArrays/test/basics.jl. The line `@eval module $(gensym())` generates a unique module name. This helps to make sure test files are fully self contained, so we can carefully control what gets brought into the namespace and make sure temporary variables from tests don't leak out into other test files.

We don't have to make all of these changes in this PR, and in terms of `export`s we are constrained by needing to keep them consistent for backwards compatibility. Mostly I'm pointing out this newer style so you have a "goal post" to aim for so we can start moving older code in that newer direction, and write new code with that newer style.

test/ITensorMPS/base/test_dmrg.jl
emstoudenmire: Great

src/ITensorMPS/ITensorMPS.jl
emstoudenmire: That's a good idea... let's do it

emstoudenmire:
I'm all for these changes to best practices around imports and exports. I remember that I think even a few years ago we had actually discussed the possibility of not using imports and just qualifying things i.e. `ModuleName.function` instead, which I like since it's very "self documenting".  So I'm supportive of that.

The other ideas about how to apply `using` and about modules for tests are new ones for me, and sound like really good ideas. I agree we should be a lot more sparing in the future both about what we import and export at all levels and in all packages. But makes sense that for ITensorMPS we are a bit constrained by backwards compatibility.

test/base/runtests.jl
```diff
@@ -14,3 +14,5 @@ ITensors.disable_threaded_blocksparse()
     @time include(filename)
   end
 end
+
+nothing
```
mtfishman, line 18: Let's remove this.

src/ITensors.jl
```diff
@@ -202,11 +177,24 @@ include("physics/site_types/tj.jl")
...
+#####################################
+# MPS/MPO
+#
+include("ITensorMPS/ITensorMPS.jl")
+@reexport using .ITensorMPS
+
+# Only include iterativesolvers.jl after ITensorMPS
+# since it defines orthogonalize! which could
+# conflict with definition in ITensorMPS
+import .ITensorMPS: orthogonalize!
```
mtfishman, line 190: This doesn't seem right to me. I have a few comments about this:
1. Can we move `iterativesolvers.jl` to `ITensorMPS`? It is only used by the MPS code anyway, and ultimately we should just deprecate it in favor of KrylovKit.jl so we may as well start that process now by moving it out of the base ITensors.jl codebase.
2. I think we should rename the `orthogonalize!` function use inside the Krylov solver to something else, like `krlyov_orthogonalize!`. It is conceptually unrelated from the MPS `orthogonalize!` function so we may as well just give them different names and avoid future namespace issues.

test/base/runtests.jl
emstoudenmire: Removed, and in other files too

src/iterativesolvers.jl
emstoudenmire: This should be very cleaned up now by the latest set of changes. See comment below.

src/ITensorMPS/exports.jl
emstoudenmire: I'll work on this and push when I've got a good solution

test/ITensorMPS/base/test_abstractprojmpo.jl
```diff
@@ -1,6 +1,7 @@
 using ITensors
 using Random
 using Test
+import ITensors.ITensorMPS
```
mtfishman, line 4:
```suggestion
using ITensors.ITensorMPS: ITensorMPS
```

test/ITensorMPS/base/test_mps.jl
```diff
@@ -1,5 +1,6 @@
 using Combinatorics
 using ITensors
+import ITensors.ITensorMPS
```
mtfishman, line 3:
```suggestion
using ITensors.ITensorMPS: ITensorMPS
```

test/ITensorMPS/base/test_abstractprojmpo.jl
emstoudenmire: Changed

test/ITensorMPS/base/test_mps.jl
emstoudenmire: Changed

mtfishman:
> The only outstanding issue is about import / export of the `contract` method in the `ITensorMPS` module. Here's my current chain of reasoning about it:
> 
>     1. Both `ITensorMPS` and `ITensors` export `contract`. This is because e.g. `ITensors` defines `contract` of `ITensor` objects and `ITensorMPS` defines `contract` of `MPO` objects among other overloads.
> 
>     2. `ITensorMPS` depends on `ITensors`
> 
>     3. So `ITensorMPS` should either: 3a) do `import ..ITensors: contract` or 3b) qualify implementations of contract as `ITensors.contract`.
> 
> 
> Currently the approach is to use `import ..ITensors: contract` in `ITensorMPS`. I'm fine with changing to the other approach. But I currently don't see an alternative to doing one of those.
> 
> Or let me know if your comment above was about something else (e.g. there's a reason that `ITensorMPS` should not be exporting `contract`?).

I think I just didn't understand the issue you were describing around the test failing.

I would vote for:

1. `ITensorMPS` overloading `ITensors.contract`, either with explicit `ITensors.contract` in the function definition (which is the preferred syntax now) or with `import ..ITensors: contract`. I'll leave that up to you to decide between the two, probably we'll want to update that style everywhere so it may not be worth doing here since that is a bigger undertaking, but chipping away little by little in PRs like this could be nice to do.
2. I would vote for `ITensorMPS` _not_ exporting `contract`, or any other function that is originally defined in `ITensors` and already exported by `ITensors`. If users want all of the functions exported by both `ITensorMPS` and `ITensors`, they can do `using ITensors; using ITensorMPS` at the top of their file.
3. In terms of the style in the tests for making use of `contract`, as I said, my approach has been to explicitly put `using M: f` to use a function `f` from ` module `M`. If `f` exists in multiple modules (say it is overloaded by a module `N` that depends on `M`), I try my best to include `f` from the module where it was originally defined. That means, in the tests, we should use `using ITensors: contract`, even if we are using the `contract` function on MPS and MPO objects.

src/ITensorMPS/exports.jl
mtfishman: See my comment. I think we should remove any exports from `ITensorMPS/exports.jl` that are already exported by `ITensors`.

mtfishman:
So the issue was that `contract` wasn't being exported from `ITensors`?

mtfishman:
@emstoudenmire it looks like the GPU tests are failing because they are trying to run `ITensorGPU/test/test_cuiterativesolvers.jl`, which has now been removed.

emstoudenmire:
Yeah! Pretty embarassing but that was the issue (i.e. contract not being exported). The `contract` issue turned out to be a combination of a silly mistake on my part plus the confusion I had about `import/export` which I've now cleared up thanks to our discussion. All it was was that I had accidentally deleted the export of `contract` from `ITensors` and then didn't understand what the error message was really saying.

I've now restored the export of `contract` from `ITensors` and removed the export of `contract` from `ITensorMPS`.

One last thing to do is to remove the `import` of `contract` and just write `ITensors.contract` instead. I don't mind going through and doing that because like you said we want to start moving toward this in chunks of work.

emstoudenmire:
I've now also removed the entry for the deleted ITensorGPU test

mtfishman:
> One last thing to do is to remove the `import` of `contract` and just write `ITensors.contract` instead. I don't mind going through and doing that because like you said we want to start moving toward this in chunks of work.

Sounds good. After that it looks good to merge from my end. Unfortunately we are importing a lot of functions right now, so it will be quite an undertaking to change over the style everywhere. This may be a case where it is worth trying to automate that process (I need to change over other repositories as well).

PR #1357, opened 2024-03-20
[GradedAxes] Add `dual` of `GradedUnitRange`

mtfishman:
This is a first attempt at adding back support for taking the `dual` of a `GradedUnitRange`.

Before, it was implemented through a boolean `isdual` stored in the type. In this new design, it is implemented as a wrapper type `UnitRangeDual`, so when you call `dual(r::GradedUnitRange)` it outputs the object `UnitRangeDual(r)` which marks it as a dual space. The goal is that a `UnitRangeDual` wrapping a `GradedUnitRange` for all intents and purposes acts like a `GradedUnitRange` (with the same blocking structure) but when labels are accessed they are the dual of the original labels.

There are some corner cases I came across with the new `GradedUnitRange` type implemented in #1355 that I am still working through, but I think the design around `UnitRangeDual` is working well so far.

Here is a small demonstration:
```julia
using NDTensors.GradedAxes: GradedAxes, dual, gradedrange
struct U1
  n::Int
end
GradedAxes.dual(c::U1) = U1(-c.n)

a = gradedrange([U1(0) => 2, U1(1) => 3])
ad = dual(a)
```
```julia
julia> typeof(a)
BlockArrays.BlockedUnitRange{Vector{LabelledInteger{Int64, U1}}}

julia> typeof(ad)
NDTensors.GradedAxes.UnitRangeDual{LabelledInteger{Int64, U1}, BlockArrays.BlockedUnitRange{Vector{LabelledInteger{Int64, U1}}}}

julia> a[Block(2)]
LabelledInteger{Int64, U1}(3, U1(1)):LabelledInteger{Int64, U1}(1, U1(1)):LabelledInteger{Int64, U1}(5, U1(1))

julia> ad[Block(2)]
LabelledInteger{Int64, U1}(3, U1(-1)):LabelledInteger{Int64, U1}(1, U1(-1)):LabelledInteger{Int64, U1}(5, U1(-1))
```


emstoudenmire:
Interesting – keep us posted about how the design works out

PR #1359, opened 2024-03-22
[NDTensorsMetalExt] Trying to add GPU testing to M1 jenkins

kmp5VT:
# Description

In this PR I am making minimal changes to get Jenkins to run CI testing on the mac M1's with `Metal` backend.  It looks like I just need to add it to Jenkins and Dylan added juliaup for us to use different versions of Julia.  Currently Metal is not supported on Julia 1.6 so I am only testing with release (1.10)

# Checklist:

- [ ] Jenkins runs NDTensors unittest on the M1 mac
- [ ] Metal Unittests pass on Jenkins


src/tensor_operations/matrix_decomposition.jl
```diff
@@ -159,6 +159,7 @@ function svd(
     AC = permute(AC, cL, cR)
   end
```
kmp5VT, line 161: The code in this file was accidentally included. I will remove it and close this comment after I do.

kmp5VT: This has all been removed.

NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -1,11 +1,14 @@
 using NDTensors: NDTensors
+using Pkg
```
mtfishman, line 2:
```suggestion
using Pkg: Pkg
```

mtfishman:
Nice to see this already up and running, I like how the Jenkins tests are now organized, the way the test results are displayed in Jenkins now is nicer.

The `Metal` tests appear to be failing because `resize!(::MtlArray, ::Integer)` can't be found. I added that in https://github.com/JuliaGPU/Metal.jl/pull/279 but it hasn't been included in a registered version of Metal.jl yet. We could using `Pkg.dev("Metal")` as a temporary workaround.

mtfishman:
@maleadt sorry to ping you, but could you register a new version of Metal.jl?

NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -1,11 +1,14 @@
 using NDTensors: NDTensors
+using Pkg: Pkg
 if "cuda" in ARGS || "all" in ARGS
   using CUDA
 end
 if "rocm" in ARGS || "all" in ARGS
+  Pkg.add("AMDGPU")
```
mtfishman, line 7: Let's add a comment here and down below to explain that we do this instead of putting it into `Project.toml` because it can't be installed on Julia 1.6.

NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -1,11 +1,14 @@
 using NDTensors: NDTensors
+using Pkg: Pkg
 if "cuda" in ARGS || "all" in ARGS
   using CUDA
```
mtfishman, line 4: I'm confused by something. I see `CUDA` wasn't in the `[extras]` section of `NDTensors/test/Project.toml`, how as it being installed?

NDTensors/test/NDTensorsTestUtils/device_list.jl
mtfishman: Oh I see, it is in `[deps]`. I guess that's a bit strange but ok.

kmp5VT: Its in [extra] but I think its a good idea to put a comment about how Metal and AMDGPU don't work on versions lower than 1.x (I have to verify which version)

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: So CUDA is in the `[deps]` section but perhaps its a better idea in general to use `Pkg.add("CUDA")` here, that way we don't have to always compile `CUDA.jl` when we run `Pkg.test("NDTensors")`.

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: Perhaps we should add back the `[extras]` section. Its not actually used in `Pkg.test` so it should be no problem

NDTensors/test/NDTensorsTestUtils/device_list.jl
mtfishman: Sure, let's try that.

NDTensors/test/NDTensorsTestUtils/device_list.jl
mtfishman: Oops my second comment was meant for the other comment thread.

I'd be ok adding back `[extras]` and listing these optional test dependencies there. I was never clear on what the `[extras]` section was for, but it could be a good way to keep track of which packages are optional test dependencies in one place.

kmp5VT: We could name the section `[extensions]` if that makes more sense? Regardless of the name its ignored by `Pkg.test` which only looks at `[deps]`

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: Okay I updated the code, thanks!

NDTensors/test/NDTensorsTestUtils/device_list.jl
mtfishman: Sounds good.

mtfishman:
Note that some of the latest test failures were caused by issues introduced in https://github.com/ITensor/ITensors.jl/pull/1356.

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: I updated this to `[weakdeps]`

PR #1365, opened 2024-03-27
Refactoring GPU extensions

kmp5VT:
# Description
In this PR I am working to address some refactoring issues. Such as removing using statements from module files, try to diagnose/fix GPU blocksparse backend failures, update adapt functions to use `storagemode` like AMDGPU, remove gpu examples from testing based on plan listed in google drive

# Checklist:

- [x] Investigate why the AMDGPU and CUDA backends don’t work for BlockSparse storage. Implement resize! for AMDGPUs to get block sparse operations on AMDGPU working (see https://github.com/JuliaGPU/Metal.jl/pull/279 for a reference on how it might be implemented).
- [x] Update the `adapt` functions to have `storagemode` as the input parameter
Update the Extensions libraries to move the bare using statements from the `NDTensorsXExt.jl` file to the files where they are necessary. 
- [x] Delete `NDTensors/ext/examples`.
- [x] Remove examples from NDTensors test
- [x] Define a MtlArrayAdaptor for adapting storage to MtlArray, analagous to CuArrayAdaptor, to be used here: https://github.com/ITensor/ITensors.jl/blob/v0.3.57/NDTensors/ext/NDTensorsMetalExt/adapt.jl
- [x] update the `generic_zeros` and `generic_rand` functions for Dense and AbstractArray
- [x] All unittests pass

### DIFF
--- a/NDTensors/src/abstractarray/generic_array_constructors.jl
+++ b/NDTensors/src/abstractarray/generic_array_constructors.jl
@@ -17,7 +17,7 @@ end
 function generic_randn(
   arraytype::Type{<:AbstractArray}, dims::Tuple; rng=Random.default_rng()
 )
-  return generic_randn(arraytype, dims...; rng=rng)
+  return generic_randn(arraytype, dims...; rng)
 end
 
 # Implementation, catches if `ndims(arraytype) != length(dims)`.


mtfishman:
Note that `resize!` is defined for `AMDGPU.ROCVector`: https://github.com/JuliaGPU/AMDGPU.jl/blob/v0.8.11/src/array.jl#L266.

kmp5VT:
I was able to get the CUDA DMRG working with blocksparse so next I will be working to verify the AMDGPU blocksparse DMRG works

NDTensors/ext/NDTensorsCUDAExt/append.jl
```diff
@@ -0,0 +1,9 @@
+## CUDA.jl has no append! function and 
+## calling base results in a scalar issue
+## This is probably super slow and I am seeing 
+## very slow blocksparse DMRG but it does work.
+function Base.append!(a::CuArray, b::CuArray)
```
kmp5VT, line 5: I made this append function but most likely this is type piracy and we need to talk to CUDA.jl about creating this function. For now this is a quick hack to get DMRG working with no scalar indexing error. Additionally we could just use `@allowscalar` when we have append. Or convert to CPU.

NDTensors/src/dense/fill.jl
```diff
@@ -22,18 +22,27 @@ function generic_randn(StoreT::Type{<:Dense}, dim::Integer=0)
...
+) where {DataT<:AbstractArray,ElT}
+  @assert ElT == eltype(DataT)
+  data = generic_zeros(DataT, dim(dims))
+  StoreT = set_datatype(StoreT, typeof(data))
+  return StoreT(data)
+end
+
+function generic_zeros(StoreT::Type{<:Dense{ElT}}, dim) where {ElT}
   return generic_zeros(default_storagetype(ElT), dim)
 end
```
mtfishman, line 44: This definition doesn't make sense to me. Should it be specifying the `datatype` of `StoreT`?

NDTensors/src/dense/fill.jl
```diff
@@ -22,18 +22,27 @@ function generic_randn(StoreT::Type{<:Dense}, dim::Integer=0)
...
 
-function generic_zeros(StoreT::Type{<:Dense}, dim::Integer=0)
+function generic_zeros(StoreT::Type{<:Dense}, dim)
   return generic_zeros(default_storagetype(), dim)
 end
```
mtfishman, line 48: Same with this one, seems like this should specify default values of `eltype` and `datatype` of `StoreT`. This seems to implicitly rely on the fact that `default_storagetype()` makes a `Dense`, but I don't think we should rely on that behavior in code like this.

NDTensors/src/dense/fill.jl
```diff
@@ -22,18 +22,27 @@ function generic_randn(StoreT::Type{<:Dense}, dim::Integer=0)
...
 
-function generic_zeros(StoreT::Type{<:Dense{ElT}}, dim::Integer=0) where {ElT}
+function generic_zeros(
+  StoreT::Type{<:Dense{ElT,DataT}}, dims
+) where {DataT<:AbstractArray,ElT}
+  @assert ElT == eltype(DataT)
+  data = generic_zeros(DataT, dim(dims))
+  StoreT = set_datatype(StoreT, typeof(data))
+  return StoreT(data)
+end
```
mtfishman, line 40: These two definitions should be combined.

I also think it probably doesn't make sense to allow `dims` to be anything besides either `Int` or `Tuple{Int}`, since `Dense` is an `AbstractVector`.

So this could be something like:
```julia
function generic_zeros(
  StoreT::Type{<:Dense{ElT,DataT}}, dim::Integer
) where {DataT<:AbstractArray,ElT}
  return generic_zeros(StoreT, (dim,))
end

function generic_zeros(
  StoreT::Type{<:Dense{ElT,DataT}}, dims::Tuple{Integer}
) where {DataT<:AbstractArray,ElT}
  @assert ElT == eltype(DataT)
  data = generic_zeros(DataT, dim(dims))
  StoreT = set_datatype(StoreT, typeof(data))
  return StoreT(data)
end

function generic_zeros(
  StoreT::Type{<:Dense{ElT,DataT}}, dim::Tuple{Integer,Integer,Vararg{Integer}}
) where {DataT<:AbstractArray,ElT}
  return error("Can't make a multidimensional `Dense` object.")
end
```
though I'm not sure if we are expecting `dim`/`dims` to be something besides integers.

Is there code where `generic_zeros` is being called to construct a `Dense` where `dims` has length greater than 1? If so I would be curious where that is happening and why (maybe it is needed for generic code where that makes sense for another storage type?).

NDTensors/ext/NDTensorsCUDAExt/append.jl
mtfishman: I see, I guess `append!` also has a scalar indexing issue for `MtlArray`:
```julia
julia> using Metal

julia> v = mtl(randn(4))
4-element MtlVector{Float32, Private}:
  0.9403535
  0.89847064
  1.4950696
 -0.33930042

julia> Metal.@allowscalar append!(v, [2, 3])
6-element MtlVector{Float32, Private}:
  0.9403535
  0.89847064
  1.4950696
 -0.33930042
  2.0
  3.0

julia> v
6-element MtlVector{Float32, Private}:
  0.9403535
  0.89847064
  1.4950696
 -0.33930042
  2.0
  3.0

julia> append!(v, [2, 3])
ERROR: Scalar indexing is disallowed.
Invocation of setindex! resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.

If you want to allow scalar iteration, use `allowscalar` or `@allowscalar`
to enable scalar iteration globally or for the operations in question.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] errorscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:155
 [3] _assertscalar(op::String, behavior::GPUArraysCore.ScalarIndexing)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:128
 [4] assertscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:116
 [5] setindex!
   @ ~/.julia/packages/GPUArrays/Hd5Sk/src/host/indexing.jl:56 [inlined]
 [6] _append!(a::MtlVector{Float32, Private}, ::Base.HasShape{1}, iter::Vector{Int64})
   @ Base ./array.jl:1201
 [7] append!(a::MtlVector{Float32, Private}, iter::Vector{Int64})
   @ Base ./array.jl:1187
 [8] top-level scope
   @ REPL[5]:1
 [9] top-level scope
   @ ~/.julia/dev/Metal/src/initialization.jl:57
```

mtfishman: A slightly better implementation could be:
```julia
function Base.append!(a::CuVector, items::AbstractVector)
  n = length(items)
  resize!(a, length(a) + n)
  copyto!(a, length(a) - n + 1, items, firstindex(items), n)
  return a
end
```
which is closer to the definition of `copyto!` for `Vector` in `Base`: https://github.com/JuliaLang/julia/blob/v1.10.2/base/array.jl#L1179-L1185. Note that version can also handle appending a CPU vector onto a GPU vector, as long as `copyto!` can handle that, which it looks like it does for CUDA.jl (https://github.com/JuliaGPU/CUDA.jl/blob/v5.2.0/src/array.jl#L534-L574) and Metal.jl (https://github.com/JuliaGPU/Metal.jl/blob/main/src/array.jl#L267-L312).

I suppose as an optimization that `Base.unsage_copyto!` could be used in exchange for `copyto!`, as you did, since we are sure that the inputs of `copyto!` are not out of bounds.

mtfishman: I would say for the sake of this PR, let's not do this kind of type piracy with `append!`, and instead call `append!` with `@allowscalar`. Then we can make PRs to the various GPU backends implementing `append!` as a followup.

kmp5VT: For this issue, I made an expose version of `Base.append!` and for the generic version I call `append!` but for `Type{<:AbstractGPUArray}` I use `@allowscalar` to work generically across GPU backends. This might be too strongly opinionated so I can push the function to each specific backend if that is a better idea. Thanks!

src/imports.jl
```diff
@@ -110,7 +110,7 @@ import LinearAlgebra:
   tr,
   transpose
 
-using ITensors.NDTensors.Expose: cpu
+using ITensors.NDTensors.GPUArraysCoreExtensions: cpu
```
mtfishman, line 113: Because this is needed, this is technically a breaking change so we should bump the minor version of `NDTensors`.

NDTensors/src/dense/fill.jl
kmp5VT: Yes in `blocksparsetensor.jl` that calls generic_zeros with a tuple of length greater than one
```
if !Rblock_exists
      # Rblock doesn't exist
      block_size = permute(size(Tblock), perm)
      # TODO: Make GPU friendly.
      DenseT = set_type_parameters(Dense, (eltype, parenttype), (eltype(R), datatype(R)))
      Rblock = tensor(generic_zeros(DenseT, block_size), block_size)
elseif !Tblock_exists
      # Tblock doesn't exist
      block_size = permute(size(Rblock), invperm(perm))
      # TODO: Make GPU friendly.
      DenseT = set_type_parameters(Dense, (eltype, parenttype), (eltype(T), datatype(T)))
      Tblock = tensor(generic_zeros(DenseT, block_size), block_size)
 end
 ```

Comment (mtfishman) [reply NDTensors/src/dense/fill.jl:40]: I see, probably it is better to change that code rather than make `generic_zeros` more general, for example:
```julia
      Rblock = tensor(generic_zeros(DenseT, prod(block_size)), block_size)
```
etc.

kmp5VT: Okay I have flattened these and it looks like it might have been the only case. I also updated your suggestion to remove the `where{}` portion of the code. Let me know if it looks reasonable, thanks!

NDTensors/src/dense/fill.jl
kmp5VT: Okay I removed this code. thanks!

NDTensors/src/dense/fill.jl
kmp5VT: Also if the code looks good I can update the `generic_randn` code too from this file

NDTensors/src/dense/fill.jl
kmp5VT: This code is also removed

NDTensors/ext/NDTensorsCUDAExt/indexing.jl
```diff
@@ -1,9 +1,15 @@
+using CUDA: CuArray
+using GPUArraysCore: @allowscalar
+using NDTensors: NDTensors
+using NDTensors.Expose: Exposed, expose, parent, unexpose
+
+using GPUArraysCore: @allowscalar
```
mtfishman, line 6: This was already written above.

NDTensors/ext/NDTensorsCUDAExt/indexing.jl
```diff
@@ -1,9 +1,15 @@
+using CUDA: CuArray
+using GPUArraysCore: @allowscalar
+using NDTensors: NDTensors
+using NDTensors.Expose: Exposed, expose, parent, unexpose
```
mtfishman, line 4: Does `NDTensors.Expose` have it's own version of `parent`?

mtfishman: Ideally we would use `Base.parent`, or if we needed a special version of `parent` I would think it would be defined in `NDTensors.TypeParameterAccessors`.

NDTensors/ext/NDTensorsCUDAExt/indexing.jl
kmp5VT: Thanks this was removed

NDTensors/ext/NDTensorsCUDAExt/indexing.jl
kmp5VT: No this is a mistake, there is just a version of `Base.parent(::Exposed)` in `Expose` it does not have its own version of the function. I removed it from the using list, thanks!

NDTensors/ext/NDTensorsCUDAExt/copyto.jl
```diff
@@ -1,3 +1,7 @@
+using CUDA: CuArray
+using NDTensors.Expose: Exposed, expose, parent, unexpose
```
mtfishman, line 2:
```suggestion
using NDTensors.Expose: Exposed, expose, unexpose
```

NDTensors/ext/NDTensorsCUDAExt/mul.jl
```diff
@@ -1,3 +1,7 @@
+using CUDA: CuArray
+using LinearAlgebra: LinearAlgebra, mul!, transpose
+using NDTensors.Expose: Exposed, expose, parent, unexpose
```
mtfishman, line 3:
```suggestion
using NDTensors.Expose: Exposed, expose, unexpose
```

NDTensors/test/ITensors/TestITensorDMRG/TestITensorDMRG.jl
```diff
@@ -17,9 +17,9 @@ is_broken(dev, elt::Type, conserve_qns::Val) = false
 
 ## Disable blocksparse GPU testing on CUDA, Metal and ROC backends.
 ## Currently Metal fails because we are waiting for `resize!` to be added. Should be in the next metal release
-## CUDA fails because there is no defined `append!`.
 ## ROC fails because TODO determine again why roc fails.
-is_broken(dev::typeof(cu), elt::Type, conserve_qns::Val{true}) = true
+
+#is_broken(dev::typeof(cu), elt::Type, conserve_qns::Val{true}) = true
```
mtfishman, line 22: Reminder to delete this before merging.

NDTensors/src/lib/GPUArraysCoreExtensions/src/gpuarrayscore.jl
```diff
@@ -13,3 +15,9 @@ function set_storagemode(type::Type, param)
 end
 
 function cpu end
+
+cpu(E::Exposed) = cpu(unexpose(E))
+
+function Base.append!(Ecollection::Exposed{<:AbstractGPUArray}, collections...)
+  return @allowscalar append!(unexpose(Ecollection), collections...)
+end
```
mtfishman, line 23: Let's define separate versions for each GPU backend so we are prepared once they start getting `append!` implementations.

NDTensors/src/dense/fill.jl
```diff
@@ -21,19 +21,26 @@ function generic_randn(StoreT::Type{<:Dense}, dim::Integer=0)
...
+
+  data = generic_zeros(DataT, dim(dims))
+  StoreT = set_datatype(StoreT, typeof(data))
+  return StoreT(data)
 end
 
-function generic_zeros(StoreT::Type{<:Dense}, dim::Integer=0)
-  return generic_zeros(default_storagetype(), dim)
+function generic_zeros(::Type{<:Dense}, ::Tuple{Integer,Integer,Vararg{Integer}})
+  return error("Can't make a multidimensional `Dense` object.")
```
mtfishman, line 45: Instead of this definition just for catching issues with `Dense`, I think we should make a more systematic fix to the generic `generic_zeros` and `generic_randn` definitions. I think they are too general and should be constrained such that `ndims(arraytype)` == length(dims)`.

NDTensors/src/dense/fill.jl
```diff
@@ -21,19 +21,26 @@ function generic_randn(StoreT::Type{<:Dense}, dim::Integer=0)
...
 end
 
-function generic_zeros(
-  StoreT::Type{<:Dense{ElT,DataT}}, dim::Integer=0
-) where {DataT<:AbstractArray,ElT}
-  @assert ElT == eltype(DataT)
-  data = generic_zeros(DataT, dim)
-  StoreT = set_datatype(StoreT, typeof(data))
-  return StoreT(data)
+function generic_zeros(StoreT::Type{<:Dense}, dim::Integer)
```
mtfishman, line 24: I think this should become a generic definition for `generic_zeros` for any `Type{<:AbstractArray}`.

NDTensors/test/ITensors/TestITensorDMRG/TestITensorDMRG.jl
kmp5VT: Metal fails because of the `resize!` problem but I removed the CUDA and ROCM functions

kmp5VT: @mtfishman should I make an `append!` for Metal which converts it to CPU and then back to the GPU?

NDTensors/ext/NDTensorsCUDAExt/append.jl
kmp5VT: I made `Base.append!` functions for each GPU backend in their extension folders. I am closing this comment

NDTensors/ext/NDTensorsCUDAExt/copyto.jl
kmp5VT: This has been removed

NDTensors/ext/NDTensorsCUDAExt/mul.jl
kmp5VT: This has been removed, thanks!

NDTensors/test/ITensors/TestITensorDMRG/TestITensorDMRG.jl
mtfishman: I'm not sure if that is possible to do in-place right now with having something like `resize!`.

NDTensors/src/lib/GPUArraysCoreExtensions/src/gpuarrayscore.jl
kmp5VT: Okay I made an append for each GPU backend and I made a comment about how metal is currently broken and removed the out of date append!! code from the library

NDTensors/src/dense/fill.jl
kmp5VT: It is not a constraint that `generic_zeros` only take `dim::Integer` 
```
julia> NDTensors.generic_zeros(Array, (2,3))
2×3 Matrix{Float64}:
 0.0  0.0  0.0
 0.0  0.0  0.0
 ``` 
 The constraint is that `Dense` only accepts Vectors.
This issue does exist for `generic_randn(::Type{<:AbstractArray})` so I fixed that.

File: NDTensors/test/ITensors/TestITensorDMRG/TestITensorDMRG.jl (hunk 10 above)
Comment (kmp5VT) [reply NDTensors/test/ITensors/TestITensorDMRG/TestITensorDMRG.jl:22]: Good point I forgot about it having to be in-place. I added a comment in-line. Thanks!

File: NDTensors/src/dense/fill.jl (hunk 12 above)
Comment (mtfishman) [reply NDTensors/src/dense/fill.jl:45]: Right, but what I mean is that right now the current version allows `NDTensors.generic_zeros(Vector, (2,3))` which I don't think it should. More generally, I don't think it makes sense to allow inputting an array type where the dimensions are specified and where `dims` doesn't match that specified number of dimensions.

Comment (mtfishman) [reply NDTensors/src/dense/fill.jl:45]: I agree that if the dimensions aren't specified, like in your example `generic_zeros(Array, (2, 3))`, then any length `dims` should be allowed.

**Comment by kmp5VT** [comment]
I updated the Metal code and made a MetalArrayAdaptor but I still need to address the `generic_randn` and `generic_zeros` code and I also need to do a review of the code to make sure its all consistent.

**Comment by kmp5VT** [comment]
I was able to fix the tests in the same way that I did on Jenkins by creating a random project, then developing first NDTensors and then ITensors. Is this sufficient? Do we want to update the other git actions to match?

File: NDTensors/ext/NDTensorsAMDGPUExt/append.jl [hunk 14]
```diff
@@ -0,0 +1,7 @@
+using GPUArraysCore: @allowscalar
+using AMDGPU: ROCArray
+using NDTensors.Expose: Exposed, unexpose
+
```
kmp5VT, line 4: TODO add warning that this is extremely slow on GPU.

NDTensors/src/dense/fill.jl
kmp5VT: Okay I was able to update AbstractArrays `generic_zeros` and `generic_randn` to verify that the length of dims = ndims(Arraytype) but it does require `TypeParameterAccessors.position(::Type, ::typeof(ndims))` to be defined

NDTensors/ext/NDTensorsAMDGPUExt/append.jl
kmp5VT: I have added these warnings in all the GPU codes

NDTensors/src/abstractarray/fill.jl
```diff
@@ -1,13 +1,38 @@
...
 
+function generic_zeros(arraytype::Type{<:AbstractArray}, dims::Tuple)
+  return generic_zeros(arraytype, dims...)
+end
+
 function generic_zeros(arraytype::Type{<:AbstractArray}, dims...)
-  arraytype_specified = specify_default_type_parameters(unwrap_array_type(arraytype))
+  ## Right now I have to call specify first because its not possible to 
+  ## call specify_default_type_parameter and overwrite a default parameter.
+  ## So if arraytypes ndim is unspecified, specify it to the length of dims...
```
mtfishman, line 30: I understand why you designed the code this way but I don't understand this comment.

kmp5VT: Sorry this was a comment reminder for me to make one on github to ask you about. 
What I mean is right now I don't believe that there is a function which can fill parameters of a struct with both default and non-default values. I was going to ask if you believe this would be a useful function to create (most likely in a subsequent PR)

mtfishman: That's a good question. We could do that by defining a `DefaultTypeParamer()` object (if we don't have one already), and have an interface like `specify_type_parameters(Array, (eltype, ndims), (DefaultTypeParamer(), 3)`.

kmp5VT: This makes sense and I think would be a good idea! I will add this comment to the google doc to implement in a subsequent PR. Thanks!

NDTensors/src/dense/fill.jl
mtfishman: Right now you have organized the code like this:
```julia
# Implementation, catches if `ndims(arraytype) != length(dims)`.
function generic_zeros(arraytype::Type{<:AbstractArray}, dims...)

# Forwards to `generic_zeros(::Type{<:AbstractArray}, ::Tuple)`
function generic_zeros(::Type{<:AbstractArray}, dims...)

# Forwards to `generic_zeros(::Type{<:Dense}, ::Tuple)`
function generic_zeros(StoreT::Type{<:Dense}, dim::Integer)

# Implementation
function generic_zeros(::Type{<:Dense}, dims::Tuple{Integer})

# Catches errors
function generic_randn(::Type{<:Dense}, ::Tuple{Integer,Integer,Vararg{Integer}})
```
You can see that it is inconsistent which signature is responsible for the implementation and which signature is responsible for catching dimension errors in the generic case and in the `Dense` case. I'm suggesting organizing it like this:
```julia
# Implementation, catches if `ndims(arraytype) != length(dims)`.
function generic_zeros(arraytype::Type{<:AbstractArray}, dims::Tuple)

# Forwards to `generic_zeros(::Type{<:AbstractArray}, ::Tuple)`
function generic_zeros(::Type{<:AbstractArray}, dims...)

# Implementation
function generic_zeros(::Type{<:Dense}, dims::Integer)
```
If my logic is correct, then `generic_zeros(Dense, (2,))` will forward to `generic_zeros(Dense, 2)` through the `generic_zeros(::Type{<:AbstractArray}, ::Tuple)` function, and calls like `generic_zeros(Dense, (2, 2))` or `generic_zeros(Dense, 2, 2)` will end up in the generic `generic_zeros(arraytype::Type{<:AbstractArray}, dims::Tuple)` version and will throw an error because of a dimension mismatch.

kmp5VT: Ahh yes that is something I didn't check. I updated the code and now it no longer throws an error! One question is that I found in this new structure is that 
`generic_zeros(Dense, (2,3)) = Error: Can't make a multidimensional `Dense` object.` which is what we expect. But when I do this `generic_zeros(Dense, 2,3)` it calls the `AbstractArray` version which ultimately fails but is unexpected behavior. To fix this I have these two catches.
```
function generic_zeros(::Type{<:Dense}, ::Integer, ::Vararg{Integer})
  return error("Can't make a multidimensional `Dense` object.")
end

function generic_zeros(::Type{<:Dense}, ::Tuple{Integer,Integer,Vararg{Integer}})
  return error("Can't make a multidimensional `Dense` object.")
end
```
Does this seem reasonable?

mtfishman: The way I would design it would be like this:
```julia
# Generic `AbstractArray` implementation, also catches if `ndims(arraytype) != length(dims)`.
function generic_zeros(arraytype::Type{<:AbstractArray}, dims::Integer...) ... end

# Forwards to `generic_zeros(::Type{<:AbstractArray}, ::Integer...)`
function generic_zeros(::Type{<:AbstractArray}, dims::Tuple{Vararg{Integer}}) ... end

# `Dense` implementation.
function generic_zeros(::Type{<:Dense}, dims::Integer) ... end
```
Does that work and have all of the behavior we want?

NDTensors/src/abstractarray/fill.jl
mtfishman, line 1: We should rename this file, it is called `fill.jl` but doesn't even have a `fill(...)` constructor. Maybe `generic_array_constructors.jl` would be more appropriate and descriptive.

kmp5VT: This seems to make sense, I have updated the file names

NDTensors/src/dense/fill.jl
mtfishman: @kmp5VT what about this final design I suggested? I think the code you wrote is still a bit more complicated than it needs to be. Also I think we should make `generic_zeros(arraytype::Type{<:AbstractArray}, dims::Integer...)` in inner implementation function and forward to that from `generic_zeros(arraytype::Type{<:AbstractArray}, dims::Tuple{Vararg{Integer}})`, that is more standard for how the analgous Base Julia functions are defined.

kmp5VT: Hi sorry I missed this comment on my last go through. The issue with the design as you have written it is that codes like
```
generic_zeros(Dense, 2,3)
generic_zeros(Dense, (2,3))
```
call the `AbstractArray` version which is incorrect behavior.  I added the extra code to catch these cases and throw an error.

I swapped the function, though this requires rewrititng the dense function like this 
```
function generic_zeros(StoreT::Type{<:Dense}, dims::Tuple{Integer})
```

mtfishman: Right, but then wouldn't it catch that `ndims(Dense) != length((2, 3))`, since `ndims(Dense) == 1`?

kmp5VT: No I am not using `ndims` I am using `type_parameter(Dense, ndims) = ERROR: Not implemented` because Dense doesn't have an `ndims` parameter

mtfishman: I see, but that seems like more an issue of how you've implemented `generic_zeros(arraytype::Type{<:AbstractArray}, dims...)`.

mtfishman: I.e. maybe it shouldn't be calling `type_parameter(Dense, ndims)`, and instead should call `ndims(Dense)`.

src/imports.jl
mtfishman: @kmp5VT please bump the minor version of `NDTensors` in `Project.toml` and then change `Project.toml` of `ITensors` accordingly, including bumping the patch version of `ITensors`.

NDTensors/src/dense/fill.jl
mtfishman: Hopefully you understand my motivation for harping on this point, if we make the generic version work better and more general it makes it easier to define specialized versions for new array types, just like our over all goal with the GPU code of making better generic code to lower the barrier for defining new GPU backends.

NDTensors/src/dense/generic_array_constructors.jl
```diff
@@ -0,0 +1,31 @@
...
+  StoreT = specify_default_type_parameters(StoreT)
+  DataT = specify_type_parameter(type_parameter(StoreT, parenttype), eltype, eltype(StoreT))
+  @assert eltype(StoreT) == eltype(DataT)
+
+  data = generic_randn(DataT, dims; rng=rng)
+  StoreT = set_datatype(StoreT, typeof(data))
+  return StoreT(data)
+end
+
+function generic_zeros(StoreT::Type{<:Dense}, dims::Tuple{Integer})
```
mtfishman, line 23:
```suggestion
function generic_zeros(StoreT::Type{<:Dense}, dims::Integer)
```
to be consistent with `generic_randn`.

mtfishman: `generic_zeros(StoreT::Type{<:AbstractArray}, dims...)` definitions should be updated accordingly.

NDTensors/src/dense/fill.jl
kmp5VT: Using `ndims(arraytype_specified)` doesn't fix the issue because the first thing that happens in both of these `generic_x(::Type{<:AbstractArray})` functions is `arraytype_specified = specify_type_parameter(unwrap_array_type(arraytype), ndims, length(dims)` and `Dense` does not have a position for `ndims` defined. And though we have a `set_ndims` function defined, we don't have a `specify_ndims` function. 

I totally understand the goal and appreciate your perspective! It helps with seeing holes in my logic and finding more robust solutions.

NDTensors/src/dense/generic_array_constructors.jl
kmp5VT: In order to have both of these lines work
```
generic_randn(Dense, 2)
generic_randn(Dense, (2,))
``` 
I updated the `generic_randn` definition to `generic_randn(StoreT::Type{<:Dense}, dims::Tuple{Integer})`. Previously this would work because the tuple would be piped to an int with the `generic_randn(arraytype, dims...)` call but now the Tuple version goes into the `AbstractArray` version of the function.

I don't understand what you mean by 
```
generic_zeros(StoreT::Type{<:AbstractArray}, dims...) definitions should be updated accordingly.
```

NDTensors/src/dense/fill.jl
mtfishman: Right, that just seems to be another indication that the logic of `generic_x(::Type{<:AbstractArray}, ...)` isn't quite correct or general enough and should be rethought, since it is a very reasonable case to input a vector or matrix type that has a hard-coded `ndims` so doesn't have an `ndims` type parameter.

NDTensors/src/dense/generic_array_constructors.jl
mtfishman: I think we are talking past each other here, why don't you do your best to implement things as I've described and let me know when it is ready to be reviewed.

NDTensors/src/dense/generic_array_constructors.jl
```diff
@@ -0,0 +1,33 @@
...
+  specify_default_type_parameters,
+  type_parameter
+##TODO replace randn in ITensors with generic_randn
+## and replace zeros with generic_zeros
+
+# This is a file to write generic fills for NDTensors.
+#  This includes random fills, zeros, ...
+
+function generic_randn(
+  StoreT::Type{<:Dense}, dims::Tuple{Integer}; rng=Random.default_rng()
```
mtfishman, line 14:
```suggestion
  StoreT::Type{<:Dense}, dims::Integer; rng=Random.default_rng()
```

NDTensors/src/dense/generic_array_constructors.jl
```diff
@@ -0,0 +1,33 @@
...
+  StoreT = specify_default_type_parameters(StoreT)
+  DataT = specify_type_parameter(type_parameter(StoreT, parenttype), eltype, eltype(StoreT))
+  @assert eltype(StoreT) == eltype(DataT)
+
+  data = generic_randn(DataT, dims; rng=rng)
+  StoreT = set_datatype(StoreT, typeof(data))
+  return StoreT(data)
+end
+
+function generic_zeros(StoreT::Type{<:Dense}, dims::Tuple{Integer})
```
mtfishman, line 25:
```suggestion
function generic_zeros(StoreT::Type{<:Dense}, dims::Integer)
```

mtfishman: My proposal was to define both of them in terms of `::Integer`, not `::Tuple{Integer}`, since it is a simpler interface for new types to overload in general.

kmp5VT: If you define with integer this code fails.
 ```
julia> NDTensors.generic_zeros(Dense, (2,))
ERROR: Not implemented
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] datatype(#unused#::Type{Dense})
   @ NDTensors ~/.julia/dev/ITensors/NDTensors/src/tensorstorage/tensorstorage.jl:5
 [3] parenttype(storagetype::Type{Dense})
   @ NDTensors ~/.julia/dev/ITensors/NDTensors/src/tensor/set_types.jl:26
 [4] is_wrapped_array(arraytype::Type{Dense})
   @ NDTensors.TypeParameterAccessors ~/.julia/dev/ITensors/NDTensors/src/lib/TypeParameterAccessors/src/base/abstractarray.jl:31
 [5] trait(#unused#::Type{NDTensors.TypeParameterAccessors.IsWrappedArray{Dense}})
   @ NDTensors.TypeParameterAccessors ~/.julia/packages/SimpleTraits/l1ZsK/src/SimpleTraits.jl:171
 [6] unwrap_array_type
   @ ~/.julia/packages/SimpleTraits/l1ZsK/src/SimpleTraits.jl:331 [inlined]
 [7] generic_zeros(arraytype::Type{Dense}, dims::Tuple{Int64})
   @ NDTensors ~/.julia/dev/ITensors/NDTensors/src/abstractarray/generic_array_constructors.jl:25
 [8] top-level scope
   @ REPL[2]:1
 ```

kmp5VT: My only concern is that if you do something like this `generic_zeros(Dense, (2,))` it calls the `AbstractArray` version of this funciton

NDTensors/src/dense/generic_array_constructors.jl
kmp5VT: Updated

NDTensors/src/dense/generic_array_constructors.jl
mtfishman: Right, so then you should rewrite the `AbstractArray` version so that doesn't happen.

kmp5VT: I am not sure I understand what you're suggesting to do. I reverted the code back the original design you suggested which works for `generic_zeros(Dense, 2)` and `generic_zeros(Dense, (2,))`

kmp5VT:
@mtfishman If you have a minute can you please take a look at this PR again? Thank you I appreciate your time!

NDTensors/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "NDTensors"
 uuid = "23ae76d9-e61a-49c4-8f12-3f1a16adf9cf"
 authors = ["Matthew Fishman <mfishman@flatironinstitute.org>"]
-version = "0.2.30"
+version = "0.2.31"
```
mtfishman, line 4: I meant this:
```suggestion
version = "0.3.0"
```
in order to indicate a breaking change according to [semantic versioning](https://pkgdocs.julialang.org/v1/compatibility/#compat-pre-1.0).

NDTensors/src/abstractarray/generic_array_constructors.jl
```diff
@@ -0,0 +1,36 @@
...
+  arraytype_specified = specify_default_type_parameters(arraytype_specified)
+  @assert length(dims) == ndims(arraytype_specified)
+  data = similar(arraytype_specified, dims...)
+  return randn!(rng, data)
+end
+
+function generic_randn(
+  arraytype::Type{<:AbstractArray}, dims::Tuple; rng=Random.default_rng()
+)
+  return generic_randn(arraytype, dims...; rng=rng)
```
mtfishman, line 20:
```suggestion
  return generic_randn(arraytype, dims...; rng)
```
I prefer that style when it can be used.

Project.toml
```diff
@@ -52,7 +52,7 @@ IsApprox = "0.1"
 KrylovKit = "0.4.2, 0.5, 0.6, 0.7"
 LinearAlgebra = "1.6"
 LinearMaps = "3"
-NDTensors = "0.2.30"
+NDTensors = "0.2.31"
```
mtfishman, line 55:
```suggestion
NDTensors = "0.3"
```

NDTensors/src/abstractarray/generic_array_constructors.jl
kmp5VT: Updated, thanks!

Project.toml
kmp5VT: Oh yeah I didn't realize this was the convention. Thank you for the explanation!

NDTensors/Project.toml
kmp5VT: Thanks for the reference! I updated this, thanks!

kmp5VT: Thank you for the reference, I updated this thanks!

mtfishman:
Looks good, thanks!

PR #1366, opened 2024-03-27
[ITensorMPS] Move ITensorTDVP code into ITensorMPS module

emstoudenmire:
This PR moves all the code from ITensorTDVP into the ITensorMPS module.
All files were included at the top level, except the AbstractProjMPO related files which were put into the abstractprojmpo/ folder.
Main changes:
- Update various `using` statements to reflect the new location of the code inside the ITensorMPS module.
- Add Observers.jl as a dependency
- Rename ITensorTDVP.dmrg to alternating_dmrg for now
- Update some dmrg code to call new `default_cutoff` interface


src/ITensorMPS/abstractprojmpo/projmpo_mps2.jl
```diff
@@ -0,0 +1,46 @@
+using ITensors: contract
```
mtfishman, line 1:
```suggestion
using ITensors: ITensors, contract
```
since we are overloading `contract` with `ITensors.contract`.

mtfishman:
I think there are two issues we should also address:
1. We should set it up so that we can run the `ITensorTDVP.jl` tests from within PRs to this repository, to make sure we catch if we break anything proactively instead of after the fact. It should be easy to make a Github Actions workflow to do that. We should probably set that up and merge it to `main` as a separate PR before merging this. We can set it up to be triggered by a comment.
2. If we do break tests in `ITensorTDVP.jl` during the development of `ITensors.ITensorMPS`, we should decide how to handle that in terms of versioning of `ITensors.jl`. I think at that point we should bump the minor version of `ITensors.jl` (say from `v0.3` to `v0.4`), to make it clear to external users which versions of `ITensors.jl` are compatible with which versions of `ITensorTDVP.jl`, related to the discussion on the ITensor Slack channel about handling the versioning of `NDTensors.jl` and `ITensors.jl`. I don't think we need to bump to ITensors to v0.4 for this PR, since if I understand correctly, this doesn't break the current version of `ITensorTDVP.jl`, i.e. the current registered version of ITensorTDVP work perfectly well with or without this PR. However, we will need to mark in ITensorTDVP.jl that the next version requires whatever the next registered version of ITensors.jl is (say v0.3.59, since right now it is v0.3.58).

emstoudenmire:
I just made both of the suggested changes, and will push those when I figure out why the tests were failing (I think it's related to an overload of `NDTensors.scalartype` made in ITensorMPS).

Definitely a good idea to start running the ITensorTDVP tests. Would it make sense as a simpler move to just copy the actual testing files to a subfolder (or not even a subfolder) of the ITensorMPS tests? Then there wouldn't need to be any Github configuration. (We could then delete the tests from ITensorTDVP.jl so there's no code duplication.) I guess one downside of that idea is that the ITensorTDVP package would then effectively have no examples or 'test documentation' associated with it.

That's right about this PR (together with the other) not making any breaking changes to ITensorTDVP. Hopefully the upcoming breaking changes to it would be pretty light, or sort of "pseudo-breaking" if they involve users who have been dipping into things like the projected MPO types themselves.

mtfishman:
> Definitely a good idea to start running the ITensorTDVP tests. Would it make sense as a simpler move to just copy the actual testing files to a subfolder (or not even a subfolder) of the ITensorMPS tests? Then there wouldn't need to be any Github configuration. (We could then delete the tests from ITensorTDVP.jl so there's no code duplication.) I guess one downside of that idea is that the ITensorTDVP package would then effectively have no examples or 'test documentation' associated with it.

I think it would be better to keep the tests as they are now, in separate repositories. We can consider the tests and examples (besides this example:  https://github.com/ITensor/ITensorTDVP.jl/blob/v0.2.1/examples/05_utils.jl, which is using too many internals) in `ITensorTDVP.jl` to be the "definition" of that package. For example, if we make a change to `ITensors.ITensorMPS` that purposefully breaks some functionality (and a corresponding test) in `ITensorTDVP.jl`, we can then handle that by bumping the minor version of `ITensors.jl` (i.e. `v0.3.x` to `v0.4.0`), updating the tests and versions of `ITensorTDVP.jl` accordingly, and registering a new version of `ITensorTDVP.jl`, also with a minor version bump. Basically I think it will make it easier to tie a certain version of the ITensorTDVP.jl tests to a version of that package, and also make it easier for users to test ITensorTDVP on it's own to make sure things are configured correctly.

> That's right about this PR (together with the other) not making any breaking changes to ITensorTDVP. Hopefully the upcoming breaking changes to it would be pretty light, or sort of "pseudo-breaking" if they involve users who have been dipping into things like the projected MPO types themselves.

Right, I think we can now mostly consider changes that don't break tests of ITensors.jl and ITensorTDVP.jl to be internal changes that just warrant changing the patch version (i.e. `v0.3.x` to `v0.3.x+1`) of the package. A package we'll have to update if we make changes to the projected MPO functionality is [ITensorParallel.jl](https://github.com/ITensor/ITensorParallel.jl) but that shouldn't be too hard. I think the changes done in #1356/ #1362 are a first test of what might break if we change a lot of internals of `ITensors.jl` MPS code, which wasn't too bad. I'm leaning towards saying we should maybe bump the minor version to indicate a breaking change if we make a big change to the projected MPO system, to make things a bit friendlier for users who may have been overloading that interface themselves, but we can see how things shake out. Ideally we will try to keep the interface of that system preserved for as long as possible before going and fully changing it, but that may be difficult to do.

One reason to do a minor version bump if we change the projected MPO system is that then users who were relying on the internals of that system can be better informed about how to pin their version of `ITensors.jl` so their code relying on that system continues to work, while they try to update to whatever new system we end up with. Again, that may be a bit tough to do in practice, but it is something to aim for. My concern is that users who try to update their code, rather than just pinning to an older version of ITensors, may have a moving target since likely we will keep iterating on the design for some time. So to mitigate that, we might want to communicate that to users, and bump the minor version of ITensors.jl with the first breakage we make to the projected MPO system, and tell users who were relying on those internals to not try to update their code for the time being.

emstoudenmire:
Sounds good. I just did a push that should hopefully have tests passing. Then the last thing for this PR will be to configure the ITensorTDVP tests to be run from this repo (using a comment trigger).

I like the plan about the versioning. I definitely see what you mean about needing to communicate to users like Ash and also giving them a way to pin versions that will work for them.

emstoudenmire:
[test ITensorMPS]

emstoudenmire:
[test ITensorMPS]

emstoudenmire:
[test ITensorMPS]

emstoudenmire:
[test ITensorTDVP]

emstoudenmire:
[test ITensorTDVP]

emstoudenmire:
[test ITensorTDVP]

emstoudenmire:
[test ITensorTDVP]

emstoudenmire:
All tests passing including the new ITensorTDVP one 🙌

I think there's still some question about if the ITensorTDVP test workflow is correctly using the PR version of ITensorMPS or the one on main. (I don't know if that's the case, I'm just not sure). Please let me know if you can think of a way to check which one is the case (I guess intentionally introducing a bug in a test PR to see whether the ITensorTDVP tests fail?).

The plan about the versioning sounds good. Perhaps one way to go is to leave most of the AbstractProjMPO system untouched and roll out a new system alongside it that we switch to as the default.

mtfishman:
> All tests passing including the new ITensorTDVP one 🙌

Great! I would like to add workflows for other downstream dependencies, such as `ITensorNetworks.jl`, `ITensorParallel.jl`, `ITensorChemistry.jl`, `PastaQ.jl`, etc. so this is a nice reference point for that.

If I'm thinking about this correctly, I don't think we need to test downstream dependencies on Julia 1.6. Presumably both `ITensors.jl` and the downstream package (`ITensorTDVP.jl`) are tested on Julia 1.6, and our only purpose is to make sure we don't make a change to `ITensors.jl` that breaks `ITensorTDVP.jl`. I think it would be unlikely for that to occur only on Julia 1.6, if both are tested separately on Julia 1.6 already.

> I think there's still some question about if the ITensorTDVP test workflow is correctly using the PR version of ITensorMPS or the one on main. (I don't know if that's the case, I'm just not sure). Please let me know if you can think of a way to check which one is the case (I guess intentionally introducing a bug in a test PR to see whether the ITensorTDVP tests fail?).

Seems reasonable to double check that by introducing a bug and making sure it properly breaks the tests, we could try that in this PR.

> The plan about the versioning sounds good. Perhaps one way to go is to leave most of the AbstractProjMPO system untouched and roll out a new system alongside it that we switch to as the default.

That seems like a reasonable thing to try. One potential downside with that kind of approach is that if we want to test the new system inside `dmrg`, etc. it may require modifying `dmrg`, so then we have to figure out how to support both systems until we get to the point of removing `AbstractProjMPO`. So the devil is in the details if that is practical to do. Of course it is worth trying to do that for as long as is practical/possible.

Project.toml
```diff
@@ -17,6 +17,7 @@ KrylovKit = "0b1a1467-8014-51b9-945f-bf0ae24f4b77"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 LinearMaps = "7a12625a-238d-50fd-b39a-03d52299707e"
 NDTensors = "23ae76d9-e61a-49c4-8f12-3f1a16adf9cf"
+Observers = "338f10d5-c7f1-4033-a7d1-f9dec39bcaa0"
```
mtfishman, line 20: This reminds me that I would like to change `Observers` to be an optional dependency using a package extension, since `Observers` depends on `DataFrames` which is a relatively big package.

The idea would be that there would be a minimal interface for new observers to overload from `ITensorMPS` (say `ITensorMPS.update!`), and then users can load and pass an observer from `Observers.jl` into a solver. We wouldn't be able to use `Observers` directly inside of the package, but I think that is ok. Any use cases where I was tempted to do that I was able to use another code pattern, such as passing a `Ref` (or some other mutable data structure) as a keyword argument and writing a result into that `Ref`.

This would also have the advantage that we would have a better defined observer interface in `ITensorMPS` independent of `Observers.jl`, so in principle users could define and pass their own observer-like object. This could also help formalize how we can more conveniently keep backwards compatibility with the old observer system, without polluting `alternating_update`.

Not sure if we should do that in this PR or a follow-up, I think a follow-up would be fine, for now we are just forcing users to install an unnecessary dependency (unless they are using `ITensorTDVP.jl` with observers) but I'm sure most users won't notice.

mtfishman: I added a note about this in the Google Doc.

mtfishman:
> > I think there's still some question about if the ITensorTDVP test workflow is correctly using the PR version of ITensorMPS or the one on main. (I don't know if that's the case, I'm just not sure). Please let me know if you can think of a way to check which one is the case (I guess intentionally introducing a bug in a test PR to see whether the ITensorTDVP tests fail?).
> 
> Seems reasonable to double check that by introducing a bug and making sure it properly breaks the tests, we could try that in this PR.

Regarding this point, note that if you read the log for the `ITensorTDVP` CI run, you can see sections titled "Checkout PR branch" where you can see it checking out the latest commit of this PR branch. I remember getting that configured properly was a difficult part of getting the other comment trigger workflows working properly.

src/ITensorMPS/alternating_update.jl
```diff
@@ -0,0 +1,147 @@
...
+  time_start=default_time_start(),
+  time_step=default_time_step(t),
+  order=default_order(),
+  (observer!)=default_observer!(),
+  (step_observer!)=default_step_observer!(),
+  outputlevel=default_outputlevel(),
+  normalize=default_normalize(),
+  maxdim=default_maxdim(),
+  mindim=default_mindim(),
+  cutoff=default_cutoff(Float64),
```
mtfishman, line 58: Seems better to make this `default_cutoff(scalartype(psi0))`, say for example it is single precision then the cutoff should be adjusted. Same comment in other places where `default_cutoff` is used.

Of course in general you may want to account for the scalar type of the operator as well, but the state seems like a reasonable stand-in and we could always make it more sophisticated as needed.

mtfishman:
This looks good to go from my end, ready to merge?

emstoudenmire:
Glad to hear the ITensorTDVP workflow is probably configured correctly. I'll still give it a test but I see what you mean about the checking out this PR part.

PR #1370, opened 2024-03-31
[ITensors] Make `scalartype` accessible as `ITensors.scalartype`

mtfishman:
`ITensorTDVP.jl` appears to rely on the function `scalartype` (which is defined in `NDTensors`) being available as `ITensors.scalartype`.

It seems like that was dropped during the `ITensorMPS` reorganization, since now only `ITensorMPS` uses `scalartype`. This fixes that by adding `using NDTensors: scalartype` inside of the `ITensors` module, and adds a test that it is accessible as `ITensors.scalartype`.

I also made some namespace improvements in some tests.

**EDIT:** I also added a test file that tests what names are exported by `ITensors`, which I think is good practice to do from now on for any package/module. Before we were implicitly testing what got exported by typing `using ITensors` and then using a function in a test, which is both finicky and we are moving away from that import/using style.


mtfishman:
[test ITensorMPS] [test ITensorTDVP] [test ITensorGaussianMPS]

emstoudenmire:
Thanks... was just about to work on this

emstoudenmire:
By the way, I meant to ask what `@eval module $(gensym())` does in the testing files? I noticed that also in the ITensorTDVP tests and saw that you are adding it here as well.

mtfishman:
> By the way, I meant to ask what `@eval module $(gensym())` does in the testing files? I noticed that also in the ITensorTDVP tests and saw that you are adding it here as well.

It generates a module with a random name, i.e.:
```julia
julia> M1 = @eval module $(gensym())
       f() = 2
       end
Main.var"##226"

julia> M2 = @eval module $(gensym())
       f() = 3
       end
Main.var"##227"

julia> M1.f()
2

julia> M2.f()
3
```
I have found that it is better to wrap test files in modules, since it makes them more self contained, i.e. you can control exactly what goes into the namespace to be used by the tests in that test file. That makes sure you aren't accidentally using a variable or function that was defined in a previous test file (or even some other code you ran, if you are running the tests with `include("runtests.jl")`), and also makes sure that when you type something like `include("runtests.jl")` it doesn't pollute the namespace with temporary variables. On top of that, using `gensym()` is just a trick to make a random unique module name so we don't have to come up with new names for modules and potentially have clashes between module names.

Maybe it is better to make a small macro:
```julia
@temp_module
# Test code here
end
```
which internally is just defined as `@eval module $(gensym())` to make the intent clearer.

This kind of thing (wrapping test files in modules) is pretty commonly suggested for avoiding potential namespace issues in tests:
https://github.com/YingboMa/SafeTestsets.jl
https://discourse.julialang.org/t/setup-and-teardown-in-test-module/68589
https://discourse.julialang.org/t/how-to-avoid-name-conflicts-in-mypackage-runtests-jl/57186
https://discourse.julialang.org/t/best-practices-for-julia-unit-testing/30858
https://discourse.julialang.org/t/my-unit-tests-are-interferring-with-each-other/21686

`@eval module $(gensym())` is essentially how `SafeTestsets.jl` works, though that package introduces a new macro `@safetestset` that is meant to be a replacement for `@testset` as opposed to wrapping the entire file in a module, but I found that package had some issues with tests that involved macros.

emstoudenmire:
Nice, thanks for explaining. I do think the idea of a macro would be a good one down the road to make it more 'self documenting'. But not a big priority of course.

mtfishman:
For example:
```julia
julia> macro temp_module(expr)
         return :(@eval module $(gensym()) $expr end)
       end
@temp_module (macro with 2 methods)

julia> M = @temp_module begin
         f() = 2
       end
Main.var"##231"

julia> M.f()
2
```
Probably that is less opaque than directly using `@eval module $(gensym())`.

emstoudenmire:
I do like that better. And perhaps even `@temporary_module` since it's not something that one would be typing too often.

mtfishman:
Ah I see an issue with the current testing setup to test against `ITensorTDVP`. I think even though the workflow is checking out this PR, `ITensorTDVP` is still configured to use the registered version of `ITensors.jl` instead of the `dev` version, so in practice it is still testing against the registered version `v0.3.59` which doesn't have this change to `scalartype`.

I'm sure there is a way to fix that, probably in the workflow we have to activate the environment of `ITensorTDVP` and add something like `using Pkg: Pkg; Pkg.develop("ITensors")`.

mtfishman:
I see the tests also generate a lot of warnings like:
```julia
WARNING: Method definition update!(ITensors.ITensorMPS.AbstractObserver) in module ITensorMPS at /Users/mfishman/.julia/dev/ITensors/src/ITensorMPS/update_observer.jl:3 overwritten in module ITensorTDVP at /Users/mfishman/.julia/packages/ITensorTDVP/OmMNj/src/update_observer.jl:4.
ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
WARNING: Method definition update!(ITensors.ITensorMPS.AbstractObserver) in module ITensorMPS at /Users/mfishman/.julia/dev/ITensors/src/ITensorMPS/update_observer.jl:3 overwritten in module ITensorTDVP at /Users/mfishman/.julia/packages/ITensorTDVP/OmMNj/src/update_observer.jl:4.
WARNING: Method definition kwcall(NamedTuple{names, T} where T<:Tuple where names, typeof(Observers.update!), ITensors.ITensorMPS.AbstractObserver) in module ITensorMPS at /Users/mfishman/.julia/dev/ITensors/src/ITensorMPS/update_observer.jl:3 overwritten in module ITensorTDVP at /Users/mfishman/.julia/packages/ITensorTDVP/OmMNj/src/update_observer.jl:4.
WARNING: Method definition contract(NDTensors.AlgorithmSelection.Algorithm{:fit, Kwargs} where Kwargs<:(NamedTuple{names, T} where T<:Tuple where names), ITensors.ITensorMPS.MPO, ITensors.ITensorMPS.MPS) in module ITensorMPS at /Users/mfishman/.julia/dev/ITensors/src/ITensorMPS/contract_mpo_mps.jl:15 overwritten in module ITensorTDVP at /Users/mfishman/.julia/packages/ITensorTDVP/OmMNj/src/contract_mpo_mps.jl:26.
WARNING: Method definition kwcall(NamedTuple{names, T} where T<:Tuple where names, typeof(NDTensors.contract), NDTensors.AlgorithmSelection.Algorithm{:fit, Kwargs} where Kwargs<:(NamedTuple{names, T} where T<:Tuple where names), ITensors.ITensorMPS.MPO, ITensors.ITensorMPS.MPS) in module ITensorMPS at /Users/mfishman/.julia/dev/ITensors/src/ITensorMPS/contract_mpo_mps.jl:15 overwritten in module ITensorTDVP at /Users/mfishman/.julia/packages/ITensorTDVP/OmMNj/src/contract_mpo_mps.jl:26.
WARNING: Method definition linsolve(ITensors.ITensorMPS.MPO, ITensors.ITensorMPS.MPS, ITensors.ITensorMPS.MPS) in module ITensorMPS at /Users/mfishman/.julia/dev/ITensors/src/ITensorMPS/linsolve.jl:25 overwritten in module ITensorTDVP at /Users/mfishman/.julia/packages/ITensorTDVP/OmMNj/src/linsolve.jl:26.
WARNING: Method definition linsolve(ITensors.ITensorMPS.MPO, ITensors.ITensorMPS.MPS, ITensors.ITensorMPS.MPS, Number) in module ITensorMPS at /Users/mfishman/.julia/dev/ITensors/src/ITensorMPS/linsolve.jl:25 overwritten in module ITensorTDVP at /Users/mfishman/.julia/packages/ITensorTDVP/OmMNj/src/linsolve.jl:26.
WARNING: Method definition linsolve(ITensors.ITensorMPS.MPO, ITensors.ITensorMPS.MPS, ITensors.ITensorMPS.MPS, Number, Number) in module ITensorMPS at /Users/mfishman/.julia/dev/ITensors/src/ITensorMPS/linsolve.jl:25 overwritten in module ITensorTDVP at /Users/mfishman/.julia/packages/ITensorTDVP/OmMNj/src/linsolve.jl:26.
WARNING: Method definition kwcall(NamedTuple{names, T} where T<:Tuple where names, typeof(KrylovKit.linsolve), ITensors.ITensorMPS.MPO, ITensors.ITensorMPS.MPS, ITensors.ITensorMPS.MPS) in module ITensorMPS at /Users/mfishman/.julia/dev/ITensors/src/ITensorMPS/linsolve.jl:25 overwritten in module ITensorTDVP at /Users/mfishman/.julia/packages/ITensorTDVP/OmMNj/src/linsolve.jl:26.
WARNING: Method definition kwcall(NamedTuple{names, T} where T<:Tuple where names, typeof(KrylovKit.linsolve), ITensors.ITensorMPS.MPO, ITensors.ITensorMPS.MPS, ITensors.ITensorMPS.MPS, Number) in module ITensorMPS at /Users/mfishman/.julia/dev/ITensors/src/ITensorMPS/linsolve.jl:25 overwritten in module ITensorTDVP at /Users/mfishman/.julia/packages/ITensorTDVP/OmMNj/src/linsolve.jl:26.
WARNING: Method definition kwcall(NamedTuple{names, T} where T<:Tuple where names, typeof(KrylovKit.linsolve), ITensors.ITensorMPS.MPO, ITensors.ITensorMPS.MPS, ITensors.ITensorMPS.MPS, Number, Number) in module ITensorMPS at /Users/mfishman/.julia/dev/ITensors/src/ITensorMPS/linsolve.jl:25 overwritten in module ITensorTDVP at /Users/mfishman/.julia/packages/ITensorTDVP/OmMNj/src/linsolve.jl:26.
```
We should fix those (say by just not overloading those functions from `ITensors.jl`) and include that along with this PR's fix in a new `ITensors.jl` v0.3.60, before we go ahead with https://github.com/ITensor/ITensorTDVP.jl/pull/67, so that we have a "clean" starting point for proceeding with https://github.com/ITensor/ITensorTDVP.jl/pull/67. Could you fix those overloading issues in a new PR?

mtfishman:
I tested this fix locally and it fixes `ITensorTDVP.jl` so I will merge and register this (since it is pretty bad that ITensorTDVP.jl is broken with the latest version of ITensors.jl) and we can include the fixes to those overloading warnings in the next version. We can also investigate fixing the ITensorTDVP test workflow in future PRs.

mtfishman:
@emstoudenmire a reference point for fixing the downstream testing of `ITensorTDVP` could be this workflow from `BlockArrays`: https://github.com/JuliaArrays/BlockArrays.jl/blob/v0.16.41/.github/workflows/downstream.yml#L49-L65

Some other references for testing a PR against dependencies are:
https://discourse.julialang.org/t/github-actions-workflow-for-reverse-dependency-integration-testing/53536
https://discourse.julialang.org/t/ann-github-action-workflow-to-test-if-your-pr-breaks-another-package-reverse-dependency-integration-testing/57635

mtfishman:
Perhaps we can just use one of those instead. I think they don't support optionally running the test based on a comment trigger but it may be better to just run the test every time anyway.

emstoudenmire:
Sounds good & I'll fix the warnings in another PR.

PR #1371, opened 2024-04-01
[ITensorMPS] Make ITensorTDVP functions into compatibility layer

emstoudenmire:
This PR changes the names of the interface functions moved from ITensorTDVP. Now they are prepended with itensortdvp_ so as to be a compatibility layer behind which we can translate arguments to be able to work on the ITensorMPS.alternating_update code however we want without breaking ITensorTDVP.

### DIFF
--- a/src/ITensorMPS/contract_mpo_mps.jl
+++ b/src/ITensorMPS/contract_mpo_mps.jl
@@ -12,8 +12,8 @@ function contractmpo_solver(; kwargs...)
   return solver
 end
 
-function ITensors.contract(
-  ::Algorithm"fit", A::MPO, psi0::MPS; init_mps=psi0, nsweeps=1, kwargs...
+function itensortdvp_contract(
+  alg::Algorithm"fit", A::MPO, psi0::MPS; init_mps=psi0, nsweeps=1, kwargs...
 )::MPS
   n = length(A)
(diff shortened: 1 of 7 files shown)


mtfishman:
@emstoudenmire I also see this warning:
```julia
WARNING: Method definition update!(ITensors.ITensorMPS.AbstractObserver) in module ITensorMPS at /Users/mfishman/.julia/packages/ITensors/jmIjr/src/ITensorMPS/update_observer.jl:3 overwritten in module ITensorTDVP at /Users/mfishman/.julia/dev/ITensorTDVP/src/update_observer.jl:4.
```
because of the overload here: https://github.com/ITensor/ITensors.jl/blob/v0.3.60/src/ITensorMPS/update_observer.jl.

Let's just remove that for now and decide how `update!` will work in a future PR.

Removing that overload aligns with the plan of moving the `Observers.jl` dependency to a package extension anyway, since that will likely be implemented by defining a new function `ITensorMPS.update!` which will be used in `alternating_update` and can be overloaded on `Observers.Observer` (in an `ITensorsObserversExt` package extension) and `ITensorMPS.AbstractObserver`.

mtfishman:
[test ITensorMPS][test ITensorGaussianMPS]

Issue #1143, opened 2023-07-10
[ITensorGPU] [BUG] cuITensor(m, n) * delta(m, n) causes BoundsError


csmaczny:
**Description of bug**

Contracting a tensor using delta(m, n) causes a BoundsError.

**Minimal code demonstrating the bug or unexpected behavior**

<details><summary>Minimal runnable code</summary><p>

```julia
using ITensors
using ITensorGPU

m = Index(2); n = Index(2)
T = cuITensor(m, n)

T * delta(m, n)
```
</p></details>

**Actual output or behavior**

<details><summary>Output of minimal runnable code</summary><p>

```julia
ERROR: BoundsError: attempt to access 1-element CUDA.CuArray{Float64, 1, CUDA.Mem.DeviceBuffer} at index [4]
Stacktrace:
  [1] throw_boundserror(A::CUDA.CuArray{Float64, 1, CUDA.Mem.DeviceBuffer}, I::Tuple{Int64})
    @ Base ./abstractarray.jl:744
  [2] checkbounds
    @ ./abstractarray.jl:709 [inlined]
  [3] copyto!
    @ ~/.julia/packages/CUDA/tVtYo/src/array.jl:440 [inlined]
  [4] copyto!
    @ ~/.julia/packages/CUDA/tVtYo/src/array.jl:447 [inlined]
  [5] contract!(C::NDTensors.DenseTensor{Float64, 0, Tuple{}, NDTensors.Dense{Float64, CUDA.CuArray{Float64, 1, CUDA.Mem.DeviceBuffer}}}, Clabels::Tuple{}, A::NDTensors.DiagTensor{Float64, 2, Tuple{Index{Int64}, Index{Int64}}, NDTensors.Diag{Float64, Float64}}, Alabels::Tuple{Int64, Int64}, B::NDTensors.DenseTensor{Float64, 2, Tuple{Index{Int64}, Index{Int64}}, NDTensors.Dense{Float64, CUDA.CuArray{Float64, 1, CUDA.Mem.DeviceBuffer}}}, Blabels::Tuple{Int64, Int64})
    @ ITensorGPU ~/.julia/packages/ITensorGPU/CqqEK/src/tensor/cudiag.jl:156
  [6] contract!
    @ ~/.julia/packages/ITensorGPU/CqqEK/src/tensor/cudiag.jl:167 [inlined]
  [7] _contract!!
[...]

PR #1372, opened 2024-04-01
Modify delta to have Bool storage

ryanlevy:
# Description

Generically, having `delta` default to Float64 storage is causing problems with GPU (see #1143  and comment from @kmp5VT ) as well as converting previous codes to different types Float32/etc. One bandaid is to swap `delta` to have Bool storage so it can select the correct type when multiplying something later. However, there are some questionable uses of `delta` in the tests that now break, but I expect these won't be in many production codes.

However, this doesn't address the separate problem of matching an ITensor eltype, notably when GPU vs not GPU, so delta should probably be converted to something that only creates storage once a contraction is called

```julia
ITensor ord=2 (dim=3|id=99) (dim=3|id=712)
NDTensors.Diag{Float64, Float64}
```
</p></details>

```julia
julia> i,j = Index.([3,3]); D1 = δ(i,j)
ITensor ord=2 (dim=3|id=427) (dim=3|id=721)
NDTensors.Diag{Bool, Bool}
```
</p></details>

# How Has This Been Tested?

- [x] Tests in `test_diagitensor`

# Checklist:

### DIFF
--- a/test/base/test_diagitensor.jl
+++ b/test/base/test_diagitensor.jl
@@ -263,8 +263,8 @@ using Test
     end
 
     @testset "Add (Diag uniform + Diag uniform)" begin
-      D1 = δ(i, j, k)
-      D2 = δ(k, i, j)
+      D1 = δ(i, j, k) * 1.0 # hack, for Float64
+      D2 = δ(k, i, j) * 1.0
 
       D3 = D1 + D2
 
@@ -391,12 +391,12 @@ using Test
     @testset "delta constructor (order 2)" begin
       D = δ(i, j)
 
-      @test eltype(D) == Float64
+      @test eltype(D) == Bool
       for ii in 1:d, jj in 1:d
         if ii == jj
-          @test D[i => ii, j => jj] == 1.0
+          @test D[i => ii, j => jj] == 1
         else
-          @test D[i => ii, j => jj] == 0.0
+          @test D[i => ii, j => jj] == 0
         end
       end
     end
@@ -404,12 +404,12 @@ using Test
     @testset "delta constructor (order 3)" begin
       D = δ(i, j, k)
 
-      @test eltype(D) == Float64
+      @test eltype(D) == Bool
       for ii in 1:dim(i), jj in 1:dim(j), kk in 1:dim(k)
         if ii == jj == kk
-          @test D[i => ii, j => jj, k => kk] == 1.0
+          @test D[i => ii, j => jj, k => kk] == 1
         else
-          @test D[i => ii, j => jj, k => kk] == 0.0
+          @test D[i => ii, j => jj, k => kk] == 0
         end
       end
     end
@@ -417,7 +417,7 @@ using Test
     @testset "Set elements" begin
       D = δ(i, j, k)
 
-      @test eltype(D) == Float64
+      @test eltype(D) == Bool
 
       # Can't set elements of uniform diag tensor
       # TODO: should we make a function that converts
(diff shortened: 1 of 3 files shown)


test/base/test_diagitensor.jl
```diff
@@ -431,18 +431,18 @@ using Test
...
         else
-          @test T[i => ii, j => jj, k => kk] == 0.0
+          @test T[i => ii, j => jj, k => kk] == 0
         end
       end
     end
 
     @testset "Add (Diag uniform + Dense)" begin
-      D = δ(i, j, k)
+      D = δ(i, j, k) * 1.0 # hack, Float64
```
mtfishman, line 445: What goes wrong if you don't add this?

test/base/test_diagitensor.jl
```diff
@@ -508,16 +508,16 @@ using Test
     end
 
     @testset "Contraction (Diag uniform * Diag uniform, all contracted)" begin
-      D1 = δ(l, i, k, j)
-      D2 = δ(j, l, i, k)
+      D1 = δ(l, i, k, j) * 1.0 # hack, convert to Float64
+      D2 = δ(j, l, i, k) * 1.0
```
mtfishman, line 512: What goes wrong if you don't add these?

test/base/test_diagitensor.jl
ryanlevy: You get
```
Expression: D3 ≈ dense(D1) + dense(D2)
  InexactError: Bool(2)
```

test/base/test_diagitensor.jl
ryanlevy: similar to abiove
```
Test threw exception
  Expression: D2 * D1 ≈ dense(D1) * dense(D2)
  InexactError: Bool(3)
```

test/base/test_diagitensor.jl
kmp5VT: you could do 
```
for elt in [Float32, Float64]
  D3 \approx dense(D1), adapt(elt, D2)
end
```

mtfishman:
Thanks @ryanlevy. I'm hesitant to do this because it is definitely a band-aid and we are working on a full rewrite of NDTensors that will help fix these kinds of issues in a deeper way.

Can you share an example of where this was causing issues for you? A fix I have used for this kind of thing is to replace uses of `δ(i, j, k)` with `δ(eltype(...), i, j, k)` (where the element type is derived from some other tensors involved in the calculation), which we can either do inside the library or you can do in your own code. That isn't much different from how other constructors in Julia work, i.e. you shouldn't use `a + randn(2, 2)` if you want code to preserve the element type of `a`.

ryanlevy:
I can understand the hesitation. I would argue though that this helps catch strange uses of delta which shouldn't really be used, like in the hacked tests. Or we should just find a way to have a better gpu friendly `dense(delta(i))` function.

As for the use case, I'm in the midst of making ITensorInfiniteMPS "type safe", and all the delta functions in the code are causing trouble, for example
```julia
subspace_expansion.jl
15:  δʳ(n) = δ(dag(r[n]), prime(r[n]))
16:  δˢ(n) = δ(dag(s[n]), prime(s[n]))
17:  δˡ(n) = δ(l[n], dag(prime(l[n])))
107:  δʳ(n) = δ(dag(r[n]), prime(r[n]))
108:  δˢ(n) = δ(dag(s[n]), prime(s[n]))
109:  δˡ(n) = δ(l[n], dag(prime(l[n])))
175:  δʳ(n) = δ(dag(r[n]), prime(r[n]))
176:  δˢ(n) = δ(dag(s[n]), prime(s[n]))
177:  δˡ(n) = δ(l[n], dag(prime(l[n])))
```
So for the immediate time I plan do as you suggest, but this would help make that code cleaner

mtfishman:
I understand it is annoying to have to specify the element type, but since it is no more annoying than how other constructors in the Julia universe work I think we'll just have to live with this for now, and keep it in mind for our `NDTensors` rewrite.

I don't understand what you mean by:

> I would argue though that this helps catch strange uses of delta which shouldn't really be used, like in the hacked tests.

test/base/test_diagitensor.jl
mtfishman: That is related to an issue in our contract code that it doesn't always handle mixtures of element types gracefully, which is being improved in our `NDTensors` rewrite.

test/base/test_diagitensor.jl
mtfishman: Same comment as below.

mtfishman:
I don't think we're going to do this at this point, it is not worth causing any breakages when we are working on a full rewrite anyway that will address issues like this. At worst, users (or `ITensorInfiniteMPS.jl`) can write `delta(Bool, i, j)` or `delta(eltype(...), i, j)` instead of `delta(i, j)`.

PR #1373, opened 2024-04-02
[ITensors] Extension implementing `VectorInterface.jl` for ITensors

mtfishman:
As of v0.7, [KrylovKit.jl](https://github.com/Jutho/KrylovKit.jl?tab=readme-ov-file#v07) now uses [VectorInterface.jl](https://github.com/Jutho/VectorInterface.jl) as the interface for enabling new types to be used as vectors inside of Krylov solvers. This adds a package extension implementing that interface.

TODO:
- [x] Add tests.
- [x] Add [PackageExtensionCompat.jl](https://github.com/cjdoris/PackageExtensionCompat.jl) for backwards compatibility to Julia version below v1.9.
- [x] Optimize `VectorInterface.add!` by fusing more operations.


mtfishman:
This will supersede #1354.

mtfishman:
@emstoudenmire this should serve as a good reference point for making an `ITensorsObserversExt` extension.

emstoudenmire:
Thanks. Good to learn about this vector interface also; I was reading the specification for it.

mtfishman:
@lkdvos @Jutho hopefully I did that right! At first I got confused about the convention of `add(w, v, α, β)` and thought it was `w * α + v * β` rather than `w * β + v * α`, the definition makes sense I just didn't read carefully enough and translated argument ordering to operation ordering.

Anyway, looks like a nice interface, definitely useful for treating more complicated data structures as vectors.

mtfishman:
[test ITensorMPS][test ITensorGaussianMPS]

mtfishman:
Maybe something that would be useful would be to make an automated set of tests that can be generated for the interface, say assuming some minimal set of simple out-of-place `Base` functions are implemented for the object. I realize that may not work for deeply nested data structures that the package is targeting, however.

lkdvos:
Hi Matthew, 

I had a quick glance and it definitely looks good!
I think in principle it should only be necessary to define 4 argument add, these should by default be filled in. We added `One` and `Zero`, which should enable automatic constant folding by the compiler for optimised implementations in these cases.

I can have a better look tomorrow if you want ☺️

mtfishman:
> Hi Matthew,
> 
> I had a quick glance and it definitely looks good! I think in principle it should only be necessary to define 4 argument add, these should by default be filled in. We added `One` and `Zero`, which should enable automatic constant folding by the compiler for optimised implementations in these cases.
> 
> I can have a better look tomorrow if you want ☺️

I only defined that one at first but then was concerned that our ITensor broadcasting wasn't be smart about those cases, thanks for the heads up about that!

mtfishman:
I guess that would mean that if we don't specify `add(w, v)`, it will call `add(w, v, One(), One())`, in which case our broadcasting will have to handle `One()` in a smart way. I worry that it isn't right now, so I think I'll leave it as is for now. Our current ITensor broadcasting is not written well and is very narrow and fragile, it is being rewritten as part of our rewrite of `NDTensors` in which case we should be able to handle things like `One()` in a smarter way.

mtfishman:
[test ITensorMPS][test ITensorGaussianMPS]

mtfishman:
[test ITensorMPS][test ITensorGaussianMPS]

mtfishman:
[test ITensorMPS][test ITensorGaussianMPS]

PR #1376, opened 2024-04-03
[ITensorMPS] Observers package extension

emstoudenmire:
This PR introduces a simple package extension that defines `ITensorMPS.update_observer!` when the `Observers` package is loaded, and a simple test for it. It has been locally tested against the `ITensorTDVP` package also.


src/ITensors.jl
```diff
@@ -225,4 +225,14 @@ end
...
+
+
+#
+# See section on "Transition from normal dependency to extension"
+# in https://pkgdocs.julialang.org/v1/creating-packages
+#
+if !isdefined(Base, :get_extension)
+  include("../ext/ITensorsObserversExt.jl")
+end
+
```
emstoudenmire, line 237: I question is the above code is really needed. When reading up on package extensions, I saw the section "Transition from normal dependency to extension" and thought since `Observers` is currently a normal dependency that maybe we need this?

emstoudenmire:
Also I could add the `PackageExtensionCompat` dependency, but since it is being added in #1373 I hesitated to do that.

mtfishman:
Thanks. We should use `PackageExtensionCompat` instead of the `if !isdefined(Base, :get_extension) ...` pattern. I'll be merging #1373 soon but the order of operations doesn't matter there.

However, before making any more changes to `ITensorsMPS` I think we should get downstream tests of `ITensorTDVP` work (https://github.com/ITensor/ITensors.jl/pull/1370#issuecomment-2028821695).

src/ITensors.jl
mtfishman: I don't think we need this. This would make `Observers` a conditional dependency in Julia v1.9 and above and a regular dependency in versions below v1.9. By using `PackageExtensionCompat` we can make it a conditional dependency for any version of Julia.

mtfishman:
Not sure what's going on there, the configuration looks good to me.

src/ITensors.jl
emstoudenmire: Removed

PR #1379, opened 2024-04-04
[ITensors] Remove `Observers` dependency

emstoudenmire:
This PR introduces a simple package extension that defines ITensorMPS.update_observer! when the Observers package is loaded, and a simple test for it. It has been locally tested against the ITensorTDVP package also.

(Another attempt, making incremental changes to see what is breaking the tests. All tests pass locally but not in the CI.)


mtfishman:
@emstoudenmire we could just merge this and then work on the extension in a followup, we may as well at least remove the dependency.

emstoudenmire:
That sounds good. In the exten@ PR my plan is to make small changes until I get that Pkg bug again, to narrow down why. It's probably to do with Project.toml (maybe even the test/ did ones?) and maybe weak deps. We will see.

PR #1381, opened 2024-04-04
[ITensors] `Observers.jl` package extension

emstoudenmire:
This PR is work in progress toward adding `Observers` support as a package extension. I am going to gradually submit commits and re-test to pinpoint (or avoid) the testing issue with previous PRs.


Project.toml
```diff
@@ -53,6 +54,7 @@ KrylovKit = "0.4.2, 0.5, 0.6, 0.7"
 LinearAlgebra = "1.6"
 LinearMaps = "3"
 NDTensors = "0.2.30"
+Observers = "0.2"
```
mtfishman, line 57: Perhaps you could try removing this `compat` entry to start out with, maybe that is confusing Julia into thinking it is a dependency instead of a weak dependency.

emstoudenmire: Good idea, I'll try

emstoudenmire: Unfortunately that didn't do it. Maybe I should remove it from extras also just to see.

mtfishman: Right, I remember that the only reason I added `VectorInterface` to `[extras]` was so that I could add it to `[compat]`, so that corroborates that might be the primary culprit.

mtfishman: Guess not.

mtfishman:
@emstoudenmire in #1365 @kmp5VT is trying out running the tests in a temporary project with `Pkg.activate(temp=true)`, I wonder if that will help here since my guess is that it is using an old poorly configured `Manifest.toml` instead of building one from scratch.

mtfishman:
@emstoudenmire seems like the updated CI workflow fixed the issue.

emstoudenmire:
That's a relief! Thanks to you and Karl for figuring out that fix.

emstoudenmire:
Still some more changes to make on this PR, but I think only some small ones left.

emstoudenmire:
Unles you prefer to just make this PR about removing `Observers`. Then I could submit a new one for the Observers extension. Otherwise I'll just change the PR name and add those commits here.

mtfishman:
Let's make the package extension in this PR.

emstoudenmire:
[test ITensorTDVP]

emstoudenmire:
[test ITensorTDVP]

emstoudenmire:
This is running the ITensorTDVP workflow on main, as I guessed it would (though the tests ran and passed, for what it's worth, though I think not against this PR). So I'm not sure if the new one I included in this PR will work better or work correctly. If you have a good suggestion for how to repeatedly test a workflow file in a PR I'd be curious. Or maybe another way is to push

.github/workflows/comment_trigger_test_itensortdvp.yml
```diff
@@ -51,6 +51,8 @@ jobs:
         run: |
           using Pkg;
           Pkg.add("ITensorTDVP");
+          Pkg.activate(temp=true);
```
mtfishman, line 54: I think you should activate the project before adding any packages (i.e. before the line `Pkg.add("ITensorTDVP");`).

.github/workflows/test_ndtensors.yml
```diff
@@ -38,3 +38,4 @@ jobs:
           Pkg.develop(path="./NDTensors");
           Pkg.develop(path=".");
           Pkg.test("NDTensors");
+
```
mtfishman, line 41: Seems like something funny going on with this extra line?

.github/workflows/comment_trigger_test_itensortdvp.yml
mtfishman, line 1: Let's just turn this into a regular workflow, not a comment triggered workflow, so it could basically just be based on https://github.com/ITensor/ITensors.jl/blob/v0.3.67/.github/workflows/test_itensors_base_ubuntu.yml.

emstoudenmire:
I'm pretty sure the new ITensorTDVP test is correctly running the code from this PR. There is the line:
```  [9136182c] ITensors v0.3.67 `~/work/ITensors.jl/ITensors.jl` ```
showing up in the output (under "Install Julia Dependencies and Run Tests") from Github Actions. Also I based it closely on the ITensors Base tests workflow.

If we want to be totally sure, though, I could add a printing statement to e.g. the `tdvp` function inside `ITensorMPS`. Then we should see the tests pass but with that printing statement running a bunch of times and showing up in the test output. So then we could know for sure that it's running using the code from this PR.

File: .github/workflows/test_itensortdvp.yml [hunk 4]
```diff
@@ -0,0 +1,49 @@
+name: Run ITensorTDVP tests from comment trigger
+
+# https://dev.to/zirkelc/trigger-github-workflow-for-comment-on-pull-request-45l2
```
mtfishman, line 3:
```suggestion
name: Run ITensorTDVP tests (Unbuntu)
```

.github/workflows/test_itensortdvp.yml
mtfishman, line 1: Let's change the file name to `test_itensortdvp_ubuntu.yml` to be consistent with other workflows that only test on Ubuntu.

emstoudenmire: Good idea, done

emstoudenmire:
I tried my idea about the printing statement, and re-ran the ITensorTDVP test. Unfortunately, though, I did not see the printing output when the tests ran. (I had put a simple println statement at the top of the `itensortdvp_tdvp` function, a version of it that I'm pretty sure gets covered by the tests.

.github/workflows/test_itensortdvp.yml
```diff
@@ -0,0 +1,47 @@
...
+          arch: ${{ matrix.arch }}
+      - name: Install Julia dependencies and run tests
+        shell: julia {0}
+        run: |
+          using Pkg;
+          Pkg.activate(temp=true);
+          Pkg.develop(path="./NDTensors");
+          Pkg.develop(path=".");
+          Pkg.add("ITensorTDVP");
+          Pkg.test("ITensorTDVP");
```
mtfishman, line 44: We should also wrap this part in a `try ... catch` statement to catch the case where we knowingly break `ITensorTDVP` through a breaking release, in which case we would bump the minor version of `ITensors.jl` and Julia's package manager wouldn't allow us to add `ITensorTDVP`. See https://github.com/JuliaDiff/ChainRulesCore.jl/blob/b5df374885dbcaacda0f22175064eaa0319cee00/.github/workflows/IntegrationTest.yml#L42-L54.

.github/workflows/test_itensortdvp.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: julia-actions/julia-uploadcodecov@latest
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
mtfishman, line 47: I think we should remove this part related to code coverage, it seems to be causing issues in the CI run.

mtfishman:
> I tried my idea about the printing statement, and re-ran the ITensorTDVP test. Unfortunately, though, I did not see the printing output when the tests ran. (I had put a simple println statement at the top of the `itensortdvp_tdvp` function, a version of it that I'm pretty sure gets covered by the tests.

The ITensorTDVP testing workflow is running the tests of the latest registered version of ITensorTDVP.jl, which has it's own implementation of `tdvp` which isn't calling `ITensorMPS.itensortdvp_tdvp`, since https://github.com/ITensor/ITensorTDVP.jl/pull/67 isn't merged and registered yet. So I think the current behavior is expected.

We could put a print statement in some other part of the code, say inside one of the `ProjMPO` functions that is being used by the `ITensorTDVP.tdvp` function.

emstoudenmire:
That makes sense, thanks. I'll move the printing statement.

mtfishman:
Putting it in the `makeR!` of `ProjMPOApply` also doesn't work since `ITensorTDVP.jl` currently has its own `ProjMPOApply` that it is using: https://github.com/ITensor/ITensorTDVP.jl/blob/v0.2.1/src/projmpo_apply.jl.

.github/workflows/test_itensortdvp.yml
emstoudenmire: Very interesting. Please check that I got the idea right.

emstoudenmire:
I temporarily moved the printing statement to the `norm` function in `abstractmps.jl` and finally saw it in the TDVP testing output. So we can be sure that it's testing from the current PR now. I've since removed that printing statement altogether and now re-running all the tests.

I believe this is ready to merge once tests pass.

PR #1382, opened 2024-04-12
[ITensors] `HDF5.jl` package extension

emstoudenmire:
Moved HDF5 support into a package extension. Updated docs accordingly.

Also fixed a couple of unrelated build issues in the docs and an ITensorMPS module import/overload issue.


emstoudenmire:
I'm seeing some warnings in the tests about loading the various package extensions, e.g.
```
 ITensors → ITensorsVectorInterfaceExt
│  ┌ Warning: Module ITensorsVectorInterfaceExt with build ID ffffffff-ffff-ffff-0000-00274c0366aa is missing from the cache.
...
```
I found this open `Pkg.jl` issue which shows that others have been seeing it, even in Julia 1.11. 
https://github.com/JuliaLang/Pkg.jl/issues/3557
It was supposed to be fixed in this PR https://github.com/JuliaLang/julia/pull/52841 but some are still reporting it.

docs/src/ITensorType.md
```diff
@@ -26,7 +26,7 @@ itensor(::Array{<:Number}, ::ITensors.Indices)
 ## QN BlockSparse Constructors
 
 ```@docs
-ITensor(::Type{<:Number}, ::QN, ::ITensors.Indices)
+ITensor(::Type{<:Number}, ::QN, ::ITensors.QNIndices)
```
emstoudenmire, line 29: The doc string in the source is associated with this version of the function. Without this change there is an error because the original version has no doc string.

emstoudenmire:
(Also sneaking in a small orthogonal update: added `directsum` to the docs as we discussed.)

ext/ITensorsHDF5Ext/itensor.jl
```diff
@@ -0,0 +1,36 @@
...
+
+function HDF5.read(
+  parent::Union{HDF5.File,HDF5.Group}, name::AbstractString, ::Type{ITensor}
+)
+  g = open_group(parent, name)
+  if read(attributes(g)["type"]) != "ITensor"
+    error("HDF5 group or file does not contain ITensor data")
+  end
+  # TODO: use Vector{Index} here?
+  inds = read(g, "inds", IndexSet)
```
mtfishman, line 21: Yeah, we should use `Vector{<:Index}` here, `IndexSet` is meant to be deprecated.

test/base/test_readwrite.jl
```diff
@@ -1,4 +1,6 @@
-using ITensors, HDF5, Test
+using HDF5
+using ITensors
+using Test
```
mtfishman, line 3: Would you mind updating the style of this test, i.e. wrapping in a module and changing to explicit using statements?

src/ITensorMPS/abstractmps.jl
```diff
@@ -1,5 +1,24 @@
...
+  swapprime,
+  swapprime!,
+  replaceprime,
+  replaceprime!,
+  addtags,
+  addtags!,
+  removetags,
+  removetags!,
+  settags,
+  settags!
```
mtfishman, line 21: Why do these imports need to be added?

docs/src/ITensorType.md
mtfishman: Sounds good, thanks.

src/ITensorMPS/abstractmps.jl
mtfishman: I see, you explained in the first comment. I think we should switch to appending `ITensors.` in the source code instead of adding these `import` statements, and handle `dag!` separately (without adding `dag!` to `ITensors`, if anything I would like to remove all of the in-place versions of these functions from `ITensors` and maybe also from `ITensorMPS`).

mtfishman:
> I'm seeing some warnings in the tests about loading the various package extensions, e.g.
> 
> ```
>  ITensors → ITensorsVectorInterfaceExt
> │  ┌ Warning: Module ITensorsVectorInterfaceExt with build ID ffffffff-ffff-ffff-0000-00274c0366aa is missing from the cache.
> ...
> ```
> 
> I found this open `Pkg.jl` issue which shows that others have been seeing it, even in Julia 1.11. [JuliaLang/Pkg.jl#3557](https://github.com/JuliaLang/Pkg.jl/issues/3557) It was supposed to be fixed in this PR [JuliaLang/julia#52841](https://github.com/JuliaLang/julia/pull/52841) but some are still reporting it.

I see, thanks for investigating. That issue seems to be related to having a system image for a package already built which triggers strange behaviors when loading package extensions relying on those pre-built packages, if I'm following the discussions in those links. That makes me think it is somewhat related to the issue in #1381, i.e. that the tests aren't being run from a clean slate but rather reusing something from a previous CI run (in this case pre-built system images, in #1381 a badly-configured Manifest.toml).

For both this issue and the issue in #1381, I wonder if running the CI using a temporary environment with `Pkg.activate(temp=true)` as @kmp5VT is trying out in #1365 might help.

mtfishman:
We should bump the minor version of `ITensors.jl` to `v0.4` since technically this is a breaking change.

test/base/test_readwrite.jl
emstoudenmire: No problem

ext/ITensorsHDF5Ext/itensor.jl
emstoudenmire: Fixed

mtfishman:
Can't the duplicated docstring just be refactored into a function that generates the docstring that can be used by both versions?

mtfishman:
I do not think we should define `dag!`, etc. in ITensors.jl.

mtfishman:
Another suggestion would be to rewrite the for loop like this:
```julia
for (fname, fname!) in (
  (:ITensors.dag, :dag!),
  (:ITensors.prime, :ITensors.prime!),
...
)
```

emstoudenmire:
The docstring funciton is a good idea – I had not thought of that.

So then with that improvement, a pretty simple fix would just be to duplicate the code in `abstractmps.jl` which is some but not a lot of duplication (maybe 10-ish lines of mostly one-liner functions). And then anyway some of that code will get deprecated later.

mtfishman:
I think I prefer this suggestion better:

> ```julia
> for (fname, fname!) in (
>   (:(ITensors.dag), :(dag!)),
>   (:(ITensors.prime), :(ITensors.prime!)),
> ...
> )
> ```

emstoudenmire:
Yes that's definitely the best idea. Very clean. I'm going to modify it to have a third entry for the docstring and otherwise will use that pattern.

ext/ITensorsHDF5Ext/ITensorsHDF5Ext.jl
```diff
@@ -0,0 +1,14 @@
+module ITensorsHDF5Ext
+
+using HDF5: HDF5, attributes, create_group, open_group, read, write
```
mtfishman, line 3: I realize it is pedantic but I would prefer if this was moved to each file to keep the style consistent with other parts of the code.

mtfishman:
I believe to get tests passing we will need to append `0.4` to the compat entries of the downstream packages like `ITensorGaussianMPS`, etc.

emstoudenmire:
Ah thanks, was just wondering the best course of action. 

By the way, I wasn't able to reproduce the `sim!` issue again locally. So maybe it was a false alarm or one-off local issue that had to do with some temporary state my REPL was in. I'll see if the tests pass without it because they are passing again locally without that change.

emstoudenmire:
This one should be ready to go, unless there are possibly other tests I didn't think to run?

mtfishman:
Could you also bump the patch versions of `ITensorGaussianMPS` and `ITensorVisualizationBase`?

emstoudenmire:
Sure thing, done. I put [no ci] at the end of the commit for that, but I can also rerun the tests if that's best.

mtfishman:
Could you also add a NEWS item to the README about the 0.4 release?

emstoudenmire:
Sure thing, I added the NEWS entry

mtfishman:
> Sure thing, I added the NEWS entry

Sorry, I meant add an entry about it here: https://github.com/ITensor/ITensors.jl/blob/main/docs/src/index.md which will show up on both the README.md and in the documentation.

I haven't been updating the `NEWS.md` file, I found it was too much work to keep it updated and doesn't add much beyond the commit history or the documentation. Probably we should just get rid of it to avoid confusion for developers and users.

emstoudenmire:
I updated the docs index file – please take a look. We could add other major changes since 0.3 such as e.g. replacing the original OpSum with Ops.OpSum and other big changes like that.

I went ahead and removed NEWS.md. Let me know if you want me to put it back.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "ITensors"
 uuid = "9136182c-28ba-11e9-034c-db9fb085ebd5"
 authors = ["Matthew Fishman <mfishman@flatironinstitute.org>", "Miles Stoudenmire <mstoudenmire@flatironinstitute.org>"]
-version = "0.3.68"
+version = "0.4.1"
```
mtfishman, line 4:
```suggestion
version = "0.4.0"
```
?

Project.toml
```diff
@@ -34,11 +33,27 @@ Zeros = "bd1ec220-6eb4-527a-9b49-e79c3db6233b"
 ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [weakdeps]
+<<<<<<< HEAD
+<<<<<<< HEAD
+HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
+VectorInterface = "409d34a3-91d5-4945-b6ec-7529ddf182d8"
```
mtfishman, line 39: Seems like the merge got messed up and `Observers` got dropped.

docs/src/index.md
```diff
@@ -29,6 +29,8 @@ Development of ITensor is supported by the Flatiron Institute, a division of the
 
 ## News
 
+- April 16, 2024: ITensors.jl v0.4 has been released. This version removes HDF5 as a dependency (a breaking change) and moves the HDF5 read and write functions for ITensor, MPS, MPO, and other associated types into a package extension. To enable ITensor HDF5 features, install the HDF5 package and put `using HDF5` before `using ITensors` in your code. Other recent changes include support for multiple GPU backends using package extensions and support for the Observers.jl package (also through a package extension).
```
mtfishman, line 32: The ordering of calling `using HDF5` and `using ITensors` shouldn't matter.

Also, is `Observers` actually supported in `ITensors.dmrg`? I don't think it is right now, that sounds misleading.

src/ITensorMPS/abstractmps.jl
mtfishman: This issue doesn't appear to be resolved.

Project.toml
emstoudenmire: Oof, that is bad. I will fix.

src/ITensorMPS/abstractmps.jl
emstoudenmire: I've removed it now.

docs/src/index.md
emstoudenmire: Please see if the update text is more accurate

Project.toml
```diff
@@ -35,9 +34,11 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [weakdeps]
 Observers = "338f10d5-c7f1-4033-a7d1-f9dec39bcaa0"
+HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
```
mtfishman, line 37:
```suggestion
HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
Observers = "338f10d5-c7f1-4033-a7d1-f9dec39bcaa0"
```
Alphabetized.

PR #1383, opened 2024-04-12
[NDTensors] Update to latest `Metal.jl`, fix `ITensorGPU` CI

kmp5VT:
# Description

In this PR I am making some modifications to deal with changes from the bump in Metals version. First, now metal has a `resize!` function so the Blocksparse DMRG code now is functional with that backend so remove the `is_broken` function. 

Next there seems to be an issue with copy and subarrays so I opened an issue with Metal. In the meantime I marked the test as broken for metal in the `Expose` library.  I am currently running the tests to determine if there is anything else that needs to be fixed with metal and if there is I will make those changes here too.

# Checklist:

- [ ] Metal DMRG tests pass
- [ ] Metal CI no longer is failing


NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -13,7 +13,10 @@ if "rocm" in ARGS || "all" in ARGS
 end
 if "metal" in ARGS || "all" in ARGS
   ## Warning Metal does not work in Julia versions below 1.8
-  Pkg.add("Metal")
+  ## Essentially compat versioning here because we can't put Metal
+  ## in the Project.toml because it causes issues during testing in 
+  ## Julia versions below 1.8
+  Pkg.add(; name="Metal", version="1.1.0")
```
kmp5VT, line 19: @mtfishman since we don't have `Metal` in a the `Project.toml` of `NDTensors/test` this is a way to version control ourselves. I found that Jenkins was failing for metal because it was installing the wrong version. Adding this fixed it but the issue is that one will have to manually change this.
Another option would be to force an update on `Metal` and its dependencies with code like this 
```
Pkg.add("Metal")
Pkg.update("Metal"; preserve=PRESERVE_DIRECT)
```
Let me know what you think. Thanks!

mtfishman: Do you know why it isn't installing the latest version in the first place? Also it seems fine to just call `Pkg.update()` if that works.

NDTensors/src/lib/Expose/test/runtests.jl
```diff
@@ -184,7 +184,11 @@ using NDTensors.GPUArraysCoreExtensions: cpu
   copyto!(expose(y), expose(x))
   @allowscalar begin
     @test y == x
-    @test copy(x) == x
+    if dev == NDTensors.mtl
+      @test_broken copy(x) == x
```
mtfishman, line 188: Does `copy(expose(x))` work? Or is that not needed since we don't `copy` this particular wrapped type in the code anyway?

kmp5VT: Yes the expose version of copy does work with metal so I switched it over to use that one. Thanks!

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: I think it has something to do with this `preserve` variable. If I do something like this you can see Metal doesn't update properly.
```
Pkg.activate(temp=true)
Pkg.add(;name="Metal", version="1.0.0")
Pkg.update("Metal")
Updating registry at `~/.julia/registries/General.toml`
  No Changes to `/tmp/jl_XzMvJ2/Project.toml`
  No Changes to `/tmp/jl_XzMvJ2/Manifest.toml`
]st
Status `/tmp/jl_XzMvJ2/Project.toml`
⌃ [dde4c033] Metal v1.0.0
Info Packages marked with ⌃ have new versions available and may be upgradable.
```
But I don't really know why Pkg isn't grabbing the latest version of metal in the temporary project. Nowhere in our code do we explicitly compat Metal to 1.0.0 and the temporary package should always be clean.

NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -1,5 +1,5 @@
 using NDTensors: NDTensors
-using Pkg: Pkg
+using Pkg: Pkg, PRESERVE_DIRECT
```
mtfishman, line 2: Seems like `PRESERVE_DIRECT` isn't needed anymore?

jenkins/Jenkinsfile
```diff
@@ -118,6 +118,9 @@ pipeline {
             sh '''
               juliaup update
             '''
+            sh '''
+              julia -e 'using Pkg; Pkg.activate("./NDTensors/test/"); Pkg.add("Metal")';
```
mtfishman, line 122: This fix seems strange to me, since the tests already execute `Pkg.add("Metal")` when `Pkg.test("NDTensors"; test_args=["metal"])` is called.

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: Thanks I removed this!

kmp5VT:
@mtfishman regarding this issue. I tried deleting everything in my `.julia` folder. Next I ran this on my mac
```
$ julia
julia> using Pkg; Pkg.activate(temp = true);
julia> Pkg.add(;name="Metal", version="1.0.0")
julia> exit()
$julia
julia> using Pkg; Pkg.activate(temp=true)
julia> Pkg.dev("/path/to/NDTensors"); Pkg.dev("/path/to/ITensors");
julia> Pkg.test("NDTensors"; test_args=["metal"])
```
and it looks like metal 1.1.0 is added like it should be.
```
Running /Users/kpierce/Workspace/julia_deps_41824/dev/ITensors/NDTensors/test/test_blocksparse.jl
    Updating registry at `~/.julia/registries/General.toml`
   Resolving package versions...
   Installed LLVMDowngrader_jll ─ v0.1.0+1
   Installed ObjectiveC ───────── v2.1.1
   Installed CodecBzip2 ───────── v0.8.2
   Installed TranscodingStreams ─ v0.10.7
   Installed Metal ────────────── v1.1.0
  Downloaded artifact: LLVMDowngrader
    Updating `/private/tmp/jl_OeKJm5/Project.toml`
  [dde4c033] + Metal v1.1.0
    Updating `/private/tmp/jl_OeKJm5/Manifest.toml`
```
So most likely the problem we experienced is something related to Jenkins.  I went and added a `[compat]` for metal in `NDTensors/test/Project.toml` just in case that does work in `Pkg.test`.  My thought is that we might just have to manually update `Metal` on `Jenkins` when there is a version bump but its hard to determine until there is another version bump.
Another thing I could try is on Jenkins before Julia is started I could  `cd ~/.julia/packages/` and `rm Metal/`?

NDTensors/test/Project.toml
```diff
@@ -25,3 +25,6 @@ Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 [weakdeps]
 AMDGPU = "21141c5a-9bdb-4563-92ae-f87d6854732e"
 Metal = "dde4c033-4e86-420c-a63e-0dd931031962"
+
+[compat]
+Metal = "^1.1.0"
```
mtfishman, line 30:
```suggestion
Metal = "1.1.0"
```
`^` is implied if left off, see https://pkgdocs.julialang.org/v1/compatibility/#Caret-specifiers.

mtfishman:
@kmp5VT I think you have to add `Metal` to `[extras]` in `test/Project.toml` in older versions of Julia in order to add a compat entry for it.

NDTensors/test/Project.toml
kmp5VT: Thanks I will fix this and I am verifying that moving Metal to `[extras]` works properly. (I will also move AMDGPU to `[extras]` if it does)

mtfishman:
@kmp5VT can you change the `ITensorGPU` `julia` compat entry to `1.6 - 1.9`?

mtfishman:
Also can you remove `HDF5` as a dependency of `ITensorGPU`?

mtfishman:
Looks good, thanks for sorting that out.

PR #1390, opened 2024-04-17
[NDTensors] HDF5 Extension

emstoudenmire:
This PR moves all NDTensors HDF5 support into a package extension and increases the minor version, from 0.3.0 to 0.4.0.


emstoudenmire:
Initial test errors are a compat issue.

NDTensors/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "NDTensors"
 uuid = "23ae76d9-e61a-49c4-8f12-3f1a16adf9cf"
 authors = ["Matthew Fishman <mfishman@flatironinstitute.org>"]
-version = "0.3.0"
+version = "0.4.0"
```
mtfishman, line 4:
```suggestion
version = "0.3.1"
```
I think a minor version bump is ok, the HDF5 functionality of NDTensors is just serving ITensors anyway.

Project.toml
```diff
@@ -55,7 +55,7 @@ IsApprox = "0.1"
 KrylovKit = "0.4.2, 0.5, 0.6, 0.7"
 LinearAlgebra = "1.6"
 LinearMaps = "3"
-NDTensors = "0.3.0"
+NDTensors = "0.3.0, 0.4"
```
mtfishman, line 58:
```suggestion
NDTensors = "0.3.0"
```
if we just bump the patch version.

NDTensors/Project.toml
emstoudenmire: I didn't think about it that way. Right – no one should be depending on that part of NDTensors directly.

mtfishman:
Thanks for the quick fix.

PR #1395, opened 2024-04-18
[NDTensors] cuTENSOR extension

kmp5VT:
# Description

The goal of this PR is to create a cuTENSOR extension to the NDTensors library. In this extension, I will make a function which converts (Dense) Tensors from NDTensors into cuTENSORs and I will create an overload of the NDTensors `contract` function that calls the `cuTENSOR` backend contract. As a reach goal, in the first effort for the `BlockSparse` code I will convert the BS tensors into dense tensors, call contract, construct the output blocksparse tensor and transfer only the non-zero blocks back into a blocksparse tensor.  This functionality will be later more robustly solved using efforts from the NVIDIA team.

# Checklist:

- [x] It is possible to convert a NDTensors.Tensor into a cuTENSOR
- [x] It is possible to call cuTENSOR based contraction code 
- [x] The result from cutensor contract is equivalent to the NDTensors contract
- [x] Create unittests for cuTENSOR extension.


NDTensors/Project.toml
```diff
@@ -32,20 +32,22 @@ TupleTools = "9d95972d-f1c8-5527-a6e0-b4b365fa01f6"
...
 HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 Metal = "dde4c033-4e86-420c-a63e-0dd931031962"
 Octavian = "6fd5a793-0b7e-452c-907f-f8bfe9c57db4"
 TBLIS = "48530278-0828-4a49-9772-0f3830dfa1e9"
-AMDGPU = "21141c5a-9bdb-4563-92ae-f87d6854732e"
 
 [extensions]
+NDTensorsAMDGPUExt = "AMDGPU"
 NDTensorsCUDAExt = "CUDA"
+NDTensorscuTENSORExt = "cuTENSOr"
```
mtfishman, line 46:
```suggestion
NDTensorscuTENSORExt = "cuTENSOR"
```

kmp5VT: Thanks!

NDTensors/src/tensoroperations/generic_tensor_operations.jl
```diff
@@ -83,12 +83,15 @@ function can_contract(tensor1::Tensor, tensor2::Tensor)
...
 
+using NDTensors.Expose: Exposed, expose, unexpose
 # Version where output labels aren't supplied
 @traitfn function contract(
   tensor1::TensorT1, labels_tensor1, tensor2::TensorT2, labels_tensor2
 ) where {TensorT1<:Tensor,TensorT2<:Tensor;CanContract{TensorT1,TensorT2}}
   labelsoutput_tensor = contract_labels(labels_tensor1, labels_tensor2)
-  return contract(tensor1, labels_tensor1, tensor2, labels_tensor2, labelsoutput_tensor)
+  return contract(
+    expose(tensor1), labels_tensor1, expose(tensor2), labels_tensor2, labelsoutput_tensor
```
kmp5VT, line 93: @mtfishman so here I am using expose to capture when the deep `datatype` is a cuarray and cutensor is loaded. When this happens the code calls cutensor based contract backend and skips the rest of the canonical NDTensors logic.

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -0,0 +1,41 @@
...
+  ## convert the ITensors into CuTensors
+  cutensorA = CuTensor(dA, [i for i in labelstensor1])
+  cutensorB = CuTensor(dB, [i for i in labelstensor2])
+
+  ## contract the CuTensors
+  cutensorC = cutensorA * cutensorB
+
+  ## TODO this is a first draft to this idea to see if the 
+  ## conversion works 
+  output_tensor = NDTensors.contraction_output(
```
kmp5VT, line 26: This isn't probably the best way to handle this especially since the next line I replace the storage/data with the data from the cutensor. But as a first draft, this does currently work and give numerically the same answers (from my small tests using dense)

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -0,0 +1,41 @@
...
+  labelstensor2,
+  labelsoutput_tensor,
+)
+  tensor1 = unexpose(Etensor1)
+  tensor2 = unexpose(Etensor2)
+  ## reshape the flat data from ITensors into the correct dimensions
+  dA = reshape(data(tensor1), dims(tensor1))
+  dB = reshape(data(tensor2), dims(tensor2))
+  ## convert the ITensors into CuTensors
+  cutensorA = CuTensor(dA, [i for i in labelstensor1])
```
kmp5VT, line 18: It looks like `cuTENSOR` changed their backend and supports only `Vector{Int32}` and does support signed numbers so we don't need to convert our label systems. This conversion to vector could be written better but this would just a proof of concept for myself

mtfishman:
Looks like a good start, nice to see it is pretty simple.

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
kmp5VT: Converted to collect instead of this and it works find. Also am using the `array` function instead of reshape

kmp5VT:
There are a few errors in the ITensor/MPS testing because of unsupported mixed type contractions in cutensor. I have opened a bug report in CUDA.jl [here](https://github.com/JuliaGPU/CUDA.jl/issues/2349)

mtfishman:
Thanks, seems like we can promote the tensors to a common type ourselves to circumvent that.

NDTensors/src/blocksparse/contract.jl
```diff
@@ -1,12 +1,14 @@
 using .AlgorithmSelection: Algorithm, @Algorithm_str
 
 function contract(
-  tensor1::BlockSparseTensor,
+  tensor1::Exposed{<:AbstractArray,<:BlockSparseTensor},
```
kmp5VT, line 4: Modifying this code to look like this enables one to utilize cuTENSOR for `BlockSparseTensors`. However currently I don't believe there is an easy way to convert the blocksparsetensors into cutensors. So in the cuTENSOR code I have a dispatched function which just uses CUDA's `mul!` function. But I am working on a solution

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -0,0 +1,93 @@
...
+  tensor1 = unexpose(Etensor1)
+  tensor2 = unexpose(Etensor2)
+
+  return cutensor_contract(
+    tensor1, labelstensor1, tensor2, labelstensor2, labelsoutput_tensor
+  )
+end
+
+## working to fix blocksparse implementation. 
+function NDTensors.contract(
```
kmp5VT, line 31: This function is necessary so no error is thrown when trying to contract blocksparse tensors with cutensor enabled.  It does not however use cuTENSOR based contraction because there is no easy way to convert `BlockSparseTensors` to `CuTensors` and simply uses the previous `CUDA.mul!` function.  I am currently working on a solution for this.

mtfishman:
@kmp5VT I think we should only send tensors with `Dense` storage wrapping `CuArray` data to the `cuTENSOR` backend, for example:
```julia
function NDTensors.contract(
  Etensor1::Exposed{<:CuArray,<:DenseTensors},
  labelstensor1,
  Etensor2::Exposed{<:CuArray,<:DenseTensor},
  labelstensor2,
  labelsoutput_tensor,
)
  # ...
end
```

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
mtfishman: I think we should only overload `contract` for `DenseTensor` in the `cuTENSOR` extension. `BlockSparseTensor`s will get contracted as usually, but when it goes to perform the contractions over blocks, if the blocks have `Dense` storage they will get sent to `cuTENSOR`, but that logic will work automatically and doesn't have to be defined explicitly in the package extension.

mtfishman:
For some more context, here is where the dense blocks are contracted when two tensors with `BlockSparse` storage are contracted: https://github.com/ITensor/ITensors.jl/blob/v0.4.0/NDTensors/src/blocksparse/contract_generic.jl#L141-L150. `R[blockR]`, `tensor1[blocktensor1]`, and `tensor2[blocktensor2]` are blocks of the block sparse tensor, which are tensors with `Dense` storage.

So ideally when block sparse contraction occurs, if the `cuTENSOR` backend is enabled those dense block contractions will use dense contraction code defined in this new `cuTENSOR` extension. For that to happen, I think we should overload this contract signature:
```julia
function NDTensors.contract!(
  exposed_tensor_dest::Exposed{<:CuArray,<:DenseTensor},
  tensor_dest_labels,
  exposed_tensor1::Exposed{<:CuArray,<:DenseTensor},
  tensor1_labels,
  exposed_tensor2::Exposed{<:CuArray,<:DenseTensor},
  tensor2_labels,
  labelsoutput_tensor,
  α::Number,
  β::Number,
)
  # ...
end
```
in the package extension as opposed to the out-of-place `NDTensors.contract` function.

NDTensors/src/blocksparse/contract.jl
kmp5VT: I am removing this and moving the CuTensor portion to the inplace `contract!` function

NDTensors/src/tensoroperations/generic_tensor_operations.jl
kmp5VT: I am reverting this and removing the function below and moving the `CuTensor` portion to `contract!`

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
kmp5VT: I am closing this comment because I am limiting the code to overload `DenseTensor` and moving it down to the `contract!` level so the code will work with blockwise contractions.

kmp5VT:
An issue with adding cuTENSOR to ITensors that I just realized again is that cuTENSOR has a compat restriction on TensorOperations to version 0.7.1. Currently TensorOperations is on version 4.1.1 . This compat restriction causes some of our tests to fail in the TensorAlgebra module.  We could do what we are doing with Metal and AMDGPU and do `Pkg.add("cuTENSOR")` when we specifically test that package and have a flag in `TensorAlgebra` module to disable the tests when `cuTENSOR \in ARGS`

mtfishman:
You say "cuTENSOR has a compat restriction on TensorOperations" but `cuTENSOR` is a lower level library that likely doesn't know anything about `TensorOperations`, maybe you mean the other way around?

mtfishman:
In the `TensorOperations.jl` `Project.toml`:
https://github.com/Jutho/TensorOperations.jl/blob/master/Project.toml
I see they have a `cuTENSOR` extension and the `[compat]` entry is set to `cuTENSOR = "1"`, while the latest `cuTENSOR.jl` version is `v2.1.0` (https://github.com/JuliaGPU/CUDA.jl/blob/master/lib/cutensor/Project.toml). Maybe that is the issue you are seeing?

mtfishman:
I see there is an open PR about upgrading `TensorOperations` to `cuTENSOR` `v2` here: https://github.com/Jutho/TensorOperations.jl/pull/160.

Also note that `cuTENSOR` `v2` only supports Julia 1.8 and onward (https://github.com/JuliaGPU/CUDA.jl/blob/v5.3.3/lib/cutensor/Project.toml#L20).

mtfishman:
I guess the latest `CUDA.jl` version now requires Julia 1.8 and up anyway so we could have the same restriction for `NDTensorsCUDAExt` and `NDTensorscuTENSORExt`.

mtfishman:
I think the best course of action would be something like what you said with manually adding and removing packages as needed in the tests instead of putting them as dependencies in the test `Project.toml`.

What we could do is surround any code that relies on `TensorOperations` in the tests with `Pkg.add("TensorOperations") ... Pkg.rm("TensorOperations")`, and then surround any code that relies on `cuTENSOR` with `Pkg.add("cuTENSOR") ... Pkg.rm("cuTENSOR")`. Then we should be able to use the latest versions of `TensorOperations` and `cuTENSOR` in the appropriate parts of the tests. That's similar to your suggestion but would allow us to test `TensorAlgebra` even if `cuTENSOR` tests are requested. Hopefully that wouldn't be too complicated to set up.

kmp5VT:
@mtfishman Sorry, I did misunderstand what was going on. Thank you for looking into that and sending me this information!
If we are supporting `CUDA` and `cuTENSOR` only in julia version 1.8 and up, does that mean we should bump the compat of `NDTensors` and `ITensors` to `julia = 1.8` ?

kmp5VT:
Right now I have a rudementary `Pkg.add` and `Pkg.rm` in the `TensorAlgebra` test file but it throws some extension compile errors which might be an issue so there might be a better way. I haven't done much yet to enforce the `Julia 1.8` version other than add a compat value to the `NDTensors` project.  I have updated the code to launch CuTensor from the `contract!` kernel and have updated the kernels across the library to be able to use CuTensor. Right now for BlockSparse, dense blockwise contractions can cal cuTENSOR based contract however in the DMRG testing I am seeing an internal cuTENSOR error
```
ERROR: CUTENSORError: an invalid value was used as an argument (code 7, CUTENSOR_STATUS_INVALID_VALUE) 
``` 
That I am working to remove. Dense contractions are working properly and using cuTENSOR

mtfishman:
Sounds good, seems like there are a few wrinkles to work out but mostly coming along.

NDTensors/src/diag/tensoralgebra/contract.jl
```diff
@@ -58,6 +58,26 @@ function _contract!!(
...
+  output_tensor::Exposed{<:AbstractArray,<:DiagTensor},
+  labelsoutput_tensor,
+  tensor1::Exposed,
+  labelstensor1,
+  tensor2::Exposed,
+  labelstensor2,
+  α::Number=one(Bool),
+  β::Number=zero(Bool),
+)
+  return contract!(
```
kmp5VT, line 71: I found that I needed this function because the `contract(::DiagTensor, any, ::DiagTensor, any, ::DiagTensor, any)` function below does not have `alpha` and `beta`.

lkdvos:
> I see there is an open PR about upgrading `TensorOperations` to `cuTENSOR` `v2` here: [Jutho/TensorOperations.jl#160](https://github.com/Jutho/TensorOperations.jl/pull/160).
> 
> Also note that `cuTENSOR.jl` `v2` only supports Julia 1.8 and onward (https://github.com/JuliaGPU/CUDA.jl/blob/v5.3.3/lib/cutensor/Project.toml#L20).

Sorry if I repeat things you already know, but let me just give you the state of affairs for that PR here: we have a package extension that works for cuTENSOR v1 in the current version of TensorOperations, but cuTENSOR v2 is actually very breaking, as it renames lots of the functions. I started doing some work on getting the update going, but it is a bit more cumbersome as TensorOperations handles generic `StridedView` objects from `Strided.jl`. In principle this is now finished, but it requires a large amount of code duplication from `cuTENSOR` itself, which specializes it's methods on `DenseCuArray`, while in principle this restriction can be loosened. I got in touch with @maleadt to maybe reorganize a bit of cuTENSORs functions, (see https://github.com/JuliaGPU/CUDA.jl/pull/2356), and once that is finalized TensorOperations should be updated soon after.

I have also struggled a bit with the restriction of cuTENSOR v2 with julia v1.8, and I think we more or less decided to also drop the support for julia v1.6-1.7 in the new TensorOperations versions, although this is not set in stone yet.

mtfishman:
Thanks for the context @lkdvos. It makes sense to start by improving the `cuTENSOR` wrapper code before going through a big refactor of `TensorOperations`. This isn't causing serious issues for us, only a few tests of ours rely on `TensorOperations`.

NDTensors/src/diag/tensoralgebra/contract.jl
kmp5VT: check here that \alpha and \beta are [1, 0]

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -0,0 +1,32 @@
...
+using cuTENSOR: CuArray, CuTensor, mul!
+
+function NDTensors.contract!(
+  R::Exposed{<:CuArray,<:DenseTensor{ElT}},
+  labelsR,
+  T1::Exposed{<:CuArray,<:DenseTensor},
+  labelsT1,
+  T2::Exposed{<:CuArray,<:DenseTensor},
+  labelsT2,
+  α::Elα=one(Bool),
```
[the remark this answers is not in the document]
kmp5VT: These have been removed

NDTensors/src/diag/tensoralgebra/contract.jl
kmp5VT: I added asserts to verify that alpha was approx 1.0 and beta was approx 0.0

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -0,0 +1,32 @@
...
+  NDTensors,
+  DenseTensor,
+  Tensor,
+  array,
+  contract!,
+  contraction_output,
+  contraction_output_type,
+  contract_inds,
+  dense,
+  inds
```
mtfishman, line 11: Seems like not all of these are needed.

NDTensors/src/diag/tensoralgebra/contract.jl
```diff
@@ -58,6 +58,29 @@ function _contract!!(
...
+  tensor1::Exposed,
+  labelstensor1,
+  tensor2::Exposed,
+  labelstensor2,
+  α::Number=one(Bool),
+  β::Number=zero(Bool),
+)
+  elt = eltype(unexpose(output_tensor))
+  @assert α ≈ one(elt)
+  @assert β ≈ zero(elt)
```
mtfishman, line 73:
```suggestion
  @assert α == one(Bool)
  @assert β == zero(Bool)
```
Was there an issue with using strict equality instead of approximate equality? It seems better to be strict here.

mtfishman: Also for the sake of comparison the element type of the number shouldn't matter.

NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -16,14 +16,18 @@ if "metal" in ARGS || "all" in ARGS
   Pkg.add("Metal")
   using Metal
 end
+if "cutensor" in ARGS || "all" in ARGS
+  Pkg.add("cuTENSOR")
```
mtfishman, line 20: I see some funny things in the CI installation process with `TensorOperations` getting downgraded to some random old version of that package. Maybe we should also add `Pkg.rm("TensorOperations")` here. Or to be safe:
```julia
if in("TensorOperations", map(v -> v.name, values(Pkg.dependencies())))
  Pkg.rm("TensorOperations")
end
```
to only remove it if it is installed (see https://discourse.julialang.org/t/new-pkg-how-to-check-if-a-package-is-installed/13141).

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
kmp5VT: Thanks I missed this cleanup. I removed the unused functions

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: Yes `TensorOperations` is downgraded which causes an issue but should be fixed when they update their code. I will try removing it and see if that introduces any issues

kmp5VT: This works I just moved `using TensorOperations` into an if statement that is unreachable if one is testing `cuTENSOR`. Thanks!

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -0,0 +1,22 @@
...
+  T2::Exposed{<:CuArray,<:DenseTensor},
+  labelsT2,
+  α::Number=one(Bool),
+  β::Number=zero(Bool),
+)
+  cuR, cuT1, cuT2 =
+    CuTensor.(
+      array.((unexpose(R), unexpose(T1), unexpose(T2))),
+      collect.((labelsR, labelsT1, labelsT2)),
+    )
```
mtfishman, line 19: I think it would be better to split this into three separate lines, I think it will be easier to read.

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -0,0 +1,22 @@
+using NDTensors: NDTensors, DenseTensor, array
+using NDTensors.Expose: Exposed, unexpose
+using cuTENSOR: CuArray, CuTensor, mul!
```
mtfishman, line 3:
```suggestion
using cuTENSOR: CuArray, CuTensor
```

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -0,0 +1,22 @@
...
+  labelsT2,
+  α::Number=one(Bool),
+  β::Number=zero(Bool),
+)
+  cuR, cuT1, cuT2 =
+    CuTensor.(
+      array.((unexpose(R), unexpose(T1), unexpose(T2))),
+      collect.((labelsR, labelsT1, labelsT2)),
+    )
+  mul!(cuR, cuT1, cuT2, α, β)
```
mtfishman, line 20:
```suggestion
  cuTENSOR.mul!(cuR, cuT1, cuT2, α, β)
```
To make it clearer that this is `cuTENSOR`'s `mul!` function.

mtfishman:
Thanks @kmp5VT, this is a great step to get to for our GPU backends!

Can you update the table entry in https://github.com/ITensor/ITensors.jl/blob/v0.5.0/docs/src/RunningOnGPUs.md#gpu-backends?

kmp5VT:
@mtfishman I added a `block sparse (cuTENSOR)` row in the table and put `In progress` there since I am working on that `view` issue that I found before still and will be making an issue with cutensor when I can isolate the problem.

docs/src/RunningOnGPUs.md
```diff
@@ -60,9 +60,10 @@ The table below summarizes each backend's current capabilities.
 |                              | CUDA | ROCm   | Metal  | oneAPI |
```
mtfishman, line 60: Maybe instead of the current design it would be better to add a new `cuTENSOR` row:
```suggestion
|                              | CUDA | cuTENSOR | ROCm   | Metal  | oneAPI |
```
where we can fill with `✓ (cuLAPACK)` for the factorizations. The idea could be that the table tells users that, when they load a package extension, if those operations are implemented and additional information about how the operations are run where relevant.

We could also specify `✓ (cuBLAS)` for `CUDA` `Contractions` and `✓ (cuLAPACK)` for `CUDA` factorizations.

kmp5VT: @mtfishman Sure I can do this. Should I write something up about how when you enable `cuTENSOR` you cannot revert to using `cuBLAS` for dense contractions? And that the cuTENSOR library is an extension to the CUDA.jl library?

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
kmp5VT: I updated this code

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
kmp5VT: This has been removed, thanks!

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
kmp5VT: That makes sense! I split these into three lines, thanks!

docs/src/RunningOnGPUs.md
mtfishman: Sure, a short summary about `cuTENSOR` would be reasonable.

docs/src/RunningOnGPUs.md
```diff
@@ -45,24 +45,25 @@ ITensor currently provides
 package extensions for the following GPU backends:
 
 * [CUDA.jl](https://github.com/JuliaGPU/CUDA.jl) (NVIDIA GPUs)
+* [cuTENSOR.jl] (https://github.com/JuliaGPU/CUDA.jl/tree/master/lib/cutensor) (Submodule of `CUDA.jl`) 
 * [Metal.jl](https://github.com/JuliaGPU/Metal.jl) (Apple GPUs)
 * [AMDGPU.jl](https://github.com/JuliaGPU/AMDGPU.jl) (AMD GPUs)
 
 Our goal is to support all GPU backends which are supported by the [JuliaGPU organization](https://juliagpu.org).
 
+Notice that `cuTENSOR.jl` is an optional contraction backend for `CUDA.jl` and relies primarily on the `CuArray`. If `cuTENSOR` is loaded nearly all `CuArray` based contraction will utilize the `cuTENSOR.mul!` implementation instead of `cuBLAS.mul!`. It is not possible, as of now, to revert to `cuBLAS` if `cuTENSOR` is loaded. Blocksparse support with `cuTENSOR` is an active area of development and has a known bug for certain block-contractions. It is advised as of now to not use `cuTENSOR` with `BlockSparse`.
```
mtfishman, line 54: I wouldn't describe `cuTENSOR.jl` as a "contraction backend for `CUDA.jl`", rather that it is based on `CUDA.jl` and provides new functionality for accelerated binary tensor contractions.

Also, I think:

> If `cuTENSOR` is loaded nearly all `CuArray` based contraction will utilize the `cuTENSOR.mul!` implementation instead of `cuBLAS.mul!`.

is too low level (and also sounds misleading, it makes it sound like `cuBLAS` has a function `mul!` that performs tensor contraction, which it doesn't). I would just say that when `cuTENSOR` is loaded then ITensors with `CuArray` data are contracted using `cuTENSOR`, while without it contractions are mapped to matrix multiplications which are performed using `cuBLAS`.

kmp5VT: I updated the docs with your suggestions. Let me know if there is more I should change. Thanks!

docs/src/RunningOnGPUs.md
```diff
@@ -45,24 +45,25 @@ ITensor currently provides
 package extensions for the following GPU backends:
 
 * [CUDA.jl](https://github.com/JuliaGPU/CUDA.jl) (NVIDIA GPUs)
+* [cuTENSOR.jl] (https://github.com/JuliaGPU/CUDA.jl/tree/master/lib/cutensor) (Submodule of `CUDA.jl`) 
 * [Metal.jl](https://github.com/JuliaGPU/Metal.jl) (Apple GPUs)
 * [AMDGPU.jl](https://github.com/JuliaGPU/AMDGPU.jl) (AMD GPUs)
 
 Our goal is to support all GPU backends which are supported by the [JuliaGPU organization](https://juliagpu.org).
 
+Notice that `cuTENSOR.jl` is based on `CUDA.jl` and provides new functionality for accelerated binary tensor contractions. If the `cuTENSOR` library is loaded then ITensors with `CuArray` data are contracted using `cuTENSOR` and if the `cuTENSOR` library is not loaded then binary tensor contractions are mapped to a matrix multiplication and performed using `cuBLAS`. It is advised as of now to not use `cuTENSOR` with `BlockSparse`.
```
mtfishman, line 54:
```suggestion
Notice that `cuTENSOR.jl` is an extension of `CUDA.jl` that provides new functionality for accelerated binary tensor contractions. If the `cuTENSOR.jl` library is loaded then ITensors with `CuArray` data are contracted using `cuTENSOR` and if the `cuTENSOR.jl` library is not loaded but `CUDA.jl` is loaded then binary tensor contractions are mapped to a matrix multiplication and performed using `cuBLAS`.
```
I don't think we have have to say "It is advised as of now to not use `cuTENSOR` with `BlockSparse`." since we already have a warning about block sparse operations on GPU below.

NDTensors/src/diag/tensoralgebra/contract.jl
```diff
@@ -58,6 +58,28 @@ function _contract!!(
...
+  labelsoutput_tensor,
+  tensor1::Exposed,
+  labelstensor1,
+  tensor2::Exposed,
+  labelstensor2,
+  α::Number=one(Bool),
+  β::Number=zero(Bool),
+)
+  @assert α == one(Bool)
+  @assert β == zero(Bool)
```
mtfishman, line 72:
```suggestion
  @assert isone(α)
  @assert iszero(β)
```
The same thing but simpler and easier to read. I like to use `isone` and `iszero` when appropriate.

docs/src/RunningOnGPUs.md
```diff
@@ -45,24 +45,25 @@ ITensor currently provides
 package extensions for the following GPU backends:
 
 * [CUDA.jl](https://github.com/JuliaGPU/CUDA.jl) (NVIDIA GPUs)
+* [cuTENSOR.jl] (https://github.com/JuliaGPU/CUDA.jl/tree/master/lib/cutensor) (Submodule of `CUDA.jl`) 
 * [Metal.jl](https://github.com/JuliaGPU/Metal.jl) (Apple GPUs)
 * [AMDGPU.jl](https://github.com/JuliaGPU/AMDGPU.jl) (AMD GPUs)
 
 Our goal is to support all GPU backends which are supported by the [JuliaGPU organization](https://juliagpu.org).
 
+Notice that `cuTENSOR.jl` is an extension of `CUDA.jl` that provides new functionality for accelerated binary tensor contractions. If the `cuTENSOR.jl` library is loaded then ITensors with `CuArray` data are contracted using `cuTENSOR` and if the `cuTENSOR.jl` library is not loaded but `CUDA.jl` is loaded then binary tensor contractions are mapped to a matrix multiplication and performed using `cuBLAS`.
```
mtfishman, line 54:
```suggestion
Notice that `cuTENSOR.jl` is an extension of `CUDA.jl` that provides new functionality for accelerated binary tensor contractions. If the `cuTENSOR.jl` library is loaded then ITensors with `CuArray` data are contracted using `cuTENSOR` and if the `cuTENSOR.jl` library is not loaded but `CUDA.jl` is loaded then binary tensor contractions are mapped to a matrix multiplication and performed using `cuBLAS`.

```

docs/src/RunningOnGPUs.md
```diff
@@ -45,24 +45,26 @@ ITensor currently provides
 package extensions for the following GPU backends:
 
 * [CUDA.jl](https://github.com/JuliaGPU/CUDA.jl) (NVIDIA GPUs)
+* [cuTENSOR.jl] (https://github.com/JuliaGPU/CUDA.jl/tree/master/lib/cutensor) (Submodule of `CUDA.jl`)
```
mtfishman, line 48:
```suggestion
* [cuTENSOR.jl] (https://github.com/JuliaGPU/CUDA.jl/tree/master/lib/cutensor) (`CUDA.jl` extension providing accelerated binary tensor contractions) 
```

PR #1397, opened 2024-04-25
[ITensorMPS] Rename most functions exported by ITensorTDVP

emstoudenmire:
This PR is a companion to ITensor/ITensorTDVP.jl#67 which renames most functions prepended with `itensortdvp_` back to their original name. Only `itensortdvp_dmrg` is left as-is.

### DIFF
--- a/src/ITensorMPS/contract_mpo_mps.jl
+++ b/src/ITensorMPS/contract_mpo_mps.jl
@@ -12,7 +12,7 @@ function contractmpo_solver(; kwargs...)
   return solver
 end
 
-function itensortdvp_contract(
+function ITensors.contract(
   alg::Algorithm"fit", A::MPO, psi0::MPS; init_mps=psi0, nsweeps=1, kwargs...
 )::MPS
   n = length(A)
--- a/src/ITensorMPS/dmrg_x.jl
+++ b/src/ITensorMPS/dmrg_x.jl
@@ -10,7 +10,7 @@ function dmrg_x_solver(PH, t, psi0; current_time, outputlevel)
   return U_max, nothing
 end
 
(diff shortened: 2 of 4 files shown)


mtfishman:
An issue I see with the latest design is that it overloads `ITensors.contract` and `KrylovKit.linsolve`, which means that both `ITensors.jl` and `ITensorTDVP.jl` overload those functions with the same signature which will lead to warnings and precompilation failures (Julia's precompilation system breaks if two packages overload the same function with the same signature, perhaps for obvious reasons).

So two options are:
1. Define private versions of `ITensorMPS.contract` and `ITensorMPS.linsolve` for the time being. We could call them something else, like `ITensorMPS.alternating_update_contract` and `ITensorMPS.alternating_update_linsolve`, or go back to the previous names `ITensorMPS.itensortdvp_contract` and `ITensorMPS.itensortdvp_linsolve`.
2. Keep the current design but bump the version of `ITensors.jl` to `v0.5.0` to mark this as a breaking change to protect users from using this version of `ITensors.jl` and the current version of `ITensorTDVP.jl`.

I would prefer to avoid 2. so in the end it may be best to just go back to the names `itensortdvp_contract`, `itensortdvp_dmrg_x`, `itensortdvp_linsolve`, and `itensortdvp_tdvp`. That may give us more flexibility with designing those respective functions in `ITensorMPS.jl` anyway.

emstoudenmire:
I see, thanks for the detailed explanation. So then is the action item to close this PR? (Since it would keep the names as `itensortdvp_tdvp`, etc. ?)

mtfishman:
Yeah, looks like we can just close this and go with the previous design.

emstoudenmire:
Ok, I'll update the other PR accordingly.

PR #1400, opened 2024-04-28
[ITensors] Code reorganization

mtfishman:
This moves some code in `ITensors` into submodules. The goal is to attempt to split the library into independent parts, with an eye towards deciding what will go into the future `ITensorMPS.jl` package, and what else might get split off from `ITensors.jl`.

For example, I've moved the lattice construction code into `ITensors.ITensorMPS` since it is more related to the MPS code, and put the definition of the `QN` type into a submodule `ITensors.QuantumNumbers`. Additionally, I've started the process of splitting off the `op`, `state`, and `val` system into a submodule `ITensors.SiteTypes`, since it is possible that we might want to split it off from `ITensors.jl` and consider that code to be more of an add-on on top of `ITensors.jl` rather than part of the core `ITensors.jl` library. That code will be used by both `ITensorMPS.jl` and `ITensorNetworks.jl`, so we could think of it as sitting in between `ITensors.jl` and those packages.

@emstoudenmire


emstoudenmire:
These sound like welcome changes. We can keep reducing the "surface area" between these different libraries.

PR #1410, opened 2024-05-06
[WIP] [ITensors] Reorganize test exports

mtfishman:
Reorganize test exports into two lists, one which lists names that will be exported by `ITensorMPS` and the other will remain exported by `ITensors`.

This list is also reflected in the exports and tests of the new `ITensorMPS.jl` repository (https://github.com/ITensor/ITensorMPS.jl/pull/1).

It may be superfluous to make this change here but I found it was helpful to have a PR like this to assess which exports we should move over to `ITensorMPS.jl`.

@emstoudenmire


test/base/utils/TestITensorsExportedNames/TestITensorsExportedNames.jl
```diff
@@ -8,17 +8,186 @@ open("itensors_exported_names.jl", "w") do io
...
+  :uniqueind,
+  :uniqueindex,
+  :uniqueinds,
+  :use_combine_contract,
+  :use_debug_checks,
+  :vector,
+  :δ,
+  :⊕,
+  :⊙,
+]
```
mtfishman, line 183: @emstoudenmire please go through this list and see if there is anything we can argue could get moved to the `ITensorMPS.jl` export list, which would mean they would get removed from the `ITensors.jl` export list once we move `ITensors.ITensorMPS` to `ITensorMPS.jl`.

Many of them should be, or already are, deprecated, and I started to comment about those, but that is a separate issue. We should still try to categorize them by ones that are more associated with MPS operations vs. more general ITensor operations. Note that we may remove some things from the `ITensors.jl` export list, like `OpSum`, which we will keep in `ITensors.jl` since other packages like `ITensorNetworks.jl` will be using them.

emstoudenmire: Good question... Looking over the above list a few times, I did not see any that jumped out at me as being "MPS only". Some like `replacetags` certain have notable MPS/MPO overloads, but also make sense purely at the ITensor level. Were there any specifically you were still undecided about in terms of which package they should get exported from?

I do agree that reducing this list and carrying through any deprecations or planned deprecations is a great idea, with the aim of reducing the exports.

test/base/utils/TestITensorsExportedNames/TestITensorsExportedNames.jl
```diff
@@ -181,167 +271,83 @@ const ITENSORS_EXPORTED_NAMES = [
...
-  :unionind,
-  :unioninds,
   :unique_siteind,
   :unique_siteinds,
-  :uniqueind,
-  :uniqueindex,
-  :uniqueinds,
-  :use_combine_contract,
-  :use_debug_checks,
   :val,
```
emstoudenmire, line 346: Similarly, I would say the above list of `ITensorMPS` exported names all look to be correctly MPS related, and not needed for `ITensors.jl` alone, with maybe these exceptions:
* `state` 
* `siteind`
* `siteinds` maybe also if `siteind` goes back into ITensors.jl?
* similarly `op` and related functions

Ah but I see, you have moved these into a `lib/` with an extension (I saw the PR about that but only just remembered). So then is the idea of moving these exports to ITensorMPS more about whether they are exported by default (as ITensorMPS would) versus making them optional (as ITensors.jl now would)?

test/base/utils/TestITensorsExportedNames/TestITensorsExportedNames.jl
mtfishman: The main ones I was most uncertain about were the ones that were not obviously ITensor or MPS functionality, i.e. functions that you brought up below like `state`, `siteind[s]`, `op`, etc. which can be used with either ITensor operations or MPS/MPO operations but are more often used in association with MPS algorithms in practice. I'm biasing towards moving them to `ITensorMPS` (at least their exports, for the time being their implementation will live in `ITensors.jl`) but it is helpful to get a second pair of eyes on that decision as a sanity check. Also it is a long list so I wanted to double check I wasn't making an obvious mistake somewhere.

test/base/utils/TestITensorsExportedNames/TestITensorsExportedNames.jl
mtfishman: Yes, that's a good summary. Glad to see you were able to determine the logic behind these decisions without an explanation, that's a good sanity check!

Yes, the idea would be that we could provide some alternative for loading functionality like `state`, `siteind[s]`, `op`, etc. instead of exporting them by default from ITensors.jl. The initial suggestion will be to load `ITensorMPS.jl` (which is not much worse than the current situation anyway, in terms of burden of how much code gets loaded/compiled/etc.). In the future the goal will be to split off that functionality into separate packages that are either independent of or built on top of ITensors.jl.

For example, I could imagine something like:
- `LocalHilbertSpaces.jl`: Definitions like `SiteType`, `siteind`, etc. (maybe with one layer independent of ITensor that builds spaces and another that builds `Index` objects).
- `QuantumStates.jl`: Local state definitions (up, down, etc.) and things like `StateName`, independent of ITensor (say outputs dense or sparse vector representions of states or functions for setting elements independent of the particular data structures).
- `QuantumOperators.jl`: Local operators definitions (Sx, Sy, etc.) and things like `OpName`, independent of ITensor (say outputs dense or sparse matrix representions of operators or functions for setting elements independent of the particular data structures). Maybe combine with `QuantumStates.jl`.
- `ITensorStates.jl` and `ITensorOperators.jl`: Compatiblity layer between `QuantumStates.jl` and `QuantumOperators.jl` with the particular task of converting state and operator definitions into ITensor representations.
 
Obviously the particular modular structure and names will need to be discussed, I'm just giving a brief outline of a potential design.

So we could picture that these packages, along with `LazyApply` and `Ops`, sit between ITensors.jl and ITensorMPS.jl/ITensorNetworks.jl. But for the time being while we work out all of those details we can give users the instructions that they should load `ITensors` and `ITensorMPS` if they want customizable Hilbert space, state, and operator construction functionality. The only downside of that is that users that want Hilbert space, state, and operator definitions may find it strange that they have to load a full MPS package to get that, but that is already what we make them do with ITensors.jl so I don't think it is too bad. The benefit for us is that it will give us a lot more flexibility to break apart ITensors.jl into smaller composable pieces in the medium to long run while imposing fewer breaking changes for users during that process.

mtfishman:
The plan here is that once the initial version of [ITensorMPS.jl](https://github.com/ITensor/ITensorMPS.jl) is registered (https://github.com/JuliaRegistries/General/pull/106277), I plan to use `names(ITensorMPS)` instead of the list `ITENSORMPS_EXPORTED_NAMES`, revive the full list of exports `ITENSORS_EXPORTED_NAMES`, and then have a tests:
```julia
@test issetequal(names(ITensors), ITENSORS_EXPORTED_NAMES)
@test issetequal(names(ITensors), [ITENSORBASE_EXPORTED_NAMES; names(ITensorMPS)])
```
, i.e. test that the planned future exports of `ITensors.jl` stored in `ITENSORBASE_EXPORTED_NAMES` and the exports of `ITensorMPS` are disjoint and match the exports of `ITensors`.

mtfishman:
ITensorMPS.jl is registered and exports the list suggested in this PR and also has a test for that list: https://github.com/ITensor/ITensorMPS.jl/blob/v0.1.2/test/utils/TestITensorMPSExportedNames.jl so I'll close this.

PR #1432, opened 2024-05-11
[ITensors] Don't export ITensorMPS

mtfishman:
`Reexport.@reexport Module` apparently reexports the module name `Module` itself: https://github.com/simonster/Reexport.jl/issues/39 with no way of avoiding it, which we don't want in the case of `ITensors.ITensorMPS`. This fixes that with a custom solution using a for loop and `@eval`.


mtfishman:
Without this, if you type `using ITensorMPS; using ITensors` it results in the following warning:
```julia
julia> using ITensorMPS; using ITensors
WARNING: using ITensors.ITensorMPS in module Main conflicts with an existing identifier.

(jl_kbBmZ5) pkg> st ITensors ITensorMPS
Status `/private/var/folders/qz/q22pzwm144z9fq57mpf1hfp40000gq/T/jl_kbBmZ5/Project.toml`
  [0d1a4710] ITensorMPS v0.1.7
  [9136182c] ITensors v0.6.1
```
which is being hit in https://github.com/ITensor/ITensorParallel.jl/pull/22.

emstoudenmire:
Interesting. Glad there is a fix.

PR #1434, opened 2024-05-13
[SparseArraysDOKs] Add setindex_maybe_grow! and macro @maybe_grow 

### LINKED ISSUE #1433: [SparseArrayDOKs] [ENHANCEMENT] `resize!` and `@maybe_grow`
It would be helpful to make the size of `SparseArrayDOK` a mutable field (and maybe define an `IsResizable` trait).

Then we can have an API for setting the size of a `SparseArrayDOK` in-place with:
```julia
function resize!(a::SparseArrayDOK{<:Any,N}, new_size::NTuple{N,Integer}) where {N}
  a.dims = new_size
  return a
end
```
and additionally functionality for setting an element and allowing the array to grow as needed:
```julia
function setindex_maybe_grow!(a::SparseArrayDOK{<:Any,N}, value, index::NTuple{N,Int}...) where {N}
  if any(index .> size(a))
    resize!(a, max.(index, size(a)))
  end
  a[index...] = value
  return a
end
```
which could have an external API based around annotating calls to `setindex!` with a macro `@maybe_grow`:
```julia
a = SparseArrayDOK{Float64}(3, 3)
a[4, 4] = 44.0 # Calls `setindex!, errors as out-of-bounds.
@maybe_grow a[4, 4] = 44.0 # Calls `setindex_maybe_grow!`, resizes to `(4, 4)` and sets the element.
```
@emstoudenmire

emstoudenmire:
Adds the `setindex_maybe_grow!` function and the macro `@maybe_grow` as outlined in Issue #1433.

Fixes #1433.


NDTensors/src/lib/SparseArrayDOKs/src/sparsearraydok.jl
```diff
@@ -104,3 +104,28 @@ SparseArrayDOK{T}(a::AbstractArray) where {T} = SparseArrayDOK{T,ndims(a)}(a)
...
   return SparseArrayInterface.sparse_convert(SparseArrayDOK{T,N}, a)
 end
+
+function Base.resize!(a::SparseArrayDOK{<:Any,N}, new_size::NTuple{N,Integer}) where {N}
+  a.dims = new_size
+  return a
+end
+
+function setindex_maybe_grow!(
+  a::SparseArrayDOK{<:Any,N}, value, i1::Int, I::Int...
```
mtfishman, line 114:
```suggestion
  a::SparseArrayDOK{<:Any,N}, value, index::Vararg{Int,N}
```
to make sure it matches the number of dimensions of the array.

NDTensors/src/lib/SparseArrayDOKs/src/sparsearraydok.jl
```diff
@@ -104,3 +104,28 @@ SparseArrayDOK{T}(a::AbstractArray) where {T} = SparseArrayDOK{T,ndims(a)}(a)
...
+  a[index...] = value
+  return a
+end
+
+macro maybe_grow(ex)
+  arr_name = esc(ex.args[1].args[1])
+  index = esc(ex.args[1].args[2:end])
+  value = esc(ex.args[2])
+  quote
+    SparseArrayDOKs.setindex_maybe_grow!($arr_name, $value, $index...)
```
mtfishman, line 129:
```suggestion
   setindex_maybe_grow!($arr_name, $value, $index...)
```
Is the `SparseArrayDOKs.` prefix needed?

NDTensors/src/lib/SparseArrayDOKs/test/runtests.jl
```diff
@@ -94,5 +95,34 @@ using SparseArrays: SparseMatrixCSC, nnz
       end
     end
   end
+
```
mtfishman, line 98:
```suggestion
```
I've been biasing towards using fewer new lines.

NDTensors/src/lib/SparseArrayDOKs/test/runtests.jl
```diff
@@ -94,5 +95,34 @@ using SparseArrays: SparseMatrixCSC, nnz
...
+  @testset "Maybe Grow Feature" begin
+    a = SparseArrayDOK{elt,2}((0, 0))
+
+    setindex_maybe_grow!(a, 230, 2, 3)
```
mtfishman, line 102:
```suggestion
    SparseArrayDOK.setindex_maybe_grow!(a, 230, 2, 3)
```
and remove it from the list above, to make it clearer it is meant to be a private function.

NDTensors/src/lib/SparseArrayDOKs/src/sparsearraydok.jl
emstoudenmire: Good question – no not needed, so I removed it and tests are still passing.

NDTensors/src/lib/SparseArrayDOKs/src/sparsearraydok.jl
```diff
@@ -104,3 +104,28 @@ SparseArrayDOK{T}(a::AbstractArray) where {T} = SparseArrayDOK{T,ndims(a)}(a)
...
+    resize!(a, max.(index, size(a)))
+  end
+  a[index...] = value
+  return a
+end
+
+macro maybe_grow(ex)
+  arr_name = esc(ex.args[1].args[1])
+  index = esc(ex.args[1].args[2:end])
+  value = esc(ex.args[2])
```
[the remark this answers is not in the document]
emstoudenmire: Made this change. I tried adding a test for the check with `@test_throws` but it seems since the throwing behavior happens at parse time that the Test package can't catch it.

NDTensors/src/lib/SparseArrayDOKs/src/sparsearraydok.jl
```diff
@@ -110,22 +110,23 @@ function Base.resize!(a::SparseArrayDOK{<:Any,N}, new_size::NTuple{N,Integer}) w
...
+  end
   arr_name = esc(ex.args[1].args[1])
   index = esc(ex.args[1].args[2:end])
   value = esc(ex.args[2])
-  quote
-    SparseArrayDOKs.setindex_maybe_grow!($arr_name, $value, $index...)
+  e = quote
+    setindex_maybe_grow!($arr_name, $value, $index...)
   end
+  return e
```
mtfishman, line 131:
```suggestion
  return quote
    setindex_maybe_grow!($arr_name, $value, $index...)
  end
```

mtfishman: Just a bit simpler.

emstoudenmire:
Ok just made those improvements to the `@maybe_grow` macro

mtfishman:
@emstoudenmire I think we could simplify the macro implementation using a package like [MacroTools.jl](http://fluxml.ai/MacroTools.jl/dev/pattern-matching/#Pattern-Matching). There are other similar packages but I think that one is popular.

mtfishman:
It's a lightweight and widely used package so if it simplifies the code then I don't see any harm.

emstoudenmire:
Ok, made that change. The `@capture` macro is definitely nice. I read through the whole MacroTools docs and didn't so easily see a nicer way to write the part of `@maybe_grow` that checks it is the right type of expression.

emstoudenmire:
There is the [MacroTools.isexpr](http://fluxml.ai/MacroTools.jl/dev/utilities/#MacroTools.isexpr) function, but the documentation for it is rather minimal so I'm not sure if it could be used.

mtfishman:
In the [Pattern Matching section](http://fluxml.ai/MacroTools.jl/stable/pattern-matching/#Pattern-Matching-1) it says that:

> Because `@capture` doubles as a test as well as extracting values, you can easily handle unexpected input (try writing this by hand):
> ...

Can you remove the manual check you wrote and see if it errors if incorrect expressions are annotated with `@maybe_grow`?

emstoudenmire:
I was also hoping initially that `@capture` would automatically validate the expression too. Maybe it does in other cases but in this case, whenever I throw an incorrect (non-matching) expression at it I just get a very generic / cryptic error:
```
julia> @maybe_grow x+y+z
ERROR: MethodError: no method matching iterate(::Nothing)
```
(it is always that error for any expression not of the form a[I...] = v).

mtfishman:
> I was also hoping initially that `@capture` would automatically validate the expression too. Maybe it does in other cases but in this case, whenever I throw an incorrect (non-matching) expression at it I just get a very generic / cryptic error:
> 
> ```
> julia> @maybe_grow x+y+z
> ERROR: MethodError: no method matching iterate(::Nothing)
> ```
> 
> (it is always that error for any expression not of the form a[I...] = v).

Maybe we could wrap it in a `try ... catch` block and throw our own error?

emstoudenmire:
Oh, that's a good idea. I'll do that.

emstoudenmire:
Unfortunately I am not able to get the `try...catch...` to work. I still get the same exception and not the one that the catch block is trying to throw. Maybe it could be because the macro is running at parse time and so the try catch block does not work there or is inactive at parse time in some sense? Here's what I tried:
```julia
macro maybe_grow(ex)
  try
    @capture(ex, array_[indices__] = value_)
    return :(setindex_maybe_grow!($(esc(array)), $value, $indices...))
  catch
    error(
      "@maybe_grow must be used with setindex! syntax (as @maybe_grow a[i,j,...] = value)"
    )
  end
end
```

mtfishman:
That's too bad, I guess it requires some more investigation.

Let's at least refactor the check into a function `is_siteindex!_expr(expr)` to make the macro a bit cleaner and more readable.

emstoudenmire:
I ended up blending the ideas of a checking function and using a try..catch block. The reason is that there are some expressions (for example `@maybe_grow v`) where my check itself errors, since some expressions have no `.head` field, so then the error message is again bad. A `try..catch` block ensures a good error message in all failing cases here. 

Also apparently the `try...catch` not working when wrapped around the `@capture` macro seems to be due to something about MacroTools itself, versus whether `try..catch` is used inside a function or a macro. I say this because I also tried moving the `@capture` macro itself out to another function and was still not able to catch the exception that it threw.

mtfishman:
> I ended up blending the ideas of a checking function and using a try..catch block. The reason is that there are some expressions (for example `@maybe_grow v`) where my check itself errors, since some expressions have no `.head` field, so then the error message is again bad. A `try..catch` block ensures a good error message in all failing cases here.

You should be able to catch that kind of issue in a more elegant way using dispatch by specializing on whether or not the type of the input is `Expr`:
```julia
function is_setindex!_expr(expr::Expr)
  return expr.head === :(=) && is_getindex_expr(first(expr.args))
end
is_setindex!_expr(x) = false

is_getindex_expr(expr::Expr) = (expr.head === :ref)
is_getindex_expr(x) = false
```
Note that I'm using `===` because it is a stricter check. `Expr` always has a `head` field and an `args` field so these checks should never error and always just output `true` or `false`.

emstoudenmire:
That is a nicer pattern, especially since the type constraints guarantee the code inside can't error. I just updated the code to use it.

NDTensors/Project.toml
```diff
@@ -65,6 +66,7 @@ HDF5 = "0.14, 0.15, 0.16, 0.17"
 HalfIntegers = "1"
 InlineStrings = "1"
 LinearAlgebra = "1.6"
+MacroTools = "0.5"
```
mtfishman, line 69: Not needed now.

NDTensors/Project.toml
```diff
@@ -18,6 +18,7 @@ GPUArraysCore = "46192b85-c4d5-4398-a991-12ede77f4527"
 HalfIntegers = "f0d1745a-41c9-11e9-1dd9-e5d34d218721"
 InlineStrings = "842dd82b-1e85-43dc-bf29-5d0ee9dffc48"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
```
mtfishman, line 21: Not needed now.

NDTensors/src/lib/SparseArrayDOKs/src/sparsearraydok.jl
```diff
@@ -1,10 +1,11 @@
 using Accessors: @set
 using Dictionaries: Dictionary, set!
+using MacroTools: @capture
 using ..SparseArrayInterface:
   SparseArrayInterface, AbstractSparseArray, getindex_zero_function
 
 # TODO: Parametrize by `data`?
-struct SparseArrayDOK{T,N,Zero} <: AbstractSparseArray{T,N}
+mutable struct SparseArrayDOK{T,N,Zero} <: AbstractSparseArray{T,N}
```
mtfishman, line 8: What I don't like about marking the entire struct as `mutable` is that makes all of the fields mutable, while we only want `dims` to be mutable.

In Julia 1.8 and above we could do the following:
```julia
mutable struct SparseArrayDOK{T,N,Zero} <: AbstractSparseArray{T,N}
  const data::Dictionary{CartesianIndex{N},T}
  dims::NTuple{N,Int}
  const zero::Zero
end
```
as described in the Julia docs here: https://docs.julialang.org/en/v1/manual/types/#Mutable-Composite-Types but I want to support Julia 1.6 for now se we'll have to hold off on that design.

For now what we can do is the following:
```julia
struct SparseArrayDOK{T,N,Zero} <: AbstractSparseArray{T,N}
  data::Dictionary{CartesianIndex{N},T}
  dims::Ref{NTuple{N,Int}}
  zero::Zero
end
```
i.e. not mark the struct as `mutable` but instead make the `dims` field a reference to the tuple of dimension lengths.

Then, we can define:
```julia
Base.size(a::SparseArrayDOK) = a.dims[]

function Base.resize!(a::SparseArrayDOK{<:Any,N}, new_size::NTuple{N,Integer}) where {N}
  a.dims[] = new_size
  return a
end
```
etc. Some of the constructors would of course need to be updated, however I think we could just do that with an inner constructor:
```julia
struct SparseArrayDOK{T,N,Zero} <: AbstractSparseArray{T,N}
  data::Dictionary{CartesianIndex{N},T}
  dims::Ref{NTuple{N,Int}}
  zero::Zero
  function SparseArrayDOK{T,N,Zero}(data, dims::NTuple{N,Int}, zero)
    return new{T,N,Zero}(data, Ref(dims), zero)
  end
end
```
and ideally keep the rest of the constructors the same.

NDTensors/src/lib/SparseArrayDOKs/src/sparsearraydok.jl
```diff
@@ -104,3 +105,44 @@ SparseArrayDOK{T}(a::AbstractArray) where {T} = SparseArrayDOK{T,ndims(a)}(a)
...
+  a[I...] = value
+  return a
+end
+
+function check_siteindex!_expr(expr, macroname=:macro)
+  try
+    @assert expr.head == :(=) && expr.args[1] isa Expr && expr.args[1].head == :(ref)
+  catch
+  end
+end
```
mtfishman, line 127: Seems like this can be removed now?

emstoudenmire: Yes, my mistake

NDTensors/src/lib/SparseArrayDOKs/src/sparsearraydok.jl
emstoudenmire: Yes, I sort of cringed having to put `mutable` on the type, so this a nice compromise. Good to know about the better option post Julia 1.8. I just made the above changes.

NDTensors/Project.toml
emstoudenmire: Actually, I am still using the `@capture` macro inside the `@maybe_grow` macro implementation, just not using it for checking the validity of the expression.

mtfishman: Oh right, never mind then.

mtfishman:
Looks like a few more constructors need to be defined.

NDTensors/src/lib/SparseArrayDOKs/src/sparsearraydok.jl
```diff
@@ -15,6 +15,10 @@ struct SparseArrayDOK{T,N,Zero} <: AbstractSparseArray{T,N}
 end
 
 # Constructors
+function SparseArrayDOK(data, dims::NTuple{N,Int}, zero) where {N}
+  return SparseArrayDOK{eltype(data),N,typeof(zero)}(data, dims, zero)
+end
```
mtfishman, line 20:
```suggestion
function SparseArrayDOK(data, dims::Tuple{Vararg{Int}}, zero)
  return SparseArrayDOK{eltype(data),length(dims),typeof(zero)}(data, dims, zero)
end
```
just to avoid the `where` syntax.

mtfishman:
Thanks! This will be helpful.

Issue #1430, opened 2024-05-10
[NDTensors] [ITensors] [BUG] `directsum` with `EmptyStorage`


mtfishman:
See https://itensor.discourse.group/t/directsum-issue/1699.

PR #1443, opened 2024-05-14
[ITensors] Fix issue with directsum involving EmptyStorage

kmp5VT:
# Description

In this I attempt to fix [issue 1430](https://github.com/ITensor/ITensors.jl/issues/1430) where there is an issue in directsum when an `EmptyTensor` is fed in as the first argument instead of the second. In `tensor_algebra.jl` the directsum code attempts to call `one(EmptyNumber)`.  

To fix this code I am modifying this function 
```julia
function directsum_projectors(
  elt1::Type{<:Number}, elt2::Type{<:Number}, i::Index, j::Index, ij::Index
)
  D1 = zeros_itensor(elt1, dag(i), ij)
  D2 = zeros_itensor(elt1, dag(j), ij)
  directsum_projectors!(tensor(D1), tensor(D2))
  return D1, D2
end
```
I have 2 fix ideas, the first is slightly intrusive `elt = promote_type(elt1, elt2)` and use `elt` instead of `elt1` in both `D1` and `D2`. The second is less intrusive, it is 
```julia 
directsum_projectors(
  elt1::Type{<:EmptyNumber}, elt2::Type{<:Number}, i::Index, j::Index, ij::Index) = directsum_projector(elt2, elt1, i, j, ij) 
```
which is what I have implemented. 

# How Has This Been Tested?

I added some tests in the `test_itensors.jl` I test an empty tensor in front and in back.

### DIFF
--- a/src/tensor_operations/tensor_algebra.jl
+++ b/src/tensor_operations/tensor_algebra.jl
@@ -261,10 +261,14 @@ function directsum_projectors(
   return D1, D2
 end
 
-function directsum_projectors(::Type{<:EmptyNumber}, ::Type{<:EmptyNumber}, ::Index, ::Index, ::Index)
-  error("It is not possible to call directsum on two tensors with element type EmptyNumber.
-  If you are inputting ITensors constructor like ITensor(i, j), try specifying the element type, 
-  e.g. ITensor(Float64, i, j), or filling them with zero value, e.g. ITensor(0.0, i, j).")
+function directsum_projectors(
+  ::Type{<:EmptyNumber}, ::Type{<:EmptyNumber}, ::Index, ::Index, ::Index
+)
+  return error(
+    "It is not possible to call directsum on two tensors with element type EmptyNumber.
+If you are inputting ITensors constructor like ITensor(i, j), try specifying the element type, 
+e.g. ITensor(Float64, i, j), or filling them with zero value, e.g. ITensor(0.0, i, j).",
+  )
 end
 
 function check_directsum_inds(A::ITensor, I, B::ITensor, J)
--- a/src/tensor_operations/tensor_algebra.jl
+++ b/src/tensor_operations/tensor_algebra.jl
@@ -254,16 +254,11 @@ function directsum_projectors(
   # Or with new notation:
   # D1 = zeros(elt1, dag(i), ij)
   # D2 = zeros(elt1, dag(j), ij)
-  D1 = zeros_itensor(elt1, dag(i), ij)
-  D2 = zeros_itensor(elt1, dag(j), ij)
+  elt = promote_type(elt1, elt2)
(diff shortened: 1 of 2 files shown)


mtfishman:
I would prefer a solution based on `elt = promote_type(elt1, elt2)`, it is more general (for example it doesn't require code logic that is hard-coded to `EmptyNumber` like the current solution) and `promote_type(elt1, elt2)` should be the resulting element type of the final tensor of the operation anyway.

As another corner case, what happens if both inputs are `EmptyStorage` with `EmptyNumber`, i.e.:
```julia
using ITensors
i, j, k = Index.(2, ("i", "j", "k"))
A = ITensor(i, j)
B = ITensor(i, k)
directsum(A=> j, B=> k)
```
? Probably the result should be `EmptyStorage` but it may be tricky to make that work with the current code logic. Something to keep in mind, but best to leave that for a future PR to keep this PR simpler and more focused.

kmp5VT:
@mtfishman Would it not be reasonable to have this function 
```
directsum_projectors(
  elt1::Type{<:EmptyNumber}, elt2::Type{<:EmptyNumber}, i::Index, j::Index, ij::Index
) = error("It is not possible to call directsum on two tensors with element type EmptyNumber")
```

mtfishman:
> @mtfishman Would it be reasonable to have this function for now
> 
> ```
> directsum_projectors(
>   elt1::Type{<:EmptyNumber}, elt2::Type{<:EmptyNumber}, i::Index, j::Index, ij::Index
> ) = error("It is not possible to call directsum on two tensors with element type EmptyNumber")
> ```

Sure, that sounds like a good idea. Maybe add a sentence like:

> "If you are inputting ITensors constructor like `ITensor(i, j)`, try specifying the element type, e.g. `ITensor(Float64, i, j)`, or filling them with zero value, e.g. `ITensor(0.0, i, j)`".

src/tensor_operations/tensor_algebra.jl
```diff
@@ -254,12 +254,23 @@ function directsum_projectors(
...
   return D1, D2
 end
 
+function directsum_projectors(
+  ::Type{<:EmptyNumber}, ::Type{<:EmptyNumber}, ::Index, ::Index, ::Index
+)
+  return error(
+    "It is not possible to call directsum on two tensors with element type EmptyNumber.
+If you are inputting ITensors constructor like ITensor(i, j), try specifying the element type, 
+e.g. ITensor(Float64, i, j), or filling them with zero value, e.g. ITensor(0.0, i, j).",
```
mtfishman, line 270:
```suggestion
If you are inputting ITensors constructed like `ITensor(i, j)`, try specifying the element type, 
e.g. `ITensor(Float64, i, j)`, or fill them with zero values, e.g. `ITensor(zero(Float64), i, j)`.",
```

mtfishman:
Thanks!

Issue #1450, opened 2024-05-17
[NDTensors] [BUG] `tr` on GPU


mtfishman:
Reported here: https://itensor.discourse.group/t/problem-performing-trace-calculation-with-gpu/1728

```julia
using ITensors: Index, random_itensor
using LinearAlgebra: tr
using Metal: mtl
i = Index(2)
x = random_itensor(i', i)
tr(mtl(x))
```

gives:

<details><summary>output</summary><p>

```julia
Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.

If you want to allow scalar iteration, use `allowscalar` or `@allowscalar`
to enable scalar iteration globally or for the operations in question.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] errorscalar(op::String)
    @ GPUArraysCore ~/.julia/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:155
  [3] _assertscalar(op::String, behavior::GPUArraysCore.ScalarIndexing)
    @ GPUArraysCore ~/.julia/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:128
  [4] assertscalar(op::String)
    @ GPUArraysCore ~/.julia/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:116
  [5] getindex
    @ ~/.julia/packages/GPUArrays/OKkAu/src/host/indexing.jl:48 [inlined]
  [6] getindex
    @ ~/.julia/packages/NDTensors/5rDm2/src/lib/Expose/src/functions/abstractarray.jl:13 [inlined]
  [7] getindex
    @ ~/.julia/packages/NDTensors/5rDm2/src/dense/densetensor.jl:90 [inlined]
  [8] getindex
    @ ~/.julia/packages/NDTensors/5rDm2/src/dense/densetensor.jl:95 [inlined]
  [9] getdiagindex
    @ ~/.julia/packages/NDTensors/5rDm2/src/tensor/tensor.jl:361 [inlined]
[...]

PR #1453, opened 2024-05-20
[NDTensors] Fix contracting dense with diag on GPU

kmp5VT:
# Description
In reference to this [bug report](https://github.com/ITensor/ITensors.jl/issues/1450)

This bug is two parted. First I am working to make it possible to call `tr` on GPU based Tensors. The issue here is a scalar indexing problem where the code tries to `getdiagindex` of the tensor. For now I have tried to use `@allowscalar` and `expose` to solve the issue. Timing wise I have found CUDA based `tr` with `@allowscalar` is roughly 2x faster than CPU based trace.

The second bug is that delta functions are not being properly constructed on GPU. This is, impart, due to delta being a `UniformDiag` which does not carry information about where the diag will be allocated and assumes CPU. I am not sure yet how to fix this problem. Potentially we could replace the datatype of `UniformDiag` from `<:Number` to `<:UnallocatedFill{<:Number}`

# Checklist:

- [ ] Trace works on CPU and GPU. 
- [ ] Performance of trace on GPU is equal to or better than CPU
- [ ] Delta functions can be constructed on GPU
- [ ] All current unittests pass
- [ ] Unittests are made for the failing cases which prompted this PR.
- [ ] Ensure that all solutions also work with quantum numbers and block sparsity


mtfishman:
> I am not sure yet how to fix this problem. Potentially we could replace the datatype of `UniformDiag` from `<:Number` to `<:UnallocatedFill{<:Number}`

I would prefer not going that route since that is a much more involved change that would likely require rewriting a lot of the `UniformDiag` code. It would be better to try to solve it in a more narrow way.

NDTensors/src/diag/tensoralgebra/contract.jl
```diff
@@ -129,7 +129,7 @@ function contract!(
       # Assumes C starts set to 0
       c₁ = zero(ElC)
       for i in 1:min_dim
-        c₁ += getdiagindex(A, i) * getdiagindex(B, i)
+        c₁ += getdiagindex(expose(A), i) * getdiagindex(expose(B), i)
```
mtfishman, line 132: Instead of this approach I think it would be better to catch cases when dense * diag contractions are occurring on GPU (say with `expose`) and the convert the diag tensor to dense and then perform a dense contraction.

kmp5VT: I will get started on that! I was actually surprised that this `@allowscalar` was still faster than the CPU version. Thanks!

mtfishman: Yeah that is surprising... It was good that you tried it out but I think converting to dense will be more robust.

kmp5VT: @mtfishman I wrote a draft that fixes the problem and goes through `contract!`. Right now, because `delta` is by default constructed on CPU, converting the `DiagTensor` to `DenseTensor` constructs it as a CPU array. Right now I am using adapt to move it to GPU but this will be updated with the changes in `delta`. I am working on adding a test now

mtfishman: You say:
> Right now, because delta is by default constructed on CPU, converting the DiagTensor to DenseTensor constructs it as a CPU array.
I don't really understand what you mean. Which delta is being constructed on CPU? Do you mean the ones in the `tr` operation?

kmp5VT: Sorry that was confusing I updated the comment. I am talking about the delta in trace the operation 
```
Tᶜ *= δ(eltype(T), dag((cᴸ, cᴿ)))
```
Converting this delta or any `UniformDiag` to `DenseTensor` using `dense` uses `Vector` as the datatype of the new `Dense` object. See this example
```
d = tensor(δ(eltype(T), dag((cᴸ, cᴿ))))
typeof(d) = NDTensors.DiagTensor{Float32, 2, Tuple{Index{Int64}, Index{Int64}}, NDTensors.Diag{Float32, Float32}}

typeof(dense(d)) = NDTensors.DenseTensor{Float32, 2, Tuple{Index{Int64}, Index{Int64}}, NDTensors.Dense{Float32, Vector{Float32}}}
```

mtfishman: Makes sense, thanks for the clarification.

mtfishman: So what is your proposed plan for constructing those tensors on GPU when that is needed? I can imagine that happening in different parts of the code, with potential costs and benefits to those different choices.

kmp5VT: Yes I totally agree and I do not yet have a solid plan. I am going to try out different solutions and will let you know when I have more ideas!

NDTensors/ext/NDTensorsCUDAExt/indexing.jl
```diff
@@ -1,6 +1,6 @@
 using CUDA: CuArray
 using GPUArraysCore: @allowscalar
-using NDTensors: NDTensors
+using NDTensors: NDTensors, getdiagindex
```
mtfishman, line 3:
```suggestion
using NDTensors: NDTensors
```
Seems like that's not needed now.

kmp5VT: Thanks!

kmp5VT:
@mtfishman So I am pushing an idea on how to fix the issue with `\delta` it involves reworking the `dense` function to look like this 
```
dense(T::Tensor) = dense(unwrap_array_type(T), T)
dense(datat::Type{<:AbstractArray}, T::Tensor) = setstorage(T, adapt(datat, dense(storage(T))))
```
Since the dense for `Diag` already looks like this
```
function dense(T::DiagTensor)
  return dense(unwrap_array_type(T), T)
end
```

kmp5VT:
@mtfishman For the dispatch of the `dense` function?

mtfishman:
> @mtfishman For the dispatch of the `dense` function?

Yes.

NDTensors/ext/NDTensorsAMDGPUExt/append.jl
```diff
@@ -1,5 +1,4 @@
-using GPUArraysCore: @allowscalar
-using AMDGPU: ROCArray
+using AMDGPU: @allowscalar, ROCArray
```
mtfishman, line 1: I would prefer loading `@allowscalar` from `GPUArraysCore` if possible to make it clearer where it is defined, was there an issue with that?

kmp5VT: Hi Matt, I was writing you a message out about this. I am running into an issue that perhaps you know how to solve. My understanding is that when we use `GPUArraysCore: @allowscalar` in our extensions, Julia requires that we have `GPUArraysCore` as a `[deps]` of `NDTensors`. 
If I have `GPUArraysCore` in `[deps]`, `[weakdeps]`, `[extensions]` and `[extras]` header and have `using GPUArraysCore: @allwoscalar` in the extensions, Julia [ignores the package](https://pkgdocs.julialang.org/dev/creating-packages/#Transition-from-normal-dependency-to-extension) as a `[dep]` and I runs into an issue compiling the extension
```
Precompiling NDTensorsMetalExt
        Info Given NDTensorsMetalExt was explicitly requested, output will be shown live 
ERROR: LoadError: ArgumentError: Package NDTensorsMetalExt does not have GPUArraysCore in its dependencies:
```
If I remove `GPUArraysCore` from `[weakdeps]` and `[extras]` but keep it in the `[extensions]` category,  the package manager fails to resolve the manifest
```
(jl_RMaKFD) pkg> resolve
  No Changes to `/private/tmp/jl_RMaKFD/Project.toml`
ERROR: KeyError: key "GPUArraysCore" not found
Stacktrace:
```
and if I remove it from `[extensions]` then the `NDTensorsGPUArraysExt` file does not load. 
I could keep it as a `[deps]` only, keep the extensions folder and just include the module loke we do with the `src/lib` modules?
The proposed solution in this PR is the opposite of this where I remove `GPUArraysCore` as a dependency and make it only an extension.

mtfishman: If you are only loading it inside of a package extension my understanding was that it does not need to be listed under `[deps]`.

mtfishman: That's definitely a solution but it just feels like circumventing an issue that shouldn't be an issue and we should try to figure out a proper solution since I'm sure this will come up again in the future. I'm certain I've defined package extensions with multiple weak dependencies but I can't recall which ones right now.

I believe you can list multiple weak dependencies of a package extension though I think I tried that and may have run into trouble with it. Maybe this post helps: https://discourse.julialang.org/t/using-multiple-packages-in-extensions-for-julia-1-9/98237?

It seems like the conclusion there was to list `GPUArraysCore` under `[deps]`, which is ok since it will be installed but not loaded unless the package extension using it is loaded. That is a bit unsatisfying but may be the best we can do if it doesn't work to list both `GPUArraysCore` and `AMDGPU`, `CUDA`, etc. as weakdeps of the corresponding package extension.

mtfishman: Note that in the example code in this section of the docs: https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions) it says:
```
[extensions]
# name of extension to the left
# extension dependencies required to load the extension to the right
# use a list for multiple extension dependencies
PlottingContourExt = "Contour"
```
so you can list multiple extension dependencies but I'm not sure what happens if you do that. Have you tried that?

kmp5VT: Ahh thank you, this was something I had meant to try but I had forgotten! I added `GPUArraysCore` to the `weakdeps` list for the GPU extensions and that fixed the problem
```
NDTensorsMetalExt = ["GPUArraysCore", "Metal"]
```
Now I can include `GPUArraysCore` in the extensions and I do not have it listed in the `[deps]` section of `NDTensors.

NDTensors/ext/NDTensorsGPUArraysCoreExt/contract.jl
```diff
@@ -0,0 +1,47 @@
...
+using NDTensors: NDTensors, DenseTensor, DiagTensor, contract!, dense, inds, Tensor
+using NDTensors.Expose: Exposed, expose, unexpose
+using NDTensors.TypeParameterAccessors: parenttype, set_ndims
+
+## In this function we convert the DiagTensor to a dense tensor and
+## Feed it back into contract
+function NDTensors.contract!(
+  output_tensor::Exposed{<:AbstractGPUArray,<:DenseTensor},
+  labelsoutput_tensor,
+  tensor1::Exposed{<:Any,<:DiagTensor},
```
mtfishman, line 12: One concern I have about making the type signature this general is that this appears to allow contracting Dense GPU tensors with Diag CPU tensors, which I think we don't want to allow right now since we don't allow contracting Dense GPU tensors with Dense CPU tensors (since it isn't a very well defined operation, i.e. should the result end up on GPU or CPU, and also may indicate someone making a mistake).

What happens with the current PR if you contract a Dense GPU tensor with a Diag CPU tensor? Does that end up in this function, go to another function, throw an error, etc.? Probably to be safe we should restrict this signature more and make it only work for uniform Diag tensors, either with:
```julia
tensor1::Exposed{<:Number,<:DiagTensor},
```
or:
```julia
tensor1::Exposed{<:Number,<:UniformDiagTensor},
```
but I'm curious what the behavior is now without making that change.

kmp5VT: I updated this to the `Exposed{<:Number, <:DiagTensor}` version. Thanks!

NDTensors/ext/NDTensorsGPUArraysCoreExt/contract.jl
```diff
@@ -0,0 +1,47 @@
...
+    β,
+  )
+end
+
+function NDTensors.contract!(
+  output_tensor::Exposed{<:AbstractGPUArray,<:DenseTensor},
+  labelsoutput_tensor,
+  tensor1::Exposed{<:AbstractGPUArray,<:DenseTensor},
+  labelstensor1,
+  tensor2::Exposed{<:Any,<:DiagTensor},
```
mtfishman, line 39:
```suggestion
  tensor2::Exposed{<:Number,<:DiagTensor},
```

kmp5VT: Sorry I missed that, thank you for catching it!

NDTensors/test/test_diag.jl
```diff
@@ -56,13 +58,19 @@ using .NDTensorsTestUtils: devices_list, is_supported_eltype
...
+  t = tensor(Diag(elt[1.0, 1.0, 1.0]), (3, 3))
+  A = randomTensor(Dense{elt}, (3, 3))
 
   @test contract(t, (1, -2), t, (-2, 3)) == t
   @test contract(A, (1, -2), t, (-2, 3)) == A
   @test contract(A, (-2, 1), t, (-2, 3)) == transpose(A)
+
+  ## Testing sparse contractions on GPU
+  t = tensor(Diag(one(elt)), (3, 3))
+  @test contract(t, (-1, -2), dev(A), (-1, -2))[] == dot(t, A)
```
mtfishman, line 73: One of the GPU tests randomly failed, I think this test is too strict, best to add a tolerance bound such as `rtol = sqrt(eps(elt))`.

mtfishman:
@kmp5VT this looks good to me.

The only remaining thing I can see to do is fix tensor contractions involving block sparse tensors with diagonal blocks. What's the status of that? I think we can fix that in a follow-up PR, I guess that can be fixed in a similar way (say by calling `denseblocks` and then moving the tensor to GPU).

kmp5VT:
> @kmp5VT this looks good to me.
> 
> The only remaining thing I can see to do is fix tensor contractions involving block sparse tensors with diagonal blocks. What's the status of that? I think we can fix that in a follow-up PR, I guess that can be fixed in a similar way (say by calling `denseblocks` and then moving the tensor to GPU).

@mtfishman Thats a good point. I hadn't checked the blocksparse implementation. I ran this code
```
using ITensorMPS: MPO, siteinds
using LinearAlgebra: tr
using Metal: mtl
s = siteinds("S=1/2", 2; conserve_qns=true)
o = MPO(s, "Id")
tr(mtl(o))
```
and it currently fails because of scalar indexing (because its fed into the cpu dense * diag code). I found that in `diagblocksparse` I can expose the block calls to contract which fixes the behavior

mtfishman:
Great, glad to see it was a simple fix, simpler than I was picturing.

Seems like the only thing left is to add a test for the block sparse case, after that is this good to go?

NDTensors/test/test_blocksparse.jl
```diff
@@ -284,39 +284,44 @@ using Test: @test, @test_throws, @testset
       A = dev(BlockSparseTensor{elt}([(2, 1), (1, 2)], [2, 2], [2, 2]))
       randn!(A)
       U, S, V = svd(A)
-      @test @allowscalar array(U) * array(S) * array(V)' ≈ array(A)
+      Asvd = contract(contract(U, (1, -2), S, (-2, 3)), (1, -2), V, (3, -2))
```
kmp5VT, line 287: @mtfishman Here U is a blocksparse tensor and S is a diagblocksparse tensor so this now tests the GPU BS * DiagBS. But I guess it doesn't check GPU BS * UniformDiag

mtfishman: I think it is better to keep these tests how they were, since then they are testing the SVD more in isolation and not testing both SVD and contract. Could you write devoted tests for block sparse (uniform) diag * dense contraction?

kmp5VT: Makes sense, I can do that. I found an issue with directly contracting block sparse with block sparse uniform diag when trying to make a test and reported it here https://github.com/ITensor/ITensors.jl/issues/1474.

mtfishman: Hopefully it is easy enough to find an example contraction that works and circumvents #1474, maybe that issue is particular to inner product-like contractions and a matrix multiplication-like contraction would work.

mtfishman: See https://github.com/ITensor/ITensors.jl/issues/1474#issuecomment-2138447085 and #1475.

mtfishman: @kmp5VT thanks for reporting issues you're finding with DiagBlockSparse contraction, for the sake of this PR I think we can just make a test based on one of the DiagBlockSparse contractions that works and worry about the ones that are failing in future PRs.

NDTensors/ext/NDTensorsGPUArraysCoreExt/contract.jl
```diff
@@ -0,0 +1,48 @@
+using Adapt: adapt
+using GPUArraysCore: AbstractGPUArray
+using NDTensors: NDTensors, DenseTensor, DiagTensor, contract!, dense, inds, Tensor
```
mtfishman, line 3: Please sort alphabetically with capitalized names grouped first (specifically, `Tensor` should be grouped with the other types).

mtfishman:
Besides the final comment, this looks good, thanks.

NDTensors/ext/NDTensorsGPUArraysCoreExt/contract.jl
kmp5VT: Oh thanks I wasn't thinking about types. It should be fixed now. Thanks!

PR #1454, opened 2024-05-20
[UnallocatedArrays] Fix UnallocatedArrays constructor

kmp5VT:
# Description

Theres an issue with a constructor line in UnallocatedArrays which can only be seen if the parameters are partially provided to construct the type. I added a test for these constructors and fixed the issue.


mtfishman:
Looks good, thanks!

PR #1479, opened 2024-05-30
[BlockSparseArrays] Fix eachindex(::BlockSparseArray) with dual axes

mtfishman:
This is a temporary hack to fix `eachindex(a::BlockSparseArray)` when `a` has a mixture of dual and non-dual axes.

This will be fixed in a better way once GradedAxes is rewritten using BlockArrays v1 but for now I'm implementing some temporary fixes for issues like this and other ones being tracked in #1336.

@ogauthe

### DIFF
--- a/NDTensors/src/lib/BlockSparseArrays/ext/BlockSparseArraysGradedAxesExt/test/runtests.jl
+++ b/NDTensors/src/lib/BlockSparseArrays/ext/BlockSparseArraysGradedAxesExt/test/runtests.jl
@@ -5,7 +5,6 @@ using BlockArrays: Block, blocksize
 using NDTensors.BlockSparseArrays: BlockSparseArray, block_nstored
 using NDTensors.GradedAxes: GradedAxes, GradedUnitRange, dual, gradedrange
 using NDTensors.LabelledNumbers: label
-## using NDTensors.Sectors: U1
 using NDTensors.SparseArrayInterface: nstored
 using NDTensors.TensorAlgebra: fusedims, splitdims
 using Random: randn!
--- a/NDTensors/src/lib/BlockSparseArrays/ext/BlockSparseArraysGradedAxesExt/test/runtests.jl
+++ b/NDTensors/src/lib/BlockSparseArrays/ext/BlockSparseArraysGradedAxesExt/test/runtests.jl
@@ -83,10 +83,9 @@ const elts = (Float32, Float64, Complex{Float32}, Complex{Float64})
     a[Block(1, 1)] = randn(size(a[Block(1, 1)]))
     a[Block(2, 2)] = randn(size(a[Block(2, 2)]))
     a_dense = Array(a)
-    if VERSION ≥ v"1.10"
-      for I in eachindex(a)
-        @test a[I] == a_dense[I]
-      end
+    @test eachindex(a) == CartesianIndices(size(a))
+    for I in eachindex(a)
+      @test a[I] == a_dense[I]
     end
   end
 end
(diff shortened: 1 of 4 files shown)


NDTensors/src/lib/BlockSparseArrays/ext/BlockSparseArraysGradedAxesExt/test/runtests.jl
```diff
@@ -3,9 +3,9 @@ using Compat: Returns
 using Test: @test, @testset, @test_broken
 using BlockArrays: Block, blocksize
 using NDTensors.BlockSparseArrays: BlockSparseArray, block_nstored
-using NDTensors.GradedAxes: GradedUnitRange, gradedrange
+using NDTensors.GradedAxes: GradedAxes, GradedUnitRange, dual, gradedrange
 using NDTensors.LabelledNumbers: label
-using NDTensors.Sectors: U1
+## using NDTensors.Sectors: U1
```
mtfishman, line 8:
```suggestion
```

NDTensors/src/lib/GradedAxes/src/unitrangedual.jl
```diff
@@ -48,6 +48,17 @@ end
...
+# This fixes an issue in `eachindex(::BlockSparseArray)`
+# when it has axes that are a mixture of non-dual
+# and dual axes.
+# TODO: This is a hack that won't be needed once we rewrite
+# GradedAxes using BlockArrays v1, delete this once it is
+# not needed anymore.
+Base.AbstractUnitRange{T}(a::UnitRangeDual{T}) where {T} = a
+function Base.AbstractUnitRange{T}(a::UnitRangeDual) where {T}
+  return AbstractUnitRange{T}(nondual(a))
+end
```
mtfishman, line 60:
```suggestion
if VERSION ≥ v"1.10"
  # This fixes an issue in `eachindex(::BlockSparseArray)`
  # when it has axes that are a mixture of non-dual
  # and dual axes.
  # TODO: This is a hack that won't be needed once we   rewrite
  # GradedAxes using BlockArrays v1, delete this once it is
  # not needed anymore.
  Base.AbstractUnitRange{T}(a::UnitRangeDual{T}) where {T} = a
  function Base.AbstractUnitRange{T}(a::UnitRangeDual) where {T}
    return AbstractUnitRange{T}(nondual(a))
  end
end
```

NDTensors/src/lib/BlockSparseArrays/ext/BlockSparseArraysGradedAxesExt/test/runtests.jl
```diff
@@ -70,5 +77,15 @@ const elts = (Float32, Float64, Complex{Float32}, Complex{Float64})
...
   end
+  @testset "dual axes" begin
+    r = gradedrange([U1(0) => 2, U1(1) => 2])
+    a = BlockSparseArray{elt}(dual(r), r)
+    a[Block(1, 1)] = randn(size(a[Block(1, 1)]))
+    a[Block(2, 2)] = randn(size(a[Block(2, 2)]))
+    a_dense = Array(a)
+    for I in eachindex(a)
+      @test a[I] == a_dense[I]
+    end
```
mtfishman, line 88:
```suggestion
    if VERSION ≥ v"1.10"
      for I in eachindex(a)
        @test a[I] == a_dense[I]
      end
    end
```

mtfishman:
I think this isn't a good fix since it is too hacky, I'll try overloading `eachindex(::BlockSparseArray)` directly instead.

mtfishman:
@ogauthe in the latest commit I added a temporary fix for printing BlockSparseArray with mixed dual and non-dual axes.

PR #1481, opened 2024-05-31
[BlockSparseArrays] Improve the design of block views

mtfishman:
This fixes a few issues related to block views listed in #1336. The strategy here is to change the design such that calls like `@view a[Block(1, 1)]` return something like `SubArray(a, Block(1, 1))`, while before this PR it returned the block of the `BlockSparseArray` directly, which causes problems when getting views of blocks that aren't stored. There are still issues to work out with that new design but they should be easy to fix.

To summarize, this fixes:

1. Initializing blocks with broadcasting, such as `a[Block(2, 2)] .= 1` or `a[Block(2, 2)] .= randn(2, 2)`.
2. Setting elements of blocks with `@view(a[Block(1, 1)])[1:1, 1:2] = rand(1, 2)` or `@views a[Block(1, 1)][1:1, 1:2] .= rand(1, 2)`.
3. Setting blocks with the syntax `a[Block(2), Block(2)] = randn(3, 3)` (before only `a[Block(2, 2)] = randn(3, 3)` worked).
4. Catching trying to set a block with an array of the wrong size, i.e. `a[Block(1, 1)] = randn(3, 3)` of `size(a[Block(1, 1)]) != (3, 3)`.


mtfishman:
@ogauthe merging this now. Please try it out and report any issues you come across.

emstoudenmire:
Nice to see this

Issue #1476, opened 2024-05-30
[NDTensors] Update tests to use TensorOperations v5 and cuTENSOR v2


kmp5VT:
**Description of bug**

Recently updates have been made to tensor operations to make it compatible with cuTENSOR v2.0. Updates are need for the unittests to reflect this version update.

kmp5VT:
This issue was fixed with #1483

PR #1483, opened 2024-06-04
[NDTensors] Use TensorOperations.jl v5 in tests

kmp5VT:
# Description
TensorOperations.jl just updated their code to work with cuTENSOR 2.0. The package is not yet registered but it is possible to use the latest version if we add the TensorOperations git repo.  This PR is in response to https://github.com/ITensor/ITensors.jl/issues/1476


mtfishman:
Thanks, I think it is better to wait for the new version to get registered so we can let the package manager install it automatically but this is a good starting point so we're ready for when it is registered.

kmp5VT:
Okay then I will close this PR and wait until there is an update. Thanks!

mtfishman:
We could keep this PR open, but design it so that it will work once the new version of TensorOperations.jl is registered.

kmp5VT:
@mtfishman I checked recently and we should be able to update cuTENSOR and TensorOperations to get this PR merged. I will look into it today

mtfishman:
Thanks!

PR #1484, opened 2024-06-04
[WIP] [BlockSparseArrays] Change behavior of non-blocked slicing

mtfishman:
Addresses one of the issues being tracked in #1336, based on a design discussed in https://github.com/JuliaArrays/BlockArrays.jl/issues/347.

To-do:
- [ ] Fix printing blocked views, for example `show(stdout, MIME("text/plain"), @view(a[[Block(2), Block(1)], [Block(2), Block(1)]]))`, and add a test.
- [x] Raise an issue in `BlockArrays` about `x = randn(2, 2); blocks(x)[1, 1]` not returning `x`, but instead returning a view object which is more challenging to work with (https://github.com/JuliaArrays/BlockArrays.jl/issues/403).
- [x] Get tests passing.
- [x] Add new tests.


mtfishman:
I've decided to go with a different design, closing in favor of #1486.

PR #1485, opened 2024-06-05
[NDTensorscuTENSORExt] Temporary fix for block sparse contractions 

kmp5VT:
# Description

Quick fix by copying all the viewed blocks and rewriting the block back to fix cuTENSOR problem.
# Checklist:

- [x] BlockSparse cutensor contractions work.
- [x] Check that the performance isn't too slow compared to cuBLAS implementation.
- [x] Avoid copying when it isn't needed, for example when the offset field of the CuArray is 0 (https://github.com/JuliaGPU/CUDA.jl/blob/v5.4.2/src/array.jl#L61).

### DIFF
--- a/NDTensors/ext/NDTensorscuTENSORExt/contract.jl
+++ b/NDTensors/ext/NDTensorscuTENSORExt/contract.jl
@@ -1,4 +1,4 @@
-using NDTensors: NDTensors, DenseTensor, array, data
+using NDTensors: NDTensors, DenseTensor, array
 using NDTensors.Expose: Exposed, unexpose
 using cuTENSOR: cuTENSOR, CuArray, CuTensor
 
--- a/NDTensors/ext/NDTensorscuTENSORExt/contract.jl
+++ b/NDTensors/ext/NDTensorscuTENSORExt/contract.jl
@@ -1,4 +1,4 @@
-using NDTensors: NDTensors, DenseTensor, array
+using NDTensors: NDTensors, DenseTensor, array, data
 using NDTensors.Expose: Exposed, unexpose
 using cuTENSOR: cuTENSOR, CuArray, CuTensor
 
@@ -12,9 +12,12 @@ function NDTensors.contract!(
   α::Number=one(Bool),
   β::Number=zero(Bool),
 )
-  cuR = CuTensor(array(unexpose(R)), collect(labelsR))
-  cuT1 = CuTensor(array(unexpose(T1)), collect(labelsT1))
-  cuT2 = CuTensor(array(unexpose(T2)), collect(labelsT2))
+  ## TODO for now a hack to get cuTENSOR working with blocksparse
+  R = unexpose(R)
+  cuR = CuTensor(copy(array(R)), collect(labelsR))
+  cuT1 = CuTensor(copy(array(unexpose(T1))), collect(labelsT1))
+  cuT2 = CuTensor(copy(array(unexpose(T2))), collect(labelsT2))
   cuTENSOR.mul!(cuR, cuT1, cuT2, α, β)
+  copyto!(data(R), cuR.data)
   return R
 end
--- a/NDTensors/ext/NDTensorscuTENSORExt/contract.jl
+++ b/NDTensors/ext/NDTensorscuTENSORExt/contract.jl
@@ -22,9 +22,7 @@ function NDTensors.contract!(
   cuT2 = CuTensor(arrayT2, collect(labelsT2))
   cuTENSOR.mul!(cuR, cuT1, cuT2, α, β)
   if !zoffR
-    ## use vec to flatten cuR.data which could be multidimensional but
-    ## tensor data is currently a vector
-    data(R) .= vec(cuR.data)
+    array(R) .= cuR.data
   end
   return R
 end
--- a/NDTensors/ext/NDTensorscuTENSORExt/contract.jl
+++ b/NDTensors/ext/NDTensorscuTENSORExt/contract.jl
@@ -2,69 +2,29 @@ using NDTensors: NDTensors, DenseTensor, array, data
 using NDTensors.Expose: Exposed, unexpose
 using cuTENSOR: cuTENSOR, CuArray, CuTensor
 
-struct CuArrayOffset{N} end
-function CuArrayOffset(V)
-  return CuArrayOffset{V}()
-end
-
 function NDTensors.contract!(
-  R::Exposed{<:CuArray,<:DenseTensor},
+  exposedR::Exposed{<:CuArray,<:DenseTensor},
   labelsR,
-  T1::Exposed{<:CuArray,<:DenseTensor},
+  exposedT1::Exposed{<:CuArray,<:DenseTensor},
   labelsT1,
-  T2::Exposed{<:CuArray,<:DenseTensor},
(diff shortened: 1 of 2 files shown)


NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -12,9 +12,12 @@ function NDTensors.contract!(
...
-  cuR = CuTensor(array(unexpose(R)), collect(labelsR))
-  cuT1 = CuTensor(array(unexpose(T1)), collect(labelsT1))
-  cuT2 = CuTensor(array(unexpose(T2)), collect(labelsT2))
+  ## TODO for now a hack to get cuTENSOR working with blocksparse
+  R = unexpose(R)
+  cuR = CuTensor(copy(array(R)), collect(labelsR))
+  cuT1 = CuTensor(copy(array(unexpose(T1))), collect(labelsT1))
+  cuT2 = CuTensor(copy(array(unexpose(T2))), collect(labelsT2))
   cuTENSOR.mul!(cuR, cuT1, cuT2, α, β)
+  copyto!(data(R), cuR.data)
```
mtfishman, line 21: Can we use:
```julia
copyto!(array(R), cuR.data)
```
here? It would be nice to avoid using `data(...)` which is pretty low level. Also we could use broadcasting syntax:
```julia
array(R) .= cuR.data
```
which is a bit more compact.

Finally, is there a more official API for converting a `CuTensor` to a `CuArray` besides `t.data`? Maybe `convert(CuArray, t)`? It seems like we shouldn't be using `t.data` since that seems like an internal interface but maybe there isn't anything better.

kmp5VT: That is a good recommendation. This primary push was a validation that all the unittests would work as I was seeing on my workstation. I will update the code to use a more official function for grabbing the cuTENSOR data. Thanks!

mtfishman: Makes sense, thanks for letting me know this is for testing at first. There may not be an official function for unwrapping the `CuArray` besides `t.data`, but worth looking into (and maybe making a PR with a definition like `Base.convert(::Type{CuArray}, t::CuTensor) = t.data`).

kmp5VT: I updated the code as we discussed and am closing this comment.

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -1,7 +1,12 @@
-using NDTensors: NDTensors, DenseTensor, array
+using NDTensors: NDTensors, DenseTensor, array, data
 using NDTensors.Expose: Exposed, unexpose
 using cuTENSOR: cuTENSOR, CuArray, CuTensor
 
+struct CuArrayOffset{N} end
+function CuArrayOffset(V)
```
kmp5VT, line 6: @mtfishman  I made this struct to be a trait for `contract!` launch so we don't always copy the `CuArrays`. In the copy based `contract!` function I copy all of the `CuArrays` but I could see it where I create multiple functions where I only copy the necessary `CuArrays`. Would this more refined approach be preferred or is it okay to copy them all for now.

mtfishman:
I think using a single function with if-statements would be better here:
```julia
arrayR = array(R)
is_R_zero_offset = iszero(arrayR.offset)
if is_R_zero_offset
  arrayR = copy(arrayR)
end
cuR = CuTensor(arrayR, collect(labelsR))

# ...

if is_R_zero_offset
  arrayR .= cuR.data
end
```
Then we can remove the if-statements once https://github.com/JuliaGPU/CUDA.jl/issues/2407 is fixed.

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -12,9 +17,54 @@ function NDTensors.contract!(
   α::Number=one(Bool),
   β::Number=zero(Bool),
 )
-  cuR = CuTensor(array(unexpose(R)), collect(labelsR))
-  cuT1 = CuTensor(array(unexpose(T1)), collect(labelsT1))
-  cuT2 = CuTensor(array(unexpose(T2)), collect(labelsT2))
+  R, T1, T2 = unexpose.((R, T1, T2))
```
mtfishman, line 20: To make this code clearer let's call the arguments to the function `exposedR`, `exposedT1`, `exposedT2` and then change this line to:
```suggestion
  R, T1, T2 = unexpose.((exposedR, exposedT1, exposedT2))
```

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -1,20 +1,30 @@
...
+  arrayT1 = iszero(array(T1).offset) ? array(T1) : copy(array(T1))
+  arrayT2 = iszero(array(T2).offset) ? array(T2) : copy(array(T2))
+  cuR = CuTensor(arrayR, collect(labelsR))
+  cuT1 = CuTensor(arrayT1, collect(labelsT1))
+  cuT2 = CuTensor(arrayT2, collect(labelsT2))
   cuTENSOR.mul!(cuR, cuT1, cuT2, α, β)
+  if !zoffR
+    ## use vec to flatten cuR.data which could be multidimensional but
+    ## tensor data is currently a vector
+    data(R) .= vec(cuR.data)
```
mtfishman, line 27: What about:
```suggestion
    array(R) .= cuR.data
```
?

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
kmp5VT: This has been updated. Thanks!

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
kmp5VT: Oh good idea, I am running now to see if that throws any errors. I had one issue in `DiagBlockSparse` before but we will see. Thanks!

kmp5VT: That worked, thanks again!

mtfishman:
Looks good, thanks. Is this ready to merge once tests pass?

kmp5VT:
> Looks good, thanks. Is this ready to merge once tests pass?

Yes ready to merge when tests pass!

mtfishman:
Glad to see the timings are comparable, I'll be curious to see what the timings are once we can remove the calls to `copy`. Pretty soon we'll have the entire table in https://itensor.github.io/ITensors.jl/dev/RunningOnGPUs.html#GPU-backends supported!

Have you tried profiling or timing (https://cuda.juliagpu.org/stable/development/profiling) to get an estimate of how much time the calls to `copy` are taking up?

kmp5VT:
> Have you tried profiling or timing (https://cuda.juliagpu.org/stable/development/profiling) to get an estimate of how much time the calls to `copy` are taking up?
No I haven't yet don't any indepth timings of the function just a quick dmrg calculation to see if it was significantly effecting performance in a known case where GPU finds improvement over CPU. 
I will run more and get back on timings/performance

Issue #1482, opened 2024-06-03
[NDTensors] [BUG] Scalar indexing issue when broadcasting tensors with `Diag` storage on GPU


mtfishman:
See: https://itensor.discourse.group/t/broadcast-sqrt-operation-on-gpu/1777

PR #1497, opened 2024-06-13
[NDTensors] Fix scalar indexing issue for Diag broadcast on GPU

kmp5VT:
# Description

Fixes https://github.com/ITensor/ITensors.jl/issues/1482.

# Checklist:

- [x] Test for new `NDTensors.map_diag!` and `NDTensors.map_diag` functions.
- [x] Fix Jenkins CI (see comment below for a suggestion).
- [x] Rename `NDTensorsGPUArraysCoreExt/diag.jl` to `NDTensorsGPUArraysCoreExt/blocksparsetensor.jl`.
- [x] The code can correctly output broadcast functions for Diag datatype using GPU backends
- [x] Create unittests in `Diag` and `DiagBlockSparse`


NDTensors/ext/NDTensorsGPUArraysCoreExt/permutedims.jl
```diff
@@ -0,0 +1,15 @@
...
+function NDTensors.permutedims!(
+  Rexposed::Exposed{<:AbstractGPUArray,<:DiagTensor},
+  texposed::Exposed{<:AbstractGPUArray,<:DiagTensor},
+  perm::NTuple{N,Int},
+  f::Function=(r, t) -> t,
+) where {N}
+  R = unexpose(Rexposed)
+  t = unexpose(texposed)
+
+  array(R) .= f.(array(R), array(t))
```
kmp5VT, line 13: No scalar indexing is thrown but also the data does not properly get moved to R and the output is numerically incorrect. Working on that

kmp5VT: Using data fixed this issue

mtfishman:
Thanks for looking into this @kmp5VT. It seems better to me to just change the definition of:
```julia
function permutedims!(
  R::DiagTensor{<:Number,N},
  T::DiagTensor{<:Number,N},
  perm::NTuple{N,Int},
  f::Function=(r, t) -> t,
) where {N}
  # ...
end
```
to the one you've defined for GPU so we don't have to support two implementations of that function.

Also I see for loops/scalar indexing in a number of other functions:
```julia
function diag(tensor::DiagTensor)
  # ...
end

# I see this one is using some ad-hoc dispatch
# for GPU but it seems like we should either use
# `expose` or come up with a better code pattern,
# say use a slicing operation.
function dense(::Type{<:Array}, T::DiagTensor)
  # ...
end

function permutedims!(
  R::DenseTensor{ElR,N}, T::DiagTensor{ElT,N}, perm::NTuple{N,Int}, f::Function=(r, t) -> t
) where {ElR,ElT,N}
  # ...
end
```
Can you look into those as well?

mtfishman:
One recommendation I have would be to look into the code patterns I came up with in NDTensors.DiagonalArrays for more conveniently working with diagonal values. For example: https://github.com/ITensor/ITensors.jl/blob/v0.6.11/NDTensors/src/lib/DiagonalArrays/src/diaginterface/diaginterface.jl.

As a demonstration:
```julia
using Compat: allequal

function diaglength(a::AbstractArray)
  return minimum(size(a))
end
diaglength(a::AbstractArray{<:Any,0}) = 1

function diagstride(a::AbstractArray)
  s = 1
  p = 1
  for i in 1:(ndims(a) - 1)
    p *= size(a, i)
    s += p
  end
  return s
end

function diagindices(a::AbstractArray)
  maxdiag = LinearIndices(a)[CartesianIndex(ntuple(Returns(diaglength(a)), ndims(a)))]
  return 1:diagstride(a):maxdiag
end
function diagindices(a::AbstractArray{<:Any,0})
  return Base.OneTo(1)
end

function diagview(a::AbstractArray)
  return @view a[diagindices(a)]
end

using LinearAlgebra: Diagonal
function diagview(a::Diagonal)
  return a.diag
end

x = Diagonal(randn(4))
y = randn(4, 4)

using Metal: mtl
x = mtl(x)
y = mtl(y)

diagview(y) .= diagview(x)
```
That appears to avoid scalar indexing but I'm not sure if it is actually fast on GPU. But I think if `diagview` was defined in a similar way for `Tensor` objects, such as `DiagTensor`, `DenseTensor`, etc. we could use a code pattern like that to implement most of the functions that currently use scalar indexing/explicit for loops that I listed above in a simple way, just in terms of broadcasting over `diagview`.

NDTensors/test/test_diag.jl
```diff
@@ -31,22 +31,30 @@ using LinearAlgebra: dot
...
     @allowscalar begin
       @test Da == NDTensors.LinearAlgebra.diagm(0 => vr)
       @test Da == NDTensors.LinearAlgebra.diagm(0 => vr)
 
-      ## TODO Currently this permutedims requires scalar indexing on GPU.
-      Da = permutedims(D, (2, 1))
       @test Da == D
     end
 
+    if (dev == NDTensors.mtl && elt != ComplexF32)
```
kmp5VT, line 49: I reported a bug in metal that causes this code to fail.
https://github.com/JuliaGPU/Metal.jl/issues/364

mtfishman: Could you add a link to that issue as a comment in the code?

kmp5VT: Yes I added that comment. Thanks!

mtfishman:
@kmp5VT I think it could simplify the code to define overloads of `DiagonalArrays.diagview` for `DenseTensor` and `DiagTensor`.

NDTensors/src/diag/diagtensor.jl
```diff
@@ -33,6 +31,19 @@ function Array(T::DiagTensor{ElT,N}) where {ElT,N}
   return Array{ElT,N}(T)
 end
 
+function DiagonalArrays.diagview(T::NonuniformDiagTensor)
+  return data(T)
```
kmp5VT, line 35: @mtfishman Should these go somewhere else like a `diagview.jl` file?

NDTensors/src/linearalgebra/linearalgebra.jl
```diff
@@ -369,7 +369,6 @@ matrix is unique. Returns a tuple (Q,R).
 function qr_positive(M::AbstractMatrix)
   sparseQ, R = qr(M)
   Q = convert(typeof(R), sparseQ)
-  nc = size(Q, 2)
```
kmp5VT, line 372: @mtfishman Miles asked me to remove this line, it is unused

NDTensors/src/diag/diagtensor.jl
mtfishman: I don't think that is necessary.

NDTensors/src/diag/diagtensor.jl
```diff
@@ -33,6 +31,19 @@ function Array(T::DiagTensor{ElT,N}) where {ElT,N}
...
+end
+
+function DiagonalArrays.diagview(T::UniformDiagTensor)
+  return array(T)
```
mtfishman, line 39: What does this output?

NDTensors/src/diag/diagtensor.jl
```diff
@@ -193,9 +202,8 @@ end
 function permutedims!(
   R::DenseTensor{ElR,N}, T::DiagTensor{ElT,N}, perm::NTuple{N,Int}, f::Function=(r, t) -> t
 ) where {ElR,ElT,N}
-  for i in 1:diaglength(T)
-    @inbounds setdiagindex!(R, f(getdiagindex(R, i), getdiagindex(T, i)), i)
-  end
+  rview = diagview(array(R))
+  rview .= f.(rview, diagview(T))
```
mtfishman, line 206:
```suggestion
  diagview(R) .= f.(diagview(R), diagview(T))
```

NDTensors/src/diag/diagtensor.jl
```diff
@@ -145,16 +156,14 @@ function permutedims!(
   f::Function=(r, t) -> t,
 ) where {N}
   # TODO: check that inds(R)==permute(inds(T),perm)?
-  for i in 1:diaglength(R)
-    @inbounds setdiagindex!(R, f(getdiagindex(R, i), getdiagindex(T, i)), i)
-  end
+  data(R) .= f.(diagview(R), diagview(T))
```
mtfishman, line 159:
```suggestion
  diagview(R) .= f.(diagview(R), diagview(T))
```

NDTensors/src/diag/diagtensor.jl
kmp5VT: Oh this is a typo I meant to put
```
return diagview(array(T))
```
To force it to have the correct length. But I am not sure if that is necessary.

mtfishman: We could consider returning `Fill(getdiagindex(T, 1), diaglength(T))` here if that mostly "just works" in operations where it is used.

NDTensors/src/diag/diagtensor.jl
kmp5VT: Updated, thanks!

NDTensors/src/diag/diagtensor.jl
kmp5VT: For some reason the first time I wrote that as `diagview` the code was not working properly. But now it is, perhaps some old bit of code was in my repl when I was testing. Updated now, thanks!

NDTensors/src/diag/diagtensor.jl
kmp5VT: It doesn't look like like any of our tests were hitting this function so I think either definition would be sufficient, just let me know which you prefer!

mtfishman: `diagview(array(T))` would of course technically work but is a pretty inefficient way of doing it so it would be best to avoid that if possible, the ideal thing in terms of memory would be `Fill` but if that turns out to be too tricky to get working then we could use `fill(getdiagindex(T, 1), diaglength(T))`. Again that isn't ideal and partially ruins the point of using a uniform diagonal tensor to begin with.

mtfishman: If it isn't required to get tests passing then we can just not define it for the time being.

NDTensors/src/diag/diagtensor.jl
```diff
@@ -33,6 +31,19 @@ function Array(T::DiagTensor{ElT,N}) where {ElT,N}
...
 
+function DiagonalArrays.diagview(T::NonuniformDiagTensor)
+  return data(T)
+end
+
+function DiagonalArrays.diagview(T::UniformDiagTensor)
+  return fill(getdiagindex(T, 1), diaglength(T))
+end
+
+## Should this go in dense.jl or here since its related to diag?
```
mtfishman: Yes, I think this should go in `densetensor.jl`.

NDTensors/src/diag/diagtensor.jl
kmp5VT: Makes sense I removed it and we will see what happens when tests run. Thanks!

NDTensors/src/diag/diagtensor.jl
```diff
@@ -193,9 +197,7 @@ end
 function permutedims!(
   R::DenseTensor{ElR,N}, T::DiagTensor{ElT,N}, perm::NTuple{N,Int}, f::Function=(r, t) -> t
 ) where {ElR,ElT,N}
-  for i in 1:diaglength(T)
-    @inbounds setdiagindex!(R, f(getdiagindex(R, i), getdiagindex(T, i)), i)
-  end
+  diagview(array(R)) .= f.(diagview(array(R)), diagview(T))
```
mtfishman, line 200:
```suggestion
  diagview(R) .= f.(diagview(R), diagview(T))
```

kmp5VT: Updated thanks

NDTensors/src/diag/diagtensor.jl
kmp5VT: I moved it to densetensor.jl, thanks!

NDTensors/src/tensor/tensor.jl
```diff
@@ -361,16 +361,15 @@ function getdiagindex(T::Tensor{<:Number,N}, ind::Int) where {N}
...
 # block sparse vector instead of dense.
 function diag(tensor::Tensor)
   ## d = NDTensors.similar(T, ElT, (diaglength(T),))
   tensordiag = NDTensors.similar(
     dense(typeof(tensor)), eltype(tensor), (diaglength(tensor),)
   )
-  for n in 1:diaglength(tensor)
-    tensordiag[n] = tensor[n, n]
-  end
+  data(tensordiag) .= diagview(tensor)
```
mtfishman, line 372: Can this just be:
```suggestion
  tensordiag .= diagview(tensor)
```
? If not I think `array(tensordiag) .= diagview(tensor)` would be better, it is best to avoid `data(...)` whenever there is a reasonable alternative since `data(...)` is meant to be a lower level function.

kmp5VT: This works, I will keep that in mind for data, thanks!

NDTensors/src/diag/diagtensor.jl
kmp5VT: @mtfishman The remaining tests that fail in ITensor are attempting to call `diagview` on a `UnitaryDiagTensor` so I will implement the `fill` version which you suggested here

kmp5VT:
@mtfishman The code right now fails during an ITensor tests where we are calling `diag` on a `BlockSparseTensor`. To get the code to work properly `diagview` needs to be defined for `BlockSparseTensor` I am working on that now and trying to add a test to `test_blocksparse.jl`

NDTensors/test/test_diag.jl
mtfishman: The logic of this if-statement doesn't seem right to me. Are you sure this will run on CPU and also on other GPU backends?

mtfishman: From what I can tell it will only run on Metal if the element type is `Float32`, but not on other device backends.

mtfishman: Maybe it should be:
```julia
if !(dev == NDTensors.mtl && elt == ComplexF32)
  # ...
end
```

kmp5VT: Oh good catch thank you for fixing that logic. I have only been testing on my mac and didn't catch that. Fixed now!

mtfishman:
I find these definitions to be pretty odd: https://github.com/kmp5VT/ITensors.jl/blob/7e2ce770ff9308f353265dbddb28b8bd2d1c4192/NDTensors/src/diag/diagtensor.jl#L114-L131.

With the change in this PR can we rewrite that as something like:
```julia
function dense(T::DiagTensor)
  R = zeros(dense(typeof(T)), inds(T))
  diagview(R) .= diagview(T)
  return R  
end
```
?

src/tensor_operations/matrix_algebra.jl
```diff
@@ -86,15 +88,7 @@ function exp(A::ITensor; kwargs...)
...
 end
 map_diag(f::Function, it::ITensor) = map_diag!(f, copy(it), it)
-
-function map_diag!(f::Function, t_destination::Tensor, t_source::Tensor)
-  for i in 1:diaglength(t_destination)
-    NDTensors.setdiagindex!(t_destination, f(NDTensors.getdiagindex(t_source, i)), i)
-  end
-  return t_destination
-end
-map_diag(f::Function, t::Tensor) = map_diag!(f, copy(t), t)
```
mtfishman, line 100: In order to make CI happy and also for a smoother upgrade path for users, we could leave ITensors.jl as-is in this PR (i.e. don't remove `ITensors.map_diag!` and don't use `NDTensors.map_diag!` in ITensors.jl yet). Then we can make a follow-up PR where we make these change to ITensors.jl, after a new version of NDTensors.jl is registered that defines `NDTensors.map_diag!`.

kmp5VT: Okay i added these definitions back and removed the `using NDTensor: map_diag!` from this file for now

kmp5VT:
> I find these definitions to be pretty odd: https://github.com/kmp5VT/ITensors.jl/blob/7e2ce770ff9308f353265dbddb28b8bd2d1c4192/NDTensors/src/diag/diagtensor.jl#L114-L131.
> 
> Using the new functionality in this PR can we rewrite that as something like:
> 
> ```julia
> function dense(T::DiagTensor)
>   R = zeros(dense(typeof(T)), inds(T))
>   diagview(R) .= diagview(T)
>   return R  
> end
> ```
> 
> i.e. get rid of the dispatch on the unwrapped array type?

I think it is a good idea to remake the function like you are suggesting. One previous point about these functions is that `UniformDiag` storage types always convert to `Array` which could cause issues with GPU code. Right now we have that fixed in other places. But I think it would be a good idea to create a function like
```julia
function dense(T::DiagTensor, ArrayT::Type{<:AbstractArray})
  R = adapt(ArrayT, zeros(dense(typeof(T)), inds(T))
  diagview(R) .= diagview(T)
  return R
end
```

mtfishman:
What if in those places in the code where we were using that more complicated version of `dense`, instead we called `adapt` explicitly, i.e. change calls like `dense(array_type, T)` to `adapt(array_type, dense(T))`?

kmp5VT:
@mtfishman Actually we don't need to use adapt, the code regardless of the datatype
```julia
julia> using NDTensors, Metal
julia> A = tensor(Diag(3,3), (3,3))
julia> dense(A)
Dim 1: 3
Dim 2: 3
Dense{Int64, Vector{Int64}}
 3×3
 3  0  0
 0  3  0
 0  0  3
jullia> dense(mtl(A))
Dim 1: 3
Dim 2: 3
Dense{Int64, MtlVector{Int64, Private}}
 3×3
 3  0  0
 0  3  0
 0  0  3
 ```

**Comment by mtfishman** [comment]
@kmp5VT could you try adding `Pkg.Registry.update(); Pkg.update();` to the Jenkins script, like I did in https://github.com/ITensor/ITensorGPU.jl/pull/9? That may solve the test failures, since it will force Julia to update the registry, I think for some reason it is just seeing an old registry and therefore only seeing old package versions.

File: NDTensors/ext/NDTensorsGPUArraysCoreExt/diag.jl
Comment (mtfishman) [inline NDTensors/ext/NDTensorsGPUArraysCoreExt/diag.jl:1]: I think this file should be named `blocksparsetensor.jl`, since these are definitions particular to the `BlockSparseTensor` type, similar to how we name and organize files in NDTensors.jl.

**Comment by mtfishman** [comment]
Besides the last two comments I left, this looks good to me.

I think with this change (and also https://github.com/JuliaGPU/Metal.jl/pull/374), we are probably getting very close to having every NDTensors/ITensors operation working on GPU, which is a great step to get to! I'm sure a few lingering issues will come up, and there will be continued work improving performance (particularly for block sparse operations), but it definitely seems like the support for GPU operations is quite good now, assuming the lack of user reports of issues is a good indicator of that.

File: NDTensors/ext/NDTensorsGPUArraysCoreExt/diag.jl
Comment (kmp5VT) [reply NDTensors/ext/NDTensorsGPUArraysCoreExt/diag.jl:1]: Ah that makes sense. I have renamed the file. And I am adding the comment above to jenkins to see if that will help. Thanks!

**Comment by mtfishman** [comment]
Not sure why the `ITensorGaussianMPS` downstream test timed out (maybe just a random failure in Github Actions) but it looks like that change fixed the Jenkins tests.

**Comment by mtfishman** [comment]
Can you add a test for `NDTensors.map_diag!`?

File: NDTensors/test/test_blocksparse.jl [hunk 12]
```diff
@@ -52,6 +53,8 @@ using Test: @test, @test_throws, @testset
     @test isblocknz(A, (1, 2))
     @test !isblocknz(A, (1, 1))
     @test !isblocknz(A, (2, 2))
+    dA = diag(A)
+    @allowscalar dA ≈ diag(dense(A))
```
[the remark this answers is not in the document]
kmp5VT: Thanks I added this test

NDTensors/test/test_blocksparse.jl
```diff
@@ -104,6 +108,10 @@ using Test: @test, @test_throws, @testset
     @allowscalar for I in eachindex(C)
       @test C[I] == A[I] + B[I]
     end
+    Cp = NDTensors.map_diag((i -> i * 2), C)
+    @allowscalar for i in 1:diaglength(Cp)
+      @test Cp[i, i] == 2.0 * C[i, i]
+    end
```
mtfishman, line 114:
```suggestion
    Cp = NDTensors.map_diag(i -> 2 * i, C)
    @allowscalar for i in 1:diaglength(Cp)
      @test Cp[i, i] == 2 * C[i, i]
    end
```

NDTensors/test/test_diag.jl
```diff
@@ -34,19 +34,39 @@ using LinearAlgebra: dot
...
+      @allowscalar begin
+        for i in 1:diaglength(S)
+          @test S[i, i] ≈ sqrt(D[i, i])
+        end
+      end
+    end
+    S = NDTensors.map_diag((i -> i * 2), dev(D))
+    @allowscalar for i in 1:diaglength(S)
+      @test S[i, i] == 2.0 * D[i, i]
+    end
```
mtfishman, line 68:
```suggestion
    S = NDTensors.map_diag(i -> 2 * i, dev(D))
    @allowscalar for i in 1:diaglength(S)
      @test S[i, i] == 2 * D[i, i]
    end
```

PR #1505, opened 2024-06-20
[NDTensorsMappedArraysExt] Support for using MappedArrays as data of ITensors

mtfishman:
To-do:
- [x] Add tests.


mtfishman:
@kmp5VT you can see I've decided to make generic definitions:
```julia
position(type::Type{<:AbstractArray}, ::typeof(eltype)) = Position(1)
position(type::Type{<:AbstractArray}, ::typeof(ndims)) = Position(2)
```
which we've discussed previously. When we discussed that last we decided to hold off on those generic definitions since they may be incorrect for certain `AbstractArray`s, but I think the convenience of defining them so they don't have to be defined for every `AbstractArray` subtype outways that issue, and unconventional `AbstractArray`s can make custom definitions of them as needed.

PR #1508, opened 2024-06-21
[NDTensors] JLArrays Extension

kmp5VT:
# Description

Introduce the JLArrays extension to test GPU for ITensors

# Checklist:

- [x] Introduce JLArrays and all NDTensors tests pass with array backend
- [ ] Potentially move function from GPU array specific libraries to GPUArraysCoreExt
- [ ] Update ITensors suite for JLArray to test on GPU adjacent array on CPU


kmp5VT:
There is an issue in JLArrays, there is a missing `resize!` function

NDTensors/ext/NDTEnsorsJLArraysExt/NDTensorsJLArraysExt.jl
mtfishman, line 1: The directory is spelled wrong, it should be `NDTensorsJLArraysExt`, not `NDTEnsorsJLArraysExt`.

NDTensors/ext/NDTEnsorsJLArraysExt/set_types.jl
```diff
@@ -0,0 +1,10 @@
+# TypeParameterAccessors definitions
+using NDTensors.TypeParameterAccessors: TypeParameterAccessors, Position
+using NDTensors.GPUArraysCoreExtensions: storagemode
+using JLArrays: JLArrays, JLArray
+
+function TypeParameterAccessors.default_type_parameters(::Type{<:JLArray})
+  return (Float64, 1)
+end
+TypeParameterAccessors.position(::Type{<:JLArray}, ::typeof(eltype)) = Position(1)
+TypeParameterAccessors.position(::Type{<:JLArray}, ::typeof(ndims)) = Position(2)
```
mtfishman, line 10: These shouldn't be needed since I made generic definitions in #1505.

NDTensors/Project.toml
```diff
@@ -47,6 +48,7 @@ NDTensorsAMDGPUExt = ["AMDGPU", "GPUArraysCore"]
 NDTensorsCUDAExt = ["CUDA", "GPUArraysCore"]
 NDTensorsGPUArraysCoreExt = "GPUArraysCore"
 NDTensorsHDF5Ext = "HDF5"
+NDTensorsJLArraysExt = ["JLArrays", "GPUArraysCore"]
```
mtfishman, line 51:
```suggestion
NDTensorsJLArraysExt = ["GPUArraysCore", "JLArrays"]
```
to alphabetize.

NDTensors/ext/NDTEnsorsJLArraysExt/set_types.jl
```diff
@@ -0,0 +1,10 @@
+# TypeParameterAccessors definitions
+using NDTensors.TypeParameterAccessors: TypeParameterAccessors, Position
+using NDTensors.GPUArraysCoreExtensions: storagemode
+using JLArrays: JLArrays, JLArray
+
+function TypeParameterAccessors.default_type_parameters(::Type{<:JLArray})
+  return (Float64, 1)
+end
```
mtfishman, line 8: We could consider making this a generic `AbstractArray` definition.

NDTensors/ext/NDTEnsorsJLArraysExt/NDTensorsJLArraysExt.jl
kmp5VT: Thanks I thought something was off but couldn't track it down. Fixed now

NDTensors/ext/NDTEnsorsJLArraysExt/set_types.jl
kmp5VT: Good point, I removed these definitions for all GPU backends. Thanks!

NDTensors/ext/NDTEnsorsJLArraysExt/set_types.jl
kmp5VT: This is something I was thinking about. I think one issue with that is wrappers don't always have ndims second for example 
```julia
julia> a = Matrix{Float32}(undef, (2,2))
julia> typeof(transpose(a))
LinearAlgebra.Transpose{Float32, Matrix{Float32}}
```

mtfishman: Right, so then for those types we would require making a custom overload of `TypeParameterAccessors.default_type_parameters`, which would be required anyway.

NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -23,6 +23,10 @@ if "cutensor" in ARGS || "all" in ARGS
   Pkg.add("cuTENSOR")
   using CUDA, cuTENSOR
 end
+if "jlarrays" in ARGS || "all" in ARGS
+  Pkg.add("JLArrays")
+  using JLArrays
+end
```
mtfishman, line 29: I don't think there is a reason to not run the tests for JLArrays since they run on CPU anyway. So I think it should just be added as a normal test dependency, and inserted into the device list by default.

NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -1,5 +1,7 @@
-using NDTensors: NDTensors
 using Pkg: Pkg
+using JLArrays
```
mtfishman, line 2:
```suggestion
using JLArrays: JLArrays
```

kmp5VT: Updated, thanks!

NDTensors/ext/NDTensorsJLArraysExt/NDTensorsJLArraysExt.jl
```diff
@@ -0,0 +1,8 @@
+module NDTensorsJLArraysExt
+include("copyto.jl")
```
kmp5VT, line 2: I mostly copied these from CUDA definitions. I am working to determine which functions are actually necessary.

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: So I am seeing an issue with the GPU tests when I have `JLArrays` is always in the `Project.toml`. To fix the problem I `isempty(ARGS) || "base" in ARGS` so this would run into a problem if the test args are `["base","metal"]` but right now we aren't testing in that way. Though if you would prefer, I can just include `JLArrays` if no GPUs are being tested, i.e. only allow `JLArrays` if `isempty(ARGS)`. Let me know, thanks!

mtfishman: What's the issue you see? It would be best to try to address that issue and have `JLArrays` as a dependency in `Project.toml` and run by default in the tests as we planned, so I would prefer to try to aim for that rather than work around an issue with a more convoluted solution.

kmp5VT: The issue is that GPUArrays does not compile during `Pkg.test`
```
ERROR: LoadError: UndefVarError: `SimplifyCFGPassOptions` not defined
Stacktrace:
 [1] top-level scope
   @ ~/.julia/packages/GPUCompiler/nWT2N/src/optim.jl:57
 [2] include(mod::Module, _path::String)
   @ Base ./Base.jl:495
 [3] include(x::String)
   @ GPUCompiler ~/.julia/packages/GPUCompiler/nWT2N/src/GPUCompiler.jl:1
 [4] top-level scope
   @ ~/.julia/packages/GPUCompiler/nWT2N/src/GPUCompiler.jl:37
 [5] include
   @ Base ./Base.jl:495 [inlined]
 [6] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::String)
   @ Base ./loading.jl:2216
 [7] top-level scope
   @ stdin:3
in expression starting at /home/jenkins/workspace/ITensor_ITensors.jl_PR-1508@tmp/.julia/packages/GPUCompiler/nWT2N/src/optim.jl:56
in expression starting at /home/jenkins/workspace/ITensor_ITensors.jl_PR-1508@tmp/.julia/packages/GPUCompiler/nWT2N/src/GPUCompiler.jl:1
in expression starting at stdin:3
ERROR: LoadError: Failed to precompile GPUCompiler [61eb1bfa-7361-4325-ad38-22787b887f55] to "/home/jenkins/workspace/ITensor_ITensors.jl_PR-1508@tmp/.julia/compiled/v1.10/GPUCompiler/jl_xlhZZG".
```
However, I think I was able to fix the problem by moving `using JLArrays` to after the GPU `using` statements.

kmp5VT:
For some reason I am having an issue with `SparseArrays` with Julia 1.6. All 1.6 tests fail with a compat issue
```julia
   Resolving package versions...
Downstream tests for ITensor DMRG: Error During Test at /Users/kpierce/.julia/packages/SafeTestsets/raUNr/src/SafeTestsets.jl:30
  LoadError: LoadError: LoadError: LoadError: Unsatisfiable requirements detected for package SparseArrays [2f01184e]:
   SparseArrays [2f01184e] log:
   ├─possible versions are: 0.0.0 or uninstalled
   ├─SparseArrays [2f01184e] is fixed to version 0.0.0
   └─restricted to versions 1.6.0-1 by NDTensors [23ae76d9] — no versions left
     └─NDTensors [23ae76d9] log:
       ├─possible versions are: 0.3.39 or uninstalled
       ├─restricted to versions 0.3.34-0.3 by ITensors [9136182c], leaving only versions 0.3.39
       │ └─ITensors [9136182c] log:
       │   ├─possible versions are: 0.6.16 or uninstalled
       │   └─ITensors [9136182c] is fixed to version 0.6.16
       └─NDTensors [23ae76d9] is fixed to version 0.3.39
```

mtfishman:
> For some reason I am having an issue with `SparseArrays` with Julia 1.6. All 1.6 tests fail with a compat issue

That's strange, do you see that locally as well?

Since `CUDA.jl` doesn't support Julia 1.6 now anyway, maybe if there isn't an obvious solution to that we could just stop testing against Julia 1.6 in the Jenkins workflow and not worry about Julia 1.6 support for the GPU backends.

The latest version of Julia compatible with CUDA.jl v5 is Julia 1.8, what if we test against Julia 1.8 instead of Julia 1.6 in the Jenkins tests? If that works, I would vote for just doing that.

kmp5VT:
> > For some reason I am having an issue with `SparseArrays` with Julia 1.6. All 1.6 tests fail with a compat issue
> 
> That's strange, do you see that locally as well?

 Yes I was seeing this locally on CPU and GPU. I am not sure why NDTensors tests work on CI with version 1.6. I bumped the version for the CUDA test to 1.8

kmp5VT:
Also JLArrays does not have `JLMatrix` defined in julia v1.6
```julia
julia> versioninfo()
Julia Version 1.6.7
Commit 3b76b25b64 (2022-07-19 15:11 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.4.0)
  CPU: Apple M1 Max
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, westmere)

julia> using JLArrays

julia> JLArray.JLMatrix
ERROR: type UnionAll has no field JLMatrix
Stacktrace:
 [1] getproperty(x::Type, f::Symbol)
   @ Base ./Base.jl:28
 [2] top-level scope
   @ REPL[7]:1
```

kmp5VT:
And `JLArrays` also has no `resize!` function in Julia 1.8 
```julia
Running /Users/kpierce/.julia/dev/ITensors/NDTensors/test/test_blocksparse.jl
test device: jl, eltype: Float32: Error During Test at /Users/kpierce/.julia/dev/ITensors/NDTensors/test/test_blocksparse.jl:32
  MethodError: no method matching resize!(::JLArrays.JLArray{Float32, 1}, ::Int64)
  Closest candidates are:
    resize!(::Vector, ::Integer) at array.jl:1233
    resize!(::BitVector, ::Integer) at bitarray.jl:814
    resize!(::NDTensors.SmallVectors.SubMSmallVector, ::Integer) at ~/.julia/dev/ITensors/NDTensors/src/lib/SmallVectors/src/subsmallvector/subsmallvector.jl:69
```

mtfishman:
@kmp5VT which issues are blocking this from getting merged?

Note that `resize!` should be defined for `JLVector` as of https://github.com/JuliaGPU/GPUArrays.jl/pull/533 and https://github.com/JuliaGPU/GPUArrays.jl/pull/546.

mtfishman:
Note that I don't see that issue with JLArrays.jl and Julia 1.8:
```julia
(JLArrays.jl) pkg> st
Status `~/Simons Foundation Dropbox/Matthew Fishman/Documents/workdir/JLArrays.jl/Project.toml`
  [27aeb0d3] JLArrays v0.1.5

julia> using JLArrays

julia> JLArrays.JLMatrix
JLArray{T, 2} where T

julia> versioninfo()
Julia Version 1.8.5
Commit 17cfb8e65ea (2023-01-08 06:45 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin21.5.0)
  CPU: 10 × Apple M1 Max
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, apple-m1)
  Threads: 1 on 8 virtual cores
```
I think your issue is that you wrote `JLArray.JLMatrix` instead of `JLArrays.JLMatrix`.

Besides that, I'm curious if the other issues you've cited get fixed if we just test the GPU backends against Julia 1.8, and implicitly drop support for Julia versions below 1.8 if users are using GPU backends.

mtfishman:
@kmp5VT it looks like the compat error is the same as the issue discussed here: https://discourse.julialang.org/t/psa-compat-requirements-in-the-general-registry-are-changing/104958#update-november-9th-2023-2. That post proposes a fix which would be to add `<0.0.1` to the compat entry of SparseArrays, which would allow version 0 of SparseArrays to be installed on older versions of Julia where standard libraries didn't have versions. Could you try that out and see if it fixes the CI failure on Julia 1.8?

kmp5VT:
@mtfishman It looks like modifying the `SparseArrays` compat like you suggested does work. Unfortunately, the Julia 1.8 tests still fail for `resize` with the JLArrays because there is a compat restriction for GPUArrays and JLArrays which prevent this code from being used. Heres the compat restrictions for the JLArrays version
```julia
(@v1.8) pkg> add JLArrays@0.1.5
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package Adapt [79e6a3ab]:
 Adapt [79e6a3ab] log:
 ├─possible versions are: 0.3.0-4.0.4 or uninstalled
 ├─restricted to versions 3.7.0-4 by NDTensors [23ae76d9], leaving only versions 3.7.0-4.0.4
 │ └─NDTensors [23ae76d9] log:
 │   ├─possible versions are: 0.3.42 or uninstalled
 │   └─NDTensors [23ae76d9] is fixed to version 0.3.42
 ├─restricted by compatibility requirements with GPUArrays [0c68f7d7] to versions: 4.0.0-4.0.4
 │ └─GPUArrays [0c68f7d7] log:
 │   ├─possible versions are: 0.3.0-10.3.1 or uninstalled
 │   ├─restricted by compatibility requirements with JLArrays [27aeb0d3] to versions: 10.0.0-10.3.1
 │   │ └─JLArrays [27aeb0d3] log:
 │   │   ├─possible versions are: 0.1.0-0.1.5 or uninstalled
 │   │   └─restricted to versions 0.1.5 by an explicit requirement, leaving only versions 0.1.5
 │   ├─restricted by julia compatibility requirements to versions: 0.3.0-10.2.3 or uninstalled, leaving only versions: 10.0.0-10.2.3
 │   └─restricted by compatibility requirements with CUDA [052768ef] to versions: [8.6.0-9.1.0, 10.0.1-10.3.1], leaving only versions: 10.0.1-10.2.3
 │     └─CUDA [052768ef] log:
 │       ├─possible versions are: 0.1.0-5.4.3 or uninstalled
 │       ├─restricted by compatibility requirements with StridedViews [4db3bf67] to versions: 4.0.0-5.4.3
 │       │ └─StridedViews [4db3bf67] log:
 │       │   ├─possible versions are: 0.1.0-0.3.1 or uninstalled
 │       │   └─restricted to versions 0.2.2-0.3 by NDTensors [23ae76d9], leaving only versions 0.2.2-0.3.1
 │       │     └─NDTensors [23ae76d9] log: see above
 │       └─restricted by compatibility requirements with GPUArrays [0c68f7d7] to versions: 5.2.0-5.4.3 or uninstalled, leaving only versions: 5.2.0-5.4.3
 │         └─GPUArrays [0c68f7d7] log: see above
 └─restricted by compatibility requirements with ArrayInterface [4fba245c] to versions: 3.0.0-3.7.2 — no versions left
   └─ArrayInterface [4fba245c] log:
     ├─possible versions are: 0.0.1-7.16.0 or uninstalled
     ├─restricted by compatibility requirements with LinearAlgebra [37e2e46d] to versions: 0.0.1-7.5.1 or uninstalled
     │ └─LinearAlgebra [37e2e46d] log:
     │   ├─possible versions are: 1.8.5 or uninstalled (package in sysimage!)
     │   └─restricted to versions 1.6.0-1 by NDTensors [23ae76d9], leaving only versions 1.8.5
     │     └─NDTensors [23ae76d9] log: see above
     ├─restricted by compatibility requirements with Compat [34da2185] to versions: [0.0.1-3.1.32, 6.0.2-7.16.0] or uninstalled, leaving only versions: [0.0.1-3.1.32, 6.0.2-7.5.1] or uninstalled
     │ └─Compat [34da2185] log:
     │   ├─possible versions are: 1.0.0-4.16.0 or uninstalled
     │   └─restricted to versions 4.9.0-4 by NDTensors [23ae76d9], leaving only versions 4.9.0-4.16.0
     │     └─NDTensors [23ae76d9] log: see above
     └─restricted by compatibility requirements with StaticArrayInterface [0d7ed370] to versions: 7.0.0-7.16.0, leaving only versions: 7.0.0-7.5.1
       └─StaticArrayInterface [0d7ed370] log:
         ├─possible versions are: 1.0.0-1.8.0 or uninstalled
         └─restricted by julia compatibility requirements to versions: 1.0.0-1.6.0 or uninstalled, leaving only versions: 1.0.0-1.6.0
```
and here are the current packages I have 
```julia
(@v1.8) pkg> st
Status `~/.julia/environments/v1.8/Project.toml`
⌅ [46192b85] GPUArraysCore v0.1.5
⌃ [27aeb0d3] JLArrays v0.1.2
  [23ae76d9] NDTensors v0.3.42 `../../dev/ITensors/NDTensors`
Info Packages marked with ⌃ and ⌅ have new versions available, but those with ⌅ are restricted by compatibility constraints from upgrading. To see why use `status --outdated`
```

NDTensors/Project.toml
```diff
@@ -70,6 +72,7 @@ GPUArraysCore = "0.1"
 HDF5 = "0.14, 0.15, 0.16, 0.17"
 HalfIntegers = "1"
 InlineStrings = "1"
+JLArrays = "0.1"
 LinearAlgebra = "1.6"
```
mtfishman, line 76:
```suggestion
LinearAlgebra = "<0.0.1, 1.6"
```

NDTensors/Project.toml
```diff
@@ -70,15 +72,16 @@ GPUArraysCore = "0.1"
...
 InlineStrings = "1"
-LinearAlgebra = "1.6"
+JLArrays = "0.1"
+LinearAlgebra = "<0.0.1, 1.6"
 MacroTools = "0.5"
 MappedArrays = "0.4"
 Metal = "1"
 Octavian = "0.3"
 PackageExtensionCompat = "1"
 Random = "1.6"
```
mtfishman, line 82:
```suggestion
Random = "<0.0.1, 1.6"
```

mtfishman:
Gotcha, thanks for checking.

I made similar changes to the compat entry of LinearAlgebra and Random since those are also standard libraries and the CI log showed a similar issue for LinearAlgebra.

If after those changes it still shows that error you are seeing with compatibility of JLArrays, let's just stop testing the GPU backends (including JLArrays.jl) against Julia 1.8. That issue may be caused by the fact that GPUArrays.jl v10 requires Julia 1.10: https://github.com/JuliaGPU/GPUArrays.jl/blob/v10.3.1/Project.toml#L26, and JLArrays.jl v0.1.5 requires GPUArrays v10: https://github.com/JuliaGPU/GPUArrays.jl/blob/v10.3.1/lib/JLArrays/Project.toml#L13.

mtfishman:
@kmp5VT I'm not sure what's going on with the CI in Julia 1.8, it's showing the same compat issue with LinearAlgebra even though I changed the compat bound to allow LinearAlgebra v0. Let's just stop testing against Julia 1.8 in the Jenkins GPU CI workflow.

Beyond that, are there issues that need to be addressed before this can be merged? How is the JLArrays extension being tested, as part of the Github Actions CPU test or as a Jenkins test? I don't see it run in Jenkins so I assume it is the former but if you could give a summary of the current status of this PR including how the tests are working that would be helpful since I've forgotten at this point.

NDTensors/ext/NDTensorsJLArraysExt/NDTensorsJLArraysExt.jl
mtfishman: I guess this comment still needs to be addressed, i.e. checking if all of the definitions are necessary?

NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -1,9 +1,10 @@
-using NDTensors: NDTensors
 using Pkg: Pkg
+using NDTensors: NDTensors
+
 if "cuda" in ARGS || "all" in ARGS
   ## Right now adding CUDA during Pkg.test results in a
   ## compat issues. I am adding it back to test/Project.toml
-  # Pkg.add("CUDA")
+  Pkg.add("CUDA")
   using CUDA
```
mtfishman, line 8:
```suggestion
  using CUDA: CUDA
```
to only bring `CUDA` into the namespace.

NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -18,13 +19,21 @@ if "metal" in ARGS || "all" in ARGS
 end
 if "cutensor" in ARGS || "all" in ARGS
   Pkg.add("cuTENSOR")
+  Pkg.add("CUDA")
   using CUDA, cuTENSOR
```
mtfishman, line 23:
```suggestion
  using CUDA: CUDA
  using cuTENSOR: cuTENSOR
```

NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -18,13 +19,21 @@ if "metal" in ARGS || "all" in ARGS
 end
 if "cutensor" in ARGS || "all" in ARGS
   Pkg.add("cuTENSOR")
+  Pkg.add("CUDA")
```
mtfishman, line 22:
```suggestion
  Pkg.add("CUDA")
  Pkg.add("cuTENSOR")
```
Alphabetical ordering.

NDTensors/test/NDTensorsTestUtils/device_list.jl
mtfishman: Or maybe it should be:
```julia
  using CUDA: cu
```

NDTensors/test/NDTensorsTestUtils/device_list.jl
mtfishman: Or maybe it should be:
```suggestion
  using cuTENSOR: cuTENSOR
  using CUDA: cu
```

mtfishman: Or maybe it should be:
```suggestion
  using cuTENSOR: cuTENSOR
  using CUDA: cu
```

jenkins/Jenkinsfile
```diff
@@ -7,7 +7,7 @@ pipeline {
   stages {
     stage('GPU Testing') {
       parallel {
-        stage('NDTensorsCUDAExt julia-1.6') {
+        stage('NDTensorsCUDAExt julia-1.8') {
```
mtfishman, line 10: Reminder to delete this section of the workflow to not test against Julia 1.8.

NDTensors/test/NDTensorsTestUtils/device_list.jl
```diff
@@ -18,13 +19,21 @@ if "metal" in ARGS || "all" in ARGS
...
 end
 
+using JLArrays: JLArrays, jl
+
 function devices_list(test_args)
   devs = Vector{Function}(undef, 0)
   if isempty(test_args) || "base" in test_args
     push!(devs, NDTensors.cpu)
+    ## Skip jl on lower versions of Julia for now
+    ## all linear algebra is failing on Julia 1.6 with JLArrays
```
mtfishman, line 33: I was wondering how you were handling testing JLArrays on older Julia versions where you were seeing issues, this seems like a reasonable solution. I guess JLArrays installs fine on Julia 1.6, but the tests fail?

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: I wasn't sure if this was going to work. I have previously run into trouble in `Pkg.test` when I tried to add things to the manifest during a test. So far this seems to work though it may be causing the issues we see on Jenkins. Regardless I have reorganized and fixed the `using` calls

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: I am working to fix my workstation currently I cannot test CUDA to verify this code

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: JLArrays does install fine and I was able to fix a symbol issue (JLArrays doesn't have `JLMatrix` defined but they do have `AnyJLMatrix` which is a larger union definition which includes some wrapper types). 
The issue with the JLArrays code in 1.6 is mainly that the `resize!` function is not included in the `JLArrays` version which is compatible with Julia 1.6. So i think its safer to just not test it on lower version especially because almost all GPU implementations don't support lower Julia versions

kmp5VT:
For some reason Jenkins did not launch during this test, I do not know why

NDTensors/ext/NDTensorsJLArraysExt/NDTensorsJLArraysExt.jl
kmp5VT: I believe that I have gone through all of the functions to verify that they are necessary. 
What I have found is that JLArrays suffers from its own issues require the implementation of these specialized function.

NDTensors/ext/NDTEnsorsJLArraysExt/set_types.jl
kmp5VT: Should we make this default definition in this PR?

jenkins/Jenkinsfile
kmp5VT: This section has been deleted. Thanks!

NDTensors/ext/NDTEnsorsJLArraysExt/set_types.jl
mtfishman: You mean the default for all `AbstractArray` subtypes? Let's do it.

kmp5VT: I updated the code with a default `AbstractArray` version and call the default function for GPU tensors, except for Metal because it requires Float32.

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: These are working now from what I can tell, I am closing this comment

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: I updated these and I bring in the `CUDA` name instead of `cu` because we use `NDTensors.cu` in the code not the opinionated (F32 preferred) `CUDA.cu` version. I am closing this comment. Thanks!

NDTensors/test/NDTensorsTestUtils/device_list.jl
kmp5VT: Updated per the comment below. Thanks!

kmp5VT:
@mtfishman All of the JLArrays code seems to work now! A questions for this PR: 
1. Should I spend time making an `NDTensorsGPUArraysCoreExt` module and generalize functions which are duplicated across the GPU extension libraries?
2. Should I try to introduce JLArrays to the ITensor tests in this PR or should we merge this and I start a new one for that process?

mtfishman:
>     1. Should I spend time making an `NDTensorsGPUArraysCoreExt` module and generalize functions which are duplicated across the GPU extension libraries?

I already see: https://github.com/ITensor/ITensors.jl/tree/main/NDTensors/ext/NDTensorsGPUArraysCoreExt. I guess you mean adding more functionality to that extension? Let's hold off on that and focus on wrapping up this PR and keeping it more self-contained.

>     2. Should I try to introduce JLArrays to the ITensor tests in this PR or should we merge this and I start a new one for that process?

Isn't it already running as part of the tests? I.e. the `JLArrays.jl` adaptor is added to the device list here: https://github.com/kmp5VT/ITensors.jl/blob/14922b1063676c02d8876d3403aaae8d7fa022f2/NDTensors/test/NDTensorsTestUtils/device_list.jl#L36.

kmp5VT:
@mtfishman oh I had forgotten that I created a GPUArraysCore extension. But yes this is what I mean, adding functionality to `NDTensorsGPUArraysCoreExt` to replace overlapping function definitions in the other GPU extensions. But I understand your response and leaving that for work in a future PR was going to be my suggestion.

Regarding testing: Yes JLArrays is being used in the `NDTensors` test suite, but I was wondering if we want to modify the `ITensors` test suite to test on CPU and with `JLArrays.jl`? This idea comes from the understanding that the `ITensors` test suite is currently more far reaching than the NDTensors tests so we may catch GPU issues which we don't see in `NDTensors` tests.

Thanks!

mtfishman:
> @mtfishman oh I had forgotten that I created a GPUArraysCore extension. But yes this is what I mean, adding functionality to `NDTensorsGPUArraysCoreExt` to replace overlapping function definitions in the other GPU extensions. But I understand your response and leaving that for work in a future PR was going to be my suggestion.

Sounds good, sounds like we are in agreement there. One comment about the philosophy there is that even if there are overlapping definitions across extensions, it doesn't necessarily mean we should always turn them into generic definitions in `NDTensorsGPUArraysCoreExt`. For example, multiple GPU extensions might have a certain `Exposed` workaround for an operation that isn't working, but I think it is better to have repeated definitions in the different backends in that case since that makes it easier to customize the extensions as the backends change (say, if one GPU backend fixes an operation but it remains broken in the other backends). So it is a judgement call about what operations we should turn into generic definitions.

> Regarding testing: Yes JLArrays is being used in the `NDTensors` test suite, but I was wondering if we want to modify the `ITensors` test suite to test on CPU and with `JLArrays.jl`? This idea comes from the understanding that the `ITensors` test suite is currently more far reaching than the NDTensors tests so we may catch GPU issues which we don't see in `NDTensors` tests.

I see, thanks for clarifying. Let's hold off on incorporating it into the ITensors test suite, I think those tests already take too long to run, and would prefer to make more tests at a lower level (i.e. at the NDTensors level, and more specifically at the submodule level) going forward anyway, since that is a better strategy for having more systematic and efficient test coverage.

NDTensors/ext/NDTensorsAMDGPUExt/set_types.jl
```diff
@@ -1,11 +1,12 @@
...
-using NDTensors.TypeParameterAccessors: TypeParameterAccessors, Position
+using NDTensors.TypeParameterAccessors:
+  TypeParameterAccessors, Position, default_type_parameters
 using NDTensors.GPUArraysCoreExtensions: storagemode
 using AMDGPU: AMDGPU, ROCArray
+using GPUArraysCore: AbstractGPUArray
 
 function TypeParameterAccessors.default_type_parameters(::Type{<:ROCArray})
-  return (Float64, 1, AMDGPU.Mem.HIPBuffer)
+  return (default_type_parameters(AbstractGPUArray)..., AMDGPU.Mem.HIPBuffer)
```
mtfishman, line 9:
```suggestion
  return (default_type_parameters(AbstractArray)..., AMDGPU.Mem.HIPBuffer)
```
and same thing elsewhere. Then we wouldn't have to load `GPUArraysCore` above.

NDTensors/ext/NDTensorsCUDAExt/set_types.jl
```diff
@@ -1,18 +1,14 @@
...
-function TypeParameterAccessors.position(::Type{<:CuArray}, ::typeof(ndims))
-  return Position(2)
-end
 function TypeParameterAccessors.position(::Type{<:CuArray}, ::typeof(storagemode))
   return Position(3)
 end
 
 function TypeParameterAccessors.default_type_parameters(::Type{<:CuArray})
-  return (Float64, 1, CUDA.Mem.DeviceBuffer)
+  return (default_type_parameters(AbstractGPUArray)..., CUDA.Mem.DeviceBuffer)
```
mtfishman, line 13:
```suggestion
  return (default_type_parameters(AbstractArray)..., CUDA.Mem.DeviceBuffer)
```

mtfishman: (And if so, don't load `GPUArraysCore`.)

NDTensors/ext/NDTensorsAMDGPUExt/set_types.jl
kmp5VT: This makes sense, I updated it. Thanks!

NDTensors/ext/NDTensorsCUDAExt/set_types.jl
kmp5VT: Same as above, thanks!

mtfishman:
Thanks! This will be helpful for testing the GPU backends.

emstoudenmire:
Great to see this

PR #1510, opened 2024-06-23
[ITensors] Generalize `diag_itensor` constructors to AbstractVector

mtfishman:
Fixes https://itensor.discourse.group/t/diag-itensor-constructor-on-gpu-arrays/1827.


mtfishman:
@kmp5VT it would be nice to start testing things like this using #1508, looks like right now ITensors.jl tests are only set up to run on CPU unless I'm missing something.

PR #1511, opened 2024-06-23
[NDTensorsGPUArraysCoreExt] Fix nonuniform Diag-Dense contractions on GPU

mtfishman:
This fixes:
```julia
using ITensors: Index, random_itensor, diag_itensor
using Metal: mtl
device = mtl
d = 3
i, j = Index.((d, d))
a = device(random_itensor(i, j))
d = device(diag_itensor(randn(d), i, j))
a * d
```
whereas before that contraction failed with a scalar indexing error. The fix is fairly crude and is based on converting the Diag tensor to Dense, similar to the fix for contracting uniform Diag with Dense in #1453.

I've also added some definitions for `mapreduce` for `Tensor`s with storage types `Dense`, `Diag`, and `BlockSparse`, and defined `sum(::ITensor)` and `prod(::ITensor)` which make use of those definitions and therefore work on GPU and take advantage of sparsity.

This fixes the issues raised in https://itensor.discourse.group/t/gpu-issues-scalar-indexing-is-disallowed-on-sum-itensor-and-traces-of-products-involving-diagonal-itensors/1828.

### DIFF
--- a/NDTensors/Project.toml
+++ b/NDTensors/Project.toml
@@ -1,7 +1,7 @@
 name = "NDTensors"
 uuid = "23ae76d9-e61a-49c4-8f12-3f1a16adf9cf"
 authors = ["Matthew Fishman <mfishman@flatironinstitute.org>"]
-version = "0.3.34"
+version = "0.3.35"
 
 [deps]
 Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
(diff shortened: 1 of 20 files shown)


mtfishman:
@kmp5VT I also went ahead and started using `NDTensors.map_diag!` to implement the definition of that function for `ITensor`, as a followup to #1497.

mtfishman:
@kmp5VT another thing I added in this PR is a generic definition of `==` for `Tensor`s which avoids scalar indexing and therefore works on GPU. I switched over some tests to use that new definition. It looks like we could remove a few `@allowscalar` calls in the tests now that we have that.

kmp5VT:
> @kmp5VT another thing I added in this PR is a generic definition of `==` for `Tensor`s which avoids scalar indexing and therefore works on GPU. I switched over some tests to use that new definition. It looks like we could remove a few `@allowscalar` calls in the tests now that we have that.

This is great, thanks!!

PR #1517, opened 2024-07-09
[NDTensors] Scalar diag calls `expose` to skip scalar indexing

kmp5VT:
# Description

When Diag is a scalar we should call the `expose` version of getindex to allow this to work on CPU and GPU without scalar indexing issue. Related to this issue on discourse https://itensor.discourse.group/t/scalar-on-a-zero-rank-tensor-on-gpu/1863/2


emstoudenmire:
Looks good

emstoudenmire (approved):
Reviewed this with Karl in person, and seems good and is consistent with other uses of `expose` and other places where we allow scaling indexing for `getindex` of GPU arrays.

mtfishman:
Looks good to me, can you add a test?

PR #1529, opened 2024-09-12
[NDTensors] split dual into UnitRangeDual and GradedUnitRangeDual

ogauthe:
This PR aims to simplify the use of `GradedAxes.dual`, especially to be used in a `BlockSparseArray`. It proposes to split the `UnitRangeDual` type into two different types:
- a generic `UnitRangeDual{<:AbstractUnitRange} <: AbstractUnitRange`
- a labelled `GradedUnitRangeDual{<:LabelledInteger,<:AbstractGradedUnitRange} <: AbstractGradedUnitRange`

This should simplify slicing a `BlockSparseArray` with dual axes and allows to preserve labels or take their dual.

This branch is built on top of #1363 and should only be merged once #1363 is completed. It does not make any change in `Sectors`.

PR #1363, opened 2024-03-25
[SymmetrySectors] Non-abelian fusion

ogauthe:
This is a work in progress for a full implementation of non-abelian fusion rules in `Sectors`. It also contains several patches on other components of `Sectors`. A core aspect is that the dependency order between `GradedAxes` and `Sectors` has been reversed: `Sectors` now depends on `GradedAxes`.

To do:
- [x] Continue to refactor and simplify `SectorProduct` fusion, which is implemented in `sectors_fusion_rule`. The code logic handled by `recover_style`/`recover_sector_product_type` for handling different data formats (`Tuple` vs. `NamedTuple`, abelian vs. non-abelian) is still fairly complicated and hard to follow, we should try to simplify that.
- [x] Change output format of `label_fusion_rule` to `Vector{<:Pair{<:AbstractSector,Int}}`.
- [x] Define `abelian_label_fusion_rule` that can be used for abelian groups as a simpler interface for defining fusion rules just in terms of the labels and without the need for specifying degeneracies/block lengths. `label_fusion_rule` by default can call `abelian_label_fusion_rule` and specify that there is one block of length 1.
- [X] Rename `sectors(...)` to `product_sectors(...)` for accessing the storage of a `SectorProduct`. Additionally, rename functions that act on `SectorProduct` storage from `sectors_*(...)` to `product_sectors_*(...)` (for example, rename `sectors_isless(...)` to `product_sectors_isless(...)`).
- [x] Change `SectorProduct` equality so that `SectorProduct(SU2(1), U1(0)) == SectorProduct(SU2(1))` so that it returns `true` (i.e. we should assume trailing trivial sectors, like the fusion code does), along with changing any other related corner cases of equality and inequality of category products.
- [x] Change the name of the module from `Sectors` to `SymmetrySectors`.
- [x] Rename `AbstractCategory` to `AbstractSector` and `CategoryProduct` to `SectorProduct` (also `categories` to `sectors` throughout the code).
- [x] Introduce `struct TrivialSector <: AbstractSector end`, remove as many special code definitions for `SectorProduct{Tuple{}}` and `SectorProduct{NamedTuple{()}}` as possible.
- [x] implement fusion rules for `SectorProduct`.
- [x] implement `SymmetryStyle` trait with values `AbelianGroup`, `NonAbelianGroup` and `NonGroupSymmetry` (this has been simplified to just `AbelianStyle` and `NotAbelianStyle`).
[...]

PR #1531, opened 2024-09-16
[GradedAxes] Introduce GradedUnitRangeDual

ogauthe:
This PR aims to simplify the use of `GradedAxes.dual`, especially to be used in a `BlockSparseArray`. It proposes to remove the `UnitRangeDual` type: 
- a generic `AbstractUnitRange` is now self dual
- a labelled `GradedUnitRange` or `GradedOneTo` are cast to `GradedUnitRangeDual{<:LabelledInteger,<:AbstractGradedUnitRange} <: AbstractGradedUnitRange`

This should simplify slicing a `BlockSparseArray` with dual axes and allows to preserve labels or take their dual.

This branch is similar to #1529 but is based on main.

Currently, this branche solves:
- preserving labels in `blocklengths(dual(...))`
```julia
g = gradedrange([U1(1) => 2, U1(2) => 2])
@test GradedAxes.blocklengths(dual(g)) isa Vector{<:LabelledInteger}  #    # broken in main and in non-abelian_fusion
```

- preserving `GradedUnitRange` behavior when slicing:
```julia
blocklengths(dual(gradedrange([U1(0) => 2, U1(1) => 2])[1:3])) isa Vector  # broken in main and in non-abelian_fusion
```

- manipulating arrays with `DualGradedUnitRange`
```julia
r = gradedrange([U1(0) => 2, U1(1) => 2])[1:3]
a = BlockSparseArray{Float64}(r, r)
b = BlockSparseArray{Float64}(dual(r),dual(r))
@test a[:, :] isa BlockSparseArray   # broken in main and in non-abelian_fusion
@test b[:, :] isa BlockSparseArray   # broken in main and in non-abelian_fusion
```
Meaning one can create and manipulate `BlockSparseArray` with `GradedUnitRange` or its dual (not just `BlockedOneTo`). However `axes(a[:,:])` are still `Base.IdentityUnitRange(GradedUnitRangeDual{...})`, which creates other issues.

It also adds many tests, including broken ones.

- [ ] The current plan is to have multiple dual unit range types, such as `UnitRangeDual <: UnitRange`, `GradedUnitRangeDual <: AbstractGradedUnitRange`, and maybe `BlockedUnitRangeDual <: AbstractBlockedUnitRange`. In that case, it would be useful to define an `IsDual` trait based on the `isdual` function (say using [SimpleTraits.jl](https://github.com/mauro3/SimpleTraits.jl)). See https://github.com/ITensor/ITensors.jl/pull/1363/files#r1779660676 and https://github.com/ITensor/ITensors.jl/pull/1363/files#r1779708817 for examples of where that would be useful, say for generalizing `blocklabels(a::UnitRangeDual)` and `flip_dual(r::UnitRangeDual)`.


ogauthe:
Overall I still find there are too many kind of `UnitRange`. I wonder if we could avoid the doublet `GradedOneTo` and `GradedUnitRange`. `gradedrange` creates a `GradedOneTo`, but when slicing it one gets a `GradedUnitRange`.

mtfishman:
For dispatch it is helpful to have special ranges that start at one since that is used as axes by most arrays, like how Base has `UnitRange` and `Base.OneTo`. Handling both shouldn't be too difficult if we write good generic code.

mtfishman:
To be more specific, if we write most code in terms of `AbstractGradedUnitRange` and `AbstractUnitRange`, that should allow us to have a lot of subtypes (such as `GradedOneTo`, `GradedUnitRage`, `GradedUnitRangeDual`, etc.) without adding too much code complexity. Though perhaps one question is, is there a need for a `GradedOneToDual`? That does start to sound excessive and it is best to try to design it without that for now.

One thing to consider is that dual only really makes sense for graded unit ranges, so maybe we could consider dropping `UnitRangeDual`, and just define `dual(r::AbstractUnitRange) = r` as a generic fallback (i.e. if the unit range isn't graded, the dual is itself).

ogauthe:
I have no strong opinion whether there should be a `UnitRangeDual` type or not. This is mostly a design question for `BlockSparseArray`, about how much complexity you want there. In the specific case of `FusionTensor`, I will always use a graded axis. My objective here is to make `dual(GradedUnitRange)` an `AbstractGradedUnitRange` to deal with labels in a generic way.

However I believe dual and not dual axes should be handled on equal footing: deciding who is dual and who is not is a pure convention and they should always appear with a 1:1 ratio for contraction to be possible. If there is some gain in having both `GradedOneTo` and `GradedUnitRangeDual`, then I think there should be a `GradedOneToDual`.

ogauthe:
I realize that with the current design, there are already `GradedUnitRangeDual{<:LabelledInteger, BlockArrays.BlockedOneTo` and `GradedUnitRangeDual{<:LabelledInteger, BlockArrays.BlockedUnitRange}`, so there should be no need for an explicit `GradedOneToDual`.

I will remove `UnitRangeDual` and use self-dual axes as the default.

mtfishman:
> I realize that with the current design, there are already `GradedUnitRangeDual{<:LabelledInteger, BlockArrays.BlockedOneTo` and `GradedUnitRangeDual{<:LabelledInteger, BlockArrays.BlockedUnitRange}`, so there should be no need for an explicit `GradedOneToDual`.
> 
> I will remove `UnitRangeDual` and use self-dual axes as the default.

Sounds good, that's what I was going to suggest (i.e. `GradedOneToDual` can be a type alias, if needed).

ogauthe:
I think this is now ready for `GradedAxes`. I will focus on `BlockSparseArrays` in another PR.

mtfishman:
@ogauthe I realized that one reason to keep `UnitRangeDual` is that it may be a reasonable return type when slicing blocks of a `GradedUnitRangeDual`, i.e. it might make sense for `dual(gradedrange([U1(0) => 2, U1(1) => 2]))[Block(1)]` to return a `UnitRangeDual`, to preserve the fact that the block was taken from a dual space.

That could be helpful since then it could make it easier to create a dual graded unit range out of the blocks of another dual graded unit range (say if you have a dual graded unit range and want to remove one of the blocks, it could make it easier to write that kind of code).

ogauthe:
Indeed slicing a `GradedUnitRangeDual` with a `Block` was an issue I spotted (this is associated with the slicing issues in #1336). Currently it yields a `LabelledUnitRange` and the dual information is lost.

Now, how generic `UnitRangeDual` should be?  I guess it should be a subtype of `AbstractUnitRange` and should be valid axis for a `BlockSparseArray`. Should we come back to "any `AbstractUnitRange` can be dualed and become a `UnitRangeDual`" or should it be more specific, as "obtained from slicing a `GradedUnitRangeDual`"?

I favor the second as it would make it simpler to use a `BlockSparseArray` without caring about mathematical details, but I think both options carry the same amount of code complexity (mostly coming from being valid `BlockSparseArray` axes).

mtfishman:
I agree with your assessment that `UnitRangeDual` should be a subtype of `AbstractUnitRange`, and also that we should be thoughtful about which operations create `UnitRangeDual`. I'll try to summarize that view:

I think it should be valid to make a `BlockSparseArray` with any kind of unit range as long as it satisfies the [BlockArrays.jl axis interface](https://juliaarrays.github.io/BlockArrays.jl/stable/man/abstractblockarrayinterface/#The-block-axis-interface), and maybe some additional interface functions we require in `BlockSparseArrays`, so axes may be dual or non-dual axes of whatever kind.

The question to me more comes down to how and when dual axes get constructed, depending on the unit range type and operation. So to be concrete, if you have a block sparse matrix `a`, in general `a'` should have axes: `axes(a') == dual.(reverse(axes(a)))`, so that operations like `a' * a` make sense (i.e. involve a contraction of a dual axis with a non-dual axis).

However, I think it would be surprising and unnecessarily complicated if `a` started with non-graded axes, say `OneTo` or `BlockOneTo`, and `a'` had dual axes, say `UnitRangeDual{<:Any,<:OneTo}` or `UnitRangeDual{<:Any,<:BlockOneTo}`, since they are self-dual. So that leads me to think that we should define `dual(r::OneTo) = r` and `dual(r::BlockOneTo) = r` (since `adjoint(::BlockSparseMatrix)` will be defined in terms of `dual` to take the dual of the axes where appropriate), and if a user wants to create a `BlockSparseArray` with dual axes of type `UnitRangeDual{<:Any,<:BlockOneTo}`, they'll have to do that manually in another way besides the `dual(...)` function.

We may want to define that `dual(r::LabelledUnitRange)` outputs a `UnitRangeDual(r)`, though in that case it may be better to rename `LabelledUnitRange` to something else to indicate that it is more specifically the block of a `GradedUnitRange`.

mtfishman:
This is copied from this comment: https://github.com/ITensor/ITensors.jl/pull/1363#discussion_r1774410848:

It would be nice to think through how functions like:
```julia
flip(g::AbstractGradedUnitRange) = dual(gradedrange(label_dual.(blocklengths(g))))
```
could be written in a simpler way.

One proposal would be to make this work as an implementation:
```julia
flip(g::AbstractGradedUnitRange) = gradedrange(flip.(blocklengths(g)))
```
In order to make that work, we would need to have `flip` work on sectors, define the concept of dual sectors, and make it so that `gradedrange` outputs a dual unit range when constructed from dual sectors.

An implementation could be the following:
```julia
flip(l::LabelledInteger) = labelled(unlabel(l), flip(label(l)))

struct CategoryDual{NondualCategory<:AbstractCategory} <: AbstractCategory
  nondual_category::NondualCategory
end
flip(c::AbstractCategory) = CategoryDual(dual(c))
nondual(c::AbstractCategory) = c
nondual(c::CategoryDual) = c.nondual_category
dual(c::CategoryDual) = nondual(c)
```
In general, `dual(flip(x))` can be the canonical/generic way of instantiating a dual sector/space/range, and could supersede `label_dual` (which I never liked but didn't have a better idea for at the time).

Additionally, `gradedrange` would have to be able to detect that all of the labels are `CategoryDual` and in that case construct a `UnitRangeDual`/`GradedUnitRangeDual`.

NDTensors/src/lib/GradedAxes/src/gradedunitrangedual.jl
```diff
@@ -0,0 +1,148 @@
+struct GradedUnitRangeDual{
+  T<:LabelledInteger,NondualUnitRange<:AbstractGradedUnitRange{T}
+} <: AbstractGradedUnitRange{T,Vector{T}}
```
mtfishman, line 3:
```suggestion
struct GradedUnitRangeDual{
  T,NondualUnitRange<:AbstractGradedUnitRange{T,CS}
} <: AbstractGradedUnitRange{T,CS}
```
to make it a bit more general.

mtfishman:
@ogauthe probably at this point it makes more sense to define:
```julia
abstract type AbstractGradedUnitRange{T,CS} = AbstractBlockedUnitRange{T,CS}
```
as opposed to making it a type alias. Also I would drop the `T<:LabelledInteger` constraint for now, and think of it more as an interface requirement, i.e. the elements of an `AbstractGradedUnitRange` should act like a `LabelledInteger` (integer element of the range and associated sector), but it might not be a `LabelledInteger`.

ogauthe:
Indeed we would have more freedom with `abstract type AbstractGradedUnitRange{T,CS} `. Just to be clear, that would be a significant change in the implementation. It seems to me that implies rewriting several parts of `gradedunitrange.jl`, as we would need to explicitly write code path that were calling/assuming `BlockedUnitRange` and `BlockedOneTo`. I am working on it.

mtfishman:
> Indeed we would have more freedom with `abstract type AbstractGradedUnitRange{T,CS} `. Just to be clear, that would be a significant change in the implementation. It seems to me that implies rewriting several parts of `gradedunitrange.jl`, as we would need to explicitly write code path that were calling/assuming `BlockedUnitRange` and `BlockedOneTo`. I am working on it.

I see, I wasn't sure if that would affect other parts of the code, but I think it is the right design so hopefully those changes are not too involved.

ogauthe:
Here is a first draft passing `test_basics.jl`.

Concerning `dual`, if I remember correctly, the rational behind `UnitRangeDual` was precisely to have `GradedUnitRange = BlockedUnitRange{LabelledInteger}`, at a time when `BlockArrays` did not define `AbstractBlockedUnitRange`.
Now we have custom `GradedOneTo` and `GradedUnitRange`, which are `AbstractBlockedUnitRange` with extra logic for label manipulation. Therefore I believe the best for `dual` would be to have a `isdual::Bool` field inside them, preserved over slicing. That would add a few more lines to the label logic, but would completely avoid to introduce a new type `GradedUnitRangeDual`. I think that would considerably simplify `BlockSparseArrays`, where many issues comes from `UnitRangeDual` being too complex.

We may however still need a `UnitRangeDual<:AbstractUnitRange` for non-blocked ranges. Or maybe a `DualCategory` to be used only for that purpose, although I am worried introducing `DualCategory` would cause other issues.

mtfishman:
> Concerning `dual`, if I remember correctly, the rational behind `UnitRangeDual` was precisely to have `GradedUnitRange = BlockedUnitRange{LabelledInteger}`, at a time when `BlockArrays` did not define `AbstractBlockedUnitRange`. Now we have custom `GradedOneTo` and `GradedUnitRange`, which are `AbstractBlockedUnitRange` with extra logic for label manipulation. Therefore I believe the best for `dual` would be to have a `isdual::Bool` field inside them, preserved over slicing. That would add a few more lines to the label logic, but would completely avoid to introduce a new type `GradedUnitRangeDual`. I think that would considerably simplify `BlockSparseArrays`, where many issues comes from `UnitRangeDual` being too complex.

I'm open to considering switching to an `isdual` flag if there is a convincing argument for it or get convinced that the current design really isn't working, but for the time being why don't we do the best we can improving on the current design and see how it works out. I started designing `GradedAxes` with an `isdual` flag but I felt like it got very awkward having to check that flag at runtime within a lot of functions. I thought that putting it at the type level would allow us to separate out the abstraction layers, so one layer of code doesn't have to worry too much about duals, and then duals can be handled in a separate layer of code using dispatch (which I think is working out pretty well as shown in this PR but I agree there seems to be a lot of code duplication). I think it is working out ok and I think that we can probably simplify the code once the dust settles a bit on all the PRs and improvements we need to make to `BlockSparseArray`, but of course we haven't done that yet so we'll have to see how it works out.

Something I think could help with having multiple dual unit range types is having an `IsDual` trait and using trait dispatch to define common functionality on the various dual types, but again that is speculation right now and we would have to try it out and see if that helps.

> We may however still need a `UnitRangeDual<:AbstractUnitRange` for non-blocked ranges. Or maybe a `DualCategory` to be used only for that purpose, although I am worried introducing `DualCategory` would cause other issues.

One reason we may still want to have `UnitRangeDual` is because I think that should be the output of slicing operations like `r[Block(2)]` when `r::GradedUnitRangeDual`.

ogauthe:
`GradedAxes` and `SymmetrySectors ` tests are passing. I am now working on issues with `BlockSparseArrays`. I spotted several cases where the labels are dropped:

```julia
d1 = gradedrange([U1(0) => 2, U1(1) => 2])
d2 = gradedrange([U1(0) => 2, U1(1) => 2])
a = BlockSparseArray{Float64}(d1, d2, d1, d2);

@test axes(2 * a, 1) isa AbstractGradedUnitRange
@test_broken axes(a + a, 1) isa AbstractGradedUnitRange
@test axes(a + a, 1) isa BlockedUnitRange{Int64, Vector{Int64}}

b = a[2:3, 2:3, 2:3, 2:3];
@test_broken axes(b, 1) isa AbstractGradedUnitRange 
@test axes(b, 1) isa BlockedOneTo{Int64, Vector{Int64}}

bv = BlockVector([Block(1), Block(2)], [1, 1])
b = a[bv, bv, bv, bv];
@test_broken axes(b, 1) isa AbstractGradedUnitRange
@test axes(b, 1) isa BlockedOneTo{Int64, Vector{Int64}}
```

It is not clear to me where the problem lies as slicing the axes is fine. Also, `blockedrange([labelled(2,U1(0)), labelled(2,U1(1))])` still returns a `GradedOneTo` so we can still use generic code. @mtfishman do you have an idea?

mtfishman:
I don't have a good guess without digging into the code... I had to track down a lot of issues like that before, usually it is just a missing overload.

ogauthe:
Indeed the problems come from missing overload for `BlockArrays.combine_blockaxes`. Fixing it in a generic way should solve may related issues in #1336 .

mtfishman:
> Indeed the problems come from missing overload for `BlockArrays.combine_blockaxes`. Fixing it in a generic way should solve many related issues in #1336 .

Great! Glad to hear you tracked it down, that makes sense since it is used in block sparse array broadcasting to make the blocks commensurate, and scalar multiplication, addition, and some slicing operations ultimately call broadcasting.

ogauthe:
The other problems where due to the change `blockedunitrange_getindices/gradedunitrange_getindices`. Maybe I should avoid defining `gradedunitrange_getindices` and just specialize `blockedunitrange_getindices` for `AbstractGradedUnitRange`.

mtfishman:
> The other problems where due to the change `blockedunitrange_getindices/gradedunitrange_getindices`. Maybe I should avoid defining `gradedunitrange_getindices` and just specialize `blockedunitrange_getindices` for `AbstractGradedUnitRange`.

That seems simpler, for context the primary purpose of `blockedunitrange_getindices` was to circumvent issues in `BlockArrays.jl`.

NDTensors/src/lib/BlockSparseArrays/ext/BlockSparseArraysGradedAxesExt/test/runtests.jl
```diff
@@ -130,41 +140,149 @@ const elts = (Float32, Float64, Complex{Float32}, Complex{Float64})
         @test a[I] == a_dense[I]
       end
       @test axes(a') == dual.(reverse(axes(a)))
-      # TODO: Define and use `isdual` here.
-      @test typeof(axes(a', 1)) === typeof(dual(axes(a, 2)))
-      @test typeof(axes(a', 2)) === typeof(dual(axes(a, 1)))
+
+      @test isdual(axes(a', 1)) ⊻ isdual(axes(a, 2))
+      @test isdual(axes(a', 2)) ⊻ isdual(axes(a, 1))
```
mtfishman, line 145:
```suggestion
      @test isdual(axes(a', 1)) ≠ isdual(axes(a, 2))
      @test isdual(axes(a', 2)) ≠ isdual(axes(a, 1))
```

NDTensors/src/lib/GradedAxes/src/gradedunitrange.jl
```diff
@@ -17,31 +17,51 @@ using BlockArrays:
...
   LabelledInteger,
   LabelledUnitRange,
   label,
+  label_type,
   labelled,
   labelled_isequal,
   unlabel
 
-const AbstractGradedUnitRange{T<:LabelledInteger} = AbstractBlockedUnitRange{T}
+abstract type AbstractGradedUnitRange{T,CS} <: AbstractBlockedUnitRange{T,CS} end
```
mtfishman, line 34:
```suggestion
abstract type AbstractGradedUnitRange{T,BlockLasts} <: AbstractBlockedUnitRange{T,BlockLasts} end
```
since that's what we use elsewhere in `GradedAxes`. I know I suggested `CS` previously but I was just copying that from `BlockArrays.jl` and I don't know why they use that, it looks very mysterious to me.

NDTensors/src/lib/GradedAxes/src/gradedunitrange.jl
```diff
@@ -17,31 +17,51 @@ using BlockArrays:
...
+  function GradedOneTo(lasts::CS) where {T<:Integer,CS<:AbstractVector{T}}
+    Base.require_one_based_indexing(lasts)
+    isempty(lasts) || first(lasts) >= 0 || throw(ArgumentError("blocklasts must be >= 0"))
+    return new{T,CS}(lasts)
+  end
+  function GradedOneTo(lasts::CS) where {T<:Integer,CS<:Tuple{T,Vararg{T}}}
+    first(lasts) >= 0 || throw(ArgumentError("blocklasts must be >= 0"))
+    return new{T,CS}(lasts)
+  end
+end
```
mtfishman, line 54:
```suggestion
  function GradedOneTo(lasts::BlockLasts) where {T<:Integer,BlockLasts<:AbstractVector{T}}
    Base.require_one_based_indexing(lasts)
    isempty(lasts) || first(lasts) >= 0 || throw(ArgumentError("blocklasts must be >= 0"))
    return new{T,BlockLasts}(lasts)
  end
  function GradedOneTo(lasts::BlockLasts) where {T<:Integer,BlockLasts<:Tuple{T,Vararg{T}}}
    first(lasts) >= 0 || throw(ArgumentError("blocklasts must be >= 0"))
    return new{T,BlockLasts}(lasts)
  end
end
```

NDTensors/src/lib/GradedAxes/src/gradedunitrangedual.jl
```diff
@@ -0,0 +1,117 @@
+struct GradedUnitRangeDual{T,CS,NondualUnitRange<:AbstractGradedUnitRange{T,CS}} <:
+       AbstractGradedUnitRange{T,CS}
+  nondual_unitrange::NondualUnitRange
+end
```
mtfishman, line 4:
```suggestion
struct GradedUnitRangeDual{T,BlockLasts,NondualUnitRange<:AbstractGradedUnitRange{T,BlockLasts}} <:
       AbstractGradedUnitRange{T,BlockLasts}
  nondual_unitrange::NondualUnitRange
end
```

mtfishman:
@ogauthe this PR looks good to me, given the current type hierarchy that we are aiming for right now. I definitely get your point about there being a lot of code that needs to be defined to make all of the different unit range types work. I think there are a few potential solutions, some of which are more benign like reassessing what code can be made more generic and can get shared across unit range types or even simplified by making upstream changes to `BlockArrays.jl`, and some more extreme like making an `isdual` flag or moving the information about duality into the sectors. As I said let's stick with this design for now and see how far it takes us and reassess once more of the `FusionTensors` and `BlockSparseArrays` code is fleshed out.

From what I can tell, at least with the current code design and organization a lot of the complexity for dealing with the different unit range types is living in `GradedAxes` while `BlockSparseArrays` and `FusionTensors` can remain relatively agnostic about those different types, so that should make it easier to try out different designs within `GradedAxes` in the future without affecting those libraries too much.

Regarding slicing `GradedUnitRangeDual`, do you plan to fix that in this PR or in a follow-up? I'm fine either way, though this PR has gotten pretty long already so maybe it would be nicer to merge this and work on that in a future PR, for the sake of making reviewing easier.

NDTensors/src/lib/GradedAxes/src/gradedunitrange.jl
ogauthe: Indeed, I did not understand what `CS` means.

ogauthe:
Thanks for the review. I already started working on a slicing `GradedUnitRangeDual`, but this is non trivial.

I am pretty happy that many issues appearing in #1336 actually come from `GradedAxes` and could be fixed here. I added many tests such that test failure would appear clearly in `GradedAxes` test suit instead of buried deep below a `BlockSparseArrays` error. So I agree it would be best to merge this PR as it is now.

mtfishman:
Oops, it looks like https://github.com/ITensor/ITensors.jl/pull/1562 and https://github.com/ITensor/ITensors.jl/pull/1563 caused test failures, I'll revert those. They passed separately but together they allowed an upgrade to Functors.jl v0.5 which breaks something in NDTensors.jl.

ogauthe:
All tests are passing locally. The failures come from unchanged `UnallocatedArrays`, I suppose they are not related to this PR.

mtfishman:
> I am pretty happy that many issues appearing in #1336 actually come from `GradedAxes` and could be fixed here. I added many tests such that test failure would appear clearly in `GradedAxes` test suit instead of buried deep below a `BlockSparseArrays` error. So I agree it would be best to merge this PR as it is now.

That's a good sign about the design, my goal with designing `BlockSparseArrays` was to make it as generic as possible about the axes (and also is the design goal of `BlockArrays.jl`) so I'm glad to see that seems to be working out so far. Those tests will be helpful, thanks!

mtfishman:
@ogauthe looks like the test failures are fixed now, is this ready to merge once the rest of the tests pass?

ogauthe:
> @ogauthe looks like the test failures are fixed now, is this ready to merge once the rest of the tests pass?

yes it is

mtfishman:
Thanks for the PR @ogauthe. Can you check off which issues listed in #1336 that this fixes?

PR #1536, opened 2024-10-08
[ITensorMPS] Added option for MPO-MPS zipup

corbett5:
# Description

Minimal changes necessary to extend the "zipup" algorithm for MPO-MPO products to MPO-MPS products. The main request I foresee is to bring it inline with the "naive" algorithm which also has a single implementation for MPO-MPO(S) products. Once those details are worked out I'll add docs and tests.

# How Has This Been Tested?

It has not.

# Checklist:

### DIFF
--- a/src/lib/ITensorMPS/src/mpo.jl
+++ b/src/lib/ITensorMPS/src/mpo.jl
@@ -836,15 +836,14 @@ function ITensors.contract(
   N = length(A)
   N != length(B) &&
     throw(DimensionMismatch("lengths of MPOs A ($N) and B ($(length(B))) do not match"))
-  ResultType = typeof(B)
   # Special case for a single site
-  N == 1 && return ResultType([A[1] * B[1]])
+  N == 1 && return typeof(B)([A[1] * B[1]])
   A = orthogonalize(A, 1)
   B = orthogonalize(B, 1)
   A = sim(linkinds, A)
   sA = siteinds(uniqueinds, A, B)
   sB = siteinds(uniqueinds, B, A)
-  C = ResultType(N)
+  C = typeof(B)(N)
   lCᵢ = Index[]
   R = ITensor(true)
   for i in 1:(N - 2)
--- a/src/lib/ITensorMPS/src/mpo.jl
+++ b/src/lib/ITensorMPS/src/mpo.jl
@@ -971,10 +971,7 @@ C = apply(A, B; alg="naive", truncate=false)
    for the `"naive"` version, if you just want to contract the tensors pairwise
    exactly. This can be useful if you are contracting MPOs that have diverging
    norms, such as MPOs originating from sums of local operators.
-- `alg="zipup"`: the algorithm to use for the contraction. Supported algorithms are
-    - "naive": The MPO tensors are contracted exactly at each site and then a truncation 
-    of the resulting MPO is performed.
-    $(zipup_docstring(false))
+  $(zipup_docstring(true))
 
 See also [`apply`](@ref) for details about the arguments available.
 
--- a/src/lib/ITensorMPS/src/mpo.jl
+++ b/src/lib/ITensorMPS/src/mpo.jl
@@ -826,7 +826,7 @@ function ITensors.contract(
   cutoff=1e-14,
   maxdim=maxlinkdim(A) * maxlinkdim(B),
   mindim=1,
-  truncate_kwargs=(;),
+  truncate_kwargs=(; cutoff, maxdim, mindim),
   kwargs...,
 )
   if hassameinds(siteinds, A, B)
--- a/src/lib/ITensorMPS/src/mpo.jl
+++ b/src/lib/ITensorMPS/src/mpo.jl
@@ -875,7 +875,7 @@ function ITensors.contract(
     mindim,
     kwargs...,
   )
-  truncate!(C, truncate_kwargs...)
+  truncate!(C; truncate_kwargs...)
   return C
 end
 
--- a/src/lib/ITensorMPS/src/mpo.jl
+++ b/src/lib/ITensorMPS/src/mpo.jl
@@ -826,6 +826,7 @@ function ITensors.contract(
   cutoff=1e-14,
   maxdim=maxlinkdim(A) * maxlinkdim(B),
   mindim=1,
+  truncate_kwargs=(;),
   kwargs...,
 )
   if hassameinds(siteinds, A, B)
@@ -874,6 +875,7 @@ function ITensors.contract(
     mindim,
     kwargs...,
   )
+  truncate!(C, truncate_kwargs...)
   return C
 end
 
--- a/src/lib/ITensorMPS/src/mpo.jl
+++ b/src/lib/ITensorMPS/src/mpo.jl
@@ -662,11 +662,16 @@ Choose the method with the `method` keyword, for example
 - `mindim::Int=1`: the minimal bond dimension of the resulting MPS.
 - `normalize::Bool=false`: whether or not to normalize the resulting MPS.
 - `method::String="densitymatrix"`: the algorithm to use for the contraction.
-   Currently the options are "densitymatrix", where the network formed by the
-   MPO and MPS is squared and contracted down to a density matrix which is
-   diagonalized iteratively at each site, and "naive", where the MPO and MPS
-   tensor are contracted exactly at each site and then a truncation of the
-   resulting MPS is performed.
+    - "densitymatrix": The network formed by the MPO and MPS is squared and contracted down to
+    a density matrix which is diagonalized iteratively at each site.
+    - "naive": The MPO and MPS tensor are contracted exactly at each site and then a truncation 
+    of the resulting MPS is performed.
+    - "zipup": The MPO and MPS tensors are contracted then truncated at each site without enforcing
+    the appropriate orthogonal gauge. Once this sweep is complete a call to `truncate!` occurs.
+    Because the initial truncation is not locally optimal it is recommended to use a loose
+    `cutoff` and `maxdim` and then pass the desired truncation parameters to the locally optimal
+    `truncate!` sweep via the additional keyword argument `truncate_kwargs`.
+    Suggested use is `contract(A, ψ; method="zipup", cutoff=cutoff / 10, maxdim=2 * maxdim, truncate_kwargs=(; cutoff, maxdim))`.
 
 See also [`apply`](@ref).
 """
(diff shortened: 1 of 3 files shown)


src/lib/ITensorMPS/src/mpo.jl
```diff
@@ -836,14 +836,15 @@ function ITensors.contract(
   N = length(A)
   N != length(B) &&
     throw(DimensionMismatch("lengths of MPOs A ($N) and B ($(length(B))) do not match"))
+  ResultType = typeof(B)
```
mtfishman, line 839:
```suggestion
```

src/lib/ITensorMPS/src/mpo.jl
```diff
@@ -836,14 +836,15 @@ function ITensors.contract(
...
   # Special case for a single site
-  N == 1 && return MPO([A[1] * B[1]])
+  N == 1 && return ResultType([A[1] * B[1]])
```
mtfishman, line 841:
```suggestion
  N == 1 && return typeof(B)([A[1] * B[1]])
```

src/lib/ITensorMPS/src/mpo.jl
```diff
@@ -836,14 +836,15 @@ function ITensors.contract(
...
   A = orthogonalize(A, 1)
   B = orthogonalize(B, 1)
   A = sim(linkinds, A)
   sA = siteinds(uniqueinds, A, B)
   sB = siteinds(uniqueinds, B, A)
-  C = MPO(N)
+  C = ResultType(N)
```
mtfishman, line 847:
```suggestion
  C = typeof(B)(N)
```
Just a style preference.

src/lib/ITensorMPS/src/mpo.jl
```diff
@@ -874,7 +875,6 @@ function ITensors.contract(
     mindim,
     kwargs...,
   )
-  truncate!(C; kwargs...)
```
mtfishman, line 877: Is there a reason why this line was deleted?

corbett5: Oh yeah, sorry I meant to add a comment here. I'm not really sure what the point of this extra call to truncate is especially seeing as the `maxdim` and `cutoff` kwargs are already spoken for and can't get passed to `truncate`. From looking at the code it appears that `truncate!(psi)` with no extra arguments is equivalent to `orthogonalize!(psi, 1)`, is this the intention?

src/lib/ITensorMPS/src/mpo.jl
corbett5: Made these changes.

src/lib/ITensorMPS/src/mpo.jl
mtfishman: Hmm it's a good question, it may be performing an extra orthogonalization while also making sure the bonds get truncated as much as possible. I think the idea is that zip up isn't an optimal truncation, so `truncate!` can do some extra truncation using the proper orthogonal gauge. @emstoudenmire does that sound right?

mtfishman: It looks like that "return sweep" of truncations is mentioned on page 6 of https://arxiv.org/abs/1002.1305.

corbett5: What I get from reading that section is that the initial "zipup" application should be done with a loose tolerance (or large bond dimension) and then since "zipup" leaves the result left orthogonalized it is efficient to do a further truncation sweep down to the target bond dimension.

This makes sense to me, but I don't get how that connects to the ITensor implementation. From what I can tell if you call

```julia
apply(H, psi; maxdim, cutoff)
```

that will perform the "zipup" algorithm with the given parameters and then call `truncate!(C)` which will do no further truncation and "waste" time right orthogonalizing the state. My preference would be to get rid of this call to `truncate!` and document that the "zipup" algorithm is not the most precise and should be done with a large tolerance and then truncated. Alternatively we could add more keyword arguments like `zipup_maxdim` and `zipup_cutoff` to specify the parameters of the "zipup" sweep but that seems a bit clunky to me.

mtfishman: It's a good point about maybe having different `cutoff` and `maxdim` options for the zip up sweep and the `truncate!` call. We could have a keyword argument `truncate_kwargs` which is a `NamedTuple` of keyword arguments to pass to `truncate!`, so it would be called like this:
```julia
contract(A, B; alg="zipup", cutoff=1e-9, truncate_kwargs=(; cutoff=1e-5, maxdim=50))
```

However, in your comment you are assuming the call to `truncate!` won't do further truncation even if the same `cutoff` is used by the zip up sweep and the truncation sweep. I don't think that's necessarily true, since the truncation performed by `truncate!` uses the proper orthogonal gauge of the MPS/MPO, unlike the zip up sweep, so it has more information about what states to truncate and can therefore truncate to a smaller bond dimension given a certain cutoff.

mtfishman: Oh I see, I didn't notice that the keyword arguments like `cutoff` and `maxdim` weren't being passed to `truncate!`, I think that was a mistake and we meant to pass them.

src/lib/ITensorMPS/src/mpo.jl
```diff
@@ -822,10 +822,11 @@ end
 function ITensors.contract(
   ::Algorithm"zipup",
   A::MPO,
-  B::MPO;
+  B::AbstractMPS;
   cutoff=1e-14,
   maxdim=maxlinkdim(A) * maxlinkdim(B),
   mindim=1,
+  truncate_kwargs=(;),
```
mtfishman, line 829:
```suggestion
  truncate_kwargs=(; cutoff, maxdim, mindim),
```
since I think it makes sense to use those by default in `truncate!`.

corbett5: I agree. Done.

mtfishman:
Looks good, thanks @corbett5. Could you add a test for the new functionality? Also could you add something about `truncate_kwargs` in the docstring?

corbett5:
@mtfishman I updated the docs and tests, but I seem to be having trouble with the formatting. For one the formatter updated 100+ files which I did not touch, but I also can't get the format check to pass. It passes for me locally with the same JuliaFormatter version (v2.0.0).

Also now none of the other tests pass after the extensive format changes.

mtfishman:
@corbett5 it looks like JuliaFormatter.jl v2 either accidentally or deliberately changed certain formatting rules, including introducing some bugs into the code by erroneously changing positional arguments to keyword arguments.

Please format the code using JuliaFormatter.jl v1, and I can change the formatting CI checks to use JuliaFormatter.jl v1 for now while those issues get worked out.

mtfishman:
See:
https://github.com/domluna/JuliaFormatter.jl/pull/870
https://github.com/domluna/JuliaFormatter.jl/issues/878
https://github.com/domluna/JuliaFormatter.jl/issues/879
https://github.com/domluna/JuliaFormatter.jl/issues/880

corbett5:
All passing 🎉 

Thanks for fixing the formatting problems.

mtfishman:
I actually didn't do anything, it looks like the JuliaFormatter.jl v2 release was reverted because there were too many bugs (https://github.com/domluna/JuliaFormatter.jl/issues/878).

src/lib/ITensorMPS/src/mpo.jl
```diff
@@ -662,11 +662,16 @@ Choose the method with the `method` keyword, for example
...
+    - "densitymatrix": The network formed by the MPO and MPS is squared and contracted down to
+    a density matrix which is diagonalized iteratively at each site.
+    - "naive": The MPO and MPS tensor are contracted exactly at each site and then a truncation 
+    of the resulting MPS is performed.
+    - "zipup": The MPO and MPS tensors are contracted then truncated at each site without enforcing
+    the appropriate orthogonal gauge. Once this sweep is complete a call to `truncate!` occurs.
+    Because the initial truncation is not locally optimal it is recommended to use a loose
+    `cutoff` and `maxdim` and then pass the desired truncation parameters to the locally optimal
+    `truncate!` sweep via the additional keyword argument `truncate_kwargs`.
+    Suggested use is `contract(A, ψ; method="zipup", cutoff=cutoff / 10, maxdim=2 * maxdim, truncate_kwargs=(; cutoff, maxdim))`.
```
mtfishman, line 674:
```suggestion
    For example, `contract(A, ψ; method="zipup", cutoff=cutoff / 10, maxdim=2 * maxdim, truncate_kwargs=(; cutoff, maxdim))`.
```
I think "suggested" is a bit strong, users often take suggestions in docs very seriously and don't play around with things themselves, and doubling the bond dimension could lead to very large bond dimensions in some cases.

src/lib/ITensorMPS/src/mpo.jl
```diff
@@ -945,14 +951,20 @@ C = apply(A, B; alg="naive", truncate=false)
...
    norms, such as MPOs originating from sums of local operators.
+- `alg="zipup"`: the algorithm to use for the contraction. Supported algorithms are
+    - "naive": The MPO tensors are contracted exactly at each site and then a truncation 
+    of the resulting MPO is performed.
+    - "zipup": The MPO and MPS tensors are contracted then truncated at each site without enforcing
+    the appropriate orthogonal gauge. Once this sweep is complete a call to `truncate!` occurs.
+    Because the initial truncation is not locally optimal it is recommended to use a loose
+    `cutoff` and `maxdim` and then pass the desired truncation parameters to the locally optimal
+    `truncate!` sweep via the additional keyword argument `truncate_kwargs`.
+    Suggested use is `contract(A, ψ; method="zipup", cutoff=cutoff / 10, maxdim=2 * maxdim, truncate_kwargs=(; cutoff, maxdim))`.
```
mtfishman, line 967: It's unfortunate that we are repeating the same docstring in two places.

Can we factorize out the parts that are common to `contract` and `apply` in a new string `contract_or_apply_mpo_mpo_doc` that is used by both of them?

src/lib/ITensorMPS/src/mpo.jl
corbett5: This was scheme was suggested in [Time-evolution methods for matrix-product states](https://doi.org/10.1016/j.aop.2019.167998) just below equation 39

>
> For the relaxed truncation scheme a maximal growth factor
of the MPS bond dimension m′ = 2m and a truncated weight of 1/10 of the target weight turns out
to be a suitable choice. The contraction then has complexity 2m3wσ + 2m2w2σ 2. The main cost is
in the SVD of a 2mσ × mw matrix at cost O(m3σ w), i.e. linear in w. The MPS is in left canonical
form now. A subsequent compression as described in Section 2.6 should be applied to obtain the
resulting MPS with the target bond dimension.

So I do think there is some value to these specific parameters beyond simply demonstrating how to pass different truncation arguments. But of course I will defer to your preference.

mtfishman: In that case, we could cite that paper and say: "A set of parameters suggested in [1] is ...".

corbett5: I like that.

corbett5: Done.

src/lib/ITensorMPS/src/mpo.jl
corbett5: I just pulled out the `zipup` docs, but I could do more to bring them in line if you want.

mtfishman:
@corbett5 could you move this PR over to: https://github.com/ITensor/ITensorMPS.jl?

mtfishman:
I'll close this since it needs to be moved to the ITensorMPS.jl repository.

PR #1544, opened 2024-10-14
Port VUMPS functions 

ryanlevy:
This PR adds a few of the functions living in https://github.com/ITensor/ITensorInfiniteMPS.jl/blob/main/src/ITensors.jl

The one thing missing at the moment is the more efficient `sqrt(T::DiagTensor)` which I couldn't get working, but the eigen implementation seems sane at the moment so there may be no need

### DIFF
--- a/src/tensor_operations/matrix_decomposition.jl
+++ b/src/tensor_operations/matrix_decomposition.jl
@@ -590,6 +590,27 @@ function sqrt_decomp(D::ITensor, u::Index, v::Index)
   return sqrtDL, prime(δᵤᵥ), sqrtDR
 end
 
+# Take the square root of T assuming it is Hermitian
+# TODO: add more general index structures
+function Base.sqrt(T::ITensor; ishermitian=true, atol=1e-15)
+  @assert ishermitian
+  # TODO diagonal version
+  #if isdiag(T) && order(T) == 2
+  #  return itensor(sqrt(tensor(T)))
+  #end
+  D, U = eigen(T; ishermitian=ishermitian)
+  sqrtD = D
+  for n in 1:mindim(D)
+    Dnn = D[n, n]
+    if Dnn < 0 && abs(Dnn) < atol
+      sqrtD[n, n] = 0
+    else
+      sqrtD[n, n] = sqrt(Dnn)
+    end
+  end
+  return U' * sqrtD * dag(U)
+end
+
 function factorize_svd(
   A::ITensor,
   Linds...;
--- a/src/itensor.jl
+++ b/src/itensor.jl
@@ -840,6 +840,7 @@ dirs(A::ITensor, is) = dirs(inds(A), is)
 
 # TODO: add isdiag(::Tensor) to NDTensors
 isdiag(T::ITensor)::Bool = (storage(T) isa Diag || storage(T) isa DiagBlockSparse)
+LinearAlgebra.isdiag(T::ITensor) = isdiag(T)
 
 diaglength(T::ITensor) = diaglength(tensor(T))
 
--- a/src/lib/ITensorMPS/src/mps.jl
+++ b/src/lib/ITensorMPS/src/mps.jl
@@ -1032,3 +1032,5 @@ end
 function expect(psi::MPS, op1::Matrix{<:Number}, ops::Matrix{<:Number}...; kwargs...)
   return expect(psi, (op1, ops...); kwargs...)
 end
+
+Base.getindex(ψ::MPS, r::UnitRange{Int}) = MPS([ψ[n] for n in r])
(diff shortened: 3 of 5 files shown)


ryanlevy:
Looks like there's some assumption being made that for a sliced mps e.g. `psi[2:4]` returns a vector of itensors instead of a MPS, not sure if we want to adjust that here or in ITensorInfiniteMPS

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -590,6 +590,27 @@ function sqrt_decomp(D::ITensor, u::Index, v::Index)
...
+  #end
+  D, U = eigen(T; ishermitian=ishermitian)
+  sqrtD = D
+  for n in 1:mindim(D)
+    Dnn = D[n, n]
+    if Dnn < 0 && abs(Dnn) < atol
+      sqrtD[n, n] = 0
+    else
+      sqrtD[n, n] = sqrt(Dnn)
+    end
```
mtfishman, line 609: Unfortunately this won't be good on GPU (and even on CPU will have bad efficiency since it is performing scalar indexing operations on ITensors).

ryanlevy:
Re the sliced MPS, seems the assumption is setting/getting a unit range
```julia
function deepcopy_ortho_center(M::AbstractMPS)
  M = copy(M)
  c = ortho_lims(M)
  # TODO: define `getindex(::AbstractMPS, I)` to return `AbstractMPS`
  M[c] = deepcopy(typeof(M)(M[c]))
  return M
end
```
What are your thoughts given the TODO Matt?

mtfishman:
I think slicing an MPS/MPO with a unit range should return an MPS/MPO, if we make that change we can just update other code like the one you quote accordingly.

ryanlevy:
@mtfishman now that ITensorMPS is out of ITensors I think this PR is complete pending your review and I'll move over there to fix the slice issue

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -590,6 +590,19 @@ function sqrt_decomp(D::ITensor, u::Index, v::Index)
   return sqrtDL, prime(δᵤᵥ), sqrtDR
 end
 
+# Take the square root of T assuming it is Hermitian
+# TODO: add more general index structures
+function Base.sqrt(T::ITensor; ishermitian=true, atol=1e-15)
```
mtfishman, line 595:
```suggestion
function Base.sqrt(T::ITensor; ishermitian=true, atol=eps(real(eltype(T)))
```
That is more general to other element types (like single precision).

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -590,6 +590,19 @@ function sqrt_decomp(D::ITensor, u::Index, v::Index)
...
+  @assert ishermitian
+  # TODO diagonal version
+  #if isdiag(T) && order(T) == 2
+  #  return itensor(sqrt(tensor(T)))
+  #end
```
mtfishman, line 600: Let's delete this, I don't think this would be a good implementation anyway.

src/tensor_operations/matrix_decomposition.jl
```diff
@@ -590,6 +590,19 @@ function sqrt_decomp(D::ITensor, u::Index, v::Index)
...
+  D, U = eigen(T; ishermitian=ishermitian)
```
mtfishman, line 601:
```suggestion
  D, U = eigen(T; ishermitian)
```

mtfishman:
Looks like `NDTensors.map_diag` needs to be generalized to handle the element type widening.

mtfishman:
It's a bit hacky, but one strategy could be be to change [map_diag](https://github.com/ITensor/ITensors.jl/blob/v0.9.5/NDTensors/src/tensor/tensor.jl#L450-L454) to:
```julia
# Should go in `NDTensors.Exposed`, and specialized definitions may need to be added
# for certain GPU backends.
LinearAlgebra.copy_oftype(exposed_t::Exposed, elt::Type) = LinearAlgebra.copy_oftype(unexpose(exposed_t), elt)

function map_diag(f::Function, exposed_t::Exposed)
  new_elt = Base._return_type(f, Tuple{eltype(exposed_t)})
  t_dest = LinearAlgebra.copy_oftype(exposed_t, new_elt)
  map_diag!(f, expose(t_dest), exposed_t)
  return t_dest
end
```

ryanlevy:
Thanks Matt, it seems that I'm not familiar enough with the NDTensors internals to determine what should be changed. That snippet doesn't seem to work as it gets a call to `copyto!` which is not implemented
https://github.com/ITensor/ITensors.jl/blob/main/NDTensors/src/tensor/tensor.jl#L207

PR #1546, opened 2024-10-17
[ITensorMPS] Expose new `KrylovKit.realeigsolve` features to `dmrg`

corbett5:
# Description
As of `v0.8` KrylovKit supports finding the eigenvectors of real non-symmetric operators with real eigenvalues. This eliminates the need for complex arithmetic. For my problem it produces the same results as the standard `hermitian=false` algorithm.

I added an option to specify an early return from the eigensolve. I haven't tested this out but the documentation suggests it could significantly improve performance when the initial guess is very good, which I think could be a common case in later sweeps of DMRG so worth trying out.

I also forwarded the `eigsolve_verbosity` on to `eigensolve`. Previously this argument was ignored.

# How Has This Been Tested?

It has not.

# Checklist:


mtfishman:
I didn't know about [`KrylovKit.realeigsolve`](https://jutho.github.io/KrylovKit.jl/stable/man/eig/#KrylovKit.realeigsolve), that's really cool! That's something I've wanted in multiple projects and wasn't even sure if it was possible/practical to do. @JoeyT1994 this is something we were discussing.

I'll take a look at this PR when I get a chance.

corbett5:
I measured the impact of `eigsolve_eager` and it is small but noticeable. The following is with `eigensolve_krylovdim=10` and `eigensolve_maxiter=10` with an already well converged MPS (what I imagine the optimal conditions to measure the impact is). For eight sweeps `eigensolve_eager=false` took 3726.6s ± 10s and with `eigensolve_eager=true` took 3592.8s ± 55s

mtfishman:
> I measured the impact of `eigsolve_eager` and it is small but noticeable. The following is with `eigensolve_krylovdim=10` and `eigensolve_maxiter=10` with an already well converged MPS (what I imagine the optimal conditions to measure the impact is). For eight sweeps `eigensolve_eager=false` took 3726.6s ± 10s and with `eigensolve_eager=true` took 3592.8s ± 55s

The difference should get larger as you make `eigsolve_krylovdim` larger.

corbett5:
> 
> The difference should get larger as you make `eigsolve_krylovdim` larger.

That's what I expect as well, but 10 with 10 iterations is already 33x larger than ITensors` default.

mtfishman:
> > The difference should get larger as you make `eigsolve_krylovdim` larger.
> 
> That's what I expect as well, but 10 with 10 iterations is already 33x larger than ITensors` default.

`eager` only relates to the number of Krylov states per iteration/restart, not the number of iterations/restarts. So it won't reach `maxiter` even if `eager=false` if it is converged to `tol` by the end of an earlier iteration/restart.

corbett5:
> 
> `eager` only relates to the number of Krylov states per iteration/restart, not the number of iterations/restarts. So it won't reach `maxiter` even if `eager=false` if it is converged to `tol` by the end of an earlier iteration/restart.

Gotcha, that makes sense. This is the what slowed down `tdvp` too right? https://github.com/ITensor/ITensorTDVP.jl/issues/93

mtfishman:
@corbett5 could you move this PR over to https://github.com/ITensor/ITensorMPS.jl? All of the MPS/MPO code is over there now.

mtfishman:
I'll close this PR since it needs to be moved to the ITensorMPS.jl repository.

Issue #1548, opened 2024-10-18
[ITensors] [BUG] Julia v1.11 precompile fails when PackageCompiler.jl is also part of the project


jtschneider:
**Description of bug**

I encounter a strange bug when precompiling a project environment where the `PackageCompiler.jl` package is also installed.
This only happens in the current Julia version 1.11, and does not occur in v1.10.

**Minimal code demonstrating the bug or unexpected behavior**

```julia

$ mkdir test_project

$ cd test_project

$ julia                                                                                                                                                   
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.11.1 (2024-10-16)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using Pkg

julia> Pkg.activate(".")
  Activating project at `~/Documents/Code/test_project`

julia> Pkg.add(["ITensors", "PackageCompiler"])
   Resolving package versions...
    Updating `~/Documents/Code/test_project/Project.toml`
  [9b87118b] + PackageCompiler v2.1.20
    Updating `~/Documents/Code/test_project/Manifest.toml`
  [c27321d9] + Glob v1.3.1
  [9b87118b] + PackageCompiler v2.1.20
  [05181044] + RelocatableFolders v1.0.1
  [6c6a2e73] + Scratch v1.2.1
  [4af54fe1] + LazyArtifacts v1.11.0
Precompiling project...
  ✗ ITensors → ITensorsPackageCompilerExt
  0 dependencies successfully precompiled in 15 seconds. 130 already precompiled.
  1 dependency errored.
  For a report of the errors see `julia> err`. To retry use `pkg> precompile`





```

The precompilation fails with the following error message:
<details><summary>Error message in v1.11</summary><p>

```julia
julia> versioninfo()
Julia Version 1.11.1
Commit 8f5b7ca12ad (2024-10-16 10:53 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 32 × AMD Ryzen 9 7950X 16-Core Processor
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, znver4)
[...]

PR #1549, opened 2024-10-18
[ITensors] Fix `ITensorsPackageCompilerExt` precompile issue

kmp5VT:
# Description
Fixing the issue #1548 .
The issue is related to the `precomile_itensor.jl` file's dependency on `VectorInterface`. During precompile, `ITensorsVectorIntefaceExt` is not being added which throws an error due to a missing function.

### DIFF
--- a/ext/ITensorsPackageCompilerExt/precompile_itensors.jl
+++ b/ext/ITensorsPackageCompilerExt/precompile_itensors.jl
@@ -1,4 +1,6 @@
 using ITensors.ITensorMPS: MPO, OpSum, dmrg, random_mps, siteinds
+using VectorInterface: VectorInterface
+include("$(@__DIR__)/../ITensorsVectorInterfaceExt/ITensorsVectorInterfaceExt.jl")
 
 # TODO: This uses all of the tests to make
 # precompile statements, but takes a long time
--- a/ext/ITensorsPackageCompilerExt/precompile_itensors.jl
+++ b/ext/ITensorsPackageCompilerExt/precompile_itensors.jl
@@ -1,5 +1,4 @@
 using ITensors.ITensorMPS: MPO, OpSum, dmrg, random_mps, siteinds
-using VectorInterface: VectorInterface
 include("$(@__DIR__)/../ITensorsVectorInterfaceExt/ITensorsVectorInterfaceExt.jl")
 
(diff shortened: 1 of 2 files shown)


ext/ITensorsPackageCompilerExt/precompile_itensors.jl
```diff
@@ -1,4 +1,5 @@
 using ITensors.ITensorMPS: MPO, OpSum, dmrg, random_mps, siteinds
+include("$(@__DIR__)/../ITensorsVectorInterfaceExt/ITensorsVectorInterfaceExt.jl")
```
mtfishman, line 2: This doesn't look like a good idea to do. Does it work if you load `VectorInterface.jl` to trigger the extension to load, i.e. add the line:
```julia
using VectorInterface: VectorInterface
```
It seems weird that we would have to do that though.

mtfishman: I see in https://github.com/ITensor/ITensors.jl/issues/1548#issuecomment-2422883017 you said you already tried that and it didn't work.

The thing is that `VectorInterface.jl` is a dependency of `KrylovKit.jl`, and `KrylovKit.jl` is a dependency of `ITensors.jl`, so `ITensorsVectorInterfaceExt` should get loaded automatically. So there is something strange happening with the order that the packages/extensions get loaded, I suspect it is a bug in Julia and/or Pkg.

mtfishman: Something you could try is adding `VectorInterface.jl` as a dependency of `ITensors.jl` directly and see if that forces `ITensorsVectorInterfaceExt` to get loaded. It's implicitly a dependency anyway, as I described, so it shouldn't be a problem adding it as a dependency, though it seems like a bug that we would have to do that.

One reason I made `ITensorsVectorInterfaceExt` is because when we move `ITensors.ITensorMPS` to `ITensorMPS.jl`, `KrylovKit.jl` and `VectorInterface.jl` will no longer be dependencies of `ITensors.jl`, so that extension was anticipating that code reorganization. But if it is causing problems that we can't circumvent right now we could move `ITensorsVectorInterfaceExt` into `ITensors.jl` and turn `VectorInterface.jl` into a dependency.

PR #1551, opened 2024-10-22
[ITensors] Excise ITensors.ITensorMPS

mtfishman:
Remove the `ITensors.ITensorMPS` module and move it to `ITensorMPS.jl`.

See also https://github.com/ITensor/ITensorMPS.jl/pull/31.

To-do:
- [ ] Make an announcement on the ITensor discourse page.
- [x] Assess export list.
- [x] Delete NDTensors.jl GPU DMRG tests (https://github.com/ITensor/ITensors.jl/tree/ITensors_excise_ITensorMPS/NDTensors/test/ITensors) and move them to ITensorMPS.jl. (It's getting removed in #1553.)
- [x] Add note in README announcement about `siteinds`, `OpSum`, `op`, `OpName`, `state`, etc. being removed from the exports of ITensors.jl and instead users should load ITensorMPS.jl if they want access to them.
- [x] Make an announcement on the ITensors.jl README/docs.
- [x] Remove unnecessary package dependencies from `Project.toml` (maybe check with [Aqua.jl](https://github.com/JuliaTesting/Aqua.jl)).
- [x] Define `to_nameddimsarray(x::AbstractMPS)` in `ITensorMPS.jl`.
- [x] ~~Move `ITensors.ITensorMPS` docs over to ITensorMPS.jl` and aggregate documentation with [MultiDocumenter.jl](https://github.com/JuliaComputing/MultiDocumenter.jl).~~ **EDIT:** I'd like to use [MultiDocumenter.jl](https://github.com/JuliaComputing/MultiDocumenter.jl) eventually but for now I just used: https://github.com/JuliaDocs/Documenter.jl/issues/1734.
- [x] Update exports test to reflect the exports from `ITensors.ITensorMPS` that are being removed.
- [x] Decide about exporting `dot` and `inner`.
- [x] Decide about exporting `ITensors.SiteTypes`.


src/lib/ITensorsNamedDimsArraysExt/src/to_nameddimsarray.jl
```diff
@@ -57,10 +57,3 @@ end
 ## end
 
 to_nameddimsarray(t::ITensor) = ITensor(to_nameddimsarray(t.tensor))
-
-using ITensors: ITensors
-# TODO: Delete once `TensorStorage` is removed.
-# TODO: Move to `ITensors`?
-function to_nameddimsarray(x::ITensors.ITensorMPS.AbstractMPS)
-  return to_nameddimsarray.(x)
-end
```
mtfishman, line 66: Reminder that this needs to be defined in `ITensorMPS.jl`.

emstoudenmire:
This will be the PR after which one must put `using ITensorMPS` to access that module, correct?

test/base/utils/TestITensorsExportedNames/TestITensorsExportedNames.jl
mtfishman, line 1: @emstoudenmire this is the most controversial part of this PR.

As we had discussed a while ago, my goal here is to use this code reorganization as an opportunity to cut down on the number of names being exported by ITensors.jl, which has became excessive over time.

Some of these export deletions are obvious since they are types and functions that are only used by MPS/MPO functionality, such as `MPS`, `MPO`, `dmrg`, etc.

Others are not so obvious, like `op`, `OpName`, `OpSum`, `state`, etc. A vast majority of users only use those with MPS/MPO code, and they are exported from `ITensorMPS` (see https://github.com/ITensor/ITensorMPS.jl/blob/move_code/src/exports.jl), so users can access them simply by doing `using ITensorMPS` (while currently they were accessible with `using ITensors`).

My thinking is that in the longer term, those will be accessible through external libraries (they are already organized into submodules of ITensors.jl so we are moving towards that anyway). So `ITensors.jl` will be a "core" package focused much more narrowly on the `ITensor` type and its functionality, and accessing operator definitions or building symbolic operators will be "add on" functionality that you access by loading another package, such as `ITensorMPS`, `ITensorNetworks`, and maybe in the future something like `ITensorOps` if you just want operator definitions but not tensor networks.

Anyway, curious to hear your thoughts on this.

PR #1557, opened 2024-10-25
[GradedAxes] Fix labelled_isequal namespace

ogauthe:
I realized there was a conflict with `labelled_isequal`, which was not defined as a `LabelledNumbers` inside `GradedAxes`. We could also use a different function name for this method in `GradedAxes`.


mtfishman:
This looks fine for now. I think we'll have to reassess the situation with the various types of graded unit range equality anyway at some point once the dust settles on some of these PRs.

PR #1573, opened 2024-11-07
[BlockSparseArrays] Simplify and test `adjoint(::BlockSparseMatrix)`

lkdvos:
This PR simplifies both `adjoint(::BlockSparseMatrix)` and `transpose(::BlockSparseMatrix)`.

This was discussed in #1572, but was already previously mentioned by @ogauthe as well:
https://github.com/ITensor/ITensors.jl/issues/1336#issuecomment-2254260306

Regarding your comment there about the vector types, keep in mind that I (for now) only intercepted the matrix case. I can investigate if we can make this work for `Vector` as well, keeping in mind that the `Adjoint` wrapper information is still there, it is just not the outer wrapper.

### DIFF
--- a/NDTensors/src/lib/BlockSparseArrays/src/blocksparsearray/blocksparsearray.jl
+++ b/NDTensors/src/lib/BlockSparseArrays/src/blocksparsearray/blocksparsearray.jl
@@ -1,6 +1,7 @@
 using BlockArrays: BlockArrays, Block, BlockedUnitRange, blockedrange, blocklength
 using Dictionaries: Dictionary
 using ..SparseArrayDOKs: SparseArrayDOK
+using ..GradedAxes: dual
 
 # TODO: Delete this.
 ## using BlockArrays: blocks
@@ -118,3 +119,12 @@ blockstype(::Type{<:BlockSparseArray{<:Any,<:Any,<:Any,B}}) where {B} = B
 ##   # TODO: Preserve GPU data!
 ##   return BlockSparseArray{elt}(undef, axes)
 ## end
+
+# Avoid proliferating wrapper types
+function Base.adjoint(A::BlockSparseMatrix)
+  return BlockSparseMatrix(adjoint(blocks(A)), dual.(reverse(axes(A))))
+end
+
+function Base.transpose(A::BlockSparseMatrix)
+  return BlockSparseMatrix(transpose(blocks(A)), reverse(axes(A)))
+end
--- a/NDTensors/src/lib/BlockSparseArrays/test/test_basics.jl
+++ b/NDTensors/src/lib/BlockSparseArrays/test/test_basics.jl
@@ -98,6 +98,28 @@ include("TestBlockSparseArraysUtils.jl")
     a[3, 3] = NaN
     @test isnan(norm(a))
   end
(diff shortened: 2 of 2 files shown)


mtfishman:
> Regarding your comment there about the vector types, keep in mind that I (for now) only intercepted the matrix case. I can investigate if we can make this work for `Vector` as well, keeping in mind that the `Adjoint` wrapper information is still there, it is just not the outer wrapper.

I forgot that was a reason why I was concerned about pushing the wrapper further down but I agree the current design still preserves that information. Definitely something to think carefully about.

mtfishman:
I think it would be strange for `adjoint(::BlockSparseMatrix)` to return `BlockSparseMatrix{_,_,<:Adjoint}` while `adjoint(::BlockSparseVector)` returns `Adjoint{_,<:BlockSparseVector}`. However, returning `BlockSparseVector{_,_,<:Adjoint}` would mean we need to special case operations for `BlockSparseVector{_,_,<:Adjoint}`, which is slightly annoying and repeating `Base` logic. So I'm a bit torn, maybe it would be good to think through how annoying it is to deal with `BlockSparseVector{_,_,<:Adjoint}` and get things to match `Base` behavior.

lkdvos:
I think the type of what `Base.adjoint` returns should be considered an implementation detail, as already there is actually nothing that forces you to return an `Adjoint` type (and there really shouldn't). In that sense, I don't think it really matters that these two are different.

Alternatively, I think [these lines](https://github.com/ITensor/ITensors.jl/blob/4299ab4ab699787f26ac61ec73ca1c892160e08b/NDTensors/src/lib/BlockSparseArrays/src/abstractblocksparsearray/arraylayouts.jl#L5-L9) are not correct, I think if you want to have the `Adjoint` as a wrapper, the `MemoryLayout` should be `DualLayout{BlockLayout}`.

ogauthe:
Indeed I previously proposed a similar change. I understand that extra logic would be needed for the `BlockSparseVector` case, but I believe the change is worth it. It would automatically fix slicing issues in `1336`. It would also simplify dispatch in my use case in `FusionTensors`. Currently I have to use `Union{BlockSparseMatrix,LinearAlgebra.Adjoint{<:Number,<:BlockSparseMatrix}},` to match `data_matrix`.

mtfishman:
> Indeed I previously proposed a similar change. I understand that extra logic would be needed for the `BlockSparseVector` case, but I believe the change is worth it. It would automatically fix slicing issues in `1336`. It would also simplify dispatch in my use case in `FusionTensors`. Currently I have to use `Union{BlockSparseMatrix,LinearAlgebra.Adjoint{<:Number,<:BlockSparseMatrix}},` to match `data_matrix`.

I noticed you use that in certain places, but I think you could just use `AbstractMatrix` instead. So I think you are conflating issues to some extent, and we should think about the design of `FusionTensors` separately. As Lukas said, for most code this should just be an internal implementation detail if the code is written generically.

lkdvos:
After discussion, let me close this PR as "not planned", with the goal of ensuring that the wrapper types work.

mtfishman:
Sounds good. We can keep this design in mind if the wrapper type becomes too complicated to work with.

PR #1583, opened 2024-11-14
[GradedAxes] Blocked unit range nondual

ogauthe:
This PR simplifies `blockedunitrange_getindices` by calling `nondual(a)`. It also add tests to check both axes and elements have been dualled.

The last case with `indices::AbstractBlockVector{<:Block{1}}` is slightly different as it outputs a vector of `LabelledInteger` while the other return vectors of `LabelledUnitRangeDual`.

As a side note, the important information about dual is in the axes of the `BlockVector` and not in its element. Therefore we may completely remove `LabelledUnitRangeDual` (and just take the element dual of a `LabelledUnitRange`) without affecting `BlockSparseArray`.


NDTensors/src/lib/GradedAxes/src/gradedunitrangedual.jl
```diff
@@ -57,61 +57,41 @@ end
...
+  v = blockedunitrange_getindices(nondual(a), indices)
+  # v elements are not dualled by dual_axes due to different structure.
+  # take element dual here.
+  return dual_axes(dual.(v))
+end
+
+function dual_axes(v::BlockVector)
+  # dual both v elements and v axes
+  block_axes = dual.(axes(v))
+  flipped = mortar(dual.(vec.(blocks(v))), block_axes)
```
mtfishman, line 94:
```suggestion
  return mortar(dual.(vec.(blocks(v))), block_axes)
```
(and removing the line below). What is the `vec` call for?

mtfishman:
Looks good, thanks! Nice to see this code get simplified.

> As a side note, the important information about dual is in the axes of the `BlockVector` and not in its element. Therefore we may completely remove `LabelledUnitRangeDual` (and just take the element dual of a `LabelledUnitRange`) without affecting `BlockSparseArray`.

I was picturing it might be used to preserve dual information in operations like:
```julia
r = gradedrange([U1(0) => 2, U1(1) => 3])
a = BlockSparseArray{Float64}(r, dual(r))
axes(@view(a[Block(1, 1)])) # (LabelledUnitRange, LabelledUnitRangeDual)
```
But to me that is a good use case for having a `SectorDual`, so we can just use `LabelledUnitRange` with dual label instead of a devoted type `LabelledUnitRangeDual`.

NDTensors/src/lib/GradedAxes/src/gradedunitrangedual.jl
ogauthe: Indeed `vec` is not needed

PR #1595, opened 2024-11-18
[SparseArraysBase] Update README

lkdvos:
This PR pushes an initial update to the README of SparseArraysBase to elaborate a bit on design choices and philosophies.


mtfishman:
It looks like a new `GPUArrays.jl` release broke some tests, obviously unrelated to this PR. I'll merge this PR and we'll have to investigate those issues separately, or temporarily downgrade the compat of `GPUArrays.jl` if it isn't easy to fix.

PR #1603, opened 2024-11-25
[Docs] Update docs for ITensorMPS

ryanlevy:
# Description

Pointed out by Bo Xiao, some of the docs got were missed in updating to ITensorMPS. I did a quick glance and these are what I found that were missing.


mtfishman:
Thanks! That's why we should use [doctests](https://documenter.juliadocs.org/stable/man/doctests).

PR #1615, opened 2025-02-09
[DOCS] Fix QN Index docs regarding `dir` keyword argument

mtfishman:
See https://itensor.discourse.group/t/index-direction/977.

@emstoudenmire the controversial thing I did here was not show the `dir` keyword argument at all, I can't think of a reason why an end user should be using that rather than creating a new Index and calling `dag` on it.


emstoudenmire (approved):
Yes, it's a good idea to just have users wrap the constructor in `dag` or call `dag` afterward rather than use the keyword argument. So I approve this PR.

Issue #1620, opened 2025-02-20
Problem adding ITensors


gach213:
Hi, 
I am new to using ITenosrs. I am trying to do the standard add ITenosrs in Julia REPL but I am getting an error (Because of its long trace, I am just copying at the end). I tried standard commands like resolve, precompile but did not help. Not sure if it's even important, and perhaps I can still use it for my purpose. 

KeyError: key "TypeParameterAccessors" not found
Stacktrace:
 [1] getindex
    @ ./dict.jl:498 [inlined]
  [2] status_ext_info(pkg::Pkg.Types.PackageSpec, env::Pkg.Types.EnvCache)
    @ Pkg.Operations /Applications/Julia-1.10.app/Contents/Resources/julia/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:2214
  [3] print_status(env::Pkg.Types.EnvCache, old_env::Pkg.Types.EnvCache, registries::Vector{Pkg.Registry.RegistryInstance}, header::Symbol, uuids::Vector{Base.UUID}, names::Vector{String}; manifest::Bool, diff::Bool, ignore_indent::Bool, outdated::Bool, extensions::Bool, io::Base.TTY, mode::Pkg.Types.PackageMode, hidden_upgrades_info::Bool, show_usagetips::Bool)
    @ Pkg.Operations /Applications/Julia-1.10.app/Contents/Resources/julia/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:2299
  [4] print_status
    @ /Applications/Julia-1.10.app/Contents/Resources/julia/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:2241 [inlined]
  [5] status(env::Pkg.Types.EnvCache, registries::Vector{Pkg.Registry.RegistryInstance}, pkgs::Vector{Pkg.Types.PackageSpec}; header::Symbol, mode::Pkg.Types.PackageMode, git_diff::Bool, env_diff::Pkg.Types.EnvCache, ignore_indent::Bool, io::Base.TTY, outdated::Bool, extensions::Bool, hidden_upgrades_info::Bool, show_usagetips::Bool)
    @ Pkg.Operations /Applications/Julia-1.10.app/Contents/Resources/julia/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:2475
  [6] status (repeats 2 times)
[...]

Issue #1618, opened 2025-02-17
[ITensors] [BUG] Failed to precompile: "KeyError: key "TypeParameterAccessors" not found"


JaydevSR:
The package fails to precompile giving the following error:

```julia
pkg> precompile
Precompiling project...
  ✗ ITensors
  ✗ ITensors → ITensorsVectorInterfaceExt
  0 dependencies successfully precompiled in 7 seconds. 119 already precompiled.

ERROR: The following 2 direct dependencies failed to precompile:

ITensors

Failed to precompile ITensors [9136182c-28ba-11e9-034c-db9fb085ebd5] to "C:\\Users\\u0174972\\.julia\\compiled\\v1.11\\ITensors\\jl_E7E2.tmp".
ERROR: LoadError: KeyError: key "TypeParameterAccessors" not found
Stacktrace:
  [1] getindex
    @ .\dict.jl:477 [inlined]
  [2] _insert_extension_triggers(parent::Base.PkgId, extensions::Dict{String, Any}, weakdeps::Dict{String, Any})
    @ Base .\loading.jl:1471
  [3] insert_extension_triggers(env::String, pkg::Base.PkgId)
    @ Base .\loading.jl:1434
  [4] insert_extension_triggers
    @ .\loading.jl:1390 [inlined]
  [5] __require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base .\loading.jl:2309
  [6] #invoke_in_world#3
    @ .\essentials.jl:1088 [inlined]
  [7] invoke_in_world
    @ .\essentials.jl:1085 [inlined]
  [8] _require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base .\loading.jl:2287
  [9] macro expansion
    @ .\loading.jl:2226 [inlined]
 [10] macro expansion
    @ .\lock.jl:273 [inlined]
 [11] __require(into::Module, mod::Symbol)
    @ Base .\loading.jl:2183
 [12] #invoke_in_world#3
    @ .\essentials.jl:1088 [inlined]
 [13] invoke_in_world
    @ .\essentials.jl:1085 [inlined]
 [14] require(into::Module, mod::Symbol)
    @ Base .\loading.jl:2176
 [15] include(mod::Module, _path::String)
    @ Base .\Base.jl:557
 [16] include(x::String)
    @ ITensors C:\Users\u0174972\.julia\packages\ITensors\Zs2nC\src\ITensors.jl:51
 [17] top-level scope
[...]

PR #1623, opened 2025-02-22
[NDTensors] Support for older versions of Julia 1.10 and 1.11

mtfishman:
This circumvents an issue in older versions of Julia 1.10 and 1.11 that made it so that the latest versions of NDTensors.jl/ITensors.jl couldn't be installed, caused by how we have some of our package extensions were configured. Basically, some package extensions depend on TypeParamaterAccessors.jl, but so does the main NDTensors.jl package, so TypeParamaterAccessors.jl is both a dependency and a weak dependency, and in older version of Julia I couldn't find a good way to handle that situation (which I think basically was a bug in the package system in older versions of Julia 1.10 and 1.11 but the origin of the issue is hard to track down). Now, I am using `NDTensors.TypeParamaterAccessors` in the package extensions, so it is no longer seen as a dependency of the package extensions and can just be handled as a normal dependency of the NDTensors.jl package.

This issue has been brought up a few times recently: https://github.com/ITensor/ITensors.jl/issues/1618, https://github.com/ITensor/ITensors.jl/issues/1620, https://itensor.discourse.group/t/typeparameteraccessors-not-found-error-on-julia-v-1-10-0/2260.


emstoudenmire:
Updated on itensor.org News page too:
https://itensor.org/news/25_02_20_julia_versions.html

mtfishman:
Good idea, thanks @emstoudenmire!

PR #1644, opened 2025-04-07
Update test_phys_site_types.jl

andrewkhardy:
Adding Sy test to confirm it is Sy operator on tJ sites


andrewkhardy:
Ah sorry, did not mean to open new PR

PR #1646, opened 2025-04-15
[ITensors] Use `task_local_storage` for Index ID RNG

mtfishman:
This uses [`task_local_storage`](https://docs.julialang.org/en/v1/base/parallel/#Base.task_local_storage-Tuple{Any}) in the Index ID RNG to avoid thread-local states as recommended in https://julialang.org/blog/2023/07/PSA-dont-use-threadid.

Issues with the previous design were pointed out by @amilsted.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ITensors"
 uuid = "9136182c-28ba-11e9-034c-db9fb085ebd5"
 authors = ["Matthew Fishman <mfishman@flatironinstitute.org>", "Miles Stoudenmire <mstoudenmire@flatironinstitute.org>"]
-version = "0.9.3"
+version = "0.9.4"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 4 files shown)


mtfishman:
@amilsted can you test this out on cases where the previous code was failing?

amilsted:
We monkey-patched those changes into an older version and it indeed fixed the assertion errors we were seeing.

FWIW, just using the `default_rng()` also worked. Is it functionally any different than this?

mtfishman:
> We monkey-patched those changes into an older version and it indeed fixed the assertion errors we were seeing.

Great, thanks for checking.

> FWIW, just using the `default_rng()` also worked. Is it functionally any different than this?

I want to use a separate RNG stream from the one used by the `rand(elt, dims)`/`randn(elt, dims)` array constructors, since it could cause Index ID clashes if someone sets the seed of `default_rng()` with `Random.seed!(seed)` (that's particularly an issue with saving and loading ITensors). It isn't clear to me that I can "copy" the `default_rng()` for that purpose, but maybe I'm missing something.

amilsted:
Btw, credit to @wildart for finding the issue and testing the changes.

PR #1656, opened 2025-05-12
[NDTensors] Better `eigen` backend in Julia v1.12

mtfishman:
See https://itensor.discourse.group/t/issue-with-linearalgebra-and-julia-1-12/2362.

@VinceNeede

I'll wait to see what happens with https://github.com/JuliaLang/LinearAlgebra.jl/issues/1313, the default might be changed before the official Julia v1.12 release.


mtfishman:
I'll close this since it should be fixed in Julia 1.12 by https://github.com/JuliaLang/LinearAlgebra.jl/pull/1379.

PR #1659, opened 2025-06-30
Fix bug in QN where sectors sometimes aren't sorted

mtfishman:


emstoudenmire:
Very quick fix... I'm glad with the new sectors code we moved to set operations instead.

PR #1664, opened 2025-08-06
Fix exp check for fermionic ITensors

emstoudenmire:
This PR fixes the check inside the `exp` function so it is restricted to fermionic ITensors (and when the fermion system is on) only. 

Choosing `Linds` and `Rinds` to all have the same arrows within each set may be advisable for most cases, but we decided to allow mixed-arrow `exp` for now.

This PR also wraps selected tests in generated modules. Previously the `test_qnitensors.jl` file would error if run standalone, due to a missing import.

### DIFF
--- a/src/tensor_operations/matrix_algebra.jl
+++ b/src/tensor_operations/matrix_algebra.jl
@@ -65,8 +65,8 @@ function exp(A::ITensor, Linds, Rinds; ishermitian=false)
   end
 
   # <fermions>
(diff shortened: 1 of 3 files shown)


mtfishman:
Looks good, can you add a test of calling `exp` on a bosonic and/or non-QN ITensor with auto fermion enabled?

emstoudenmire:
Good suggestion – just added it

PR #1669, opened 2025-09-09
Fix sqrt decomp

JoeyT1994:
A quick fix is implemented to preserve the ITensor type in the `sqrt_decomp` function and allow it to work on GPUs.

### DIFF
--- a/src/tensor_operations/matrix_decomposition.jl
+++ b/src/tensor_operations/matrix_decomposition.jl
@@ -581,8 +581,8 @@ using NDTensors: map_diag!
 function sqrt_decomp(D::ITensor, u::Index, v::Index)
   (storage(D) isa Union{Diag,DiagBlockSparse}) ||
     error("Must be a diagonal matrix ITensor.")
-  sqrtDL = diag_itensor(u, dag(u)')
-  sqrtDR = diag_itensor(v, dag(v)')
+  sqrtDL = adapt(datatype(D))(diag_itensor(u, dag(u)'))
+  sqrtDR = adapt(datatype(D))(diag_itensor(v, dag(v)'))
   map_diag!(sqrt ∘ abs, sqrtDL, D)
   map_diag!(sqrt ∘ abs, sqrtDR, D)
   δᵤᵥ = copy(D)


src/tensor_operations/matrix_decomposition.jl
```diff
@@ -581,8 +581,8 @@ using NDTensors: map_diag!
 function sqrt_decomp(D::ITensor, u::Index, v::Index)
   (storage(D) isa Union{Diag,DiagBlockSparse}) ||
     error("Must be a diagonal matrix ITensor.")
-  sqrtDL = diag_itensor(u, dag(u)')
-  sqrtDR = diag_itensor(v, dag(v)')
+  sqrtDL = adapt(datatype(D))(diag_itensor(u, dag(u)'))
+  sqrtDR = adapt(datatype(D))(diag_itensor(v, dag(v)'))
```
mtfishman, line 585:
```suggestion
  sqrtDL = adapt(datatype(D), diag_itensor(u, dag(u)'))
  sqrtDR = adapt(datatype(D), diag_itensor(v, dag(v)'))
```
Just a style thing, normally I would only use `adapt(datatype(D))` if I am passing it as an argument to another function, for example `map(adapt(datatype(D)), itensor_list)`.

mtfishman:
Can you bump the package version as well, and also add a test?

mtfishman:
Test failures are unrelated, looks like it is hitting a known bug in the cuTENSOR backend: https://github.com/JuliaGPU/CUDA.jl/issues/2881.

test/base/test_svd.jl
```diff
@@ -206,6 +206,8 @@ include(joinpath(@__DIR__, "utils", "util.jl"))
 
     for dir in [ITensors.Out, ITensors.In]
       L, R, spec = ITensors.factorize_svd(A, l1, l2; dir, ortho="none")
+      @test datatype(L) == datatype(A)
+      @test datatype(R) == datatype(A)
       @test dir == ITensors.dir(commonind(L, R))
       @test norm(L * R - A) <= 1e-14
     end
```
mtfishman, line 213: A trick we can use here to test different data types is to add a loop over adapting to different array types, for example:
```julia
# Import at top of file
using JLArrays: JLArray

for arrayt in (Array, JLArray)
  A′ = adapt(arrayt, A)
  # factorize_svd test of A′
end
```
JLArray acts like a GPU array but runs on CPU so is a good basic test that the code should run on GPU.

PR #1670, opened 2025-09-18
Support for eigen of dual operator

emstoudenmire:
This PR fixes the `eigen` function to handle fermionic matrices or operators which have an In,Out index ordering (e.g. "dual operators") such as density matrices or density-matrix-like messages. It also includes new tests.

Specifically, in cases where the tensor components are a Hermitian positive semidefinite (HPSD) matrix, then we want eigen to return a diagonal "D" ITensor which has non-negative entries when its indices are in the In, Out ordering. 

Commits:
- **Initial fix of arrows for dual, HPSD fermionic matrix**
- **Format & remove comment**
- **Add tests for both types of HPSD fermionic matrices**
- **Bump version**

### DIFF
--- a/test/base/test_decomp.jl
+++ b/test/base/test_decomp.jl
@@ -520,8 +520,6 @@ end
     @test norm(prime(U)*D_O*dag(U)-O) < 1E-10
     @test all(>=(0.0), diag(array(D_O)))
 
-    println()
-
     #
     # HPSD Dual operator (In,Out) case
     #
--- a/test/base/test_decomp.jl
+++ b/test/base/test_decomp.jl
@@ -510,7 +510,7 @@ end
     # HPSD Operator (Out,In) case
     #
     M = random_itensor(s, dag(t))
-    O = prime(M,s)*dag(M)
+    O = prime(M, s)*dag(M)
 
     @test dir(inds(O)[1]) == ITensors.Out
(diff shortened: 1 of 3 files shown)


test/base/test_decomp.jl
```diff
@@ -500,4 +500,46 @@ end
...
+    @test dir(inds(O)[1]) == ITensors.Out
+    @test dir(inds(O)[2]) == ITensors.In
+    linds = [s']
+    rinds = [dag(s)]
+    D_O, U = eigen(O, linds, rinds; ishermitian=true)
+    @test norm(prime(U)*D_O*dag(U)-O) < 1E-10
+    @test all(>=(0.0), diag(array(D_O)))
+
+    println()
+
```
mtfishman, line 524:
```suggestion
```

emstoudenmire:
@mtfishman does this PR look ok to you now that tests are passing? (Except Jenkins which I'm not sure about.)

mtfishman:
Looks good to me.

Test failures are unrelated, it may be related to a cuTENSOR issue discussed here: https://github.com/JuliaGPU/CUDA.jl/issues/2881. If I'm following that thread correctly, an explanation might be that our Jenkins setup is running on a GPU that is not supported by the latest versions of cuTENSOR.

PR #1711, opened 2026-03-06
Use shared IntegrationTest workflow, add Tennis downstream test

mtfishman:


mtfishman:
@emstoudenmire note this adds Tennis.jl as a downstream test as well so we can be more assured that internal changes (say optimizations of the auto fermion system) don't break Tennis.jl.

emstoudenmire:
This will be great for Tennis development!

PR #1721, opened 2026-04-03
Blocksparse CuTensor contraction backend

kmp5VT:


mtfishman:
Great to see this, thanks @kmp5VT. I guess this relies on https://github.com/JuliaGPU/CUDA.jl/pull/3057? Once that is merged, would we just need to install the latest version of cuTENSOR/cuTENSOR.jl and the new backend in this PR "just works"?

emstoudenmire:
Looks nice and very minimal. Thanks Karl!

kmp5VT:
@mtfishman In NDTensors I also added a internal variable `_cutensor_blocksparse` which you can enable/disable during runtime. So if the variable is enabled plus you have the changes in my cuda branch then it will automatically work. And from my tests, the new backend produces noticeable speedups over the previous blocksparse GPU backends

NDTensors/ext/NDTensorscuTENSORExt/contract.jl
```diff
@@ -12,6 +13,66 @@ function to_zero_offset_cuarray(a::ReshapedArray)
     return copy(expose(a))
 end
 
+function block_extents(ind)
+    return ntuple(i -> ind.space[i].second, nblocks(ind))
+end
+
+#### Functions to turn Tensors into BlockSparseCuTensors for contraction
+function ITensor_to_cuTensorBS(T::BlockSparseTensor)
```
mtfishman:
```suggestion
function to_cuTensorBS(T::BlockSparseTensor)
```

kmp5VT:
Hi @mtfishman @emstoudenmire , the blocksparse cutensor wrapper is now merged into CUDA.jl main so this feature is no longer experimental. The biggest question I have is, is the current implementation (i.e. introducing a blocksparse cutensor global variable in NDTensors) sufficient or do you have an alternative implementation in mind?  I would guess that we would want to remove the legacy canonical cutensor code (for blocksparse arrays) after benchmark tests have been collected.

mtfishman:
Hi Karl, great to hear it, thanks for getting that implemented! We're very excited to try this out.

It's a good question. I think we should introduce this as "opt in" at first so we can have time to test it out, then if we're convinced it is either on par or better than the current implementation (which I'm guessing it is) we can make it the default when cuTENSOR is loaded. I'm thinking of trying out a new design based on ScopedValues for selecting algorithms/backends, would you mind if I made a PR based on this one testing out that design?

kmp5VT:
@mtfishman No I don't mind at all. I can also give you access to this branch if you want to add it here. Let me know!

PR #1754, opened 2026-05-28
[jenkins] updates for migration to new k8s-based system

dylex:
We are moving jenkins builds to a [new system](https://jenkins-new.flatironinstitute.org/job/CCQ/job/ITensor/job/ITensors.jl/) with more resources and flexibility. This commit will switch to allowing builds on the new system.

Once this is merged, I can remove the project on the old system.

(There's a couple other ITensor projects on the old system: `ITensor` and `ITensorGPU.jl`, but neither of these has been built in years, so I assume they're no longer needed.)


mtfishman:
Thanks @dylex!

PR #1767, opened 2026-09-05
CompatHelper: bump compat for InlineStrings to 2 for package NDTensors, (keep existing compat)

ITensorBot:
This pull request changes the compat entry for the `InlineStrings` package from `1` to `1, 2` for package NDTensors.
This keeps the compat entries for earlier versions.

Note: I have not tested your package with this new compat entry.
It is your responsibility to make sure that your package tests pass before you merge this pull request.

PR #1769, opened 2026-09-10
[NDTensors] Drop the unused InlineStrings dependency

mtfishman:
## Summary

NDTensors declares `InlineStrings` and loads it in `imports.jl`, but no name from it is referenced anywhere in the package. The tag and string machinery is built on `BitIntegers` instead, so the dependency is dead weight. This supersedes https://github.com/ITensor/ITensors.jl/pull/1767, which widened the compat bound on it.


mtfishman:
Folded into https://github.com/ITensor/ITensors.jl/pull/1763 to keep the NDTensors changes in one version bump.

PR #1770, opened 2026-09-10
[NDTensors] Route JLArray svd and qr through the CPU

mtfishman:
## Summary

Adds an `svd` method for `JLMatrix` and routes the existing `qr` through `cpu`, matching the Metal and AMDGPU extensions. Julia 1.13's `svd!` counts the singular values above a tolerance, which indexes them one at a time and so fails on a `JLArray`, and JLArrays 0.4 will disallow the pointer conversion that `qr` relied on.


mtfishman:
Folded into https://github.com/ITensor/ITensors.jl/pull/1763 to keep the NDTensors changes in one version bump.

